PR #285, opened 2013-05-29
Updated video tests to be more correct.

kenrussell:
The video-related WebGL conformance tests
(tex-image-and-sub-image-2d-with-video.html with variants, and
texture-npot-video.html) currently start the test upon receiving the
"playing" event. After receiving feedback from the Chrome on Android
team, and studying the HTMLVideoElement spec, it seems that this is
incorrect.
http://www.w3.org/html/wg/drafts/html/master/embedded-content-0.html#mediaevents
says for the "playing" event: "Even if this event fires, the element
might still not be potentially playing".

Watching for a "timeupdate" event where the video's currentTime is
past the start time seems to be a more reliable heuristic.

There are bugs in Safari, including current WebKit nightlies, with
dynamically created video elements, and video elements with
style="display:none". This patch improves the reliability with
dynamically created video elements, but causes reliable failures with
display:none elements (used by texture-npot-video.html). This test was
already previously failing in Safari, so I believe it is warranted to
update all versions of these video tests in the same way to generally
improve their reliability, and assume that this bug will be fixed in
Safari at some point in the future.

Tested all versions of these tests with Chrome, Firefox and Safari on
Mac OS. Chrome and Firefox pass the tests completely. The current
WebKit nightly now passes the 1.0.2 and top of tree
tex-image-and-sub-image-with-video.html tests, where it was failing
them before.

### DIFF
--- a/conformance-suites/1.0.1/conformance/textures/tex-image-and-sub-image-2d-with-video.html
+++ b/conformance-suites/1.0.1/conformance/textures/tex-image-and-sub-image-2d-with-video.html
@@ -16,9 +16,28 @@ var wtu = WebGLTestUtils;
 var gl = null;
 var textureLoc = null;
 var successfullyParsed = false;
+var video;
+var gotPlaying = false;
+var gotTimeUpdate = false;
+var runningTest = false;
 
 if (window.initNonKhronosFramework) {
     window.initNonKhronosFramework(true);
+}
+
+function playingListener() {
+    gotPlaying = true;
+    maybeRunTest(video);
+}
+
+function timeupdateListener() {
+    // Checking to make sure the current time has advanced beyond
+    // the start time seems to be a reliable heuristic that the
+    // video element has data that can be consumed.
+    if (video.currentTime > 0.0) {
+        gotTimeUpdate = true;
+        maybeRunTest(video);
+    }
 }
 
 function init()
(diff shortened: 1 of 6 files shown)


dano2:
Hello. I believe this is related. The following video will play in an Android browser, but when you click to take a snapshot, it's ignored.
My research tells me it's related to the drawImage method. Is this thread suggesting there is a workaround for this? 

Thanks.

http://www.html5multimedia.com/code/ch9/video-canvas-screenshot.html

kenrussell:
@dano2 : the reason your video can't be drawn to a canvas is due to a security bug that was fixed both in Chrome and Android L. If you run that page in Chrome on Android L, it works properly.

dano2:
Hi. I'm using the chrome browser v 40.0.2214.109 on an android 4.4.2, htc desire 510, and going to this link.
http://www.html5multimedia.com/code/ch9/video-canvas-screenshot.html
While the video is playing, clicking the snapshot button, does nothing. I'm only experiencing this issue on android. Did the fix, by chance, require a new setting or passing a different flag or something or are you saying this code (drawImage) should work exactly as is? 

Thanks

kenrussell:
@dano2 : this isn't the right place to have an extended conversation about this topic, but as I mentioned before, the reason the drawImage call doesn't work on your phone is that it isn't upgraded to Android L. Drawing videos to both 2D canvas and WebGL was disabled in Chrome on Android K and below due to a security bug. If you find a phone which is running Android L the code will work as expected again.

PR #342, opened 2013-08-13
Test for invalidated extensions after a lost context.

grorg:


jdashg:
Did we want to fix this up? It seems like it was mainly to check the `ext.invalidated` flag, which we've clarified isn't 'physical'.
If we want to test that extensions get detached properly, I think that should be done in the individual extension tests.
Can we close this? @grorg

jdashg:
I'll just close this.

PR #789, opened 2014-11-21
Promote color-buffer-float exts to Community Approved.

kdashg:
The float-texture extensions implicitly enable these extensions, so they should clearly come out from draft.


gouletr:
:+1: thanks for this!

PR #805, opened 2014-12-11
Clarify interactions with ImageBitmap per https://www.w3.org/Bugs/Public...

kenrussell:
.../show_bug.cgi?id=27513 .


ds-hwang:
nice. lgtm although too late :)

PR #810, opened 2014-12-13
Backport bug fix from pull request #803 to earlier conformance suite sna...

kenrussell:
...pshots.


MARSMK908:
modify on wiki links

PR #811, opened 2014-12-15
Make tex-image-and-sub-image-2d-with-webgl-canvas-* cover TEXTURE_CUBE_MAP.

ds-hwang:
Make tex-image-and-sub-image-2d-with-webgl-canvas-* cover TEXTURE_CUBE_MAP.

Improve tex-image-and-sub-image-2d-with-webgl-canvas-* to check http://crrev.com/735623003/
Change from setCanvasTo1x2 to setCanvasTo2x2 because cube map allows only square source.

### DIFF
--- a/sdk/tests/conformance/resources/tex-image-and-sub-image-2d-with-webgl-canvas.js
+++ b/sdk/tests/conformance/resources/tex-image-and-sub-image-2d-with-webgl-canvas.js
@@ -37,8 +37,6 @@ function generateTest(pixelFormat, pixelType, prologue) {
             return;
         }
 
-        var program = wtu.setupTexturedQuad(gl);
-
         gl.clearColor(0,0,0,1);
         gl.clearDepth(1);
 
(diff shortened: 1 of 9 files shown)


ds-hwang:
Thank you for review and sorry for delaying reply.
I apply it to sdk and rollback in 1.0.3 snapshot.

@toji, @zhenyao could you review? because @kenrussell is ooo.

sdk/tests/conformance/resources/tex-image-and-sub-image-2d-with-webgl-canvas.js
```diff
@@ -62,33 +63,86 @@ function generateTest(pixelFormat, pixelType, prologue) {
...
-    function setCanvasTo257x257(ctx) {
+    function setCanvasTo257x257(ctx, bindingTarget) {
       ctx.canvas.width = 257;
       ctx.canvas.height = 257;
       setCanvasToRedGreen(ctx);
     }
 
-    function setCanvasTo1x2(ctx) {
-      ctx.canvas.width = 1;
+    function setCanvasTo1x2(ctx, bindingTarget) {
```
zhenyao: Maybe change function name to setCanvasToMin, because it's no longer always 1x2

sdk/tests/conformance/resources/tex-image-and-sub-image-2d-with-webgl-canvas.js
```diff
@@ -26,6 +26,9 @@ function generateTest(pixelFormat, pixelType, prologue) {
     var gl = null;
     var successfullyParsed = false;
 
+    var TEXTURE_2D = 0;
+    var TEXTURE_CUBE_MAP = 1;
+
```
zhenyao: Is it necessary to create these two names?  It seems you can just use gl.TEXTURE_2D and gl.TEXTURE_CUBE_MAP as gl is initialized right in the beginning.

ds-hwang:
> Maybe change function name to setCanvasToMin, because it's no longer always 1x2
done.

> Is it necessary to create these two names? It seems you can just use gl.TEXTURE_2D and gl.TEXTURE_CUBE_MAP as gl is initialized right in the beginning.
done. reuse gl.TEXTURE_2D and gl.TEXTURE_CUBE_MAP. gl initialization is not needed to move.

ds-hwang:
@kenrussell @zhenyao I makes the other tex-image-and-sub-image-2d-* tests cover cube map also.
Could you review again?

zhenyao:
LGTM

PR #814, opened 2014-12-30
added a test for a sampler in a struct passed to a function

pyalot:
This test passes on desktops and iOS, but fails on my Nexus 4.


pyalot:
> highp support in fragment shader is optional. You should use mediump instead for the test to pass anywhere.

Replaced highp with mediump

zhenyao:
LGTM.  Sorry about the delay. It slipped out of my radar.

PR #816, opened 2015-01-09
Add test for call to drawArraysInstancedANGLE with 'first' > 0

joejohndoe:
This test catches a bug that is currently present in Chrome/ANGLE DX11:
https://code.google.com/p/chromium/issues/detail?id=447140
https://code.google.com/p/angleproject/issues/detail?id=864

I'm setting up a different quad and offset buffer, because the setup used for the existing tests draws outside of the full viewport (each instance is centered at one corner of the viewport and 3/4 of it lie outside of the viewport). With that setup, it would be impossible to check if the last instance has drawn, because its second triangle lies completely outside of the viewport.


kenrussell:
Thanks for this contribution and sorry for the delay reviewing it. Looks good.

PR #817, opened 2015-01-12
Angle-instanced-arrays calls notifyFinishedToHarness multiple times

RafaelCintron:
Conformance tests notify the harness of completeness one of two ways:
1) Manually include "var successfullyParsed = true;" and
    js-test-post.js at the end of the test. js-test-post.js calls notifyFinishedToHarness
2) Call finishTest at the of the test. FinishTest performs the first method programatically.

Angle-instanced-arrays does both of these things. The "TEST COMPLETED" message shows
up twice in the final output.

Since finishTest is called as part of multiple setTimeout callbacks, we end up
notifying the harness of completeness (through the manually inserted script tag)
before the test actually finishes.

The fix is to delete js-test-post.js from the end of the test and rely on finishTest exclusively.

### DIFF
--- a/conformance-suites/1.0.3/conformance/extensions/angle-instanced-arrays.html
+++ b/conformance-suites/1.0.3/conformance/extensions/angle-instanced-arrays.html
@@ -77,6 +77,7 @@ var program;
 
 if (!gl) {
     testFailed("WebGL context does not exist");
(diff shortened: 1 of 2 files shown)


zhenyao:
Thanks for the fix.  Looks good to me.

PR #818, opened 2015-01-12
framebuffer-bindings-...-to-data-url does not report completeness to test harness

RafaelCintron:
The framebuffer-bindings-affected-by-to-data-url test needs to report completeness back to
the test harness. Otherwise, the test harness will timeout the test.

To fix this we add js-test-post.js to the end of the test.


zhenyao:
Thanks for catching this.  It slipped through in my reviewing of this test.

PR #821, opened 2015-01-23
Buffer calls to console.log to reduce amount of debugging output.

kenrussell:
Only output the buffered logs if the test fails -- i.e., calls testFailed().

This reduces the amount of debugging output for a typical run of the
conformance suite, while still remaining debuggable in case of errors.

Tested by injecting a synthetic failure in a test which previously had
debugging output; verified that the output was printed, and only then.


kenrussell:
Two further comments:

1. This is for crbug.com/451017 .
2. I'm not 100% sure what will happen in the case a test hangs / times out, but this seems to be urgently requested so I'd like to defer that investigation.

toji:
LGTM

PR #822, opened 2015-01-24
rejecting WEBGL_draw_elements_no_range_check

pyalot:
Olli indicated that this extension might be best dropped.


kenrussell:
Recent discussions in the working group have finally reached consensus that the KHR_robustness extension will be allowed to provide the security guarantees required for out-of-range indices in drawElements (and drawArrays) calls. This change has been made in https://github.com/KhronosGroup/WebGL/pull/834 and https://github.com/KhronosGroup/WebGL/issues/835 has been filed for follow-on work, including updating the ANGLE_instanced_arrays WebGL extension.

This extension is now no longer needed. Merging this pull request. Thanks for helping clean up the extension registry.

PR #824, opened 2015-01-24
rejecting WEBGL_security_sensitive_resources

pyalot:
Olli indicated that this extension should be dropped because it's difficult to implement and would introduce performance penalties.


kenrussell:
We're still going to work on this one. There's currently no better proposal for potentially integrating arbitrary HTML content into WebGL scenes.

pyalot:
The extension has been in draft for a year and a half. Please provide reasoning on the ML.

PR #827, opened 2015-01-24
reject WEBGL_debug_shader_precision

pyalot:
proposing (pending objections) to reject WEBGL_debug_shader_precision


kenrussell:
Olli should comment on this. It is definitely possible to compile ANGLE's shader translator with Emscripten, though I haven't spent the time to turn it into a nice module. If that were done then this could fairly easily be implemented as a library outside the browser.

Oletus:
I agree to this. There are also other ways to ship the same functionality, such as the --emulate-shader-precision flag that already exists in Chrome. I hope I'll find time to look into the emscripten-compiled version.

kenrussell:
OK. Thanks Olli for the feedback and Florian for the cleanup to the extension registry. Merging.

PR #829, opened 2015-01-24
reject WEBGL_texture_from_depth_video

pyalot:
Proposing to reject WEBGL_texture_from_depth video because it's not required to do it as an extension.


kenrussell:
Leaving this open for now until the depth video working group comes back with a revised proposal.

robman:
Ken, you can see what will be published as our WD soon here.
http://w3c.github.io/mediacapture-depth/#webglrenderingcontext-interface

We've taken your WG's advice and are just presenting this as guidelines for phase1.

We've also mapped out a phase2 that builds on top of some of the feedback from the 2D Canvas WG and we'll be posting info on that soon.

Thanks Ken and Florian for your help and feedback.

kenrussell:
OK. Thanks Rob for the feedback and Florian for this cleanup to WebGL's extension registry. Merging.

PR #830, opened 2015-01-26
Added OES_fbo_render_mipmap proposal

pyalot:


kenrussell:
Thanks, this looks like it fills a gap in functionality very difficult to work around. Merging.

PR #831, opened 2015-01-26
added OES_depth24

pyalot:


kenrussell:
Merging mainly to allow this to be moved to the rejected state. We don't want to distract ourselves from implementing WebGL 2.0 at this point.

PR #833, opened 2015-01-27
Published WebGL 1.0.3, ratified by Khronos Promoters during meeting of J...

kenrussell:
...anuary 23, 2015.

Updated latest spec links, last modified dates, and copyright dates on current versions of specs.


kenrussell:
Merging myself per working group resolution.

PR #834, opened 2015-01-27
Per discussion in WebGL WG, loosened index validation requirements for d...

kenrussell:
...rawArrays and drawElements calls to allow the behavior defined in the KHR_robust_buffer_access_behavior OpenGL ES 2.0 extension.

Explicitly referenced that extension (non-normatively) in the new specification text as well as the preexisting section on out-of-range array accesses in shaders.


kenrussell:
Merging per resolution in WebGL WG.

PR #836, opened 2015-01-27
Mention the lack of PRIMITIVE_RESTART_FIXED_INDEX in incompatibility section

Oletus:
It's important to have all the incompatibilities listed in one place to have an
easy reference for developers.


kenrussell:
Thanks for fixing this oversight.

PR #841, opened 2015-01-28
Added Shader Library tests.

jasomobica:
Ported gles2 and gles3 Shader Library tests from Android deqp.

### DIFF
--- a/sdk/tests/00_test_list.txt
+++ b/sdk/tests/00_test_list.txt
@@ -3,5 +3,7 @@
 
 conformance/00_test_list.txt
 conformance/more/00_test_list.txt
+deqp/data/gles2/shaders/00_test_list.txt
+deqp/data/gles3/shaders/00_test_list.txt
 --min-version 2.0.0 conformance2/00_test_list.txt
 
--- a/sdk/tests/deqp/resources/deqp-draw.js
+++ b/sdk/tests/deqp/resources/deqp-draw.js
@@ -1,0 +1,253 @@
+/*-------------------------------------------------------------------------
+ * drawElements Quality Program OpenGL ES Utilities
+ * ------------------------------------------------
+ *
+ * Copyright 2014 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+
+var deqpDraw = (function() {
+'use strict';
+
+/**
+ * Description of a vertex array binding
+ * @param {WebGLRenderingContext.GLEnum} type GL Type of data
+ * @param {string|number} location Binding location
+ * @param {number} components Number of components per vertex
+ * @param {number} elements Number of elements in the array
+ * @param {Array.<number>} data Source data
+ */
+var VertexArrayBinding = function(type, location, components, elements, data) {
+    this.type = type;
+    this.location = location;
+    this.components = components;
+    this.elements = elements;
+    this.data = data;
+};
+
+/**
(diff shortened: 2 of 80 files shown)


sdk/tests/deqp/resources/deqp-draw.js
```diff
@@ -0,0 +1,253 @@
...
+    TRIANGLE_STRIP: 1,
+    TRIANGLE_FAN: 2,
+
+    LINES: 3,
+    LINE_STRIP: 4,
+    LINE_LOOP: 5,
+
+    POINTS: 6,
+
+    PATCHES: 7
```
kenrussell: The PATCHES primitive type doesn't exist in OpenGL ES 2.0 or WebGL. Should be removed so the code below throws an exception if somebody tries to use it.

sdk/tests/deqp/resources/shader-library-case.js
```diff
@@ -0,0 +1,1144 @@
...
+};
+
+/**
+ * Initialize a test case
+ */
+var init = function() {
+    var state = deqpTests.runner.getState();
+    var test = state.currentTest;
+
+    _bufferedLogToConsole('Processing ' + test.fullName());
```
kenrussell: This is "exported" as bufferedLogToConsole -- please use the public name.

sdk/tests/deqp/resources/shader-library-case.js
```diff
@@ -0,0 +1,1144 @@
...
+    {
+        case expectResult.EXPECT_PASS:
+        case expectResult.EXPECT_VALIDATION_FAIL:
+            if (!allCompilesOk)
+                failReason = 'expected shaders to compile and link properly, but failed to compile.';
+            else if (!allLinksOk)
+                failReason = 'expected shaders to compile and link properly, but failed to link.';
+            break;
+
+        case expectResult.EXPECT_COMPILE_FAIL:
```
kenrussell: It's not possible in OpenGL ES to expect compiles to fail; it's always legal for the shader compiler to defer compiles until link time. How do the tests use these expectations? I'd assume the negative tests should always use EXPECT_COMPILE_LINK_FAIL.

sdk/tests/resources/js-test-pre.js
```diff
@@ -179,6 +190,57 @@ function testFailed(msg)
     _flushBufferedLogsToConsole();
 }
 
+/**
+ * Sets the current test name for usage within testPassedOptions/testFailedOptions.
+ * @param {string} name The name to set as the current test name.
+ */
+function setCurrentTestName(name)
+{
+	this._currentTestName = name;
```
kenrussell: You can't rely on having a valid "this" pointer here. Just define _currentTestName in the global scope in this file and assign to it, and remove the definition in the anonymous function above.

sdk/tests/resources/js-test-pre.js
```diff
@@ -179,6 +190,57 @@ function testFailed(msg)
...
+}
+
+/**
+ * Gets the current test name in use within testPassedOptions/testFailedOptions.
+ * @return {string} The name of the current test.
+ */
+function getCurrentTestName()
+{
+	return this._currentTestName;
```
kenrussell: Remove "this." per above comment.

sdk/tests/00_test_list.txt
```diff
@@ -3,5 +3,7 @@
 
 conformance/00_test_list.txt
 conformance/more/00_test_list.txt
+deqp/data/gles2/shaders/00_test_list.txt
+deqp/data/gles3/shaders/00_test_list.txt
```
kenrussell: Add "--min-version 2.0.0" at the beginning of this line, which will restrict the GLES3 shaders to only run as part of the WebGL 2.0 conformance suite.

sdk/tests/deqp/resources/deqp-draw.js
kenrussell: Actually, you can ignore this comment. It's fine to leave this in place for the time being.

kenrussell:
Thank you; this is excellent work. Let me know whether you're willing to update this pull request or whether you'd rather merge it as is and submit new pull requests for the minor comments above. I'm happy to merge as is because the review is somewhat difficult with so many new files.

jasomobica:
Let's merge as is. We can fix the issues later, while working on next tests.

Janusz
28 sty 2015 19:57 "Ken Russell" <notifications@github.com> napisał(a):

kenrussell:
Sounds good. Merging.

PR #843, opened 2015-01-29
Fix mistakes in WebGL 2.0 getParameter enum list

Oletus:
DRAW_BINDING was incorrect, it should have been DRAW_BUFFER<i>.
PACK_IMAGE_HEIGHT and PACK_SKIP_IMAGES were removed in GLES3.0.1 (see spec
section E.4), so they should not be in WebGL 2.0 either.

IMPLEMENTATION_COLOR_READ_FORMAT and IMPLEMENTATION_COLOR_READ_TYPE are not
new in WebGL 2.0 any more, since they were added to WebGL 1.0, so remove them
from the table that should only list the new enums.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -28,7 +28,7 @@
     <!--end-logo-->
 
     <h1>WebGL Specification</h1>
-    <h2 class="no-toc">Editor's Draft 26 January 2015</h2>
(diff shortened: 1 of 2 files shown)


kenrussell:
Thanks. Reviewed against the OpenGL ES 3.0.3 spec. Looks good.

PR #844, opened 2015-01-30
Using RequireJS to handle dependencies.

jasomobica:
Restored original deqp directory structure and added RequireJS to handle
dependencies between modules.


kenrussell:
Thanks. This will make it easier to understand the state of the port.

PR #845, opened 2015-01-30
Fixed conformance tests

DavidYen:
Add conformance tests to check that cached shaders take into account extensions.

Due to the complexity of caching shaders with global extensions that affect shader compilations,
it is necessary to test for these special conditions by running these shader tests twice. The tests
for the following extensions have been modified to catch cache problems:

GL_OES_standard_derivatives
GL_EXT_frag_depth
GL_EXT_draw_buffers
GL_EXT_shader_texture_lod

Additionally, an issue with the "missingPragmaFragmentShader shader" inside of
ext-shader-texture-lod.html has been fixed. This shader was not properly catching pragma
errors because the shader itself was invalid with other errors such as undeclared variables.
This has been fixed now so the only error is that the pragma extension is missing.

### DIFF
--- a/sdk/tests/conformance/extensions/ext-frag-depth.html
+++ b/sdk/tests/conformance/extensions/ext-frag-depth.html
@@ -112,8 +112,7 @@ runAllTests();
 // Run all tests against with a new context to test for any cache issues.
 debug("");
 debug("Testing new context to catch cache errors");
-var canvas2 = document.getElementById("canvas2");
-gl = wtu.create3DContext(canvas2);
+gl = wtu.create3DContext();
 ext = null;
 runAllTests();
 
--- a/sdk/tests/conformance/extensions/ext-frag-depth.html
+++ b/sdk/tests/conformance/extensions/ext-frag-depth.html
@@ -106,27 +106,40 @@ var canvas = document.getElementById("canvas");
 var gl = wtu.create3DContext(canvas);
 var ext = null;
 
-if (!gl) {
-    testFailed("WebGL context does not exist");
-} else {
-    testPassed("WebGL context exists");
+// Run all tests once.
+runAllTests();
 
-    runShaderTests(false);
+// Run all tests against with a new context to test for any cache issues.
+debug("");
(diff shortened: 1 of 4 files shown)


sdk/tests/conformance/extensions/ext-frag-depth.html
```diff
@@ -106,27 +106,40 @@
...
-    testPassed("WebGL context exists");
+// Run all tests once.
+runAllTests();
 
-    runShaderTests(false);
+// Run all tests against with a new context to test for any cache issues.
+debug("");
+debug("Testing new context to catch cache errors");
+var canvas2 = document.getElementById("canvas2");
+gl = wtu.create3DContext(canvas2);
```
kenrussell: We can eliminate both the "canvas" and "canvas2" elements and just call "wtu.create3DContext()" for both. Could you make that cleanup in these tests?

sdk/tests/conformance/extensions/ext-shader-texture-lod.html
```diff
@@ -44,9 +45,11 @@
 <script id="missingPragmaFragmentShader" type="x-shader/x-fragment">
 precision mediump float;
 varying vec2 texCoord0v;
+uniform float lod;
+uniform sampler2D tex;
 void main() {
     vec4 color = texture2DLodEXT(tex, texCoord0v, lod);
-    gl_FragColor = vec4(dx, dy, w, 1.0);
```
kenrussell: Nice catch.

kenrussell:
Thanks for this improvement. These changes look good overall. Would you mind making the cleanup mentioned above and updating this pull request? Please comment on the pull request if you update it. Brandon or Mo can help you if you need help, or an additional review or merge.

DavidYen:
I've kept the original canvas so users can see things happening on the actual webpage when these performance tests run. I've removed the canvas2 variable for all the other tests with the exception of the texture-lod test, that test requires an actual canvas of size 256x256.

kenrussell:
Thanks for the update. Looks great. Merging.

PR #846, opened 2015-01-31
rejected EXT_color_buffer_half_float and WEBGL_color_buffer_float

pyalot:
Due to implementation difficulties I propose to reject these extensions.

Google has stated that these extensions will not be implemented by chrome. Mozilla has withdrawn their initial implementation.


kenrussell:
Please get confirmation from Mozilla about their intent.

jdashg:
Firefox actively supports these, and Chrome really should to. (but presently doesn't)

PR #847, opened 2015-01-31
move to draft OES_fbo_render_mipmap

pyalot:
Proposing to move OES_fbo_render_mipmap to draft.


kenrussell:
Thanks for this suggestion. Merging per feedback on public_webgl.

PR #849, opened 2015-02-03
uniformMatrixBadArgs test: allow 1 shade of tolerance in framebuffer value

putsman:
The uniformMatrixBadArgs test culminates with a call to glReadPixels and expects to get an exact result of [1,2,3,8].

The fragment shader output color is calculated like this:
gl_FragColor = vec4(fm4[0][0]/256.0, fm4[1][1]/256.0, fm4[2][2]/256.0, fm4[3][3]*texCoord0.z/256.0);

While the vertex position is calculated like this:
gl_Position = vec4(Vertex, 1.0 + d * 0.0001);

The framebuffer value therefore depends on an iterated texture coordinate (texCoord0.z) which has been subject to perspective projection by a non 1.0f vertex W-value. These operations may introduce small errors on some implementations, along with the usual float-to-unorm rounding issues.

As such, we propose that the test allows 1 shade of tolerance in the framebuffer value, as is typical in GLES Conformance tests. We have added a utility function assertArrayEqualsWithEpsilon() for this purpose.

Thanks, 
Jonathan Putsman
(Imagination Technologies)

### DIFF
--- a/conformance-suites/1.0.3/conformance/more/functions/uniformMatrixBadArgs.html
+++ b/conformance-suites/1.0.3/conformance/more/functions/uniformMatrixBadArgs.html
@@ -120,7 +120,7 @@ Tests.testUniformf = function(gl, unwrappedGL) {
   });
   var d = new Uint8Array(4);
   gl.readPixels(0,0,1,1,gl.RGBA, gl.UNSIGNED_BYTE, d);
-  assertArrayEquals([1,2,3,8], d);
+  assertArrayEqualsWithEpsilon([1,2,3,8], d, [1,1,1,1]);
   sh.destroy();
 }
 
--- a/conformance-suites/1.0.3/conformance/more/unit.js
+++ b/conformance-suites/1.0.3/conformance/more/unit.js
@@ -391,6 +391,38 @@ function assertArrayEquals(name, v, p) {
   return true;
 }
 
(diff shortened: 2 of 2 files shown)


toji:
Looks good, thanks!

kenrussell:
Thanks, these test fixes look fine, but they must be merged to the top-of-tree conformance tests under sdk/tests/conformance/. Brandon, could you take care of that?

putsman:
Thanks Brandon and Ken.

Do you update legacy conformance suites? We noticed this issue when running Android CTS which has imported WebGL Conformance 1.0.1 (or thereabouts) as far as I understand.

kenrussell:
We can certainly apply this fix to the 1.0.1 and 1.0.2 conformance suites; it seems non-controversial. Would you open a separate pull request for that? I'm trying to find out the process for updating the WebGL test snapshot that's in the Android CTS, but updating this repository would be a good first step.

kenrussell:
After talking with the people responsible for the Android CTS, please just open a new pull request applying these diffs to the 1.0.1 and 1.0.2 WebGL conformance suites. A suppression's already been added for this failing test in the Android CTS. If we apply this diff to the 1.0.1 and 1.0.2 suites, it'll get picked up the next time they pull an update from this repository.

PR #851, opened 2015-02-05
rejecting OES_depth24

pyalot:


toji:
Merging as per mailing list conversations.

PR #854, opened 2015-02-07
Removed, and documented removal of, texture swizzles per decision at rec...

kenrussell:
...ent F2F.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -27,7 +27,7 @@
     <!--end-logo-->
 
     <h1>WebGL 2 Specification</h1>
-    <h2 class="no-toc">Editor's Draft 26 January 2015</h2>
+    <h2 class="no-toc">Editor's Draft 06 February 2015</h2>
     <dl>
         <dt>This version:
             <dd>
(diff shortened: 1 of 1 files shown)


specs/latest/2.0/index.html
```diff
@@ -650,10 +650,6 @@ <h2 class="no-toc">Table of contents</h2>
   const GLenum ANY_SAMPLES_PASSED_CONSERVATIVE               = 0x8D6A;
   const GLenum SAMPLER_BINDING                               = 0x8919;
   const GLenum RGB10_A2UI                                    = 0x906F;
-  const GLenum TEXTURE_SWIZZLE_R                             = 0x8E42;
-  const GLenum TEXTURE_SWIZZLE_G                             = 0x8E43;
-  const GLenum TEXTURE_SWIZZLE_B                             = 0x8E44;
-  const GLenum TEXTURE_SWIZZLE_A                             = 0x8E45;
   const GLenum GREEN                                         = 0x1904;
   const GLenum BLUE                                          = 0x1905;
```
Oletus: GREEN and BLUE enums are only used for texture swizzle and so they could also be removed from here.

specs/latest/2.0/index.html
```diff
@@ -2198,6 +2190,10 @@ <h2 class="no-toc">Table of contents</h2>
 
     <h3>New Features Supported in the WebGL 2 API</h3>
 
+    <div class="note editor">
+        Needs many updates for OpenGL ES 3.0 features
+    </div>
```
Oletus: I don't think that this note is needed - this is under "4 Other differences between WebGL 2.0 and WebGL 1.0", so this section does not need to cover anything that's covered by the above API documentation, and section 4 already has a similar comment at the top.

Oletus:
A couple of small comments, but other than that looks good!

jdashg:
LGTM

specs/latest/2.0/index.html
kenrussell: Thanks very much; will fix.

specs/latest/2.0/index.html
kenrussell: Thanks, I'd missed the nesting of this section. I'll remove this note.

kenrussell:
Thanks for the reviews Olli and Jeff. Addressed review feedback. Merging now.

PR #855, opened 2015-02-07
moved WEBGL_color_buffer_float and EXT_color_buffer_half_float to draft

pyalot:
Rejection of these extensions has proven impossible on objections to rejection. Implementation has proven impossible because of vendor objections. Ratification will prove impossible because of a lack of implementations. These extensions should move back to draft.


kenrussell:
I don't support moving these extensions back to draft status and implicitly forbidding Firefox from exposing them. Firefox followed all of the rules when requesting these be moved to community approved status. My concerns have been detailed on the public mailing list. I'm not going to close this pull request, but request that the other committers to the WebGL repository not merge it until the discussion has concluded.

jdashg:
We discussed this at one of the F2Fs, I think.
We have a pretty good handle on how this stuff works now, and how and why the different extensions should-or-should-not be exposed.

PR #856, opened 2015-02-09
Pass -no-remote option to Firefox in test runner

Oletus:
This makes sure that a new instance of the browser is opened with newer
versions of Firefox.


kenrussell:
Thanks for this fix.

PR #858, opened 2015-02-10
Add conformance test for deferred shader compilations with extensions.

DavidYen:
Add conformance test for deferred shader compilations with extensions.

If shader compilation is deferred until the link time, a test has been added to be sure the shader compiler uses its state as seen during the compilation request, not the link request. The added conformance test does the following to test this:

1. Create a new GL context which does not have the tested extension enabled.
2. Compile the vertex and fragment shaders which would fail if the extension were disabled.
3. Enable the extension by querying if the extension is supported.
4. Link the shader and check the link status, this should fail since at the time of the compilation the extension was disabled.

### DIFF
--- a/sdk/tests/conformance/extensions/ext-frag-depth.html
+++ b/sdk/tests/conformance/extensions/ext-frag-depth.html
@@ -116,6 +116,9 @@ gl = wtu.create3DContext();
 ext = null;
 runAllTests();
 
+// Run deferred link tests.
+runDeferredLinkTests();
+
 function runAllTests() {
     if (!gl) {
         testFailed("WebGL context does not exist");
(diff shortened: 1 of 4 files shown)


zhenyao:
LGTM

sdk/tests/conformance/extensions/ext-frag-depth.html
```diff
@@ -116,6 +116,9 @@
 ext = null;
 runAllTests();
 
+// Run deferred link tests.
+runDeferredLinkTests();
```
zhenyao: nit: this should be part of the runAllTests(), here and below.

sdk/tests/conformance/extensions/ext-frag-depth.html
```diff
@@ -244,6 +247,66 @@
...
+    gl_deferred.getExtension("EXT_frag_depth");
+
+    var program = gl_deferred.createProgram();
+    gl_deferred.attachShader(program, vertex_shader);
+    gl_deferred.attachShader(program, fragment_shader);
+    gl_deferred.linkProgram(program);
+
+    var linked = gl_deferred.getProgramParameter(program, gl_deferred.LINK_STATUS);
+    if (linked) {
+        testFailed("Compilation with extension disabled then linking with extension enabled caused successful compile.");
```
toji: This is a slightly awkward failure message. Might be more natural to say "Compilation with extension disabled then linking with extension enabled should have failed."?

kenrussell:
Thanks for this contribution. Once Brandon's comments are addressed this looks good to me.

One suggestion: it might be worth adding another optional argument to loadShader in sdk/tests/conformance/resources/webgl-test-utils.js like "opt_ignoreCompileStatus" which, if set to true, just skips the query of the shader's COMPILE_STATUS. A lot of boilerplate code could then be removed from these new tests.

kenrussell:
Sorry, I should have said "Brandon's and Mo's comments".

There's one more issue which I'll detail in the diffs.

sdk/tests/conformance/extensions/ext-frag-depth.html
kenrussell: This is an excellent point -- the new runDeferredLinkTests function assumes the extension is supported. It's perfectly legal for it to not be. Adding this to the list of tests in runAllTests() in this and the other extension tests will implicitly fix this problem.

DavidYen:
I have refactored the load shader functions in the utility file to have an option to skip the compilation status check. With this change I also added a check to be sure the shader compiles as expected in a separate context before we do the deferred test.

The reason I had originally placed the new deferred link test outside of RunAllTests() is because it uses it's own context and not the GL context that the RunAllTests() function utilizes. Since my test is checking if the shader links, the result is the same whether the extension is supported or not. I think it is a bit confusing to have 1 test outside of "RunAllTests()" though so I just placed it in there to make it match all the other tests. The only downside is that it will run twice unnecessarily, but that is not a big deal for this test.

zhenyao:
LGTM

sdk/tests/conformance/extensions/ext-frag-depth.html
```diff
@@ -140,6 +140,9 @@
             runUniqueObjectTest();
         }
     }
+
+    // Run deferred link tests.
+    runDeferredLinkTests();
```
kenrussell: This should have been added in the block above where EXT_frag_depth was successfully enabled. This test can't run properly if the extension isn't supported. See below.

sdk/tests/conformance/extensions/ext-frag-depth.html
```diff
@@ -244,6 +247,62 @@
...
+    debug("");
+    debug("Testing deferred shader compilation tests.");
+
+    // Test for compilation failures that are caused by missing extensions
+    // do not succeed if extensions are enabled during linking. This would
+    // only happen for deferred shader compilations.
+
+    // First test if link succeeds with extension enabled.
+    var glEnabled = wtu.create3DContext();
+    var extEnabled = glEnabled.getExtension("EXT_frag_depth");
```
kenrussell: EXT_frag_depth might not be supported by the WebGL implementation. In this case, the call to getExtension will fail.

sdk/tests/conformance/extensions/ext-frag-depth.html
```diff
@@ -244,6 +247,62 @@
...
+
+    var vertexShader = wtu.loadShaderFromScript(glEnabled, "goodVertexShader");
+    var fragmentShader = wtu.loadShaderFromScript(glEnabled, "macroFragmentShader");
```
kenrussell: If the above call to getExtension failed, the macroFragmentShader won't compile successfully; GL_EXT_frag_depth will not be defined, and the #error directive will be hit.

sdk/tests/conformance/extensions/ext-frag-depth.html
```diff
@@ -244,6 +247,62 @@
...
+
+    if (!vertexShader || !fragmentShader) {
```
kenrussell: fragmentShader will be null here if compilation happens eagerly in the WebGL implementation and EXT_frag_depth wasn't supported.

PR #860, opened 2015-02-11
Expect extension to be supported when running deferred link tests.

kenrussell:


kenrussell:
Please review and merge if this looks good. Thanks.

DavidYen:
LGTM, thanks for fixing this

PR #863, opened 2015-02-11
Ported ANGLE drawInstancedArrays offset test to WebGL conformance suite

### LINKED ISSUE #862: Adapt ANGLE DrawArraysWithOffset test into WebGL conformance suite
http://crbug.com/457269 and http://crbug.com/447140 describe a bug in ANGLE's instancing implementation for which a test was written in https://chromium.googlesource.com/angle/angle/+/master/tests/angle_tests/InstancingTest.cpp (DrawArraysWithOffset). This test should be ported into sdk/tests/conformance/extensions/angle-instanced-arrays.html .

toji:
Fixes #862

### DIFF
--- a/sdk/tests/conformance/extensions/angle-instanced-arrays.html
+++ b/sdk/tests/conformance/extensions/angle-instanced-arrays.html
@@ -59,6 +59,21 @@ void main() {
 }
 </script>
 
+<script id="drawArraysTestVertexShader" type="x-shader/x-vertex">
+attribute vec3 aPosition;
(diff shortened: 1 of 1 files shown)


kenrussell:
Thanks for porting this. I note the data is organized slightly differently than the original test and this makes it a little hard to judge it side-by-side, but assuming you've tested this, looks good.

toji:
The difference in organization is simply to better match the organization of the existing conformance test. The test vertex data has also been altered slightly for easier readability, but tests the same behavior.

PR #864, opened 2015-02-12
Test runner: Support Chrome on Android

Oletus:
This patch adds support for running tests remotely on a connected Android
device. It adds support for Chrome on Android and the framework can be
reused for other browsers on Android.

For this, the test runner is refactored so that browsers running locally
and browsers running on an Android device are abstracted behind a
BrowserInstance interface.

To run the tests on Android, you need the following:
    - Android device connected to linux host machine
    - Chrome application installed on the Android device
    - Android SDK installed on your host linux machine
    - 'adb' in your system path
    - The 'ant' build system installed in case you want to modify the
      prebuilt port forwarder apk.

Android support uses adb port forwarding from the connected device
to the host linux machine. When running the tests on Android, the test
runner starts a port forwarder service on the device, and then launches
Chrome on Android using adb to start the tests.

As the port is being forwarded from device to host, all browser requests
to localhost:<port> on the Android device will be forwarded to
localhost:<port> on host linux machine where the local server is set up
by test runner to host the tests.

Along with the source, prebuilt port forwarder service apk is also added
to the tree.

The platform argument needs to be explicitly passed as 'remote-android'
to run tests remotely on android.
Example:
  node ./test-runner --browser Chrome --platform remote-android

Once the tests finish running, results are posted back to the host linux
machine, port forwarder service and browser are terminated, and other
necessary cleanup is done.

This commit also fixes a bug where a wrong directory name was used for
dumped shaders when the browser was not specified on the command line.

This patch was developed in collaboration with Arun Patole
(apatole@nvidia.com).


other/test-runner/index.js
```diff
@@ -83,24 +87,24 @@ function main() {
 }
 
 function process_args(config) {
-  if(config.args.browser) {
+  if (config.args.browser) {
```
kenrussell: For the future, it would have been ideal to make these whitespace changes in a separate pull request.

kenrussell:
Thanks for this substantial contribution. This is a major functionality improvement to the test runner. The new code looks good overall. It might have been easier to review if some of the changes, like the bug fix to the directory containing the generated shaders, had been factored into separate pull requests, but this isn't a big deal. Merging.

Oletus:
Thanks for the speedy review! Were you able to test the new functionality? I just noticed that it doesn't seem to be working with my non-rooted Nexus 5. I'm hoping that anything specific to our Android development setup didn't slip in, and that this is only a restriction concerning non-rooted devices in general.

kenrussell:
Sorry, no, haven't tried it yet. Will try tomorrow.

apatole:
Olli,
adb has a restricted access permissions on non-rooted devices and as all of our test runner setup and tear down happens over adb, it may not run properly. Also one more problem I found is that explicit user argument with user id 0 will have to be passed to start service on non-rooted device.

I think we should explicitly disable test runner support for non-rooted android devices now and add some extra check and try to make it working for non-rooted devices only if it is really required in future.

Adding below check in BrowserInstanceRemoteAndroid constructor disables it for non-rooted devices, if it looks ok, I can just make a commit for it:

  } else {
    var adb_root_output = shell.exec("adb root", {silent:true}).output;
    if (adb_root_output.match(/adbd cannot run as root/)) {
      console.error("\n ERROR: Test runner cannot be run on non-rooted android device.\n");
      process.exit(1);
    }

Oletus:
Arun,

that change to detect rooted Android looks good. It may be a bit fragile if adb is changed at some point, but it's better than nothing. We should also add a mention that the device needs to be rooted in docs.

Oletus:
List formatting on the other/test-runner/README.md when it's displayed on the web also ended up wonky, that should be fixed.

apatole:
Ok, will upload a change fixing this on Monday.

apatole:
Addressed above comments in: https://github.com/KhronosGroup/WebGL/pull/866

PR #865, opened 2015-02-12
Fix broken links in specs

Oletus:
The WebGL 1 spec had a link to a W3C page that has moved, and the WebGL 2
spec had some missing references and one broken internal link.


kenrussell:
Thanks, looks good.

PR #866, opened 2015-02-16
Fix bugs in test runner on android implementation

apatole:
1. Chrome command line --no-first-run doesn't seem to be working on
   android, replaced it with --disable-fre which works.

2. adb has a restricted access permissions on non-rooted devices and
   as all of our test runner setup and tear down happens over adb, it
   may not run properly. This change explicitly disables test runner
   support for non-rooted android device.

3. Fix list formatting in README files.


kenrussell:
Thanks for updating the documentation and the detection of whether the test-runner can be supported on the device.

This LGTM, but it would be much better if we could figure out how to make this work -- or at least mostly work -- on non-rooted devices. Is the operation which fails the writing of the command line options to /data/local/chrome-command-line? We might be able to figure out a different mechanism, but also we may not need it. I think that --disable-gesture-requirement-for-media-playback is no longer needed.

Merging now, but would appreciate your continued work on this.

apatole:
Ok, Thanks. I will try to get it working on non-rooted devices. Writting to chrome-command-line doesn't look to be a problem. One problem I saw was about starting service, other was about killing process using adb and thought there might be some other issues, will recheck it.

PR #867, opened 2015-02-17
Propose WEBGL_render_to_float.

jdashg:
This might be a more portable way to allow render-to-float behavior in WebGL. It's more restricted than WEBGL_color_buffer_float or EXT_color_buffer_half_float. It's implementable everywhere that supports render-to-float, including D3D9.

BLEND with floating-point is disallowed because of lack of support. EXT_color_buffer_float does not allow blending with float32s, and D3D9 only supports normalized uint8 glBlendColor-equivalent. 

clear() is disallowed because many implementations clamp clearColor, (driver bugs) and/or limit clearColor to normalized uint8 values. (D3D9. Notably, 0.5f is not representable in normalized uint8) WebGL implementations can work around this by doing a full-screen blit, but this is much more expensive, and current best-practice is to no longer hide performance caveats.

For textures, the initial contents can be set at upload-time. If there is strong support for floating-point clears on drivers that can't support the full WEBGL_color_buffer_float/EXT_color_buffer_half_float extensions, we can add an extension just for floating-point clears. (similar in nature to OES_texture_float_linear)

We could split this into WEBGL_render_to_float and WEBGL_render_to_half_float, but we should only do so if there are platforms or configurations which require this differentiation. (That is, FLOAT and HALF_FLOAT_OES textures are supported, but only one of the two is renderable)

### DIFF
--- a/extensions/proposals/WEBGL_render_to_float/extension.xml
+++ b/extensions/proposals/WEBGL_render_to_float/extension.xml
@@ -1,0 +1,115 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<proposal href="WEBGL_render_to_float/">
+  <name>WEBGL_render_to_float</name>
+
+  <contact>
+    <a href="https://www.khronos.org/webgl/public-mailing-list">WebGL working
+    group</a> (public_webgl 'at' khronos.org)
+  </contact>
+
+  <contributors>
+    <contributor>Jeff Gilbert, Mozilla Corporation</contributor>
+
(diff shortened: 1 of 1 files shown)


kenrussell:
The working group discussed this at length last week and we're in agreement that we'll work on advancing the EXT_color_buffer_half_float, WEBGL_color_buffer_float and EXT_color_buffer_float extensions instead of proposing a least-common denominator extension which restricts all implementations to D3D9-class functionality.

PR #868, opened 2015-02-18
Removed imageSize from compressedTex{Sub}Image3D

toji:
Size should be implied by the ArrayBufferView size, which is consistent with compressedTex{Sub}Image2D.


kenrussell:
Looks good. Thanks for rebuilding the IDL; I forgot to do that during my last commit.

PR #869, opened 2015-02-23
Verify behavior of texImage2D and texSubImage2D with no WebGLTexture bou...

kenrussell:
...nd.

This test previously tested this behavior, albeit accidentally.


toji:
LGTM

PR #870, opened 2015-02-23
Backported fixes to tex-image-with-invalid-data to 1.0.1, 1.0.2, and 1.0...

kenrussell:
....3 conformance suites.

https://github.com/KhronosGroup/WebGL/pull/853
https://github.com/KhronosGroup/WebGL/pull/869

Thanks to Roger Fong for pointing out the need for these fixes.

Verified all three backports manually.


toji:
LGTM

PR #873, opened 2015-03-09
Update the list of removed enums related to MapBufferRange

Oletus:
Also update the test to verify buffer mapping related constants and
methods correctly.


zhenyao:
LGTM

PR #874, opened 2015-03-09
Bring WebGL 2 methods test up-to-date with the spec

Oletus:
Add texImage3D and drawRangeElements and remove samplerParameter[fi]v.
There's no samplerParameter call that takes multiple arguments, so the
vector version is unnecessary.


sdk/tests/conformance2/context/methods-2.html
```diff
@@ -188,6 +188,7 @@
 "invalidateSubFramebuffer",
 "readBuffer",
 "renderbufferStorageMultisample",
+"texImage3D",
 "texStorage2D",
 "texStorage3D",
 "texSubImage3D",
```
zhenyao: I noticed we have multiple texSubImage3D (five of them).  Is this necessary?  If not, maybe we should clean it up in this CL.

sdk/tests/conformance2/context/methods-2.html
```diff
@@ -247,11 +249,7 @@
 "isSampler",
 "bindSampler",
 "samplerParameteri",
-"samplerParameteriv",
-"samplerParameteriv",
 "samplerParameterf",
-"samplerParameterfv",
-"samplerParameterfv",
```
kenrussell: I think we should leave these entry points in place. Even though they're not necessary right now, if for example a vendor exposes a texture border color extension, we'd have to add these entry points back in that extension. What do you think?

sdk/tests/conformance2/context/methods-2.html
kenrussell: Good point -- there are multiple duplicates in the "WebGL2 methods" part of this list. Olli, do you want to run the list through "uniq" or should I do it in a subsequent pull request?

sdk/tests/conformance2/context/methods-2.html
Oletus: texParameter[if]v were not exposed in WebGL 1, and I can't recall any calls for adding them back. I think it makes sense to stay consistent with that. I can't imagine that texture border color extension is on any developer's wishlist. A convenience feature like that is worthless if it only works on a subset of platforms.

sdk/tests/conformance2/context/methods-2.html
Oletus: I'll add a patch to this pull request.

Oletus:
Added a patch that removes the duplicate methods from the list.

sdk/tests/conformance2/context/methods-2.html
kenrussell: Fair point. I'd not double-checked the spec and realized these had already been removed there some time ago.

kenrussell:
Thanks for making these updates and cleanups.

PR #875, opened 2015-03-09
Don't print error messages from removed properties twice

Oletus:
There's code to write error messages specifically for the removed constants, they don't need to be handled later when the code is looking for extra constants.


zhenyao:
LGTM

PR #876, opened 2015-03-09
Expando verification for programs, shaders, renderbuffers, textures and ...

RafaelCintron:
...framebuffers

Ensure that any expandos set on programs, shaders, renderbuffers, textures and framebuffers
remain set across garbage collections, even when the only references are internal to WebGL.

### DIFF
--- a/sdk/tests/conformance/misc/expando-loss.html
+++ b/sdk/tests/conformance/misc/expando-loss.html
@@ -1,0 +1,234 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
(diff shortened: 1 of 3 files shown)


sdk/tests/conformance/misc/expando-loss.html
```diff
@@ -0,0 +1,234 @@
...
+
+    var fs = wtu.setupSimpleColorFragmentShader(gl);
+    setTestExpandos(fs);
+
+    var program = wtu.setupProgram(gl, [vs, fs]);
+    setTestExpandos(program);
+    assertMsg(program === gl.getParameter(gl.CURRENT_PROGRAM), "getParameter(gl.CURRENT_PROGRAM) return instance set with useProgram");
+
+    var attachedShaders = gl.getAttachedShaders(program);
+    assertMsg(vs === attachedShaders[0], "getAttachedShaders[0] returns instance set with attachShader");
```
zhenyao: To my knowledge, this order (0 -> vertex shader, 1 ->fragment shader) is not guaranteed.

kenrussell: Agreed; this order doesn't seem to be specified in the ES 2.0 spec. Would be safer to write:

assertMsg(vs === attachedShaders[0] || vs === attachedShaders[1], ...);
assertMsg(fs === attachedShaders[0] || fs === attachedShaders[1], ...);

kenrussell:
Nice test overall. Could you address the one comment above? Will gladly merge afterward. Thanks.

RafaelCintron:
@kenrussell and @zhenyao , you're right about getAttachedShaders.  I've uploaded a new change that addresses this issue. Let me know if there's more you'd like me to improve.

kenrussell:
Thanks for the update. Looks great.

PR #877, opened 2015-03-10
Backport gl_Position assignment from test.

jdashg:
1.0.3 has this, but 1.0.2 does not. The GLSL spec requires it.

It looks like this was missed when this test was updated. (Note that 1.0.3 has -1E100 for negHuge, which is obviously a bug with 1.0.2. That said, it's not hurting anything, so there's no reason to change it, according to our rules for backporting fixes)

### DIFF
--- a/conformance-suites/1.0.2/conformance/glsl/literals/float_literal.vert.html
+++ b/conformance-suites/1.0.2/conformance/glsl/literals/float_literal.vert.html
@@ -60,6 +60,7 @@ void main() {
   highp float negInRange = -4611686018427387903.;
   highp float negOutRange = -4611686018427387905.;
   highp float negHuge = 1E100;
+  gl_Position = vec4(0.0, 0.0, 0.0, 1.0);
 }
 </script>
 <script>


kenrussell:
LGTM since this fix is in 1.0.3 and top-of-tree already.

Oletus:
I think this patch is incorrect. Writing gl_Position is not required by the spec. See spec section 10.13: "Is it an error if the vertex shader doesn't write to gl_Position?" "Resolution: No error".

When writing gl_Position was added to this test in 1.0.3, it was accompanied with adding a test for an empty main function and a test for a main function that doesn't write gl_Position to maintain coverage that was already in 1.0.2. So this patch doesn't fix a bug, but arbitrarily reduces test coverage in the 1.0.2 test suite.

I put up a pull request to revert this and restore 1.0.2 test coverage: https://github.com/KhronosGroup/WebGL/pull/880

kenrussell:
Olli, thanks for catching this issue.

I am in favor of backporting this change and the associated new tests to the 1.0.2 suite to maintain parity, especially if this test is problematic. Jeff, do all of the associated tests in the 1.0.3 suite pass in Firefox? If so please submit a pull request backporting all of the changes associated with this test. Thanks.

zhenyao:
I think in 1.0.2 we should either

1) remove this line from the current test and add the test to check missing gl_Position is ok, just like in 1.0.3

or

2) leave it as is

Just removing this test case, as Olli pointed out, reduces coverage in 1.0.2, especially some drivers do fail it (we added a workaround flag in ANGLE shader translator to fix it)

Oletus:
Maybe we could also just add a comment to the 1.0.2 test that not writing gl_Position is allowed by the spec to hopefully avoid further confusion. This would be a smaller change to 1.0.2 than adding an entirely new test.

kenrussell:
We need to understand the motivation for Jeff's original change. Jeff, is Firefox failing this test?

jdashg:
We are failing this test on some hardware, so I checked the GLSL spec. Here's my reading:

Well-formed shaders "should" write to gl_Position:
All executions of a well-formed vertex shader should write a value into
[gl_Position]. [...] Compilers may generate a diagnostic message if they detect
gl_Position is not written, or read before being written, but not all such
cases are detectable."

Only well-formed programs are guaranteed to compile:
Portability is only ensured for well-formed programs, which this specification
describes.  Compilers are encouraged to detect ill-formed programs and issue
diagnostic messages, but are not required to do so for all cases.

An example of "ill-formed" meaning "reject shader":
#error will cause the implementation to put a diagnostic message into the shader
object’s information log (see the API in the platform documentation for how to
access a shader object’s information log). The message will be the tokens
following the #error directive, up to the first new-line. The implementation
must then consider the shader to be ill-formed.

On "should":
Where the specification uses the terms required, must/must not, does/does not,
disallowed or not supported, the compiler or linker is required to detect and
report any violations. Similarly when a condition or situation is an error, it
must be reported. Where the specification uses the terms should/should not or
undefined behavior there is no such requirement but compilers are encouraged to
report possible violations.

jdashg:
Specifically, the question here is whether "Well-formed shaders should write to gl_Position" means that an implementation "should" consider not writing to gl_Position to be ill-formed.

zhenyao:
Jeff,  see Olli's first comment.  Since the spec is explicit about this behavior (not writing to gl_Position) generates no error, I think we should follow that.

You could just add  SH_INIT_GL_POSITION to compile options to pass the failing platforms.

jdashg:
So here's that bit:
Is it an error if the vertex shader doesn't write to gl_Position? Whether a
shader writes to gl_Position cannot always be determined e.g. if there is
dependence on an attribute.
Option 1:  No it is not an error. The behavior is undefined in this case.
           Development systems should issue a warning in this case but the
           on-target compiler should not have to detect this.
Option 2:  It is an error if the vertex shader does not statically write to
           gl_Position
Option 3:  It is an error if there is any static path through the shader where
           gl_Position is not written.
RESOLUTION:  No error (option 1).  The nature of the undefined behavior must be
             specified.

The "must be specified" by the resolution is the bits I quoted.

There are two things at play here:
1: Is the shader considered well-formed by the compiler.
2: Is gl_Position written to.

1 can be true while 2 is false. In this case, the spec clearly says that gl_Position has undefined value.

However, I'm reading that an implementation can optionally decide 1 is false if it guarantees 2 is false. In this case, since the shader is ill-formed, it should fail to compile.

I posit that in the resolution above should be "It is not (required to be) an error".
The nVidia GLSL compiler (at least tacitly) shares this interpretation.

jdashg:
While it's good to hear that ANGLE already has a workaround to allow skipping gl_Position, I think it's technically valid to reject such shaders.

zhenyao:
I think we should talk about this in working group and reach a resolution and update the spec, so we have consistent behavior all over.

To me, using the ANGLE compile option is a simpler solution - it gets rid of undefined values also by always initializing gl_Position to (0, 0, 0, 0).  If we decide to reject such shaders, we need to specify it, i.e., select between option 2 and 3.  Then we need to write detection code, which adds another pass to the shader translator.

So if we come to a vote, I would vote for no error but requiring to init to a default value (0,0,0, 0), just to be lazy. :)

jdashg:
I can agree with that. Since it's not possible to verify that all shaders always write to gl_Position, I think we have to init to (0,0,0,0) if we can't guarantee an actual write to gl_Position. 

While my initial instinct was to allow rejection for when it's not statically written, static use doesn't guarantee the write actually happens. I don't think it's acceptable to us to allow for undefined values, so we must ensure it is actually written to.

I'll submit a PR for adding a WebGL spec errata for this.

jdashg:
I've submitted https://github.com/KhronosGroup/WebGL/pull/881.

msc-:
I agree with Jeff's analysis of the spec language. Flagging an error when gl_Position is not written is optional only because failure to write can't always be detected. There is nothing there that says the error should not be flagged, if detected.

The only thing that initializing gl_Position to (0,0,0,0) seems to accomplish is to prevent a compile error from the underlying GL in a case where ANGLE, or other shader verifier, believes gl_Position is written and the underlying compiler does not. Is such a case ever going to happen? The drawing will still be wrong, just differently wrong, possibly, than what results from the "undefined" behavior.

jdashg:
Initializing gl_Position also means its value, if unwritten, is defined instead of undefined. Since we want to minimize undefined values (for consistency and security), it's best if we specify an initial value. (Even if we should be within our rights to reject such a shader as ill-formed. However, since we define an initial value, we might as well make it a proper default and not fail a shader even if we can prove that it does not write to gl_Position)

msc-:
Not flagging an error when possible is not very developer friendly. Instead of an error telling them they have failed to write gl_Position they will get incorrect drawing and be left scratching their heads. I think it better to detect and flag the error when possible which will be in the majority of cases.

Do we have any evidence of OpenGL {,ES} implementations crashing when gl_Position is not written? I would think it extremely unlikely. What incorrect value is used for the vertex position then doesn't really matter. The drawing will be wrong. Do we really need to have consistently wrong drawing? That and slowing down the shader would be the only results of initializing to (0,0,0,0).

jdashg:
The most concerning worry with respect to undefined values is that they may contain leaked data. (Very probably not useful here, but it's a leakage channel, so it's worth patching)

We should try to issue a warning when gl_Position is not statically written and record that in the info log or warn in the web console (console.log), but I don't think it's necessary to fail the compile to alert users.

PR #878, opened 2015-03-10
Small improvements in the deqp test framework

Oletus:
Include the name of the failed/passed test when reporting the results
back to the test runner, and add a bit of documentation for the filter
parameter.


kenrussell:
Thanks, looks good.

PR #879, opened 2015-03-10
Remove illegal constructs from deqp tests used for WebGL 1

Oletus:
Remove shader functionality that's not supported according to the
restrictions in Appendix A of the OpenGL ES Shading Language 1.00 spec.

### DIFF
--- a/sdk/tests/deqp/data/gles2/shaders/functions.test
+++ b/sdk/tests/deqp/data/gles2/shaders/functions.test
@@ -2368,15 +2368,14 @@ group control_flow "Control Flow In Functions"
 
 			float func (float a)
 			{
-				int i;
-				for (i = 0; i < 6; i++) // negate a
+				for (int i = 0; i < 6; i++) // negate a
 				{
 					a = -a;
 					if (i == 4)
 						a = -a;
 				}
 
-				for (; i < 10; i++) // keep a
+				for (int i = 6; i < 10; i++) // keep a
 				{
 					if (i == 8)
 						continue;
@@ -2410,8 +2409,7 @@ group control_flow "Control Flow In Functions"
 
 			float func (float a)
 			{
-				int i;
-				for (i = 0; i < 6; i++)
+				for (int i = 0; i < 6; i++)
 				{
 					if (i == 0)
 						continue;
--- a/sdk/tests/deqp/data/gles2/shaders/scoping.test
+++ b/sdk/tests/deqp/data/gles2/shaders/scoping.test
@@ -104,33 +104,6 @@ group valid "Valid scoping and name redeclaration cases"
 		""
 	end
 
-	case while_condition_variable_hides_local_variable
-		version 100 es
-		values
-		{
-			input int in0 = [ 1 | 2 | 3 ];
-			output int out0 = [ 1 | 2 | 3 ];
-		}
-
-		both ""
-			#version 100
-			precision mediump float;
-			${DECLARATIONS}
-			void main()
-			{
-				${SETUP}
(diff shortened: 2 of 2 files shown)


zhenyao:
LGTM

kenrussell:
After talking with the drawElements folks:

 - The for-loop test restructurings look fine.
 - It would be better to keep the while-loop tests. Since while-loop support is optional, perhaps the tests can be restructured so that if they compile, they must execute successfully; if they fail compilation, that's OK too.

I think the drawElements team may be adding support for the latter behavior. If we find many of the new tests are invalidated we may want to contribute to the core dEQP AOSP project and take the fixes into the WebGL suite.

gunjambi:
A colleague pointed out that the return_after_loop_sequence test was actually
intended to do the for loop in two parts while sharing the counter and hence
correcting the for loops to conform to the strict form actually changes the
the test. Since (unlike ES) WebGL explicitly disallows exceeding any Appendix
A restriction (4.3 Supported GLSL Constructs), the test no longer serves a
purpose and could be removed instead.

Similarly, thanks to stricter rules of WebGL, instead of removing the
while-loop tests, it might be beneficial to make them negative tests instead
(with expect compile_fail).

Since the GL ES 2.0 is more relaxed on exceeding the limits, dEQP AOSP
cannot be updated to match these changes. Instead, we will annotate test cases
exceeding limits and tolerate their failures.

TL;DR:
For ES2.0 We will do:
- test does not have *full_glsl_es_100_support* tag:
  - CompileOk => test pass
  - CompileFail => test fail
- test has *full_glsl_es_100_support* tag:
  - CompileOk => test pass
  - CompileFail => test pass

For WebGL it might be beneficial to do instead:
- test does not have *full_glsl_es_100_support* tag:
  -  CompileOk => test pass
  -  CompileFail => test fail
- test has *full_glsl_es_100_support* tag:
  - CompileOk => test **FAIL**, Test exceeded a limit, which is not allowed
  - CompileFail => test pass

In the next(ish?) dEQP ASOP push, dEQP will get support for
*full_glsl_es_100_support* tag and test files be update followingly:

```
diff --git a/data/gles2/shaders/functions.test b/data/gles2/shaders/functions.test
--- a/data/gles2/shaders/functions.test
+++ b/data/gles2/shaders/functions.test
@@ -2355,6 +2355,8 @@ group control_flow "Control Flow In Functions"
 	end
 
 	case return_after_loop_sequence
+		require full_glsl_es_100_support
+
 		values
 		{
 			input float in0		= [ -0.5 | 1.5 ];
@@ -2409,8 +2411,7 @@ group control_flow "Control Flow In Functions"
 
 			float func (float a)
 			{
-				int i;
-				for (i = 0; i < 6; i++)
+				for (int i = 0; i < 6; i++)
 				{
 					if (i == 0)
 						continue;
diff --git a/data/gles2/shaders/scoping.test b/data/gles2/shaders/scoping.test
--- a/data/gles2/shaders/scoping.test
+++ b/data/gles2/shaders/scoping.test
@@ -106,6 +106,7 @@ group valid "Valid scoping and name redeclaration cases"
 
 	case while_condition_variable_hides_local_variable
 		version 100 es
+		require full_glsl_es_100_support
 		values
 		{
 			input int in0 = [ 1 | 2 | 3 ];
@@ -160,6 +161,7 @@ group valid "Valid scoping and name redeclaration cases"
 
 	case while_condition_variable_hides_global_variable
 		version 100 es
+		require full_glsl_es_100_support
 		values
 		{
 			input int in0 = [ 1 | 2 | 3 ];
```

PR #880, opened 2015-03-10
Revert "Backport gl_Position assignment from test."

Oletus:
This reverts commit 78f9793a4f327d0e2427836ee20c386d8174cfd1.

Writing gl_Position is not required by the spec (see OpenGL ES Shading
Language 1.00 section 10.13) so the patch being reverted did not fix a
bug as it was intended, but reduced test coverage in the 1.0.2 suite.


zhenyao:
LGTM

PR #881, opened 2015-03-10
Add errata for initial gl_Position value.

jdashg:
It looks like a whitespace change snuck in, oops.

This is the outcome from https://github.com/KhronosGroup/WebGL/pull/877 .


zhenyao:
LGTM

Oletus:
Unitialized varyings also show up in WebGL shaders, and we've made a conscious decision not to address this in the past. In that sense giving special treatment to gl_Position and guaranteeing that it is initialized is a bit odd. But it's not a big thing - if in practice WebGL implementations need to initialize gl_Position to get around driver differences anyway, it doesn't hurt to have it in the spec.

PR #883, opened 2015-03-11
Improve shader logging in the WEBGL_draw_buffers test

Oletus:
Always log both vertex and fragment shaders and leverage the utils to log
translated shaders when possible. This can help with diagnosing bugs.

Also remove an unused variable from the test.


kenrussell:
Looks fine.

PR #884, opened 2015-03-11
Add a test for indexing gl_FragData with gl_MaxDrawBuffers

Oletus:
The test relies on OpenGL ES Shading Language spec section 4.1.9: It is
illegal to index an array with an integral constant expression greater
than or equal to its declared size. gl_MaxDrawBuffers is declared as a
const int in section 7.4, so it qualifies as an integral constant
expression.


Oletus:
This test reproduces a curious bug in Chrome I hit while testing some refactoring of ANGLE. Firefox doesn't seem to be affected.

zhenyao:
LGTM

What is the bug like?  I suspect it might be due to our shader/program caching.

Oletus:
The bug manifests when the same shader program containing gl_FragData[gl_MaxDrawBuffers] is compiled twice in the same browser instance. On first round it generates a compile error as it should, but on the second round ANGLE seems to change the index to gl_MaxDrawBuffers - 1 so that it fits inside the array. Quite peculiar.

PR #886, opened 2015-03-11
Use null instead of undefined for optional arg omission.

jdashg:
This appears to be a bug in other browsers, since postMessage's second arg is `optional Sequence<foo>`, which null can't be converted to.

We appear to still have a bug in Firefox, but with this fix we get further in the test.


kenrussell:
Thanks. This change looks good, and correct according to http://heycam.github.io/webidl/ .

The pull request's comment is backward -- it should say "Use undefined instead of null...". Do you want to open a new pull request fixing this and close this one so that the comments are correct in the revision history? If this is too much trouble I'll merge this as is.

jdashg:
Yeah, let's get the summary right.

PR #887, opened 2015-03-11
Use undefined instead of null for optional arg omission.

jdashg:
This was https://github.com/KhronosGroup/WebGL/pull/886, but with the summary line fixed.


kenrussell:
Thanks. LGTM.

PR #888, opened 2015-03-12
Built-ins can be active uniforms

jdashg:
At least two tests (biuDepthRange_001_to_001 and gl_FragCoord_001_to_003) check that there are no active uniforms they didn't request, but since they use gl_DepthRange.far and .near, these count as active uniforms. (GLES 2.0.25 p36: "The returned uniform name can be the name of a built-in uniform state as well")

NB: The test also uses gl_DepthRange.diff, but does not return this as an active uniform. The compiler likely emits `far - near` instead of using a separate value for `diff`, thus it does not consider `diff` to be an active uniform. Whether or not this is spec-compliant is beyond the scope of this bug.

### DIFF
--- a/conformance-suites/1.0.2/conformance/ogles/ogles-utils.js
+++ b/conformance-suites/1.0.2/conformance/ogles/ogles-utils.js
@@ -500,6 +500,8 @@ function drawWithProgram(program, programInfo, test) {
   // Check for unset uniforms
   for (var name in uniforms) {
     var uniform = uniforms[name];
+    if (name.indexOf("gl_") == 0)
+      continue;
     if (!uniform.used) {
       testFailed("uniform " + name + " never set");
     }
--- a/conformance-suites/1.0.2/conformance/ogles/ogles-utils.js
+++ b/conformance-suites/1.0.2/conformance/ogles/ogles-utils.js
@@ -500,8 +500,9 @@ function drawWithProgram(program, programInfo, test) {
   // Check for unset uniforms
   for (var name in uniforms) {
     var uniform = uniforms[name];
-    if (name.indexOf("gl_") == 0)
+    if (name.indexOf("gl_") == 0) {
       continue;
+    }
     if (!uniform.used) {
       testFailed("uniform " + name + " never set");
     }
--- a/conformance-suites/1.0.3/conformance/ogles/ogles-utils.js
+++ b/conformance-suites/1.0.3/conformance/ogles/ogles-utils.js
@@ -500,6 +500,8 @@ function drawWithProgram(program, programInfo, test) {
   // Check for unset uniforms
   for (var name in uniforms) {
     var uniform = uniforms[name];
+    if (name.indexOf("gl_") == 0)
+      continue;
     if (!uniform.used) {
       testFailed("uniform " + name + " never set");
     }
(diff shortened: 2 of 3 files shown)


kenrussell:
First, thanks for looking into this. There's been an open bug about this for over a year in https://www.khronos.org/bugzilla/show_bug.cgi?id=840 , and I've been too lazy to investigate it.

Looking through Chromium's revision history, it started hiding gl_DepthRange from the list of active uniforms in https://codereview.chromium.org/1700002 . In hindsight, this was probably a mistake. As you point out, it is legal to expose gl_DepthRange's struct members as active uniforms, though their locations can't be looked up. I also see the difference in behavior between Firefox and Chrome in this area.

I think rather than sweeping these differences under the rug we should consider being more strict about this -- changing the harness and the test to say that gl_DepthRange.{near,far,diff} must be returned as active uniforms in the top of tree version of these tests, and update Chromium's implementation to stop hiding them. (It looks like Safari must be hiding these too, so this planned change should be announced on the public list.) The changes to the 1.0.2 and 1.0.3 tests look OK to smooth over the differences between the browsers. What do you think?

We could also merge this as is and I'll take responsibility for making the tests more strict. If you'd rather do that that's fine.

jdashg:
The issue with gl_DepthRange specifically is that it exposes three values but only holds two values worth of state. (any one part can be recovered by using the other two) If we require that all three are reported as active uniforms, they must then all be included in packing restriction calculations, even though the driver is only packing two of the three.

Specific to my previous NB, only `near` and `far` are claimed as active by my 331.113 NVIDIA driver on my Linux machine, so this is not a hypothetical driver optimization.

kenrussell:
These are good points. What about adding a primitive to the harness that requires a certain number of a set of uniforms to be listed as active? Something like:

        "uniforms": {
          "result": {
            "count": 1, 
            "type": "uniform4fv", 
            "value": [
              0.25, 
              0.75, 
              0.5, 
              1.0
            ]
          }
        }, 
        "builtin_uniforms": {
          "min_required": 2,
          "legal_values": [
            "gl_DepthRange.near",
            "gl_DepthRange.far",
            "gl_DepthRange.diff"
          ]
        }

and then modifying ogles-utils to obey this data.

jdashg:
This seems like scope creep for this test. I think we should test how gl_DepthRange behaves separately and exclusively.

kenrussell:
I understand your concerns but disagree on this point -- you've uncovered a longstanding discrepancy and complaint from GPU vendors about the behavior of some active uniforms. This is something that should be resolved one way or the other.

I think we should either continue to assert Chrome's and Safari's behavior of hiding gl_DepthRange's components as active uniforms, or assert Firefox's behavior that they be present in some form if referenced. I'll be happy to update Chrome to stop hiding these active uniforms and, further, modify ANGLE to report them from the shader translator if it turns out some drivers don't report them. I'll also be happy to do a follow-up commit to yours if we decide to go in the latter direction enforcing the presence of gl_DepthRange's components. Which direction do you think is best?

jdashg:
Since these vars contribute to packing restrictions, I think we must expose their existence as active uniforms. They also should be exposed by GetUniforms in ANGLE's shader validator, though Firefox doesn't use GetUniforms for this purpose at the moment.

I just think we should make a separate test for checking that gl_DepthRange vars are reported properly, since they are non-trivial, and handling them properly would needlessly complicate the test harness here. (IMO)

kenrussell:
We can certainly add a new test, but this one should also define the behavior precisely in the top of tree test.

I'll merge your change plus another one which switches the expected behavior to Firefox's for these two tests, and announce the change on the public_webgl list.

PR #889, opened 2015-03-12
Conformance add getters

Richard-Yunchao:
Add getters for WebGL 2 conformance test suite. Fixed bugs for getVertexAttrib and getFramebufferAttachmentParameter. And add GLenums supported by getRenderbufferParameter in WebGL 2.0


kenrussell:
Copying the entire gl-object-get-calls.html test from the WebGL 1.0 conformance suite to the 2.0 suite is not desirable. The 1.0 version of this test is going to be run as part of the 2.0 suite; there is no --max-version argument for this test in sdk/tests/conformance/state/00_test_list.txt.

Instead, the 2.0 version of this test should contain only those differences from the 1.0 test, of course with enough setup code to allow the test cases to run.

I've diffed your new conformance2/state/gl-object-get-calls.html with the 1.0 version and the differences are very small. Could you please edit down conformance2/state/gl-object-get-calls.html to test just those differences? Thanks.

Richard-Yunchao:
@kenrussell , I have updated the code. At first, I want to fix the bug for getter APIs for WebGL 2.0. Later, I will add new test cases.

kenrussell:
I'm sorry for asking you to redo this again, but I strongly think that there should be an .html file under sdk/tests/conformance/ for the WebGL 1.0 version of this test and one under sdk/tests/conformance2/ for the WebGL 2.0 version of the test, rather than putting the WebGL 2.0 functionality directly into the WebGL 1.0 test.

Could you cut the main script out of sdk/tests/conformance/state/gl-object-get-calls.html and put it into sdk/tests/conformance/resources/gl-object-get-calls.js, and include it from sdk/tests/conformance/state/gl-object-get-calls.html and a new sdk/tests/conformance2/state/gl-object-get-calls.html ? Leave all of the other changes as is (like the context detection, etc.); I agree they're the easiest way to add the new tests.

Also, please do an interactive rebase and squash all of the commits on your branch. We don't want to add 10+ commits to the revision history for this test. Thanks.

Richard-Yunchao:
@kenrussell, This change just fix bugs in gl-object-get-calls.html for WebGL 2.0 rendering context. As you know, when running WebGL 2.0 conformance test by https://www.khronos.org/registry/webgl/sdk/tests/webgl-conformance-tests.html?version=2.0.0 , it will run all conformance test cases under sdk/tests/conformance/ with WebGL 2 rendering context. And errors occur. This change didn't put the WebGL 2.0 functionality directly into the WebGL 1.0 test, it just fix these errors. For WebGL 2.0 functionality, I will put them into conformance2/state/.

Personally, I don't think the scripts should be put into ./conformance/resource/gl-object-get-calls.js, then include it from conformance/state/gl-object-get-call.js and conformance2/state/gl-object-get-calls.js. Because WebGL 2.0 will run all cases under conformance/state/ too(with webgl 2 rendering context), include it from conformance2/state/ will make it run the same code twice. I prefer to fix the bug to support original conformance/state/gl-object-get-call.js with minimal change for WebGL 2.0, then add WebGL 2.0 specific functionality into conformance2/state/gl-object-get-call-2.html. WDYT?

It's my fault not to squash commits into one. I am sorry... I am not familiar with github.com, although I have another account on github.com and submit patches for fun. But they are not real projects. I have squashed all these commits into one: https://github.com/Richard-Yunchao/WebGL/tree/Conformance_fixBugsInGettersForWebGL2. And I submitted one more patch to add new functionality for getRenderbufferParameter for WebGL 2.0: https://github.com/Richard-Yunchao/WebGL/tree/Conformance_updateGetRenderbufferParameterForWebGL2. Please take a look.

Richard-Yunchao:
@kenrussell and other reviewers. I admit that these getters should be wrapped into functions. I prefer to put these functions in the original file (conformance/state/gl-object-get-calls.html). They are not called by others. It is not necessary to put into ./conformance/resource/. Small change for a getter can be done directly in the original file to support WebGL 2.0, just like this one: https://github.com/KhronosGroup/WebGL/pull/898/. If a big change is necessary for a getter to support WebGL 2.0, we can rewrite that function in conformance2/state/gl-object-get-call-2.html. Then getters with simple change or without change, we can run the corresponding functions in conformance/ for all context version(WebGL 1.0, WebGL 2.0 and above), functions with big changes has different variants for different WebGL version, they will be run in conformance/  or conformance2/ respectively. 
As a result:
conformance/state/gl-objects-get-calls.html contains: functions without change(WebGL 1.0, 2.0 or above all run these functions), functions with small change(WebGL 1.0, 2.0 or above all run these functions), functions specific for WebGL 1.0 (WebGL 2.0 or above context will skip these tests)

conformance2/state/gl-objects-get-calls-2.html contains functions specific for WebGL 2.0 or above.

WDYT? Maybe I should update the code too. I will do it tomorrow.

kenrussell:
Sorry for the delay replying.

I think it's a mistake to change the behavior of the tests depending on the version of the context. There is already a mechanism for excluding tests from running on WebGL 1.0 or 2.0 -- the --min-version and --max-version flags in 00_test_list.txt. The WebGL 1.0 tests should not change behavior if they are running on WebGL 2.0. This means that conformance/state/gl-object-get-calls.html should have a --min-version of 1.9.9 and there should be a similar test in conformance2/ which does the right thing for WebGL 2.0.

I want you to have attribution in the changelogs for your work on this, so am going to merge #893 and do the refactoring I have in mind, which will involve removing the contextVersion accessor from webgl-test-utils.js. Once this is merged, please tell me whether you agree with the direction and whether you can rebase your other contributions on top of it. Thanks.

PR #890, opened 2015-03-12
these extensions are supported in core in WebGL 2.0 already

Richard-Yunchao:


Richard-Yunchao:
small change to update WebGL Extension Registry: these extensions are supported in core in WebGL 2.0 already.

Oletus:
I'm not sure if this is correct - how the extensions are specified differs from how the functionality is specified in WebGL 2.0. The float texture extensions have been under discussion lately, so there might be some changes to them as well.

kenrussell:
Olli's correct; I advised to add some undesirable behavior to the floating-point texture extensions for WebGL that isn't going to be brought forward to WebGL 2.0.

What's the motivation for making this change to the extension registry? Is it to try to document which extensions are expected to be provided by both WebGL 1.0 and 2.0 contexts, and which ones aren't expected to be exposed by WebGL 2.0 any more?

Richard-Yunchao:
@kenrussell, Yes. Just want to document which extensions are promoted into core or removed in WebGL 2.0 already, and the others are expected to be exposed in WebGL 2.0 rendering context.

kenrussell:
Sorry for the delay replying.

After looking over the ES 3.0 spec in more detail I don't think this change is appropriate. Here is one example why. OES_texture_float was written against OpenGL ES 2.0, which didn't have sized internal formats. With that extension, one can allocate a texture with format=RGBA, internalformat=RGBA, type=FLOAT. OpenGL ES 3.0 requires a sized internalformat to be specified in this case because table 3.3 doesn't support unsized internalformats for FLOAT type textures. It's likely that users' code will have to be updated in order to upgrade to WebGL 2.0 in areas like this.

Would you change this pull request to use the "removed" tag instead of the "core" tag?

Richard-Yunchao:
@kenrussell , Agree. I submitted another one: https://github.com/Richard-Yunchao/WebGL/tree/RemoveOrPromoteExts. PTAL. Thanks a lot!

kenrussell:
Closing this pull request as obsolete.

PR #893, opened 2015-03-17
fix bugs in conformance/state/gl-object-get-calls.html for WebGL 2.0

Richard-Yunchao:
Fix bugs in conformance/state/gl-object-get-calls.html for WebGL 2.0

It is the same with https://github.com/Richard-Yunchao/WebGL/tree/Conformance_addGetters, but squash all commits into one.


kenrussell:
Thanks for this contribution. Please see the comments in #889 for more background on the topic. I am going to merge this and then do a refactor to allow the WebGL 1.0 version of this test to behave unmodified when running on WebGL 2.0, and add a WebGL 2.0 version.

PR #894, opened 2015-03-17
update getRenderbufferParameter for WebGL 2.0

Richard-Yunchao:
Add WebGL 2.0 specific functionality in conformance2/state/gl-object-get-calls-2.html for GL state getters. Please take a look. If this is OK, more WebGL 2.0 specific functionalities for getters (getTexParameter, getBufferParameter, getFramebufferAttachmentParameter, etc.) will be added later.

### DIFF
--- a/sdk/tests/conformance/resources/gl-object-get-calls.js
+++ b/sdk/tests/conformance/resources/gl-object-get-calls.js
@@ -211,6 +211,9 @@ shouldBe('gl.getRenderbufferParameter(gl.RENDERBUFFER, gl.RENDERBUFFER_HEIGHT)',
 // the implementation is allowed to change it.
 shouldBeNonZero('gl.getRenderbufferParameter(gl.RENDERBUFFER, gl.RENDERBUFFER_INTERNAL_FORMAT)');
 shouldBeNonZero('gl.getRenderbufferParameter(gl.RENDERBUFFER, gl.RENDERBUFFER_DEPTH_SIZE)');
+if (contextVersion > 1) {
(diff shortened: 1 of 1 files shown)


kenrussell:
I've refactored the gl-object-get-calls.html test as described in #889 . Would you please take a look at the new structure and consider making this change in it? Please also provide feedback on the new structure. Thanks.

Richard-Yunchao:
@kenrussell , updated the code on the basis of your code refactoring. PTAL. Thanks a lot!

sdk/tests/conformance/resources/gl-object-get-calls.js
```diff
@@ -220,6 +220,10 @@ shouldBeNonZero('gl.getRenderbufferParameter(gl.RENDERBUFFER, gl.RENDERBUFFER_RE
 shouldBeNonZero('gl.getRenderbufferParameter(gl.RENDERBUFFER, gl.RENDERBUFFER_GREEN_SIZE)');
 shouldBeNonZero('gl.getRenderbufferParameter(gl.RENDERBUFFER, gl.RENDERBUFFER_BLUE_SIZE)');
 shouldBeNonZero('gl.getRenderbufferParameter(gl.RENDERBUFFER, gl.RENDERBUFFER_ALPHA_SIZE)');
+if (contextVersion > 1) {
+  gl.renderbufferStorageMultisample(gl.RENDERBUFFER, 4, gl.RGBA4, 2, 2);
```
kenrussell: Good. I see that the minimum allowed value for MAX_SAMPLES is 4 so this test is legal.

PR #895, opened 2015-03-17
Added conformance test for Chrome bug when loading SRGB textures from image element

toji:
Covers the case described in crbug.com/467590


sdk/tests/conformance/extensions/ext-sRGB.html
```diff
@@ -402,10 +406,27 @@
...
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+    callback();
+  }, function() {
+    testFailed("Image could not be loaded");
+    callback();
+  });
+}
+
 var successfullyParsed = true;
```
zhenyao: you don't need this line if all branches end up with finishTest()

toji: makeImage is async, so we have to call finishTest as a side effect of it's callback. Otherwise some tools (like Chrome's build bots) may detect that the test is complete when it hasn't run all the logic yet.

kenrussell:
The code changes LGTM.

Where did the image come from? I'd like to make sure there aren't any copyright issues.

toji:
Updated to use a flat gray image. Still had to be 1024x1024 to trigger the Chrome bug.

kenrussell:
Great. Thank you. LGTM

sdk/tests/conformance/extensions/ext-sRGB.html
zhenyao: What I mean is you don't need "var successfullyParsed = true;" because finishTest() takes care of it.  Well, no big deal.

PR #896, opened 2015-03-18
add getVersion to get current WebGL rendering context version

Richard-Yunchao:
separate from https://github.com/Richard-Yunchao/WebGL/tree/Conformance_fixBugsInGettersForWebGL2. This small change is standalone. Then my work can go on.

### DIFF
--- a/sdk/tests/conformance/resources/webgl-test-utils.js
+++ b/sdk/tests/conformance/resources/webgl-test-utils.js
@@ -1326,6 +1326,13 @@ var getUrlOptions = function() {
 };
 
 /**
+ * By default, the contextVersion is 1. If webgl rendering context 2.x (or 3.x
+ * or higher) is created by create3DContext, contextVersion should be changed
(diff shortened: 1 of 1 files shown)


Oletus:
I think that it would be better to add a function that takes a WebGL context as a parameter, and checks if the type of the context object is WebGL2RenderingContext. That way it would work even if a test uses multiple contexts with different versions of WebGL.

kenrussell:
As of the refactoring done in #901 I don't think it is necessary for the tests to query the context's version via WebGLTestUtils. I think it's a bad idea for tests to change behavior depending on the context version they're running on. Instead of this, if code sharing is needed between WebGL 1.0 and 2.0 tests, the test can be written as a separate source file, and the .html file can tell that source which version it is supposed to test.

I'm proactively closing this. Please reopen the pull request or open a new one if it's demonstrated that this primitive is absolutely necessary.

PR #897, opened 2015-03-18
Conformance test: bypass the extensions which has been promoted into core in WebGL 2

Richard-Yunchao:
Conformance test: bypass the extensions which has been promoted into core in WebGL 2.0 to fix bugs in WebGL extensions for WebGL 2. 

This simple change is based on this one: https://github.com/Richard-Yunchao/WebGL/tree/Conformance_addAPIToGetWebGLVersion/, PTAL.

### DIFF
--- a/sdk/tests/conformance/ogles/GL/biuDepthRange/biuDepthRange_001_to_002.html
+++ b/sdk/tests/conformance/ogles/GL/biuDepthRange/biuDepthRange_001_to_002.html
@@ -58,14 +58,22 @@ OpenGLESTestRunner.run({
               1.0
             ]
           }
-        }, 
+        },
         "fragmentShader": "../default/expected.frag"
       }, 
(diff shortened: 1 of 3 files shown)


Oletus:
I think we should rather use the --max-version directive in 00_test_list.txt to do this. That makes for a much smaller change.

kenrussell:
Agreed; the --max-version directive is the way to do this. See also the comments in #889 and the refactoring just done in #901. If you update this pull request to undo these changes and modify 00_test_list.txt, please squash the commits so all of these test modifications don't show up in the commit history. Thanks.

Richard-Yunchao:
@Oletus and @kenrussell , thanks for your reviewing. I am not aware --max-version directive. It is simple and direct. Thanks! Please help to review #906 when you have time. (I closed this one by mistake. Sorry... )

PR #898, opened 2015-03-18
update getBufferParameter for WebGL 2

Richard-Yunchao:
For simple change to GL state getter, I prefer to add it directly into sdk/tests/conformance/, instead of sdk/tests/conformance2/. WDYT? 

I prefer to make my changes small at the very beginning. It is easy to review. This small patch depends on this one: https://github.com/Richard-Yunchao/WebGL/tree/Conformance_addAPIToGetWebGLVersion/.


kenrussell:
I've refactored the gl-object-get-calls as described in #889. Would you please rebase this contribution on top of it? Thanks.

kenrussell:
Closing as obsolete.

PR #903, opened 2015-03-19
remove or promote extensions against WebGL 2.0

Richard-Yunchao:
Please take a look. Thanks!

### DIFF
--- a/extensions/OES_texture_float/extension.xml
+++ b/extensions/OES_texture_float/extension.xml
@@ -13,6 +13,7 @@
 
   <depends>
     <api version="1.0"/>
+    <removed version="2.0" />
   </depends>
 
   <overview>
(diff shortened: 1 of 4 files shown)


kenrussell:
I agree this is the best way to mark these extensions as no longer being advertised in the WebGL 2.0 context.

kenrussell:
I spoke too soon. WEBGL_depth_texture should be marked as removed, not promoted to core; as far as I can tell from reading the ES 3.0 spec, restrictions from WEBGL_depth_texture such as not being able to create cube map depth textures have been removed.

We need to be very careful when marking these extensions as merged into the WebGL 2 core API. I'll submit a follow-up pull request marking WEBGL_depth_texture as removed in WebGL 2.

PR #904, opened 2015-03-19
update getBufferParameter for WebGL 2.0

Richard-Yunchao:
rebase #898. I should rebase and update the original pull request... 

@kenrussell , The code refactoring is OK for me. It is clear. Differences of GL state getters between WebGLRenderingContext and WebGL2RenderingContext can be distinguished by contextVersion. And If a big change is needed, I can implement a new function for that getter to override the old one for WebGL2RenderingContext, and execute different ones according to contextVersion.


kenrussell:
Looks good. Thanks.

PR #906, opened 2015-03-20
Fix bugs in extensions for WebGL 2, they are promoted to core or removed 

Richard-Yunchao:
This change comes from #897. It skips extensions those have been promoted or removed in WebGL 2.0 by adding max-version for them.

I go through all GL extensions in WebGL Extension Registry webpage(https://www.khronos.org/registry/webgl/extensions/) , these extensions are promoted into core in WebGL 2.0:
ANGLE_instanced_arrays, 
OES_element_index_uint, 
OES_standard_derivatives,
OES_vertex_array_object,
WebGL_draw_buffers,
EXT_blend_minmax,
EXT_frag_depth,
EXT_sRGB,
EXT_shader_texture_lod. 

Aligned with #903 , WEBGL_depth_texture is also promoted into core. 
And these extensions are removed in WebGL 2.0:
OES_texture_float
OES_texture_half_float
OES_texture_half_float_linear.

Please have a look. Thanks a lot!

### DIFF
--- a/sdk/tests/conformance/extensions/00_test_list.txt
+++ b/sdk/tests/conformance/extensions/00_test_list.txt
@@ -1,35 +1,35 @@
---min-version 1.0.3 angle-instanced-arrays.html
---min-version 1.0.3 angle-instanced-arrays-out-of-bounds.html
---min-version 1.0.3 ext-blend-minmax.html
---min-version 1.0.3 ext-frag-depth.html
---min-version 1.0.3 ext-shader-texture-lod.html
---min-version 1.0.3 ext-sRGB.html
+--min-version 1.0.3 --max-version 1.9.9 angle-instanced-arrays.html
+--min-version 1.0.3 --max-version 1.9.9 angle-instanced-arrays-out-of-bounds.html
+--min-version 1.0.3 --max-version 1.9.9 ext-blend-minmax.html
+--min-version 1.0.3 --max-version 1.9.9 ext-frag-depth.html
+--min-version 1.0.3 --max-version 1.9.9 ext-shader-texture-lod.html
+--min-version 1.0.3 --max-version 1.9.9 ext-sRGB.html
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/extensions/00_test_list.txt
```diff
@@ -1,35 +1,35 @@
...
 webgl-debug-shaders.html
 --min-version 1.0.3 webgl-compressed-texture-atc.html
 --min-version 1.0.3 webgl-compressed-texture-pvrtc.html
 --min-version 1.0.2 webgl-compressed-texture-s3tc.html
 --min-version 1.0.3 webgl-compressed-texture-size-limit.html
---min-version 1.0.2 webgl-depth-texture.html
---min-version 1.0.3 webgl-draw-buffers.html
+--min-version 1.0.2 --max-version 1.9.9 webgl-depth-texture.html
+--min-version 1.0.3 --max-version 1.9.9 webgl-draw-buffers.html
 --min-version 1.0.4 webgl-draw-buffers-max-draw-buffers.html
```
kenrussell: This one should also be marked as --max-version 1.9.9 if we're going to do so for webgl-draw-buffers.html.

kenrussell:
One thought on this patch. While these tests require the extensions to be present, most of them could be refactored to also run on top of WebGL 2.0. That would avoid losing potentially valuable test coverage for WebGL 2.0. I'll merge this as is but please consider helping do this. I'm thinking that each test could be refactored into a .js file like get-object-calls.html was, and referenced from some other test name in the conformance2/ directory like conformance2/extensionsNowInCore/... .

Richard-Yunchao:
Agree. These extensions can be somehow tested against WebGL 2.0 even they are not present. Looking forward to your code refactoring, @kenrussell .

kenrussell:
@Richard-Yunchao : any contribution you would be willing to make to refactor and reuse these tests would be welcome. I don't have any immediate plans to do it.

PR #908, opened 2015-03-20
Mark WEBGL_depth_texture as removed in WebGL 2.0 instead of in core.

kenrussell:
Following up to #903, this extension's limitations have been removed in WebGL 2.0 so the extension should probably be marked as removed.


toji:
LGTM

PR #910, opened 2015-03-23
Add tests disallowing operators on certain types of structs

Oletus:
Comparing or assigning structs that contain arrays or samplers is "not
defined" in ESSL 1.00 (spec section 5.7). In addition the spec
mentions in section 5.8 that

"structures containing arrays may be passed to parameters declared as out
or inout but may not be used as the target of an assignment."

and in section 5.9:

"The equality operators equal (==), and not equal (!=) operate on all
types except arrays, structures containing arrays, sampler types and
structures containing sampler types"

It is also not allowed to use samplers as l-values, so it makes sense
that using structs containing samplers as l-values would be disallowed as
well.

Add tests to enforce that equality and assignment of structs containing
arrays or samplers are disallowed in WebGL. Spec clarification doesn't
seem necessary as the ESSL spec is already pretty clear on this.

On many platforms shaders that contain these operations are already
failing compilation:

1. Equality comparison of structs containing arrays or samplers is
failing on at least Linux/NVIDIA/Chrome, Windows/Intel/IE and
Windows/Intel/Firefox.

2. Assignment of structs containing arrays is failing on at least
Windows/Intel/IE.

3. Assignment of structs containing samplers is failing on at least
Linux/NVIDIA/Chrome, Windows/Intel/IE and Windows/Intel/Firefox.

### DIFF
--- a/sdk/tests/conformance/glsl/misc/00_test_list.txt
+++ b/sdk/tests/conformance/glsl/misc/00_test_list.txt
@@ -99,6 +99,7 @@ uniform-location-length-limits.html
 --min-version 1.0.2 shader-with-global-variable-precision-mismatch.html
 --min-version 1.0.2 large-loop-compile.html
 --min-version 1.0.3 struct-equals.html
+--min-version 1.0.4 struct-assign.html
 --min-version 1.0.3 struct-mixed-array-declarators.html
 --min-version 1.0.3 struct-nesting-of-variable-names.html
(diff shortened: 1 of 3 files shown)


Oletus:
Spec section also 4.1.7 has a bit more concerning samplers in structs.

kenrussell:
Very nice. Thanks for digging into this in detail and for making the tests more rigorous.

PR #913, opened 2015-03-26
Added buffer restriction test

matthewn4444:
Tests the functions bindBuffer, bindBufferRange and bindBufferBase
with all the valid targets in WebGL 2.0.


kenrussell:
Thanks for this contribution. The new test looks good.

PR #918, opened 2015-03-27
Added buffer copying test

matthewn4444:
Tests the function copyBufferSubData with all the valid targets in WebGL 2.0.

### DIFF
--- a/sdk/tests/conformance2/buffers/00_test_list.txt
+++ b/sdk/tests/conformance2/buffers/00_test_list.txt
@@ -1,1 +1,2 @@
 buffer-type-restrictions.html
+buffer-copying-restrictions.html
--- a/sdk/tests/conformance2/buffers/00_test_list.txt
+++ b/sdk/tests/conformance2/buffers/00_test_list.txt
@@ -1,2 +1,3 @@
-buffer-type-restrictions.html
+buffer-copying-contents.html
 buffer-copying-restrictions.html
+buffer-type-restrictions.html
--- a/sdk/tests/conformance2/buffers/buffer-copying-restrictions.html
+++ b/sdk/tests/conformance2/buffers/buffer-copying-restrictions.html
@@ -1,0 +1,107 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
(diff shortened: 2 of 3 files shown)


sdk/tests/conformance2/buffers/00_test_list.txt
```diff
@@ -1 +1,2 @@
 buffer-type-restrictions.html
+buffer-copying-restrictions.html
```
kenrussell: In general let's keep these lists alphabetized.

sdk/tests/conformance2/buffers/buffer-copying-restrictions.html
```diff
@@ -0,0 +1,107 @@
...
+  var testCopyStr = "copying from a gl." + wtu.glEnumToString(gl, srcTarget) + " buffer to a gl."
+    + wtu.glEnumToString(gl, dstTarget) + " buffer"
+
+  gl.bindBuffer(srcTarget, srcBuffer);
+  gl.bufferData(srcTarget, new Float32Array(32), gl.STATIC_DRAW);
+  gl.bindBuffer(dstTarget, dstBuffer);
+  gl.bufferData(dstTarget, new Float32Array(32), gl.STATIC_DRAW);
+  gl.copyBufferSubData(srcTarget, dstTarget, 8, 0, 4);
+  if (srcTarget == dstTarget)
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, testCopyStr + " should work.");
```
kenrussell: Do you think it would be a good idea to verify the contents of the destination buffer via getBufferSubData? Or should there be a separate test for that?

kenrussell:
Thanks for this contribution. One minor comment and one question.

Oletus:
I'd add a separate test case that verifies the buffer contents after copying. It's much easier to debug test cases when they concentrate on one specific aspect of the spec.

matthewn4444:
I'll add that test later.

Oletus:
ken: Do the tests look okay now? These have been sitting in review for quite a while.

kenrussell:
I'm sorry, I missed your earlier post. Please feel free to prod me more quickly next time if it looks like there's been an oversight. These tests look good. Thanks for contributing them.

PR #919, opened 2015-03-31
Add simple ESSL3 tests for array assignment and equality

Oletus:
Test ESSL3 array assignment and equality operators with arrays of ints
and arrays of structs. These use ESSL1-compatible code for array
initialization, which helps with developing these features independently
of array constructors. This complements dEQP array tests, which require
constructors to be supported.


kenrussell:
Very nice. Thanks for contributing these tests.

PR #920, opened 2015-03-31
Added shell script to invoke closure against the ported dEQP tests.

kenrussell:


kenrussell:
Note that this doesn't run cleanly yet -- it has about 100 warnings. We'll need to go through them one by one.

toji:
LGTM!

PR #921, opened 2015-04-01
getBufferSubData should accept null param for returnedData.

djg:
https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.7.2 says for `getBufferSubData()`:

> If returnedData is null then an INVALID_VALUE error is generated.

The current webidl doesn't allow null. Instead, in FF for example, a TypeError is caused:
```javascript
var data = null;
gl.getBufferSubData(gl.ARRAY_BUFFER, 0, data);
```
results in

```
TypeError: Argument 3 is not valid for any of the 3-argument overloads of WebGL2RenderingContext.getBufferSubData.
```

PS: I hope I updated the correct date.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -27,7 +27,7 @@
     <!--end-logo-->
 
     <h1>WebGL 2 Specification</h1>
-    <h2 class="no-toc">Editor's Draft 12 March 2015</h2>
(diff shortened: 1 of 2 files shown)


kenrussell:
Thanks for fixing this. There's another place further down in the file which needs to be updated too, but I'll take care of it in a separate pull request.

PR #922, opened 2015-04-01
Added Closure externs file for WebGL 2.0 under sdk/tests/deqp/webgl2.js ...

kenrussell:
..., based on the WebGL 1.0 externs.

Temporarily copied in all of the Closure compiler's externs, modulo webgl.js, in order to test overriding of the compiler's built-in WebGL externs.

The new externs parse, but have not been tested against the ported dEQP tests yet.

### DIFF
--- a/sdk/tests/deqp/temp_externs/chrome.js
+++ b/sdk/tests/deqp/temp_externs/chrome.js
@@ -1,0 +1,156 @@
+/*
+ * Copyright 2013 The Closure Compiler Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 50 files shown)


kenrussell:
I'd like to merge this pull request to make it easier for Mobica to type check their ported dEQP tests against the WebGL 2.0 API. Once we're convinced these externs are correct, we'll ask to merge them into the Closure compiler's workspace and delete both the webgl2.js and the copied-in externs.

toji:
There's a minor quirk in Blink that requires WebGL2RenderingContext to explicitly implement both WebGLRenderingContextBase and WebGL2RenderingContextBase or it doesn't pick up the members of the WebGLRenderingContextBase prototype. I think this is a Blink-only issue, but we should probably verify that Closure's type checking properly looks at the entire object chain. (ie: Check that WebGL 1 properties are validated correctly when used on a WebGL 2 context.)

Otherwise LGTM, with hopes that we can reduce the need for these definitions to be explicitly added here in the future.

kenrussell:
Thanks for the feedback. Added a check of a WebGL 1.0 property; it works.

This extern file will be used and tested more heavily as Mobica's dEQP port is changed to reference it, and we'll upstream it to the Closure compiler's workspace at that point. Merging this now to improve testing.

PR #926, opened 2015-04-02
Add an ESSL3 test for assigning array constructor results

Oletus:
The test covers assigning array constructor return values as opposed to
initializing arrays using array constructors. A HLSL backend of WebGL
might have very different implementations for initialization and
assignment, so it is good to have separate coverage for both. dEQP tests
have some coverage for both, but there are some limitations: dEQP
coverage for constructors of arrays of structs only uses initialization
as opposed to assignment, for example.


kenrussell:
Looks good. Thanks for this contribution.

PR #927, opened 2015-04-02
Add notes about rationale for WebGL 2 - GLES3 incompatibility

Oletus:
Add rationale for some of the WebGL 2 spec items for future reference and
to avoid repeating discussions.

Use a different CSS class for these notes to make them stand out from
other kinds of non-normative comments.

### DIFF
--- a/resources/default.css
+++ b/resources/default.css
@@ -227,6 +227,14 @@ div.note.editor {
     background: #FFDDF8;
     color: black;
 }
+div.note.rationale:before {
+    content: "Rationale";
+    background: #006;
+}
+div.note.rationale {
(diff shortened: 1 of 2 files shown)


Oletus:
Here's some rationale for spec decisions made in the past for WebGL 2.0.

Is it okay to edit the CSS file as I've done here, or do we want to keep that in sync with other specs?

kenrussell:
Thanks for putting these together. Editing this CSS file is fine -- no other specs are referencing it.

In general, please go ahead and update the last modified date at the top of the file when updating the spec. These look good as is, so thanks again and merging them now.

PR #928, opened 2015-04-03
Added WebGL 2 test for new getParameter enums

toji:
Also fixed a couple of typos in the spec found while verifying the values.


kenrussell:
Fantastic. Thanks for this contribution.

PR #929, opened 2015-04-06
Fixed erronious removal of some enums from getParameter table

toji:
I removed TRANSFORM_FEEDBACK_BUFFER_BINDING and UNIFORM_BUFFER_BINDING in a prior patch because they were indexed parameters. I didn't realize that it was also valid to query them non-indexed. Sorry for the confusion.

TRANSFORM_FEEDBACK_BINDING, on the other hand, seems to have simply been overlooked previously.


kenrussell:
LGTM. Sorry, I should have reviewed your earlier patch more carefully.

kenrussell:
P.S. We should update the spec's last modified date when making changes like this.

PR #930, opened 2015-04-07
add GLenums to getVertexAttrib in conformance test for WebGL 2

Richard-Yunchao:


sdk/tests/conformance/resources/gl-object-get-calls.js
```diff
@@ -420,12 +435,11 @@ var validArrayForVertexAttrib = new Array(
     gl.VERTEX_ATTRIB_ARRAY_STRIDE,
     gl.VERTEX_ATTRIB_ARRAY_TYPE,
     gl.VERTEX_ATTRIB_ARRAY_NORMALIZED,
-    gl.VERTEX_ATTRIB_ARRAY_STRIDE,
-    gl.VERTEX_ATTRIB_ARRAY_ENABLED,
```
kenrussell: Why were the invalid argument tests for these two enums removed? This change does not look good.

kenrussell:
The overall addition looks fine but the removal of existing tests does not.

Richard-Yunchao:
@kenrussell , the removed GLenums are redundant. They have been included in the array.

kenrussell:
Ah, sorry, I missed that. Thanks for this contribution then.

PR #931, opened 2015-04-07
update getProgramParameter in Conformance test for WebGL 2

Richard-Yunchao:
PTAL when you have time. 

The new shaders with uniform block are quite simple, we can add more code into it for other conformance tests later.

### DIFF
--- a/sdk/tests/conformance/resources/gl-object-get-calls.js
+++ b/sdk/tests/conformance/resources/gl-object-get-calls.js
@@ -182,7 +182,6 @@ testInvalidArgument(
     }
 );
 
-
 debug("");
 debug("Test getProgramParameter");
 shouldBe('gl.getProgramParameter(standardProgram, gl.DELETE_STATUS)', 'false');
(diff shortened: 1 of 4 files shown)


sdk/tests/conformance/resources/gl-object-get-calls.js
```diff
@@ -191,21 +191,52 @@ shouldBe('typeof gl.getProgramParameter(standardProgram, gl.VALIDATE_STATUS)', '
...
+  shouldBe('gl.getProgramParameter(standardProgram, gl.ACTIVE_UNIFORM_BLOCKS)', '1');
+
+  shouldBe('gl.getProgramParameter(standardProgram, gl.TRANSFORM_FEEDBACK_VARYINGS)', '2');
+  shouldBe('gl.getProgramParameter(standardProgram, gl.TRANSFORM_FEEDBACK_BUFFER_MODE)', 'gl.INTERLEAVED_ATTRIBS');
+
+  validArrayForProgramParameter += [
+      gl.ACTIVE_UNIFORM_BLOCKS,
+      gl.TRANSFORM_FEEDBACK_VARYINGS,
+      gl.TRANSFORM_FEEDBACK_BUFFER_MODE
+  ]
```
kenrussell: These need to be added to the valid array before the first call to testInvalidArgument for getProgramParameter above, or that test will fail. Correct? It should not be necessary to duplicate that test in this if-block. Did you run this test on a WebGL 2 prototype and make sure it passed?

Richard-Yunchao:
@kenrussell , code has been updated accordingly. Please have a look when you have time. Thanks a lot!

sdk/tests/conformance/resources/uniformBlockShader.frag
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (c) 2015 The Chromium Authors. All rights reserved.
```
kenrussell: Sorry I didn't notice this before, but please use the Khronos copyright header for these new files. You can use sdk/tests/conformance/resources/gl-object-get-calls.js as a template.

kenrussell:
Thanks, the new code looks great. Sorry, but one issue with the copyright headers of the new shaders. When you update this pull request please squash the commits so the intermediate commit with the old copyright header isn't in the version history. Thanks. Will merge this as soon as the update is ready.

Richard-Yunchao:
@kenrussell , the copyright headers are updated accordingly. PTAL. Thanks a lot!

sdk/tests/conformance/resources/uniformBlockShader.vert
```diff
@@ -0,0 +1,41 @@
+/*
```
kenrussell: Since this shader uses GLSL ES 3.0 syntax I wonder whether it needs the "#version 300 es" directive as the first line.

kenrussell:
Thanks for this contribution -- looks good. Assuming this has been tested on one of the WebGL 2.0 prototypes.

PR #932, opened 2015-04-07
Added test of uploading cube map faces out of order.

kenrussell:
Thanks to Gregg Tavares for the original bug report and test case.


kenrussell:
This catches the bug in Chrome, crashes Firefox, and fails to load in Safari on the affected machines.

PR #934, opened 2015-04-09
Add a conformance test for empty declarations

Oletus:
This corner case has been discussed on the Khronos bug tracker, with the conclusion that empty declarations including ones followed by declarators are valid outside structs.

IE11 fails two of the tests, NVIDIA GL driver fails one of them.


toji:
Good test, thanks! The quirks of GLSL never cease to amaze me.

Oletus:
The quirks are plentiful indeed, thanks for the quick review!

PR #937, opened 2015-04-10
fix errors for resource loading when call it from conformance2/

Richard-Yunchao:
PTAL when you have time. Thanks!

### DIFF
--- a/sdk/tests/conformance/resources/gl-object-get-calls.js
+++ b/sdk/tests/conformance/resources/gl-object-get-calls.js
@@ -311,7 +311,7 @@ testInvalidArgument(
 debug("");
 debug("Test getUniform with all variants of data types");
 // Boolean uniform variables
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
BTW, all resources (js, shader, etc) for conformance test are located in conformance/ file folder. It is reasonable when conformance2/ doesn't exist. Right now, it seems a little wired. WDYT? Could we move all files in sdk/tests/conformance/resource/ to a file folder under sdk/tests/ ?

kenrussell:
Thanks for this fix.

You have a good point about the location of the conformance tests' resources. It's only historical that there are two directories, sdk/tests/resources/ and sdk/tests/conformance/resources/ . Would you like to put together a pull request moving all the files from sdk/tests/conformance/resources/ into sdk/tests/resources/ and deleting the directory sdk/tests/conformance/resources/ ? It'll require a lot of updates to a lot of files, so having that as an independent pull request would be great. Thanks in advance for your help.

Richard-Yunchao:
@kenrussell , that's OK. I will submit a patch to do that soon.

PR #938, opened 2015-04-12
Initial version of EXT_float_blend wrapper.

msc-:


kenrussell:
Thank you for moving this forward.

PR #939, opened 2015-04-15
Add gl-vertexattribipointer.html for WebGL 2.0

legendlee1314:


legendlee1314:
@kenrussell , PTAL if have time. This patch is aimed to [chromium cl 1082203002](https://codereview.chromium.org/1082203002/).

legendlee1314:
@kenrussell , by the way, I would like to port `gl-vertexattribipointer-offsets.html` as `gl-vertexattribpointer-offsets.html` does in the directory `conformance/attrib`. However, it fails after I copy that and do some changes for api `vertexAttribIPointer`. Perhaps this is not ready at command buffer implementations. I'll follow up this. What's your opinion? Thanks!

sdk/tests/conformance2/attribs/gl-vertexattribipointer.html
```diff
@@ -0,0 +1,143 @@
...
+
+  var vertexObject = gl.createBuffer();
+  gl.bindBuffer(gl.ARRAY_BUFFER, vertexObject);
+  gl.bufferData(gl.ARRAY_BUFFER, new Int32Array(0), gl.STATIC_DRAW);
+
+  gl.vertexAttribIPointer(0, 1, gl.FLOAT, 0, 0);
+  wtu.glErrorShouldBe(gl, gl.INVALID_ENUM,
+      "vertexAttribIPointer should not support FLOAT");
+
+  var checkVertexAttribPointer = function(
```
kenrussell: It would be more clear to rename this to checkVertexAttribIPointer.

kenrussell:
Thanks for this contribution. Since it involves adding tests of a new API rather than just expanding out the range of enums that a previous test verifies, it seems fine.

Your contribution of more tests like porting gl-vertexattribpointer-offsets.html into gl-vertexattribipointer-offsets.html would be appreciated. It's expected that there are many bugs in Chrome's WebGL 2 prototype, so please also test against Firefox's, and file bugs against the browsers where you find bugs. Thanks.

sdk/tests/conformance2/attribs/gl-vertexattribipointer.html
legendlee1314: Sorry for ignoring that. Could I fix it in the next PR since it is merged?

legendlee1314:
Thanks for your comments. That is really what I need to do next. :)

sdk/tests/conformance2/attribs/gl-vertexattribipointer.html
kenrussell: Yes, certainly.

PR #940, opened 2015-04-15
Global variable initializers should be constant expressions

Oletus:
Global variable initializers need to be constant expressions (ESSL1.00
section 4.3). Global variables need to be qualified with const in order
to be constant expressions (ESSL1.00 section 5.10).

Fix the ternary operator in global initializer test by making the
other globals used in initialization constant.

Add a test that verifies that non-const global variables cannot be used
to initialize other globals.

### DIFF
--- a/conformance-suites/1.0.3/conformance/glsl/misc/ternary-operators-in-global-initializers.html
+++ b/conformance-suites/1.0.3/conformance/glsl/misc/ternary-operators-in-global-initializers.html
@@ -38,8 +38,8 @@
 <div id="console"></div>
 <script id="fragmentShader" type="text/something-not-javascript">
 precision mediump float;
-$(type) green = $(green);
-$(type) black = $(black);
+const $(type) green = $(green);
+const $(type) black = $(black);
 $(type) var = (true) ? green : black;
 void main() {
     gl_FragColor = $(asVec4);
--- a/sdk/tests/conformance/glsl/misc/00_test_list.txt
+++ b/sdk/tests/conformance/glsl/misc/00_test_list.txt
@@ -5,6 +5,7 @@ embedded-struct-definitions-forbidden.html
 --min-version 1.0.4 empty-declaration.html
 empty_main.vert.html
 --min-version 1.0.3 expression-list-in-declarator-initializer.html
+--min-version 1.0.4 global-init-with-non-const.html
 gl_position_unset.vert.html
 # this test is intentionally disabled as it is too strict and to hard to simulate
 # glsl-2types-of-textures-on-same-unit.html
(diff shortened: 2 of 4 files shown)


Oletus:
This is a pretty unfortunate little bug, since I suspect that fixing it might break a ton of real-world content. Desktop GLSL doesn't seem to have a similar restriction, though I suspect there are some unwritten rules that compilers follow in practice (like disallowing initializing globals with uniform values). Changing the spec might be a better option.

kenrussell:
This is an unfortunate oversight, and the impact of enforcing this behavior needs to be judged carefully. If it is going to break a lot of content, and if this content is already running successfully on mobile devices, then the specification should be changed.

Please submit the bug fix to the ternary operator test as a separate pull request. That way we can collectively give ourselves more time to understand whether the new test should be added.

zhenyao:
I agree with kbr that we should be careful adding the test or modifying the implementation.

I feel we can just leave it as grey area and keep allow the status quo for now. We should modify the webgl spec to allow non-constant global variable initialization only if the underlying ES2 and ES3 can also be modified.

jamiemadill:
I'd be in favour of ripping off the band-aid here and seeing what happens. If we generate clear errors hopefully it should be easy to triage issues as they come in. FWIW I don't think there will be as many instances in the wild as for the precision mediump float issue.

kenrussell:
In order to responsibly make this change we need to first test a lot of content ourselves with the new rule in place to see what will break. Who will take on that task?

jamiemadill:
Hey Ken, I'd like to take this on. It's blocking WebGL conformance already for angle with the ternary global initializers bug, and because the functionality I think we should remove was broken already in ANGLE for a long time (probably only hit stable recently), I suspect it's not a wide-spread use pattern.

kenrussell:
Thanks Jamie. Would you please do some testing with the restrictions enforced in ANGLE and report on this pull request what you find (what sites you tested, whether they worked, etc.)?

Oletus:
I don't think we can do this change, I did some testing with preventing initialization of globals with uniforms and there's some content that breaks if this restriction is enforced. Also, there's some other problematic cases we also need to test, like calling user-defined functions in global initializers. Are you okay with closing this pull request? I'll still keep working on this, I've been drafting a possible spec change and I could also contribute tests once we agree on what the exact behavior should be.

jamiemadill:
I'm fine with that, Olli.

kenrussell:
Thanks Olli for determining the impact of this change. I agree that it's necessary to preserve compatibility with WebGL 1.0 content in this area rather than adhere strictly to the ESSL 1.0 spec. Please put up a pull request with changes to the WebGL spec (and any new tests, though those can come separately) when you're ready. Thanks.

PR #941, opened 2015-04-15
Add getBufferSubData test

matthewn4444:
Tests the function getBufferSubData and its restrictions in WebGL 2.0

### DIFF
--- a/sdk/tests/conformance2/buffers/getBufferSubData.html
+++ b/sdk/tests/conformance2/buffers/getBufferSubData.html
@@ -62,15 +62,9 @@ gl.bufferData(gl.ARRAY_BUFFER, floatArray, gl.STATIC_DRAW);
 
 debug("");
 debug("Test that getBufferSubData successfully works reading buffer data from gl.ARRAY_BUFFER");
-var arrBuffer = new ArrayBuffer(vertices.length * 4);
+var arrBuffer = new ArrayBuffer(vertices.length * Float32Array.BYTES_PER_ELEMENT);
 gl.getBufferSubData(gl.ARRAY_BUFFER, 0, arrBuffer);
 wtu.glErrorShouldBe(gl, gl.NO_ERROR, "gl.getBufferSubData(gl.ARRAY_BUFFER, 0, arrBuffer) should WORK");
-
-debug("Make sure that the length of the array passed to getBufferSubData doesn't change");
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/buffers/getBufferSubData.html
```diff
@@ -0,0 +1,116 @@
...
+
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "Should be no errors from setup.");
+
+var buffer = gl.createBuffer();
+gl.bindBuffer(gl.ARRAY_BUFFER, buffer);
+gl.bufferData(gl.ARRAY_BUFFER, floatArray, gl.STATIC_DRAW);
+
+debug("");
+debug("Test that getBufferSubData successfully works reading buffer data from gl.ARRAY_BUFFER");
+var arrBuffer = new ArrayBuffer(vertices.length * 4);
```
kenrussell: This would be better written as new ArrayBuffer(floatArray.byteLength).

sdk/tests/conformance2/buffers/getBufferSubData.html
```diff
@@ -0,0 +1,116 @@
...
+gl.getBufferSubData(gl.ARRAY_BUFFER, 0, arrBuffer);
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "gl.getBufferSubData(gl.ARRAY_BUFFER, 0, arrBuffer) should WORK");
+
+debug("Make sure that the length of the array passed to getBufferSubData doesn't change");
+if (arrBuffer.byteLength / 4 == vertices.length)
+  testPassed("The returned array's length matches the original data length");
+else
+  testFailed("The returned array's length does not match the original data length");
```
kenrussell: ArrayBuffers' size is immutable except when they're transferred, so these checks don't seem useful.

sdk/tests/conformance2/buffers/getBufferSubData.html
kenrussell: (because of the constant "4". Float32Array.BYTES_PER_ELEMENT would be better at least.)

kenrussell:
Thanks for this contribution. It looks good overall. Could you please add it to 00_test_list.txt as well as address the feedback above?

Please ask toji or zhenyao to do the final review and merge. Thanks.

toji:
Test looks good! Thanks!

PR #942, opened 2015-04-15
Allow multiple checkCanvasRectColor checks with one readPixels call

hendrikwagenaar:
Quick optimization to make the test run faster.  readPixels can be slow, so call readPixels once on a union of the rects, and check each subrect for matching color.

### DIFF
--- a/sdk/tests/conformance/resources/webgl-test-utils.js
+++ b/sdk/tests/conformance/resources/webgl-test-utils.js
@@ -1071,6 +1071,79 @@ var isWebGLContext = function(ctx) {
 };
 
 /**
+ * Creates a check rect is used by checkCanvasRects.
+ * @param {number} x left corner of region to check.
+ * @param {number} y bottom corner of region to check in case of checking from
+ *        a GL context or top corner in case of checking from a 2D context.
+ * @param {number} width width of region to check.
+ * @param {number} height width of region to check.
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/resources/webgl-test-utils.js
```diff
@@ -1071,6 +1071,79 @@ var isWebGLContext = function(ctx) {
...
+ */
+var makeCheckRect = function(x, y, width, height, color, msg, errorRange) {
+  var rect = {
+    'x': x, 'y': y,
+    'width': width, 'height': height,
+    'color': color, 'msg': msg,
+    'errorRange': errorRange,
+
+    'checkRect': function (buf, l, b, w) {
+      for (var px = (x - l) ; px < (width - l) ; ++px) {
```
toji: This feels wrong. It's appears to be treating "width" as "right". A simple example is that if I test a rect with x = 10, width = 2 the loop would exit without testing any pixels.

Seems like the more accurate loop would be:
`for (var px = (x - l) ; px < (x + width - l) ; ++px) {`

Same goes for height below.

toji:
Looks good as long as the loop logic is fixed.

hendrikwagenaar:
Good catch, fixed

toji:
LGTM

PR #945, opened 2015-04-20
Modified a test to make sure MAX_DRAW_BUFFERS is equal to MAX_COLOR_ATTA...

matthewn4444:
Added a small test to draw-buffers.html to make sure that the enums match.


kenrussell:
Thanks. Looks good.

PR #946, opened 2015-04-21
update getTexParameter for WebGL 2

Richard-Yunchao:


kenrussell:

PR #947, opened 2015-04-21
Clarify error msg in expando test

Oletus:
Without this patch the expando test fails in a peculiar way in Chrome by
generating an exception, which prevents it from running into completion
and marking successfullyParsed = true. Make sure that expando2 property
is set before querying its properties to generate a better error message.


toji:
LGTM

PR #948, opened 2015-04-23
Add one more struct test to the empty declaration test

Oletus:
Empty struct declarations need some special handling since they can't
be pruned unlike other empty declarations. That makes empty struct
declarations followed by a declarator an important corner case to test.


toji:
LGTM

PR #949, opened 2015-04-23
Add WebGL version to GLSL conformance test parameters

Oletus:
This makes the WebGL 2 GLSL tests easier to use, since specifying the
context version separately with webglVersion=2 is no longer necessary.


toji:
Agreed, much nicer!

PR #950, opened 2015-04-23
Clean up small issues in some ESSL and WebGL 2 tests

Oletus:
Prefer checkCanvas to checkCanvasRect, and remove unnecessary script tags
and an unnecessary uniform that were left over to tests from copy-pasted
code.


zhenyao:
Looks good.  Thanks for the cleanup!

PR #951, opened 2015-04-23
Add a WebGL 2 test for array length expressions that have side effects

Oletus:
ESSL3 defines expressions in section 5.9 so that the length() method is
only allowed to be called on array names, not on complex expressions that
return arrays. Enforce this restriction, which simplifies constant
folding of length() method calls.

### DIFF
--- a/sdk/tests/conformance2/glsl3/array-length-side-effects.html
+++ b/sdk/tests/conformance2/glsl3/array-length-side-effects.html
@@ -86,6 +86,7 @@ void main() {
 <script type="text/javascript">
 "use strict";
 description();
+debug("These restrictions come from ESSL 3.00 section 5.9 definition of expression, which only allows length to be called on array names, not on arbitrary expressions returning an array.");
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/glsl3/array-length-side-effects.html
```diff
@@ -0,0 +1,118 @@
...
+out vec4 my_FragColor;
+
+void main() {
+    int a = (int[1](0)).length();
+    my_FragColor = vec4(float(a));
+}
+</script>
+<script type="text/javascript">
+"use strict";
+description();
```
zhenyao: Can you add a comment here, pointing to the text in the spec leading to this test? Otherwise people will very likely keep asking questions because this is a corner case and is different from Desktop GL.

Oletus:
zhenyao: I updated the pull request with the comment.

zhenyao:
Thanks.  LGTM.

PR #953, opened 2015-04-24
Fix shader in WebGL 2 draw buffers test

Oletus:
Shader outputs declared as arrays can only be indexed by constant
integral expressions, as specified in section 4.3.6 of the ESSL3 spec.
This restriction doesn't exist in the WebGL 1 extension, so only the
WebGL 2 test needs to be changed.


zhenyao:
LGTM

Issue #916, opened 2015-03-27
Ensure that all WebGL specific behavioral changes are documented.


pyalot:
WebGL introduces behavioral changes to the underlying specifications (often in areas where undefined underlying behaviors exist).

Ensure that all such changes (new errors, semantic changes, etc.) are documented at least with a reference to the underlying specification sections and text exerpt as applicable, or failing to that, with a reasoning (in case of other incompatibilities like severe performance drawbacks, widespread driver errors or D3D compatibility issues).

pyalot:
Unaddressed issue prematurely closed partaining to this issue: https://github.com/KhronosGroup/WebGL/issues/914

Issue #914, opened 2015-03-27
offset/stride multiple of type for vertexAttribPointer


pyalot:
The WebGL 1/2 specification defines an INVALID_OPERATION error for offsets/strides to vertexAttribPointer that are not multiples of the passed type.

This restriction/error does not exist in OpenGL ES 1/2/3 or Desktop GL 1/2/3/4.

It is possible this was introduced because of severe performance concerns or to keep D3D compatibility (I don't know). The WebGL 1/2 specification does not say.

* Clarify in the specification why that restriction was introduced to WebGL OR
* Remove the restriction if there isn't a good reason for it (I doubt that would be the case, there's got to be some reason right?)

kenrussell:
As has been pointed out on the public_webgl mailing list, this restriction is in fact present in section 2.9 "Buffer Objects" of the OpenGL ES 2.0 specification:

"Clients must align data elements consistent with the requirements of the client platform, with an additional base-level requirement that an offset within a buffer to a datum comprising /N/ basic machine units be a multiple of /N/."

Catching the illegal inputs during the call to vertexAttribPointer prevents the illegal state from occurring. I don't feel that additional specification text is required.

PR #954, opened 2015-04-24
Add reasoning for buffer alignment.

jdashg:
I wasn't sure on how to format this. I think this works.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -3362,6 +3362,18 @@ size of the data type passed to the call, or an <code>INVALID_OPERATION</code> e
 
 </p>
 
+<div class="note">
+    <p>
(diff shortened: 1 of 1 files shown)


specs/latest/1.0/index.html
```diff
@@ -3362,6 +3362,18 @@ <h2 class="no-toc">Table of contents</h2>
 
 </p>
 
+<div class="note">
```
kenrussell: Olli just added the ability to specify the classes "note rationale" to get slightly different formatting. Could you use that here?

kenrussell:
Could you also please update the last modified date at the top of the file?

For the record, this addresses Issue #914 which I prematurely closed and partially addresses Issue #916 .

jdashg:
Fixed?

kenrussell:
Looks good. Thanks.

PR #955, opened 2015-04-24
Refactored deqp tests to use closure library.

jasomobica:


kenrussell:
Thanks for doing this refactoring. A few issues.

 - Some errant files like deMath.js and deString.js ended up in sdk/tests/closure-library/closure/goog. I'm sure they were put there only for debugging but make me concerned about how many other changes might have been made to the closure library sources.
 - I hadn't realized how big the closure library was. It's 26 MB in size and dwarfs the size of the rest of the conformance suite. Its inclusion makes this patch very difficult to review and will permanently bloat the size of this repository. Could you please try to strip it down? Just add something like a README-Khronos.txt in the top level directory stating what was kept and/or removed, and which version of the Closure library was included. It looks like possibly the only thing that might be needed is closure/goog/base.js, in which case we should definitely remove the rest of the library. Please keep all of their top-level files like LICENSE, AUTHORS, CONTRIBUTING, README.md, etc.
  - Some new tests are included with this patch, which is great, but that isn't noted in the commit message. I suspect that since this pull request is so large that that fact might have gotten lost.

Could you please address the above comments, squash commits in this pull request, and resubmit it? Please contact me via email if there are any issues with my feedback. Thanks.

kenrussell:
This was merged as #965 . Closing.

PR #956, opened 2015-04-27
update getFramebufferAttachmentParameter for WebGL 2

Richard-Yunchao:

### DIFF
--- a/sdk/tests/conformance/resources/gl-object-get-calls.js
+++ b/sdk/tests/conformance/resources/gl-object-get-calls.js
@@ -105,7 +105,6 @@ testInvalidArgument(
     }
 );
 
-
 debug("");
 debug("Test getFramebufferAttachmentParameter");
 var texture = gl.createTexture();
@@ -122,6 +121,11 @@ gl.bindTexture(gl.TEXTURE_2D, null);
 var framebuffer = gl.createFramebuffer();
 gl.bindFramebuffer(gl.FRAMEBUFFER, framebuffer);
 gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture, 0);
+var colorAttachmentsNum = 1;
+if (contextVersion > 1) {
+  colorAttachmentsNum = gl.getParameter(gl.MAX_COLOR_ATTACHMENTS);
+  gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0 + colorAttachmentsNum - 1, gl.TEXTURE_2D, texture, 0);
+}
 var renderbuffer = gl.createRenderbuffer();
 wtu.glErrorShouldBe(gl, gl.NO_ERROR);
 gl.bindRenderbuffer(gl.RENDERBUFFER, renderbuffer);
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/resources/gl-object-get-calls.js
```diff
@@ -170,17 +217,67 @@ var validArrayForFramebufferAttachment = new Array(
...
     "attachment",
-    validArrayForFramebufferAttachment,
+    validAttachmentsForFBAttachment,
     function(attachment) {
       return gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, attachment, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE);
     }
 );
+if (contextVersion > 1) {
+  // test default framebuffer
+  gl.bindFramebuffer(gl.FRAMEBUFFER, 0);
```
Richard-Yunchao: Accrodingly to ES 3.0 spec, getFramebufferAttachmentParameter can query the information of default framebuffer. But firefox nightly build failed to handle this.

Richard-Yunchao:
Tested in Firefox nightly build. Most cases of getFramebufferAttachmentParameter are passed, except that a couple of cases is not supported by Firefox, say FRAMEBUFFER_ATTACHMENT_TEXTURE_LAYER, query OBJECT_TYPE and OBJECT_NAME for DEPTH_STENCIL_ATTACHMENT, query OBJECT_TYPE for default framebuffer.

sdk/tests/conformance/resources/gl-object-get-calls.js
```diff
@@ -122,6 +121,11 @@ gl.bindTexture(gl.TEXTURE_2D, null);
 var framebuffer = gl.createFramebuffer();
 gl.bindFramebuffer(gl.FRAMEBUFFER, framebuffer);
 gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture, 0);
+var colorAttachmentsNum = 1;
+if (contextVersion > 1) {
+  colorAttachmentsNum = gl.getParameter(gl.MAX_COLOR_ATTACHMENTS);
+  gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0 + colorAttachmentsNum - 1, gl.TEXTURE_2D, texture, 0);
```
kenrussell: I don't understand the intent of this code -- here you're reattaching the texture to the attachment point COLOR_ATTACHMENT0 + MAX_COLOR_ATTACHMENTS - 1, but below you're verifying all of the attachment points between 0 and COLOR_ATTACHMENT0 + MAX_COLOR_ATTACHMENTS - 1, inclusive. Should this code here not attach the texture to all of those attachment points too?

kenrussell:
Thanks for this contribution. Something looks wrong with the code. Please comment and address as necessary.

sdk/tests/conformance/resources/gl-object-get-calls.js
Richard-Yunchao: Hi Ken, Thanks for your reviewing. The for loop below also tests these two attachment points, because the step in the loop is MAX_COLOR_ATTACHMENTS - 1, instead of 1. I add a comment before the for loop.

kenrussell: My point is that all of the framebuffer color attachment points in between are empty, so the calls like:

shouldBeNonZero('gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0 + ii, gl.FRAMEBUFFER_ATTACHMENT_RED_SIZE')

should fail. What am I missing?

Richard-Yunchao: The color attachment points in between are all empty. That's true. But The for loop doesn't test the color attachment points in between at all. So I think the code is OK. 

Well, I want to test more color attachment point(s) except the first one for WebGL 2, but testing all color attachment points seems to be unnecessary.  So this test choose the last one. So it binds texture to the first and the last color attachment points to the framebuffer, then test these two attachment points only.

kenrussell: OK, I see. Sorry for my confusion. I didn't read the loop header carefully enough. It isn't very easy to read. Please try to keep your code simple and easy to understand.

PR #958, opened 2015-04-27
Add gl-vertexattribipointer-offsets.html for WebGL 2.0

legendlee1314:


legendlee1314:
@kenrussell , PTAL if have time. Thank you!

kenrussell:
Thanks for this contribution. Despite the fact that it replicates code in sdk/tests/conformance/attribs/gl-vertexattribpointer-offsets.html, the test is short enough and there are enough differences in the middle of the code that I think it's clearer to duplicate the test.

PR #960, opened 2015-04-27
Add test for ANGLE initializer list bug.

jamiemadill:
The bug would happen with declarator lists where a first variable
is an array, and the second is not (or a different size). Already
fixed in ANGLE.


zhenyao:
LGTM

PR #961, opened 2015-04-27
Add test for global variable init order.

jamiemadill:
In some cases, ANGLE would miss initializing variables at global
scope in the correct order. This bug is already fixed in ANGLE.

### DIFF
--- a/sdk/tests/conformance/glsl/misc/shader-with-global-varying-assign.html
+++ b/sdk/tests/conformance/glsl/misc/shader-with-global-varying-assign.html
@@ -1,0 +1,89 @@
+
+
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/glsl/misc/shader-with-global-varying-assign.html
```diff
@@ -0,0 +1,89 @@
...
+<script id="vs" type="x-shader/x-vertex">
+attribute vec4 a_position;
+varying float v;
+void main() {
+  gl_Position = a_position;
+  v = 1.0;
+}
+</script>
+<script id="fs" type="x-shader/x-fragment">
+precision highp float;
```
zhenyao: This should be mediump because highp support is optional

Oletus:
Globals must be initialized with constant expressions according to the ESSL1.00 spec section 4.3. Some other current WebGL tests already violate this rule in subtle ways (see https://github.com/KhronosGroup/WebGL/pull/940) and due to compatibility worries we haven't made a final decision on how to address this yet. However, I don't think that we should be adding a test that thinks initializing a global variable with a varying should be okay. I'd rather add a test that would require these shaders to fail compilation, as this is a clearer violation of the spec than globals being initialized with other globals.

jamiemadill:
Hm, good catch Olli. I see the same language in the 3.00 and 3.10 specs. I think we should use your test to catch errors here and remove the corresponding code in ANGLE. It's unfortunate we let this slip for so long.

PR #962, opened 2015-04-27
Removed conformance2/core/misc-parameters.html

toji:
It's logic is duplicated in conformance2/state/gl-get-calls.html


zhenyao:
You need to update the 00_test_list.txt file

toji:
Derp. Thanks for pointing that out. Fixed.

zhenyao:
LGTM

PR #963, opened 2015-04-28
Add Jeff Gilbert to spec editor(s) list.

jdashg:
As decided by the working group on the April 16 conference call.

Looks like I stripped an EOL whitespace as well.


kenrussell:
Looks good!

toji:
LGTM! :D

PR #964, opened 2015-04-28
Set dEQP-based GLES3 tests to use WebGL 2 contexts by default

Oletus:
This removes the need to supply ?webglVersion=2 in the URL when running
one of these tests by itself. A function is added to WebGLTestUtils to
set the default context version, since it is a lot simpler than passing
the context version through the fairly complex test framework.

### DIFF
--- a/sdk/tests/conformance/resources/webgl-test-utils.js
+++ b/sdk/tests/conformance/resources/webgl-test-utils.js
@@ -1398,6 +1398,17 @@ var getUrlOptions = function() {
   return options;
 };
 
+var default3DContextVersion = 1;
+
+/**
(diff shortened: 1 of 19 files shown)


kenrussell:
I'd like to hold off merging this change for a couple of days. Mobica is in the process of landing a large refactoring of the ported dEQP tests to use the Closure library rather than RequireJS, and introducing merge conflicts will only delay that work.

kenrussell:
Mobica's refactoring has landed and this pull request still applies cleanly, so merging now. Thanks for this contribution.

@jasomobica and @marcorperez : FYI.

PR #965, opened 2015-04-28
Refactored and cleaned deqp tests

marcorperez:


kenrussell:
Excellent work. Thank you all at Mobica for doing such a thorough job with this refactoring. Being able to properly type check the code with Closure will ease development of these tests going forward.

I noticed a few tiny issues in the patch, for which I'll land a pull request shortly. Mobica can then correct the authorship on a couple of the files.

PR #969, opened 2015-04-30
Fix the gl_DepthRange vertex shader test.

jamiemadill:
This test was accidentally checking the reference shader for the
presence of the gl_DepthRange built-in uniform. Modify the check
to happen on the expected shader instead.


jamiemadill:
Also cleaned up some trailing whitespace. I'm not sure, but we might also want to set the minimum expected uniforms to 4.

kenrussell:
Argh. Thanks for catching this. I confirmed that the test was failing in Firefox and passes with your change. We need to fix a bug in Chrome before it'll work there.

kenrussell:
Also -- that "min_required" property means "we require a minimum of this many of the following valid values". It basically selects 2 of the 3 gl_DepthRange uniforms, since some drivers seem to compute one of them inside the shader from the two others.

PR #970, opened 2015-05-03
Improve error handling of UNPACK_COLORSPACE_CONVERSION_WEBGL.

RafaelCintron:
Internet Explorer does not support the UNPACK_COLORSPACE_CONVERSION_WEBGL
pixelStorei parameter. To inform developers the enum is unsupported, it flags an
INVALID_VALUE WebGL error.

Various tests in the conformance suite set UNPACK_COLORSPACE_CONVERSION_WEBGL to
gl.NONE and then fail to check the error code on the context. This causes subsequent
operations be misreported as failing when they, in fact, succeed. Much frustrating time
has been devoted to debugging root causes of conformance failures due to this issue.

This change:
- Checks and clears the WebGL error code when setting the
UNPACK_COLORSPACE_CONVERSION_WEBGL pixelStoreI parameter. Errors are now correctly
attributed to the source of the failure.
- Moves pixelStoreI calls to the beginning of the test to avoid needless and duplicate
calls to testPassed/testFailed.

### DIFF
--- a/sdk/tests/conformance/context/premultiplyalpha-test.html
+++ b/sdk/tests/conformance/context/premultiplyalpha-test.html
@@ -177,6 +177,9 @@ function doNextTest() {
     shouldBeTrue('gl.getContextAttributes().preserveDrawingBuffer');
 
     wtu.log(gl.getContextAttributes());
+    // not needed as it's the default
+    // gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, false);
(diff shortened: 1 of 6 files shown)


kenrussell:
While the setting of some of these pixel storage parameters may be redundant, I think it makes the code more maintainable to keep them all together in these various tests. Does the different approach in #971 seem acceptable to you?

RafaelCintron:
The approach in #971 is acceptable to me.

PR #972, opened 2015-05-05
Added texel fetch test

matthewn4444:
Tests accessing a texel outside the texture dimensions would return the
color (0, 0, 0, 0).

This test should also be expanded in the future to consider the "lod"
parameter of the texelFetch function.


kenrussell:
Very good. Thank you for this contribution.

PR #975, opened 2015-05-07
Relax dEQP compilation error checks (consistent with other tests)

Oletus:
Compilation failures are not guaranteed to be detected at compile time,
so dEQP tests should not require that. Other tests, like ones using
tests/conformance/resources/glsl-conformance-test.js are already built
to only check for link errors in case compilation errors are expected.

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsShaderLibraryCase.js
+++ b/sdk/tests/deqp/modules/shared/glsShaderLibraryCase.js
@@ -919,13 +919,14 @@ glsShaderLibraryCase.execute = function() {
 
     if (failReason != null) {
         // \todo [2010-06-07 petri] These should be handled in the test case?
-        bufferedLogToConsole('ERROR: ' + failReason);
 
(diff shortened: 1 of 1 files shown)


kenrussell:
Thanks for tracking this down. It should certainly be allowed to compile shaders at link time per the spec. Farther down in the file (~line 925) there's a check specifically covering this case.

In the original sources https://android.googlesource.com/platform/external/deqp/+/master/modules/glshared/glsShaderLibraryCase.cpp (line 725) it looks to me like this is a warning, but not a failure.

I think that check should be modified instead to stop reporting this case as a failure. Would you consider making this change there?

CC: @jasomobica @marcorperez

Oletus:
I amended the commit to be more in line with what the C++ version of dEQP does. I did change one thing compared to the C++ version - instead of printing "ERROR" together with the failReason when there's a warning, it just appends the failReason to the "Quality warning" line. This is a lot less confusing.

kenrussell:
Thanks. Looks good.

PR #976, opened 2015-05-07
Updated constants and properties test's enums to reflect WebGL 2.0

matthewn4444:
Moved texture swizzle, and program binary enums from the WebGL 2.0
enum list to the removed enum list.


toji:
LGTM

PR #977, opened 2015-05-08
Add test case to oes-texture-float-with-canvas to verify texture is actually backed by FLOATs.

nkronlage:
Adding a test case at the request of the Chromium team. For some texture sizes, Chrome was failing to allocate the texture as FLOAT and only allocating it as UNSIGNED_BYTE.

Details on the bug at:
https://code.google.com/p/chromium/issues/detail?id=484968

The test creates textures from various 2D canvas sizes and verifies that it can read out a pixel of the texture as a FLOAT.  

Some possible concerns with this approach:
 * Could an implementation just store the texture as UNSIGNED_BYTE and only convert to FLOAT during the readPixels call?
 * Does OES_texture_float imply you should be able to readPixels with FLOAT?  Safari does not seem to support this but Chrome, FireFox and IE do support it.

Please let me know if there is a better way to test this. For example, in the original test case, I rendered an almost transparent color repeated to the texture. The FLOAT version would keep the transparency and eventually become visible but the UNSIGNED_BYTE one would stay transparent since it didn't have enough precision.

### DIFF
--- a/sdk/tests/conformance/extensions/oes-texture-float-with-canvas.html
+++ b/sdk/tests/conformance/extensions/oes-texture-float-with-canvas.html
@@ -42,7 +42,53 @@ function testPrologue(gl) {
     }
 
     testPassed("Successfully enabled OES_texture_float extension");
+
+    verifyTexturesAreFloat(gl);
     return true;
+}
+
+function verifyTexturesAreFloat(gl) {
+    var canvas = document.createElement('canvas');
+    var ctx = canvas.getContext('2d');
+
+    var framebuffer = gl.createFramebuffer();
+    gl.bindFramebuffer(gl.FRAMEBUFFER, framebuffer);
+
+    var texture = gl.createTexture();
+    gl.activeTexture(gl.TEXTURE0);
+    gl.bindTexture(gl.TEXTURE_2D, texture);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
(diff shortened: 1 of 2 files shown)


Oletus:
It's not mandatory to support float framebuffer attachments (render-to-float) if just OES_texture_float is supported.

So, another better way to test this might be to replace some of the texture contents with texSubImage2D after it has been created - set some of the contents to some suitable float value that couldn't be represented by the fixed point 0.0 - 1.0 range, like 2.0, and then sample the texture it in a shader to verify that it is correct. This way render-to-float functionality would not be required by the test.

ds-hwang:
@Oletus that's good suggestion. Currently chrome has bug that texImage2d(canvas, FLOAT) creates UNSIGNED_BYTE texture slightly. so following pseudo code probably check this issue.
```
texImage2D(TEXTURE_2D, FLOAT, empty_array);
texSubImage2D(TEXTURE_2D, FLOAT, canvas);
```

kenrussell:
@nkronlage : agree with @Oletus 's suggestion. You might look at oes-texture-float.html, which does this sampling trick in order to ensure that the texture was actually allocated with floating-point components. I think you should continue to allocate the texture from a canvas using FLOAT components, but then you could upload / modify its contents using texSubImage2D -- if that will work and will catch the bug.

conformance/extensions/oes-texture-float.html verifies the behavior of IMPLEMENTATION_COLOR_READ_FORMAT and IMPLEMENTATION_COLOR_READ_TYPE, which define the auxiliary readback format supported by readPixels. This was unfortunately added late to the WebGL spec and it is possible that Safari isn't supporting it yet. (CC @grorg )

Also, I'd like to suggest that you add the new test case to tex-image-and-sub-image-2d-with-canvas.js, which is included from oes-texture-float-with-canvas.html. You can test the pixelFormat and pixelType and only run the new test if they're "RGBA" and "FLOAT". The new test can be a separate function from runTexImageTest and be run in the same promise chain that's created at the end of the function.

PR #980, opened 2015-05-08
Add test for arrays being returned inside complex expressions

Oletus:
HLSL doesn't natively support arrays as return values. To translate
shaders to HLSL, expressions where arrays are being returned need to be
converted into a form where arrays are not being returned. There are some
cases where implementing this is somewhat tricky, and should be tested.

All the new tests already pass on Chrome's current WebGL 2 prototype on
Linux/NVIDIA.


zhenyao:
LGTM

PR #981, opened 2015-05-08
Add a test for ternary operator affected by earlier operand in sequence

Oletus:
This reveals a bug in ANGLE's HLSL translation.

The test passes in Chrome on Linux/NVIDIA and also in current version of
IE.


kenrussell:
Very nice. Thanks for this contribution.

PR #983, opened 2015-05-09
Add test case to oes-texture-float-with-canvas to verify texture is actually backed by FLOATs.

nkronlage:
Redoing the change discussed at:

https://github.com/KhronosGroup/WebGL/pull/977#issuecomment-100169188


kenrussell:
Thanks. Nice, simple addition to the test. Confirmed that it passes on Firefox and currently fails on Chrome due to https://code.google.com/p/chromium/issues/detail?id=484968 . Merging.

ds-hwang:
@kenrussell how do you test? the if statement is never executed.

kenrussell:
@ds-hwang the tests sdk/tests/conformance/extensions/oes-texture-float-with-canvas.html and sdk/tests/conformance/extensions/oes-texture-half-float-with-canvas.html cover these cases.

PR #986, opened 2015-05-12
Update gl-get-tex-parameter.html for WebGL 2.0

legendlee1314:


legendlee1314:
@kenrussell , PTAL. Thanks!

kenrussell:
Very nice. Thank you for this revised test.

PR #987, opened 2015-05-12
Fix global initializer containing an uniform in conformance tests

Oletus:
ESSL 1.00 section 4.3 says that global initializers need to be constant
expressions. It is still being discussed how strictly this rule should
apply to WebGL, but initializing globals with uniforms is clearly not
allowed by the spec, so fix the conformance test doing that.

The fix is backported all the way to 1.0.1 conformance suite.

### DIFF
--- a/conformance-suites/1.0.1/conformance/glsl/misc/shader-with-for-loop.html
+++ b/conformance-suites/1.0.1/conformance/glsl/misc/shader-with-for-loop.html
@@ -26,7 +26,7 @@ uniform float time;
 uniform vec2 resolution;
 
 // Saw-tooth function that is synced with the demo music (128bpm)
-float gBeat=fract(time*3.2/3.);
+float gBeat;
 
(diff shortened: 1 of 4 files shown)


Oletus:
Disallowing initializing globals with uniforms does create some compatibility issues. I could find one example of breakage relatively quickly: http://threejs.org/examples/#webgl_shaders_sky . Still, it might be better to clean this test up either way since I imagine this is not what it's supposed to test.

kenrussell:
OK, let's make this change to the test.

The restriction in the GLSL ES spec (which I've confirmed also exists in version 3.00.4 of that spec) seems arbitrary. As you've done here, it should be possible to pull globals' non-constant initializer expressions into the top of main(). ANGLE's shader translator could plausibly do that, and the WebGL spec could explicitly allow non-constant initializer expressions.

kenrussell:
Also, thanks for this contribution.

PR #988, opened 2015-05-13
Clarified that renderbufferStorage accepts new formats from ES 3.0.

kenrussell:


kenrussell:
@jdashg : please review. This is a request from Mobica, where it wasn't clear that WebGL 2.0 supported all of the new formats (since some prototypes weren't properly accepting them yet).

The tables could be replicated from the ES 3.0 manual page if desired.

jdashg:
LGTM

PR #989, opened 2015-05-13
Initialize variable expandoValue before using it

qkmiao:
@toji please help to review, thanks!

### DIFF
--- a/sdk/tests/conformance/misc/expando-loss.html
+++ b/sdk/tests/conformance/misc/expando-loss.html
@@ -44,12 +44,6 @@ var wtu = WebGLTestUtils;
 var canvas = document.getElementById("canvas");
 var gl = wtu.create3DContext(canvas, {antialias: false});
 
-// Run tests
-testBasicBindings();
-testProgramsAndShaders();
-testVertexAttributeBuffers();
-testFrameBufferAttachments();
-
 // Helpers that set expandos and verify they are set to the correct value.
 var expandoValue = "WebGL is awesome!"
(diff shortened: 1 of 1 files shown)


RafaelCintron:
@qkmiao , I am confused. ExpandoValue is only used in setTestExpandos and verifyTestExpandos and it is initialized before those two functions. It does not need to be initialized earlier and, in my opinion, the code reads better if it is initialized close to where it is used.

kenrussell:
Agreed -- this pull request does not make sense.

qkmiao:
@RafaelCintron @kenrussell , thanks for review. ExpandoValue is used when calling testBasicBindings() (which will call setTestExpandos and verifyTestExpandos ). So I think it should be initialized before testBasicBindings.

RafaelCintron:
Apologies @qkmiao , I now see your point.  Since testBasicBindings runs before expandoValue is set to "Webgl is awesome!", the expandos set on the object end up with the value of undefined, instead of the string. I think we should merge this pull request.

kenrussell:
@qkmiao , sorry for being hasty. Thanks for your proposed fix.

The control flow in this file is confusing. Could you please fix this instead by pulling the "// Run tests" section down to the bottom, just above "var successfullyParsed = true;"? Please update your pull request and comment here when done. Thanks.

qkmiao:
@kenrussell , the pull request has been updated. Please take another look.

kenrussell:
@qkmiao , thank you for this fix and sorry for misunderstanding it the first time. Looks good.

PR #991, opened 2015-05-13
Fix indentation in tex-image-and-sub-image-2d-with-canvas.js

ds-hwang:
PR #984 used tab instead of space.


ds-hwang:
@kenrussell could you review? I'm not sure webgl project is strict to indentation like chromium project.

toji:
It's not as strict, but consistency is always appreciated. Thanks! :)

ds-hwang:
thx for quick response  :+1:

PR #992, opened 2015-05-14
Removed test for BUFFER_MAPPED, BUFFER_MAP_POINTER enums.

toji:
Added test for MAX_CLIENT_WAIT_TIMEOUT_WEBGL. This syncs the test up with the spec.


kenrussell:
Excellent. Thanks for fixing this.

PR #993, opened 2015-05-15
Update expando test for WebGL 2.0

qkmiao:
@kenrussell , Please help to review this, thanks.

### DIFF
--- a/sdk/tests/conformance/misc/expando-loss.html
+++ b/sdk/tests/conformance/misc/expando-loss.html
@@ -61,7 +61,7 @@ function testBasicBindings() {
     debug('Basic Bindings');
 
     // Test data that describes how to create, bind, and retrieve an object off of the context
-    var glProt = WebGLRenderingContext.prototype;
+    var glProt = Object.getPrototypeOf(gl);
     var simpleData = [
         {
             creationFn: glProt.createTexture,


sdk/tests/conformance/resources/expando-loss.js
```diff
@@ -0,0 +1,225 @@
...
+    assertMsg(instance.expando2 && instance.expando2.subvalue === expandoValue, msg + ": Expect subobject expando to survive despite GC.");
+}
+
+// Tests that we don't get expando loss for bound resources where the
+// only remaining reference is internal to WebGL
+function testBasicBindings() {
+    debug('Basic Bindings');
+
+    // Test data that describes how to create, bind, and retrieve an object off of the context
+    var glProt = contextVersion > 1 ? WebGL2RenderingContext.prototype : WebGLRenderingContext.prototype;
```
kenrussell: Since this test contains no new logic for new object types, the better way to express this would be:

  var glProt = Object.getPrototypeOf(gl);

This would avoid splitting the test into two. Could you make this change and squash commits so that this pull request essentially becomes a one-line change? Thanks.

qkmiao:
@kenrussell , thanks for review. I updated the pull request as your comments.

kenrussell:
Thanks, looks good. Much simpler.

PR #994, opened 2015-05-18
Add shader logging to some tests

Oletus:
Also fix sampler-struct-function-arg.html test so that it fails rather
than runs into an exception if compilation fails.


kenrussell:
Thanks, looks good.

PR #997, opened 2015-05-19
Update tex-input-validation.html for WebGL 2.0

legendlee1314:


legendlee1314:
@kenrussell , please review this. Thanks!
This patch is just to refactor and differentiate test cases of `texParameter` & `GetTexParameter` both for WebGL and WebGL2. It is better to add more validation test cases like `tex[Sub]Image3D` and etc. But I'd like to implement those in the later patch. What do you think about this?

kenrussell:
It's fine to do this refactoring and add more test cases in the future. Thanks for this contribution.

PR #998, opened 2015-05-19
Add unfolded version of conditional discard test

Oletus:
This can expose incomplete workarounds to the same underlying bug the
other two tests in the file are testing.


zhenyao:
LGTM

PR #999, opened 2015-05-19
Add test for ANGLE short-circuiting logic unfolding issue

Oletus:
TOT ANGLE hoists statements that belong inside an if statement outside
the if statement if the if statement doesn't have braces. Test for this.


kenrussell:
Great catch. Thanks for this new test.

PR #1000, opened 2015-05-19
Added tests for draft EXT_disjoint_timer_query.

Morlunk:
The draft extension EXT_disjoint_timer_query is currently being implemented in Firefox- it would be valuable to include some simple sanity checks for it as part of the WebGL conformance tests.

Thanks!


kenrussell:
This is excellent. Thank you for contributing these tests. As Firefox's implementation of this (draft) extension progresses, it would be great if you'd add some tests which use requestAnimationFrame to do multiple frames of rendering, and actually test that the queries' results become available.

PR #1001, opened 2015-05-21
Added test for checking sequence operators returns constant values for WebGL 1.0

matthewn4444:
…

### DIFF
--- a/sdk/tests/conformance/glsl/misc/00_test_list.txt
+++ b/sdk/tests/conformance/glsl/misc/00_test_list.txt
@@ -14,7 +14,7 @@ glsl-long-variable-names.html
 non-ascii-comments.vert.html
 non-ascii.vert.html
 --min-version 1.0.2 re-compile-re-link.html
-sequence-operator-returns-constant.html
+--min-version 1.0.4 sequence-operator-returns-constant.html
 --min-version 1.0.3 shader-precision-format-obeyed.html
 --min-version 1.0.3 shader-struct-scope.html
 --min-version 1.0.2 shader-uniform-packing-restrictions.html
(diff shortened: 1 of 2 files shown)


matthewn4444:
Addressing the RHS for the const var assignment, I think from the spec here: https://www.opengl.org/registry/doc/GLSLangSpec.Full.1.20.8.pdf

4.3.2 Const says: __Initializers for const declarations must be constant expressions, as defined in Section 4.3.3 “Constant Expressions.”__

4.3.3 Constant Expressions

A constant expression is one of
* a literal value (e.g., 5 or true)
* a global or local variable qualified as const (i.e. not including function parameters)
* __an expression formed by an operator on operands that are all constant expressions, including getting an element or length of a constant array, or a field of a constant structure, or components of a constant vector.__
* a constructor whose arguments are all constant expressions
* a built-in function call whose arguments are all constant expressions, with the exception of the texture
lookup functions, the noise functions, and ftransform. The built-in functions dFdx, dFdy, and
fwidth must return 0 when evaluated inside an initializer with an argument that is a constant
expression

I am new to the spec and did not read all of it so I am not sure if this is what you want. If it is, then I can add this to the code as a comment or in the commit message.

jdashg:
Yep, you're correct. `const` is compile-time in GLSL and ESSL. Leaving a small comment wouldn't hurt.

Regardless, this should be --min-version 1.0.4 before we can merge it.

Oletus:
jdashg: Did you notice that the pull request was updated?

matthewn4444: It's good to send a comment to a pull request once you update it, I don't think there is an automatic notification if you simply push a new commit.

matthewn4444:
I added the version number to the comment

matthewn4444:
Sorry about that, I updated the comment again. Let me know if it is still wrong.

jdashg:
Thanks!

PR #1002, opened 2015-05-21
Fbo tests and reference rasterizer.

jasomobica:


sdk/tests/deqp/framework/common/tcuFuzzyImageCompare.js
```diff
@@ -125,7 +147,9 @@ var tcuTextureUtil = framework.common.tcuTextureUtil;
         var start = src.getRowPitch() * y + x * NumChannels;
         var end = start + NumChannels;
         /** @type {goog.TypedArray} */ var ptr = src.getDataPtr().subarray(start, end);
-        /** @type {Uint32Array} */ var v = new Uint32Array(ptr); //Small buffer copy
+        /** @type {Uint32Array} */ var v = new Uint32Array(ptr.buffer).subarray(
+            start * ptr.BYTES_PER_ELEMENT / 4,
+            start * ptr.BYTES_PER_ELEMENT / 4);
```
kenrussell: This is a little weird. There's no point in allocating "ptr" in this case. You might as well start with src.getDataPtr().buffer.

sdk/tests/deqp/framework/common/tcuRGBA.js
```diff
@@ -37,7 +37,7 @@ var deMath = framework.delibs.debase.deMath;
      * @constructor
      */
     tcuRGBA.RGBA = function() {
-        /** @type {Uint32Array} */ this.m_value = new Uint32Array(4);
+        /** @type {goog.NumberArray} */ this.m_value;
```
kenrussell: Why not just take the goog.NumberArray as argument to the constructor and assign this.m_value from it?

sdk/tests/deqp/framework/opengl/gluShaderUtil.js
```diff
@@ -44,6 +44,11 @@ gluShaderUtil.GLSLVersion = {
 gluShaderUtil.getGLSLVersion = function(gl) {
     var glslversion = gl.getParameter(gl.SHADING_LANGUAGE_VERSION);
 
+    // TODO: Versions are not yet well implemented... Firefox returns GLSL ES 1.0 in some cases,
```
kenrussell: Is this TODO addressed at this point in the various browsers? If not, let's please file bugs against them. Otherwise this will likely get lost.

jasomobica: Currently both mozilla and chrome report wrong GLSL versions (Mozilla reports GLSL ES 1.0, chrome reports GLSL ES 2.0). We ignore this error when testing internally. The error is not ignored in the released version of the test. Try running sdk/tests/deqp/functional/gles3/uniformbuffers.html. This TODO reminds us that we should turn the version check back on before creating a pull request. I can file bugs against the browsers, though.

kenrussell:
Thanks very much for this tremendous work. I only was able to skim through the code and didn't see any obvious problems. I'm sure the browsers' WebGL 2.0 implementations have many bugs preventing these new tests from running correctly, so we'll appreciate your continued work and guidance to finding the areas that appear broken. Merging this now.

PR #1004, opened 2015-05-25
Add tests for arrays as return values

Oletus:
The added test covers some simpler cases compared to
array-in-complex-expression.html


kenrussell:
LGTM after the fact.

PR #1005, opened 2015-05-27
Expand buffer-copying-contents.html

zhenyao:
Now the test thoroughly covers all buffer copying pairs.


kenrussell:
Excellent. Thank you for improving this test.

PR #1006, opened 2015-05-27
Convert WebGL 2 shader tests to use GLSLConformanceTester

Oletus:
This includes adding default ESSL3 shaders to the GLSLConformanceTester and converting the test files to use that.


zhenyao:
Great refactoring.  Thanks.

kenrussell:
LGTM after the fact. Thanks for this nice cleanup.

PR #1007, opened 2015-05-27
Move tests from conformance2/core to various folders and remove core.

zhenyao:
It seems to me the existence of core caused tests being dumped there where they should belong to a more meaningful sub-folder.


toji:
LGTM. I originally created the core/ folder as a place to put tests that were ported from the WebGL1 extensions/ tests. Unfortunately the generic name enabled it to be a dumping ground of sorts. I'm very much in favor of the better structure this change provides.

kenrussell:
LGTM too FWIW.

PR #1009, opened 2015-05-27
Added partial transform feedback API test. Focuses on binding.

toji:


toji:
Updated commit based on @jdashg's comments (which disappeared in a puff of smoke when I uploaded the amended commit...)

kenrussell:
Thanks, this looks good. It would be helpful if you could look at sdk/tests/deqp/functional/gles3/transformfeedback.html and see whether there's any overlap between these tests.

kenrussell:
I'll leave this to @jdashg to merge since it sounds like he was already reviewing it.

jdashg:
Cool, thanks.

PR #1011, opened 2015-05-28
move all files under sdk/tests/conformance/resources/ to sdk/tests/resources/

Richard-Yunchao:
Please take a look when you have time. This is discussed in #937


kenrussell:
Thank you for doing this renaming. I tested it locally and it looks good. I think it's a good direction so the resources can be shared between the WebGL 1.0 and 2.0 suites. Merging now.

Oletus:
This seems to have broken sdk/tests/resources/tex-image-and-sub-image-2d-with-image.js and sdk/tests/resources/tex-image-and-sub-image-2d-with-video.js. They're trying to load red-green.* files from an absolute path instead of a relative path.

Oletus:
Same issue also in sdk/tests/resources/gl-object-get-calls.js

zhenyao:
Now we are refactoring more tests into js, it seems ill-organized if we put all the js tests into resources and the originally well-organized sub folders only contain the containing html files.

What about putting such js files at the same folder with their container html files?

zhenyao:
Per discussion with kbr, we are moving js files out of resources/ and into js/, and under js/ we create utils/ and tests/

zhenyao:
I am going to prepare a pull request for this, which will also fix the issues Olli pointed out.

Richard-Yunchao:
@zhenyao , totally agree with you. This is a good idea. I also want to put all js files into a standalone file folder. Per the issue @Oletus pointed out, IIRC, the previous code also used an absolute path, so I think it was OK. You know, I use "python -m SimpleHTTPRequest" under sdk/tests in my local machine, so no test failed with this patch. If it does cause errors, please help to fix them. Thanks a lot!

PR #1012, opened 2015-05-28
Remove NUM_SAMPLE_COUNTS from WebGL 2.

zhenyao:
Since SAMPLES will just return an Int32Array, NUM_SAMPLE_COUNTS is useless;
for other parts of the WebGL spec, we always remove such redundant enums.


kenrussell:
Good catch. Thanks for fixing this. (In the future please feel free to update the last modified date of the spec at the top.)

PR #1014, opened 2015-05-28
Compare functions and transform feedback bugfixes.

jasomobica:
Fixed fuzzy and bilinear compare functions.
Fixed creating
 programs with transform feedback.
 Added automatic GL error checking.
 Improved formatting.


sdk/tests/deqp/framework/opengl/gluShaderUtil.js
```diff
@@ -37,6 +37,21 @@ gluShaderUtil.GLSLVersion = {
...
 /**
+ * glsUniformBlockCase.isGLSLVersionSupported
+ * @param {WebGL2RenderingContext|WebGLRenderingContextBase} ctx
+ * @param {gluShaderUtil.GLSLVersion} version
+ * @return {boolean}
+ */
+ gluShaderUtil.isGLSLVersionSupported = function(ctx, version) {
+    if (ctx instanceof WebGL2RenderingContext)
+        return version <= gluShaderUtil.GLSLVersion.V300_ES;
+    else if (ctx instanceof WebGLRenderingContextBase)
```
kenrussell: This should not work in any browser. WebGLRenderingContextBase has the [NoInterfaceObject] extended attribute which means that its type is not visible. This should be WebGLRenderingContext instead.

sdk/tests/deqp/functional/gles3/es3fTransformFeedbackTests.js
```diff
@@ -939,8 +938,6 @@ goog.scope(function() {
 
         this.m_transformFeedback = gl.createTransformFeedback();
 
-        es3fTransformFeedbackTests.GLU_EXPECT_NO_ERROR(gl, gl.getError(), 'es3fTransformFeedbackTests.init');
```
kenrussell: Why are the assertions around glGetError returning NO_ERROR being removed? Is it because browsers are actually generating errors? If so, then we should fix the browsers and leave the assertions in place.

sdk/tests/deqp/modules/shared/glsFboCompletenessTests.js
```diff
@@ -16,26 +15,25 @@ goog.scope(function() {
 
     glsFboCompletenessTests.initGlDependents = function(gl) {
         if (!(gl = gl || window.gl)) throw new Error('Invalid gl object');
-        
-        
+
         // The following extensions are applicable both to ES2 and ES3.
         /**
-         * GL_OES_depth_texture
+         * gl.OES_depth_texture
```
kenrussell: These comment renames look wrong. WebGL drops the GL_ prefix from extension names but doesn't define them as properties on the WebGLRenderingContext.

sdk/tests/deqp/modules/shared/glsFboCompletenessTests.js
```diff
@@ -213,120 +211,120 @@ goog.scope(function() {
...
-        
+
         /**
          * s_esExtFormats
          * @type {Array<glsFboUtil.FormatExtEntry>}
          */
         glsFboCompletenessTests.s_esExtFormats = [
             new glsFboUtil.FormatExtEntry(
-                'GL_OES_depth_texture',
+                'gl.OES_depth_texture',
```
kenrussell: These extension renames look incorrect. If this string is intended to refer to the extension name then it should be 'OES_depth_texture' rather than 'gl.OES_depth_texture'.

sdk/tests/deqp/modules/shared/glsVertexArrayTests.js
```diff
@@ -553,14 +541,12 @@ goog.scope(function() {
 
         if (this.m_storage == glsVertexArrayTests.deArray.Storage.BUFFER) {
             this.m_ctx.bindBuffer(glsVertexArrayTests.ContextArray.targetToGL(target), this.m_glBuffer);
-            glsVertexArrayTests.GLU_EXPECT_NO_ERROR(this.m_ctx.getError(), "gl.bindBuffer()");
```
kenrussell: Similar question as above regarding the removal of the error checks here. If the browsers are generating errors here, let's fix them rather than removing these checks.

kenrussell:
Thank you for this excellent work and these fixes. A few relatively minor questions about correctness in a few areas. Could you please reply before we merge this? Thanks.

sdk/tests/deqp/modules/shared/glsFboCompletenessTests.js
jasomobica: Well spotted. This code needs more thorough review. I'll add this to my TODO list.

sdk/tests/deqp/framework/opengl/gluShaderUtil.js
jasomobica: You're right, this function is broken. It is a new function and it isn't used anywhere (yet).  We'll fix this soon.

sdk/tests/deqp/functional/gles3/es3fTransformFeedbackTests.js
jasomobica: Error checking is now implemented in a context wrapper, WebGLTestUtils.create3DContextWithWrapperThatThrowsOnGLError. This makes error checking in the tests obsolete, so we can clean it up and make the code more readable.

kenrussell:
Thanks for your replies. Let's merge this and we can fix these issues in a forthcoming commit. LGTM

PR #1016, opened 2015-05-29
fix a small bug, for resources has been moved out of conformance file folder

Richard-Yunchao:
PTAL when you have time. A patch is submitted before #1011 , but merged after #1011 without code rebase. I think that patch is the first and also the last one.


kenrussell:
Thanks. LGTM

PR #1017, opened 2015-05-29
Add multisampled-renderbuffer-initialization.html

zhenyao:
Tested with local CLs applied to Chromium ToT build.


kenrussell:
Excellent test. Thanks for putting this together. LGTM

PR #1018, opened 2015-05-29
Added test for ternary operations on arrays for WebGL 1.0

matthewn4444:


kenrussell:
It looks to me like this also requires a --max-version 1.9.9 directive because the restriction on the ternary operator that it can't operate on arrays was removed in Section 5.9 "Expressions" of the GLSL ES 3.00.4 spec. Correct?

matthewn4444:
Spec says "The second and third expressions can be any type, as long their types match." I think so, I can add the max version, but to be sure should we get another person to make sure, my knowledge of the spec is rusty.

zhenyao:
The test itself looks good to me, but please add the max-version as kbr pointed out.

matthewn4444:
I think the directive is correctly added.

kenrussell:
@Oletus could you comment too?

kenrussell:
Let's merge this now and if Olli has comments we can revise it. Thanks for the test.

Oletus:
I don't think that the --max-version directive should be there. ESSL 1.00 shader compilation should work according to the ESSL 1.00 spec regardless of if the shader is compiled for WebGL 1.0 or WebGL 2.0.

zhenyao:
Olli is right.  I created a pull request to fix this and add the glsl3 version of the test.

https://github.com/zhenyao/WebGL/commit/89e568bd9887690a4b4130aef02f22f92c9493d0

PR #1020, opened 2015-06-01
Add test cases in tex-input-validation.js for WebGL 2.0

legendlee1314:
1. Adjust some functions location to make clearer.
2. Add test cases of texImage3D and texSubImage3D.

### DIFF
--- a/sdk/tests/js/tests/tex-input-validation.js
+++ b/sdk/tests/js/tests/tex-input-validation.js
@@ -42,6 +42,22 @@ function enumToString(value) {
   return wtu.glEnumToString(gl, value);
 }
 
+function testTexParameter(testCase) {
+  var msg = "paramName: " + enumToString(testCase.pname);
+  error = testCase.expectedError;
+  gl.texParameteri(testCase.target, testCase.pname, testCase.param);
+  wtu.glErrorShouldBe(gl, error, msg);
+  gl.texParameterf(testCase.target, testCase.pname, testCase.param);
+  wtu.glErrorShouldBe(gl, error, msg);
+}
+
+function testGetTexParameter(testCase) {
+  var msg = "paramName: " + enumToString(testCase.pname);
+  error = testCase.expectedError;
+  gl.getTexParameter(testCase.target, testCase.pname);
+  wtu.glErrorShouldBe(gl, error, msg);
+}
+
 function testTexImage2D(testCase) {
   var level = 0;
   var width = 16;
@@ -69,22 +85,6 @@ function testTexSubImage2D(testCase) {
 
   gl.texSubImage2D(testCase.target, level, xoffset, yoffset, width, height, testCase.format, testCase.type, array);
   error = testCase.expectedError;
-  wtu.glErrorShouldBe(gl, error, msg);
-}
-
-function testTexParameter(testCase) {
-  var msg = "paramName: " + enumToString(testCase.pname);
-  error = testCase.expectedError;
-  gl.texParameteri(testCase.target, testCase.pname, testCase.param);
-  wtu.glErrorShouldBe(gl, error, msg);
-  gl.texParameterf(testCase.target, testCase.pname, testCase.param);
-  wtu.glErrorShouldBe(gl, error, msg);
-}
-
-function testGetTexParameter(testCase) {
-  var msg = "paramName: " + enumToString(testCase.pname);
-  error = testCase.expectedError;
-  gl.getTexParameter(testCase.target, testCase.pname);
   wtu.glErrorShouldBe(gl, error, msg);
 }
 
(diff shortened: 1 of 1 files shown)


legendlee1314:
@kenrussell , please review this patch. Thank you!

sdk/tests/resources/tex-input-validation.js
```diff
@@ -394,4 +429,98 @@ for (var ii = 0; ii < testCases.length; ++ii) {
...
 }
 
+if (contextVersion > 1) {
+// Create new texture for testing api of WebGL 2.0.
+shouldBeNonNull("tex = gl.createTexture()");
+gl.bindTexture(gl.TEXTURE_3D, tex);
+wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+debug("");
+debug("Checking TexImage3D: a set of inputs that are valid in GL but invalid in WebGL");
```
kenrussell: Some of these tests look like they're invalid in OpenGL ES as well as WebGL.

sdk/tests/resources/tex-input-validation.js
```diff
@@ -394,4 +429,98 @@ for (var ii = 0; ii < testCases.length; ++ii) {
...
+    border: 0,
+    format: gl.RGBA,
+    type: gl.UNSIGNED_BYTE,
+    expectedError: gl.INVALID_ENUM },
+  { target: gl.TEXTURE_3D,
+    internalFormat: gl.RED,
+    border: 0,
+    format: gl.RED,
+    type: gl.UNSIGNED_BYTE,
+    expectedError: gl.INVALID_ENUM },
```
kenrussell: This looks like it fails because the internal format should be gl.RED8, and it looks to me like it should generate INVALID_OPERATION, not INVALID_ENUM. It also looks like it should fail in OpenGL ES as well as WebGL.

sdk/tests/resources/tex-input-validation.js
```diff
@@ -394,4 +429,98 @@ for (var ii = 0; ii < testCases.length; ++ii) {
...
+    border: 0,
+    format: gl.RGBA,
+    type: gl.UNSIGNED_BYTE,
+    expectedError: gl.INVALID_ENUM },
+  { target: gl.TEXTURE_2D,
+    internalFormat: gl.RGBA,
+    border: 0,
+    format: gl.RGBA,
+    type: gl.UNSIGNED_BYTE,
+    expectedError: gl.INVALID_ENUM },
```
kenrussell: This looks like it should fail in OpenGL ES as well as WebGL.

sdk/tests/resources/tex-input-validation.js
```diff
@@ -394,4 +429,98 @@ for (var ii = 0; ii < testCases.length; ++ii) {
...
+    border: 0,
+    format: gl.RED,
+    type: gl.UNSIGNED_BYTE,
+    expectedError: gl.INVALID_ENUM },
+  { target: gl.TEXTURE_3D,
+    internalFormat: gl.RGBA,
+    border: 1,
+    format: gl.RGBA,
+    type: gl.UNSIGNED_BYTE,
+    expectedError: gl.INVALID_VALUE },
```
kenrussell: The "border==0" restriction is in OpenGL ES as well as WebGL.

sdk/tests/resources/tex-input-validation.js
```diff
@@ -394,4 +429,98 @@ for (var ii = 0; ii < testCases.length; ++ii) {
...
+    border: 0,
+    format: gl.RGB,
+    type: gl.UNSIGNED_BYTE,
+    expectedError: gl.INVALID_OPERATION },
+  { target: gl.TEXTURE_3D,
+    internalFormat: gl.RGBA,
+    border: 0,
+    format: gl.RGBA,
+    type: gl.BYTE,
+    expectedError: gl.INVALID_ENUM },
```
kenrussell: It looks to me like this now fails not because gl.BYTE is an invalid type, but because the internal format would need to be gl.RGBA8_SNORM, which would result in an INVALID_OPERATION error.

kenrussell:
Thanks for this contribution, but it looks to me like there are several errors in the new test cases. I didn't go through all of them. Please double-check your work and the expected results. Hoping that @zhenyao can also help review this one once it's updated.

legendlee1314:
@kenrussell : Thanks for review! I have updated this patch according to the differences between [OpenGL doc sdk](https://www.opengl.org/sdk/docs/man/) and [OpenGLES doc sdk](https://www.khronos.org/opengles/sdk/docs/man3/).

sdk/tests/resources/tex-input-validation.js
```diff
@@ -260,12 +295,6 @@ var testCases = [
     expectedError: gl.INVALID_ENUM },
   { target: gl.TEXTURE_2D,
     internalFormat: gl.RGBA,
-    border: 1,
```
legendlee1314: Since `border` must be value 0 in GL, it is also wrong here so remove this case.

sdk/tests/resources/tex-input-validation.js
legendlee1314: Yes, this should aim to find out valid parameters in GL but invalid in GLES. In my opinion, the rules are the same for tex[Sub]Image3D both in GLES and WebGL. So those invalid in GLES are also invalid in WebGL.

legendlee1314:
@zhenyao , could you please help review? Thank you!

sdk/tests/resources/tex-input-validation.js
kenrussell: Upon further thought, please leave this test case in. The OpenGL compatibility profile allows border to be 1, so there should be at least one test in this file ensuring that WebGL doesn't allow that to be passed.

sdk/tests/resources/tex-input-validation.js
kenrussell: Correct.

kenrussell:
Thanks for this contribution. Please put one of the test cases back in as mentioned above, then LGTM.

legendlee1314:
@kenrussell , have put the case back now.

kenrussell:
Very good. Thank you for this contribution. LGTM.

kenrussell:
Actually, could you please rebase this to top of tree? It won't automatically merge. Please comment here once you've done that. Thanks.

legendlee1314:
Thanks! The patch is rebased.

kenrussell:
Thanks. Merging.

PR #1021, opened 2015-06-01
fix bug in conformance test in framebuffer-test.html under WebGL 2 rendering context

Richard-Yunchao:

### DIFF
--- a/sdk/tests/conformance2/renderbuffers/framebuffer-test.html
+++ b/sdk/tests/conformance2/renderbuffers/framebuffer-test.html
@@ -1,7 +1,7 @@
 <!--
 
 /*
-** Copyright (c) 2012 The Khronos Group Inc.
+** Copyright (c) 2015 The Khronos Group Inc.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
 ** copy of this software and/or associated documentation files (the
@@ -29,7 +29,7 @@
 <html>
 <head>
 <meta charset="utf-8">
-<title>WebGL Framebuffer Test</title>
+<title>WebGL Framebuffer Test Against WebGL 2</title>
 <link rel="stylesheet" href="../../resources/js-test-style.css"/>
 <script src="../../resources/js-test-pre.js"></script>
 <script src="../../resources/webgl-test-utils.js"></script>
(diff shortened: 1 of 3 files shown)


Richard-Yunchao:
Run this updated test case in Chromium, a couple of cases failed too. Because there are small bugs in Chromium. I will fix them later.

sdk/tests/resources/framebuffer-test.js
```diff
@@ -0,0 +1,212 @@
...
+
+  debug("");
+  debug("Checking framebuffer/renderbuffer stuff.");
+
+  var value = gl.getFramebufferAttachmentParameter(
+     gl.FRAMEBUFFER,
+     gl.COLOR_ATTACHMENT0,
+     gl.FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE);
+  if (contextVersion >= 2) {
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
```
kenrussell: Per section 6.1.13 of the GLSL ES 3.0.3 spec: "If the default framebuffer is bound to target, then attachment must be BACK, identifying the color buffer; DEPTH, identifying the depth buffer; or STENCIL, identifying the stencil buffer." Therefore this should still produce an error in WebGL 2.0.

kenrussell:
Thanks for this contribution. It looks good overall but there seems to be one bug. Could you please look again and update the test if necessary?

Richard-Yunchao:
@kenrussell , agree. The bug is fixed. PTAL.

Richard-Yunchao:
@kenrussell , when I run the test in chromium, I found a small issue here. calling getFramebufferAttachmentParameter on the default framebuffer should generate INVALID_OPERATION. But in chromium, it will report INVALID_ENUM if the parameters are not correct. Seems that browsers should raise INVALID_OPERATION if default FB is bound and query or set FB, before validate the parameters: https://codereview.chromium.org/1166663003/. Well, maybe the current code is OK. It is not a big deal. I will close that simple change in that situation.

sdk/tests/js/tests/framebuffer-test.js
```diff
@@ -0,0 +1,212 @@
...
+} else {
+  testPassed("context exists");
+
+  debug("");
+  debug("Checking framebuffer/renderbuffer stuff.");
+
+  var value = gl.getFramebufferAttachmentParameter(
+     gl.FRAMEBUFFER,
+     gl.BACK,
+     gl.FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE);
```
kenrussell: Two issues:

1) Please leave the original call and test in place:
gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE);

Tests should not arbitrarily be removed from the WebGL conformance suite.

2) This call can generate two potential errors when run on top of WebGL 1.0: INVALID_OPERATION, because the default framebuffer is bound; and INVALID_ENUM, because gl.BACK is not a valid argument for "attachment" in WebGL 1.0. The conformance tests must check for exactly one error at a time. Please do this particular call only if contextVersion >= 2.

kenrussell:
Thanks for revising this but please see above. Once those issues are revised this will look OK.

I note that there are a lot of differences between WebGL 1.0 and 2.0 in this file. Basically every call has an if-test immediately following it. I think it would be easier to understand in this case if the entire test were duplicated into 1.0 and 2.0 versions of framebuffer-test.html, rather than trying to share code between them. What do you think?

Richard-Yunchao:
@kenrussell , agree with you. If there are lots of if-tests, I will create a new file for WebGL 2.0. Sharing code with WebGL 1.0 seems to be a little messy in this situation. 

Code has been updated, PTAL.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -0,0 +1,203 @@
...
+} else {
+  testPassed("context exists");
+
+  debug("");
+  debug("Checking framebuffer/renderbuffer stuff.");
+
+  gl.getFramebufferAttachmentParameter(
+     gl.FRAMEBUFFER,
+     gl.COLOR_ATTACHMENT0,
+     gl.FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE);
```
kenrussell: Later, please add another test which calls getFramebufferAttachmentParameter against the default framebuffer with gl.BACK as the attachment.

kenrussell:
Thanks for updating the test. Looks good. Please add another test to it later per above. Merging now.

Richard-Yunchao:
Sure, this one just fix bugs against WebGL 2.0. There is no error right now. I will update this test to add read/write binding points to framebuffer related APIs for WebGL 2.0. In that patch, I will add the test on default framebuffer with gl.BACK as the attachment.

PR #1022, opened 2015-06-01
Add test for pow with constant vector exponent

Oletus:
This reveals a bug in NVIDIA 331 series OpenGL drivers.


zhenyao:
Looks good.  Thanks for the test.

kenrussell:
LGTM too after the fact. Thanks for contributing this test.

PR #1023, opened 2015-06-01
Fix default shaders in ESSL3 shaders with WebGL 1.0 test

Oletus:
GLSLConformanceTester was recently updated to support default ESSL3
shaders. This unintentionally affected some subtests in
essl3-shaders-with-webgl1.html. Restore the intended testing of
mixing ESSL1 shaders with ESSL3 shaders.


zhenyao:
Looks good.

PR #1024, opened 2015-06-01
Correct ternary-operator-on-arrays.html version

zhenyao:
Also add the glsl3 version of the test.

### DIFF
--- a/sdk/tests/conformance/glsl/misc/00_test_list.txt
+++ b/sdk/tests/conformance/glsl/misc/00_test_list.txt
@@ -108,6 +108,6 @@ uniform-location-length-limits.html
 --min-version 1.0.3 struct-nesting-of-variable-names.html
 --min-version 1.0.3 struct-specifiers-in-uniforms.html
 --min-version 1.0.3 struct-unary-operators.html
---min-version 1.0.4 --max-version 1.9.9 ternary-operator-on-arrays.html
+--min-version 1.0.4 ternary-operator-on-arrays.html
 --min-version 1.0.3 ternary-operators-in-global-initializers.html
 --min-version 1.0.3 ternary-operators-in-initializers.html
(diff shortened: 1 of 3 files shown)


sdk/tests/conformance/glsl/misc/00_test_list.txt
```diff
@@ -108,6 +108,6 @@ uniform-location-length-limits.html
 --min-version 1.0.3 struct-nesting-of-variable-names.html
 --min-version 1.0.3 struct-specifiers-in-uniforms.html
 --min-version 1.0.3 struct-unary-operators.html
---min-version 1.0.4 --max-version 1.9.9 ternary-operator-on-arrays.html
+--min-version 1.0.4 ternary-operator-on-arrays.html
```
kenrussell: Is removing the --max-version correct? I'm pretty sure that the test will be run with a WebGL 2.0 context if it's removed.

kenrussell:
LGTM overall; one question. Please merge yourself once it's answered. Thanks.

kenrussell:
Oops, hadn't meant to merge that. :)

zhenyao:
Yes, WebGL 2 should just accept glsl1 shaders.

Oletus:
The ESSL3 spec is a bit ambiguous on whether ternary operator is actually allowed on arrays and structs - according to Khronos bugzilla discussion the intent was to support this, but the changes didn't make it to the table in section 5.7 of ESSL3.00. It's unknown if all actual ESSL3 implementations support the ternary operator on arrays, so driver workarounds for this might also be required in WebGL implementations.

Because the ESSL3 spec is unclear, I think we should add an explicit mention of the expected ESSL3 behavior in the WebGL 2 spec. Do you agree?

kenrussell:
@Oletus , thanks for pointing that out. There's clearly a discrepancy between the definition of the ternary operator in ESSL 3.00.4 Section 5.9 "Expressions", and section 5.7 "Structure and Array Operations". I filed #1031 about this issue. Could you put the Khronos bug ID in that issue?

PR #1025, opened 2015-06-01
Move js tests from resources/ to js/tests/

zhenyao:
I didn't move the util js files because that will involve almost every html file and I prefer to do it in a separate CL.


sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image.js
```diff
@@ -43,7 +43,7 @@ function generateTest(pixelFormat, pixelType, pathToTestRoot, prologue) {
         gl.clearColor(0,0,0,1);
         gl.clearDepth(1);
 
-        wtu.loadTexture(gl, "/resources/red-green.png", runTest);
+        wtu.loadTexture(gl, "../../resources/red-green.png", runTest);
```
kenrussell: Do these URLs really change? I thought they were relative to the containing HTML file. The relative URLs are getting a little confusing.

sdk/tests/resources/iterable-test.js
```diff
@@ -1,173 +0,0 @@
-/*
```
kenrussell: Is this file recreated in the new directory? It's not showing up in this pull request.

sdk/tests/resources/tex-input-validation.js
```diff
@@ -1,397 +0,0 @@
-/*
```
kenrussell: Is this file recreated in the new directory? It looks like it's missing from this pull request.

kenrussell:
Agree with moving the utility files in a separate pull request.

This LGTM overall if you've tested it locally; a few questions and comments. Please feel free to merge yourself once they're addressed.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image.js
zhenyao: Thanks for catching that.  These two files are moved locally but not in the git commit.  Maybe for these two I accidentally used mv instead of git mv.  Add them back. I tested locally, with this, the conformance suites work fine, including the few bugs Olli pointed out after a previous merge.

zhenyao:
Merge myself with kbr's blessing.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image.js
zhenyao: The above comment is for the two missing files in the commit.

As for the URL change, the wtu.loadTexture simply put the URL as Image.src, so it has to be relative.  I agree it's confusing because now the js code and the html are in different locations.  Maybe we can modify the wtu.loadTexture() code and always load from resources/ folder? So even the "resources" in the path can be avoided.

kenrussell: It's a good suggestion. Let's see how difficult this new structure is to maintain. If we can avoid adding too much intelligence to the resource finding code, that's probably better in the long run.

PR #1026, opened 2015-06-01
Add test for comparing structs containing arrays

matthewn4444:
This is a special case for HLSL implementation


kenrussell:
Thanks for this new test. Looks good.

PR #1028, opened 2015-06-02
Fix flakiness of video test.

ds-hwang:
Video tests are flaky on Chrome Android. This PR fixes it.
Two changes:
1. reuse one video element for all variants of tests.
2. clear listerners before calling callback because callback can call listerners again.

BUG=http://crbug.com/492383


ds-hwang:
@kenrussell  could you review?
I think we can re-enable video-related WebGL conformance tests on Android after this PR.

ds-hwang:
still flaky

kenrussell:
I think the problem is in Chrome's WebMediaPlayer implementation on Android. It's probably dispatching certain events too early, given that this test works reliably across other browsers and platforms. Could you look into Chrome's code and when it dispatches the various events that this test relies on?

PR #1029, opened 2015-06-02
Simplify GLSLConformanceTester

Oletus:
Clean up unused functionality.

Make it more explicit that no settings need to be made for vShaderSuccess, vShaderSource or the equivalent fragment shader settings if the default shader is used.

Remove setting of unused vShader / fShader keys in the test info structures and streamline the setupTest() function.


kenrussell:
Thanks for this nice cleanup. LGTM after the fact.

Issue #1027, opened 2015-06-02
WebGL 2 Conformance - buffers/buffer-copying-contents.html


djg:
sdk/tests/conformance2/buffers/buffer-copying-contents.html contains a number of logic errors:
1. Reports failed copies because of element data/other data mismatch as an error,
2. When copying with the same buffer:
  * Setup logic fails when binding gl.COPY_WRITE_BUFFER. (Wrong buffer is used as source)
  * Tries to copy overlapping ranges and reports failure as an error.

zhenyao:
Thanks for reporting this.

For 1, I don't think I get it.  Can you explain?
For 2, I can't find the text in the spec saying COPY_READ can't be dst and COPY_WRITE can't be src.  If you know where it is, could you please point to me?
Overlapping ranges, I'll fix it.

djg:
For 1. For example, the test says that a copy from gl.ARRAY_BUFFER to gl.ELEMENT_ARRAY_BUFFER should succeed. The spec in section 5.2 says:
> Attempting to use copyBufferSubData to copy between buffers that have element array and other data  WebGL buffer types as specified in section Buffer Object Binding generates an INVALID_OPERATION error and no copying is performed.

My understanding of section 5.1 is that binding to gl.ARRAY_BUFFER sets the contents to `other data` and to gl.ELEMENT_ARRAY_BUFFER to `element data` so the copyBufferSubData should fail.

For 2, my explanation probably wasn't clear enough; it's just a bug in the order of setting up the buffer bindings in the test.

In the following code from the test, if `srcTarget == dstTarget` and `copyRead == false` and `copyWrite == true` then when copyBufferSubData is called `dstBuffer` is bound to `srcTarget` as well as `gl.COPY_WRITE_BUFFER` which means that the copy happens in the same buffer and the overlapping range check fails.

```
gl.bindBuffer(srcTarget, srcBuffer);
gl.bufferData(srcTarget, originalData, gl.STATIC_DRAW);
if (copyRead)
  gl.bindBuffer(gl.COPY_READ_BUFFER, srcBuffer);

gl.bindBuffer(dstTarget, dstBuffer);
gl.bufferData(dstTarget, new Float32Array(length), gl.STATIC_DRAW);
if (copyWrite)
  gl.bindBuffer(gl.COPY_WRITE_BUFFER, dstBuffer);

gl.copyBufferSubData(copyRead ? gl.COPY_READ_BUFFER : srcTarget,
[...]

PR #1032, opened 2015-06-03
Fix incorrect behavior in buffer-copying-contents.html

zhenyao:
Per discussion in https://github.com/KhronosGroup/WebGL/issues/1027


kenrussell:
Very good. LGTM

PR #1035, opened 2015-06-03
Change Buffer Object Binding rule

zhenyao:
So that binding a buffer object has been bound to COPY_READ_BUFFER or COPY_WRITE_BUFFER to ELEMENT_ARRAY_BUFFER will also generate INVALID_OPERATION.

This is because we can upload data directly to a buffer object bound to these
two binding points, so they are no different from other non ELEMENT_ARRAY_BUFFER
binding points.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -2390,8 +2390,10 @@ WebGL2RenderingContext implements WebGL2RenderingContextBase;
         In the WebGL 2 API, buffers have their WebGL buffer type initially set to <em>undefined</em>. Calling
         <code>bindBuffer</code>, <code>bindBufferRange</code> or <code>bindBufferBase</code> with the
         <code>target</code> argument set to any buffer binding point except <code>COPY_READ_BUFFER</code> or
-        <code>COPY_WRITE_BUFFER</code> will then set the WebGL buffer type of the buffer being bound according
-        to the table above.
+        <code>COPY_WRITE_BUFFER</code> will then set the WebGL buffer type of the buffer being bound
+        according to the table above. Binding a buffer with type <em>undefined</em> to
+        the <code>COPY_READ_BUFFER</code> or <code>COPY_WRITE_BUFFER</code> binding points will set
+        its type to <em>other data</em>.
     </p>
 
     <p>


zhenyao:
@jdashg @kenrussell Please review.  Thanks.

Oletus:
This has a somewhat unfortunate consequence that ELEMENT_ARRAY_BUFFERs can't be copied at all anymore. Maybe that isn't too bad, but we could also consider the alternative of setting the type of a buffer into "other data" if something gets uploaded into it while its bound to COPY_READ_BUFFER/COPY_WRITE_BUFFER.

The tests that already exist for this need to be updated as well.

zhenyao:
From the current wording, ELEMENT_ARRAY_BUFFERs still can be copied, just it needs to be bound to ELEMENT_ARRAY_BUFFER first, then to COPY_READ_BUFFER or COPY_WRITE_BUFFER; but not the other way around where it binds to COPY_READ_BUFFER or COPY_WRITE_BUFFER first, and then try to bind to ELEMENT_ARRAY_BUFFER, even though COPY_READ_BUFFER or COPY_WRITE_BUFFER is not considered as "other data".  The reason is that we can still upload data to a buffer that's only bound to these two binding points.

I think the alternative is also good, that we disallow uploading data to a buffer that doesn't have a "data type" yet, i.e., it's only bound to COPY_READ_BUFFER or COPY_WRITE_BUFFER.

kenrussell:
The intent of this change was still to allow copying between element array buffers, but looking at the wording it is a little confusing. Saying "bind a WebGLBuffer which has been bound to a binding point other than ELEMENT_ARRAY_BUFFER to ELEMENT_ARRAY_BUFFER will generate an INVALID_OPERATION error" implies that if a buffer is bound in this order:
  1) ELEMENT_ARRAY_BUFFER
  2) COPY_READ_BUFFER
  3) ELEMENT_ARRAY_BUFFER
that (3) will generate an error.

Mo, I'll email you a slightly different suggested wording.

zhenyao:
Thanks Ken.  Updated.

kenrussell:
Thanks Mo. @Oletus does that address your concern?

@jdashg do you have any feedback?

LGTM -- merge at your convenience from my standpoint, though would be good to get feedback from Olli and Jeff.

zhenyao:
@Oletus @jdashg Any further comments?

djg:
Just recently, I authored the checks to handle this in Firefox. Just to make it painfully obvious, if a buffer of `unknown type` is bound to COPY_READ_BUFFER, it becomes `other data`. Thus, the only way to get `element data` is to bind to ELEMENT_ARRAY_BUFFER before anything else?

Oletus:
Looks reasonable to me. I believe @djg interpreted the spec correctly.

kenrussell:
@djg yes, that's the intent.

zhenyao:
Rebased

kenrussell:
Merging now.

PR #1037, opened 2015-06-04
add test cases for read/draw binding points for framebuffer related APIs against webgl 2.0

Richard-Yunchao:
Chromium failed on several tests. My patch (https://codereview.chromium.org/1120953002/) can fix some (NOT all) failures in chromium. The other failed cases can be passed in firefox nightly build, although firefox has some other failures too. 

I think the conformance test itself is OK. We should fix the failures in browsers. PTAL.

### DIFF
--- a/sdk/tests/conformance2/renderbuffers/framebuffer-test.html
+++ b/sdk/tests/conformance2/renderbuffers/framebuffer-test.html
@@ -33,7 +33,6 @@
 <link rel="stylesheet" href="../../resources/js-test-style.css"/>
 <script src="../../resources/js-test-pre.js"></script>
 <script src="../../resources/webgl-test-utils.js"></script>
-<script src="../../resources/desktop-gl-constants.js"></script>
 </head>
 <body>
 <div id="description"></div>
@@ -63,6 +62,12 @@ if (!gl) {
      gl.FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE);
   wtu.glErrorShouldBe(gl, gl.INVALID_ENUM,
             "calling getFramebufferAttachmentParameter with attachment = COLOR_ATTACHMENT0 on the default framebuffer should generate INVALID_ENUM.");
+  gl.getFramebufferAttachmentParameter(
+     gl.FRAMEBUFFER,
+     gl.BACK,
+     gl.FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "calling getFramebufferAttachmentParameter with attachment = GL_BACK on the default framebuffer should succeed.");
   gl.checkFramebufferStatus(gl.FRAMEBUFFER);
   wtu.glErrorShouldBe(gl, gl.NO_ERROR,
             "calling checkFramebufferStatus on the default framebuffer should succeed");
@@ -122,7 +127,7 @@ if (!gl) {
   wtu.glErrorShouldBe(gl, gl.NO_ERROR,
             "binding a newly created framebuffer should succeed.");
 
-  var target = desktopGL.READ_FRAMEBUFFER
+  var target = gl.READ_FRAMEBUFFER
   gl.getFramebufferAttachmentParameter(
      target,
      gl.COLOR_ATTACHMENT0,
@@ -153,18 +158,20 @@ if (!gl) {
 
   var colorAttachmentsNum = gl.getParameter(gl.MAX_COLOR_ATTACHMENTS);
   if (colorAttachmentsNum >= 2) {
-    var attachment = desktopGL.COLOR_ATTACHMENT1
+    var attachment = gl.COLOR_ATTACHMENT1
     gl.framebufferTexture2D(gl.FRAMEBUFFER, attachment, gl.TEXTURE_2D, fbtex, 0);
     wtu.glErrorShouldBe(gl, gl.NO_ERROR,
             "calling framebufferTexImage2D with attachment = COLOR_ATTACHMENT1 should succeed.");
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, attachment, gl.TEXTURE_2D, null, 0);
     gl.framebufferRenderbuffer(gl.FRAMEBUFFER, attachment, gl.RENDERBUFFER, rb);
     wtu.glErrorShouldBe(gl, gl.NO_ERROR,
             "calling framebufferRenderbuffer with attachment = COLOR_ATTACHMENT1 should succeed.");
+    gl.framebufferRenderbuffer(gl.FRAMEBUFFER, attachment, gl.RENDERBUFFER, null);
   }
 
   gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER,
                                        gl.COLOR_ATTACHMENT0,
-                                       desktopGL.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
+                                       gl.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
   wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
             "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING without any attachment should generate INVALID_OPERATION.");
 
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
   wtu.glErrorShouldBe(gl, gl.NO_ERROR,
             "detaching a renderbuffer from a framebuffer should succeed.");
 
+  // attach a 2d texture to read framebuffer, while no attachment to draw framebuffer.
+  gl.framebufferTexture2D(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, fbtex, 0);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "attaching a texture to read framebuffer should succeed.");
+  gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER,
```
kenrussell: How about making this DRAW_FRAMEBUFFER to make the test more obvious?

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
...
+                                       gl.COLOR_ATTACHMENT0,
+                                       desktopGL.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
```
kenrussell: FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING is part of the WebGL 2.0 spec so you could just as well get it from the "gl" object instead of the "desktopGL" object.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
...
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on draw framebuffer without any attachment should generate INVALID_OPERATION.");
+  gl.getFramebufferAttachmentParameter(gl.READ_FRAMEBUFFER,
+                                       gl.COLOR_ATTACHMENT0,
+                                       // desktopGL.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
```
kenrussell: Remove commented out code.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
...
+                                       gl.FRAMEBUFFER_ATTACHMENT_RED_SIZE);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING should generate INVALID_OPERATION.");
```
kenrussell: This comment doesn't match the code.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
...
+
+  // attach different textures to read and draw framebuffer.
+  var fbtex1 = gl.createTexture();
+  gl.bindTexture(gl.TEXTURE_2D, fbtex1);
+  gl.texImage2D(gl.TEXTURE_2D, 0, gl.ALPHA, canvas.width, canvas.height, 0, gl.ALPHA, gl.UNSIGNED_BYTE, null);
+  gl.framebufferTexture2D(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, fbtex1, 0);
```
kenrussell: It might be clearer to create a new framebuffer and attach it to DRAW_FRAMEBUFFER right here. It's initialized much higher up in the test so it's difficult to understand the current state just by reading the code.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
...
+                                       gl.COLOR_ATTACHMENT0,
+                                       // desktopGL.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
+                                       gl.FRAMEBUFFER_ATTACHMENT_RED_SIZE);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING should generate INVALID_OPERATION.");
+
+  // attach different textures to read and draw framebuffer.
+  var fbtex1 = gl.createTexture();
+  gl.bindTexture(gl.TEXTURE_2D, fbtex1);
+  gl.texImage2D(gl.TEXTURE_2D, 0, gl.ALPHA, canvas.width, canvas.height, 0, gl.ALPHA, gl.UNSIGNED_BYTE, null);
```
kenrussell: ALPHA type textures aren't required to be supported. See "Required Texture Formats" in section 3.8.3 "Texture Image Specification". I suggest you use something like an RG8 texture and check that the red channel has non-zero bits but the blue channel has zero bits instead.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
...
+            "attaching a texture to draw framebuffer should succeed.");
+  shouldBeNonZero('gl.getFramebufferAttachmentParameter(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_RED_SIZE)');
+  shouldBe('gl.getFramebufferAttachmentParameter(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_RED_SIZE)', '0');
+
+  gl.framebufferTexture2D(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, null, 0);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "detaching a texture from read framebuffer should succeed.");
+  gl.getFramebufferAttachmentParameter(gl.READ_FRAMEBUFFER,
+                                       gl.COLOR_ATTACHMENT0,
+                                       desktopGL.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
```
kenrussell: Since FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING is used so many times throughout this test please get it from the gl object rather than desktopGL.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
...
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on read framebuffer without any attachment should generate INVALID_OPERATION.");
+  gl.getFramebufferAttachmentParameter(gl.DRAW_FRAMEBUFFER,
+                                       gl.COLOR_ATTACHMENT0,
+                                       desktopGL.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on draw framebuffer should generate INVALID_OPERATION.");
+  gl.framebufferTexture2D(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, null, 0);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "detaching a texture from draw framebuffer should succeed.");
+  gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER,
```
kenrussell: Use DRAW_FRAMEBUFFER for clarity.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
...
+                                       gl.COLOR_ATTACHMENT0,
+                                       desktopGL.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on draw framebuffer without any attachment should generate INVALID_OPERATION.");
+
+  // attach a renderbuffer to read framebuffer, while no attachment to draw framebuffer.
+  gl.framebufferRenderbuffer(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.RENDERBUFFER, rb);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "attaching a renderbuffer to a read framebuffer should succeed.");
+  gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER,
```
kenrussell: Use DRAW_FRAMEBUFFER for clarity.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
...
+                                       gl.COLOR_ATTACHMENT0,
+                                       desktopGL.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on draw framebuffer without any attachment should generate INVALID_OPERATION.");
+  gl.getFramebufferAttachmentParameter(gl.READ_FRAMEBUFFER,
+                                       gl.COLOR_ATTACHMENT0,
+                                       desktopGL.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING should generate INVALID_OPERATION.");
```
kenrussell: "should succeed".

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
...
+  // attach different renderbuffers to read and draw framebuffer respectively.
+  var rb1 = gl.createRenderbuffer();
+  gl.bindRenderbuffer(gl.RENDERBUFFER, rb1);
+  gl.renderbufferStorage(gl.RENDERBUFFER, gl.DEPTH_COMPONENT16, canvas.width, canvas.height);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "allocating renderbuffer storage of a newly created renderbuffer should succeed.");
+  gl.framebufferRenderbuffer(gl.DRAW_FRAMEBUFFER, gl.DEPTH_ATTACHMENT, gl.RENDERBUFFER, rb1);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "attaching a renderbuffer to draw framebuffer should succeed.");
+  shouldBeNonZero('gl.getFramebufferAttachmentParameter(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_RED_SIZE)');
```
kenrussell: The state for this check is set up much further above, so it is confusing to understand why this assertion is true.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
...
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on draw framebuffer without any attachment should generate INVALID_OPERATION.");
+  gl.getFramebufferAttachmentParameter(gl.READ_FRAMEBUFFER,
+                                       gl.COLOR_ATTACHMENT0,
+                                       desktopGL.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING should generate INVALID_OPERATION.");
+
+  // attach different renderbuffers to read and draw framebuffer respectively.
+  var rb1 = gl.createRenderbuffer();
```
kenrussell: How about a more descriptive name for this variable: at least something like "depthRB"?

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -188,6 +196,107 @@
...
+  shouldBeNonZero('gl.getFramebufferAttachmentParameter(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_RED_SIZE)');
+  gl.getFramebufferAttachmentParameter(gl.READ_FRAMEBUFFER,
+                                       gl.DEPTH_ATTACHMENT,
+                                       gl.FRAMEBUFFER_DEPTH_SIZE);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_DEPTH_SIZE on read framebuffer without depth attachment should generate INVALID_OPERATION.");
+  shouldBeNonZero('gl.getFramebufferAttachmentParameter(gl.DRAW_FRAMEBUFFER, gl.DEPTH_ATTACHMENT, gl.FRAMEBUFFER_ATTACHMENT_DEPTH_SIZE)');
+  gl.getFramebufferAttachmentParameter(gl.DRAW_FRAMEBUFFER,
+                                       gl.COLOR_ATTACHMENT0,
+                                       gl.FRAMEBUFFER_RED_SIZE);
```
kenrussell: Again, it's difficult to remember the state of the draw and read framebuffers at any particular point in the code. The code for these tests should be close together and relatively self-contained. Delete and re-allocate framebuffer, renderbuffer, and texture objects as necessary to avoid leaking too many resources for too long periods of time.

kenrussell:
Thanks for this contribution. On the whole these new tests are difficult to understand just from looking through them. They could be better structured. If two of the tests are mostly identical -- for example, they test the same basic thing against READ_FRAMEBUFFER and DRAW_FRAMEBUFFER -- then they should be factored out into a separate function which just takes in the target. The setup and testing of the various framebuffers' state should be kept close together so that it's possible to read just a small section of the code and understand what's going on. Please fix the above comments and restructure this pull request to be clearer. Thanks.

Richard-Yunchao:
Thanks for your reviewing, Ken. The code has been updated accordingly, PTAL.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -191,6 +198,127 @@
...
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on read/draw framebuffer without any attachment should generate INVALID_OPERATION.");
+    gl.framebufferTexture2D(targetA, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, null, 0);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "detaching a texture from read/draw framebuffer should succeed.");
+  }
+
+  var readDrawFB = gl.createFramebuffer();
+  gl.bindFramebuffer(gl.FRAMEBUFFER, readDrawFB);
+  attachAndDetachTexture(gl.READ_FRAMEBUFFER, gl.DRAW_FRAMEBUFFER);
+  attachAndDetachTexture(gl.DRAW_FRAMEBUFFER, gl.READ_FRAMEBUFFER);
```
kenrussell: These tests look wrong. The readDrawFB object is being bound to both the read and draw attachment points, so when you modify the object attached to the READ_FRAMEBUFFER attachment point and then query the DRAW_FRAMEBUFFER attachment point, you're talking about the same framebuffer object and the tests above should fail.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -191,6 +198,127 @@
...
+
+  // attach different textures as color attachment to read and draw framebuffer respectively, then detach these attachments.
+  var fbtex1 = gl.createTexture();
+  gl.bindTexture(gl.TEXTURE_2D, fbtex1);
+  gl.texImage2D(gl.TEXTURE_2D, 0, gl.RG8, canvas.width, canvas.height, 0, gl.RG, gl.UNSIGNED_BYTE, null);
+  gl.framebufferTexture2D(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, fbtex1, 0);
+  gl.framebufferTexture2D(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, fbtex, 0);
```
kenrussell: Similarly, the same framebuffer object is still bound to both the draw and read framebuffer attachment points, so the second of these calls overwrites the result of the first.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -191,6 +198,127 @@
...
+  gl.getFramebufferAttachmentParameter(gl.READ_FRAMEBUFFER,
+                                       gl.COLOR_ATTACHMENT0,
+                                       gl.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on read framebuffer without any attachment should generate INVALID_OPERATION.");
+  gl.getFramebufferAttachmentParameter(gl.DRAW_FRAMEBUFFER,
+                                       gl.COLOR_ATTACHMENT0,
+                                       gl.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on draw framebuffer should succeed.");
```
kenrussell: The draw framebuffer is identical to the read framebuffer, so this query will fail.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -191,6 +198,127 @@
...
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on read/draw framebuffer without any attachment should generate INVALID_OPERATION.");
+    gl.framebufferRenderbuffer(targetA, gl.COLOR_ATTACHMENT0, gl.RENDERBUFFER, null);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "detaching a renderbuffer from a read/draw framebuffer should succeed.");
+  }
+
+  var readDrawFB1 = gl.createFramebuffer();
+  gl.bindFramebuffer(gl.FRAMEBUFFER, readDrawFB1);
+  attachAndDetachRenderbuffer(gl.READ_FRAMEBUFFER, gl.DRAW_FRAMEBUFFER);
+  attachAndDetachRenderbuffer(gl.DRAW_FRAMEBUFFER, gl.READ_FRAMEBUFFER);
```
kenrussell: Same basic problem here, where the same framebuffer object is bound to both the READ and DRAW attachment points.

kenrussell:
Thanks for revising the test, but there are several errors as far as I can tell by reading the code.

Richard-Yunchao:
Sorry, I made the same mistake again. There are read framebuffer and draw framebuffer. But within a FBO, there are no read or write attachments. Code has been updated accordingly. PTAL.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -191,6 +198,131 @@
...
+  gl.framebufferTexture2D(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, fbtex, 0);
+  shouldBeNonZero('gl.getFramebufferAttachmentParameter(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_BLUE_SIZE)');
+  shouldBe('gl.getFramebufferAttachmentParameter(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_BLUE_SIZE)', '0');
+
+  gl.framebufferTexture2D(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, null, 0);
+  gl.getFramebufferAttachmentParameter(gl.READ_FRAMEBUFFER,
+                                       gl.COLOR_ATTACHMENT0,
+                                       gl.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on read framebuffer without any attachment should generate INVALID_OPERATION.");
```
kenrussell: This test's redundant -- it's already done in attachAndDetachTexture, above.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
kenrussell: This is resolved in the current patch set.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -191,6 +198,131 @@
...
+                                       gl.COLOR_ATTACHMENT0,
+                                       gl.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on draw framebuffer should succeed.");
+  gl.framebufferTexture2D(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, null, 0);
+  gl.getFramebufferAttachmentParameter(gl.DRAW_FRAMEBUFFER,
+                                       gl.COLOR_ATTACHMENT0,
+                                       gl.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING on draw framebuffer without any attachment should generate INVALID_OPERATION.");
```
kenrussell: These tests are redundant -- they're already done in attachAndDetachTexture. Please remove them.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -191,6 +198,131 @@
...
+                                       gl.FRAMEBUFFER_RED_SIZE);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_RED_SIZE on draw framebuffer without color attachment should generate INVALID_OPERATION.");
+
+  gl.framebufferRenderbuffer(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.RENDERBUFFER, null);
+  gl.getFramebufferAttachmentParameter(gl.READ_FRAMEBUFFER,
+                                       gl.COLOR_ATTACHMENT0,
+                                       gl.FRAMEBUFFER_RED_SIZE);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_RED_SIZE on read framebuffer without any attachment should generate INVALID_OPERATION.");
```
kenrussell: This getFramebufferAttachmentParameter test is redundant and should be deleted.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -191,6 +198,131 @@
...
+  shouldBeNonZero('gl.getFramebufferAttachmentParameter(gl.DRAW_FRAMEBUFFER, gl.DEPTH_ATTACHMENT, gl.FRAMEBUFFER_ATTACHMENT_DEPTH_SIZE)');
```
kenrussell: This test is redundant and should be removed.

sdk/tests/conformance2/renderbuffers/framebuffer-test.html
```diff
@@ -191,6 +198,131 @@
...
+  gl.framebufferRenderbuffer(gl.DRAW_FRAMEBUFFER, gl.DEPTH_ATTACHMENT, gl.RENDERBUFFER, null);
+  gl.getFramebufferAttachmentParameter(gl.DRAW_FRAMEBUFFER,
+                                       gl.DEPTH_ATTACHMENT,
+                                       gl.FRAMEBUFFER_DEPTH_SIZE);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+            "calling getFramebufferAttachmentParameter with pname = GL_FRAMEBUFFER_ATTACHMENT_DEPTH_SIZE on draw framebuffer without any attachment should generate INVALID_OPERATION.");
```
kenrussell: Again, redundant test which should be removed.

kenrussell:
Please make the cleanups pointed out above.

Richard-Yunchao:
@kenrussell , they are a little different. attachAndDetach functions just attach one binding point at a time, then detach it. They are easy and direct. However, the functions you marked as redundant attach different textures/renderbuffers to read and draw framebuffer respectively, test them, then remove attachment from one binding point, test again, then remove the other, test once more. So these test cases help to make sure WebGL 2.0 implementations correctly attach textures/renderbuffers to specified binding points, and detach them from specified binding points in correct sequence. So attachments and framebuffers in browser implementations are not in a mess. 

I prefer to keep them, WDYT?

kenrussell:
I'm doubtful that the additional test cases increase the code coverage of the overall test, and make the test harder to understand and maintain. That having been said, we've been arguing over this pull request for nearly a week and I'm eager to incorporate it so we can start running it, so am merging it now.

Going forward, please try to structure your tests so they're minimal, and focused on covering exactly the intended code. Thanks for this contribution.

Richard-Yunchao:
OK. I will revisit the code after that all test cases for renderbuffer/framebuffer are there.

PR #1039, opened 2015-06-05
WebGL2 Conformance - state/gl-get-calls fixes

djg:
I found a couple of bugs in this test:
1. The default value for gl.getParameter(gl.READ_BUFFER) should be gl.BACK not gl.NONE 
```
The initial value of the read framebuffer for the default framebuffer is BACK
if there is a default framebuffer associated with the context, otherwise it is 
NONE. (GL ES 3 3.0.4, Section 4.3.1, page 191)
```
2. An instance of `context.MAX_ARRAY_TEXTURE_LAYERS` that should have been `context.MAX_CLIENT_WAIT_TIMEOUT_WEBGL`


zhenyao:
Thanks for the fix.

kenrussell:
Yes, thank you!

PR #1040, opened 2015-06-05
Add different versions of constant-precision-qualifier test

Oletus:
After fixing the bug in ANGLE that made this test fail, it was discovered
that some NVIDIA mobile drivers also have a bug making this test fail.
Add two more versions of the test, one that doesn't use constants and one
that only uses highp variables, so it's easier to narrow down
platform-specific failures of this test.

### DIFF
--- a/sdk/tests/conformance/glsl/bugs/constant-precision-qualifier.html
+++ b/sdk/tests/conformance/glsl/bugs/constant-precision-qualifier.html
@@ -67,6 +67,31 @@ void main() {
     gl_FragColor = vec4(0.0, 2.0 * a, 0.0, 1.0);
 }
 </script>
+<script id="fshaderNoConstants" type="x-shader/x-fragment">
+// This shader has the same functionality as the one above, but it doesn't contain
+// operations that can be constant folded at compile-time.
+// It's here to provide a point of comparison.
+uniform mediump float uTest;
+
+void main() {
+    highp float c = 4096.5 + uTest;
+    mediump float a = 0.0;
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/glsl/bugs/constant-precision-qualifier.html
```diff
@@ -67,6 +67,30 @@
...
+void main() {
+    highp float c = 4096.5 + uTest;
+    mediump float a = 0.0;
+    a = fract(c + uTest);
+    gl_FragColor = vec4(0.0, 2.0 * a, 0.0, 1.0);
+}
+</script> <script id="fshaderAllHighp" type="x-shader/x-fragment">
+// This shader has the same functionality as the one above, but it only uses highp.
+// It's here to provide a point of comparison.
+uniform highp float uTest;
```
zhenyao: highp is optional for fragment shaders in GLSL 1. You should guard it using GL_FRAGMENT_PRECISION_HIGH.

Oletus:
This test calls gl.getShaderPrecisionFormat(gl.FRAGMENT_SHADER, gl.HIGH_FLOAT) and only gets run if the returned precision value is not 0, meaning that highp is supported. I did notice one formatting issue in the test though and fixed that now.

kenrussell:
Ah, that query was in part of the diff that was collapsed, which was a little confusing. @zhenyao should re-review. Thanks for this contribution.

zhenyao:
Ah I missed that part.

PR #1041, opened 2015-06-05
Add a test for invalid default precision statements

Oletus:
The test covers a small selection of invalid types.

It reveals a bug in ANGLE which allows "precision mediump uint;"


zhenyao:
Great!

PR #1043, opened 2015-06-05
Add test for indexing complex array expressions

matthewn4444:
Related Discussion: https://cvs.khronos.org/bugzilla/show_bug.cgi?id=13850


zhenyao:
Looks great. Thank you.

PR #1044, opened 2015-06-05
Fixed shader code generation.

jasomobica:
Strip leading whitespace from generated shaders to make sure that '#version' is on the first line.


kenrussell:
LGTM. Assuming this has been tested.

PR #1046, opened 2015-06-09
The function framebufferTextureLayer has an error.

djg:
Instead of taking third param, texture, as a GLuint it should accept
WebGLTexture? instead.

This change brings framebufferTextureLayer inline with framebufferTexture2D.


zhenyao:
Looks good.

kenrussell:
Thanks for this fix.

PR #1048, opened 2015-06-09
Make depth/stencil/antialias drawing buffer attributes required

zhenyao:
With an ES3 capable GPU/driver, I don't think it's possible that such requests can't be honored.


toji:
LGTM, thanks!

PR #1050, opened 2015-06-09
Fix the context creation steps to properly set the actual context att…

kenrussell:
…ributes.

Thanks to Janusz Sobczak from Mobica for pointing out this longstanding error.


kenrussell:
Please review, @zhenyao in particular.

Thanks again to @jasomobica for pointing out this longstanding error.

zhenyao:
LGTM

PR #1052, opened 2015-06-11
Modify inheritance hierarchy test to be ES6 compliant

RafaelCintron:
As of ES6, the prototypes for typed array constructors point to an intrinsic object whose internal
prototype is Object.prototype. Relevant spec section is 22.2.5.2: TypedArray.prototype.

For test purposes, instead of testing for Object.getPrototypeOf(Int8Array.prototype) being equal
to Object.prototype, we now test Object.getPrototypeOf(Object.getPrototypeOf(Int8Array.prototype)) being equal to Object.prototype. Same goes for the rest of the typed array types.

### DIFF
--- a/sdk/tests/conformance/typedarrays/array-unit-tests.html
+++ b/sdk/tests/conformance/typedarrays/array-unit-tests.html
@@ -195,15 +195,18 @@ function testInheritanceHierarchy() {
   // case, assert the new behavior.
   shouldBe('new Uint8ClampedArray(1) instanceof Uint8Array', 'false');
 
-  shouldBe('Object.getPrototypeOf(Int8Array.prototype)', 'Object.prototype');
-  shouldBe('Object.getPrototypeOf(Uint8Array.prototype)', 'Object.prototype');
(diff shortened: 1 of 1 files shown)


toji:
Have you tried this with all the major browsers? Are any failing with the old code? I'm not super familiar with ES6 semantics yet, and just want to make sure that this doesn't break browsers that aren't fully ES6 compliant yet (since that's not what we're testing for.)

kenrussell:
The V8 team provided some input about this yesterday. As browsers upgrade from ES5 to ES6 semantics, the new inheritance hierarchy will be used. In the limit, it would be good if the typed array tests here verified just the ES6 behavior. We will look into adding a test to the forthcoming test262 suite to verify just that.

For the time being, I'm going to merge this change and then follow up with another one which allows the test to pass on both ES5 and ES6 implementations, both on top of tree and in the earlier snapshots.

kenrussell:
Also, thanks Rafael for this contribution.

RafaelCintron:
To answer @toji 's question about pass rates.  Before my change, Edge and Firefox failed the test, Chrome and IE passed.  After my commit, the opposite is true. With @kenrussell 's pull request #1059, all four browsers pass.  I did not try Safari.

PR #1054, opened 2015-06-11
Add text in WebGL 2 spec describing FBO attachment behaviors.

zhenyao:
It is different from WebGL 1.


specs/latest/2.0/index.html
```diff
@@ -2442,6 +2442,19 @@ <h2 class="no-toc">Table of contents</h2>
         See section <a href="#NO_PRIMITIVE_RESTART_FIXED_INDEX">PRIMITIVE_RESTART_FIXED_INDEX is always enabled</a>.
     </p>
 
+    <h4>Framebuffer Object Attachments</h4>
```
kenrussell: Please name this section like <a name="FRAMEBUFFER_OBJECT_ATTACHMENTS">.

kenrussell:
Thanks. This looks good. Could you please make a couple of other changes?

 - Could you update the documentation for getFramebufferAttachmentParameter to refer to this new section, and mention the restriction about querying DEPTH_STENCIL_ATTACHMENT?
 - Could you add documentation for checkFramebufferStatus in the WebGL 2.0 spec, mention the new rule, and refer to this new section?

Also, could you please update the spec's last modified date?

Thanks.

zhenyao:
Updated. Please take another look.

specs/latest/2.0/index.html
```diff
@@ -1063,8 +1063,17 @@ <h2 class="no-toc">Table of contents</h2>
     <h4>Framebuffer objects</h4>
 
     <dl class="methods">
-      <dt class="idl-code">any getFramebufferAttachmentParameter(GLenum target, GLenum attachment,
-                                        GLenum pname)
+      <dt class="idl-code">[WebGLHandlesContextLoss] GLenum checkFramebufferStatus(GLenum target)
+          <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.3.pdf#nameddest=section-4.4.5">OpenGL ES 3.0.3 &sect;4.4.4</a>, <a class="nonnormative" href="http://www.khronos.org/opengles/sdk/docs/man3/html/glCheckFramebufferStatus.xhtml">man page</a>)</span>
+      <dd>
+        <p>Only differences from <a href="https://www.khronos.org/registry/webgl/specs/latest/1.0/#CHECK_FRAMEBUFFER_STATUS">checkFramebufferStatus</a> in WebGL 1 are described here.</p>
```
kenrussell: I'm pretty sure this is the first reference from the WebGL 2.0 to the 1.0 spec. It's a little complicated to handle these because theoretically we should use relative URLs. However this seems fine as a start. Please give a little more thought to how we might handle these references better.

kenrussell:
Looks excellent. Merging now.

PR #1056, opened 2015-06-11
Add readPixels in WebGL 2 spec.

zhenyao:
Dealing with PIXEL_PACK_BUFFER situation.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -1818,6 +1818,29 @@ match the <em>type</em> according to the <a href="#TEXTURE_PIXELS_TYPE_TABLE">ab
 
 <!-- ======================================================================================================= -->
 
+    <h4><a name="readpixels">Reading back pixels</a></h4>
+
+    <p>Pixels in the current framebuffer can be read back into an ArrayBufferView object or a WebGLBuffer bound to the <code>PIXEL_PACK_BUFFER</code> target.</p>
+    <p>Only differences from <a href="../1.0/index.html#readpixels">Reading back pixels</a> in WebGL 1 are described here.</p>
+    <dl class="methods">
+        <dt class="idl-code">void readPixels(GLint x, GLint y, GLsizei width, GLsizei height, GLenum format, GLenum type, ArrayBufferView? pixels)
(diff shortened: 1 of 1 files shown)


specs/latest/2.0/index.html
```diff
@@ -1818,6 +1818,29 @@ <h2 class="no-toc">Table of contents</h2>
...
+        <dd>
+          <p>If a WebGLBuffer is bound to the <code>PIXEL_PACK_BUFFER</code> target, generates an <code>INVALID_OPERATION</code> error.</p>
+        </dd>
+
+        <dt class="idl-code">void readPixels(GLint x, GLint y, GLsizei width, GLsizei height, GLenum format, GLenum type, GLintptr offset)
+            <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.3.pdf#nameddest=section-4.3.2">OpenGL ES 3.0 &sect;4.3.2</a>, <a class="nonnormative" href="http://www.khronos.org/opengles/sdk/docs/man3/html/glReadPixels.xhtml">man page</a>)</span>
+        </dt>
+        <dd>
+          <p>If no WebGLBuffer is bound to the <code>PIXEL_PACK_BUFFER</code> target, generates an <code>INVALID_OPERATION</code> error.</p>
+          <p><em>offset</em> is the byte offset into the WebGLBuffer's data store; generates an <code>INVALID_VALUE</code> error if it's less than 0.</p>
```
kenrussell: Should an INVALID_OPERATION error should be generated if the offset + the number of bytes that would be read back is > the length of the bound buffer?

kenrussell:
LGTM -- if you could please squash commits and address the question above, please feel free to merge yourself.

specs/latest/2.0/index.html
zhenyao: Out-of-bound memory access case is well defined in the spec where a INVALID_OPERATION will be generated.  It is among a bunch of other error cases that we skip because it's in the ES3 spec.

However, I specifically mentioned offset < 0 cases here to be consistent with VertexAttribPointer situation where a negative case generates an INVALID_VALUE instead of INVALID_OPERATION.

PR #1059, opened 2015-06-12
Allow both ES5 and ES6 behavior in typed array unit test.

kenrussell:
While the WebGL conformance suite doesn't focus on ECMAScript
behavior, this test is still useful, and should work with both ES
versions for the time being.

Related to #1052.


kenrussell:
Please review @toji @zhenyao @RafaelCintron . Thanks.

toji:
LGTM

RafaelCintron:
LGTM after the fact.

PR #1060, opened 2015-06-17
Unbind PIXEL_UNPACK_BUFFER after testing.

djg:
Leaving PIXEL_UNPACK_BUFFER bound causes gl.texImage2D(...) to fail in
`getFramebufferAttachmentParameter` section.


kenrussell:
Thanks for catching this and for this fix.

PR #1063, opened 2015-06-18
framebufferTexture2D can support mipmap level in WebGL 2 rendering context

Richard-Yunchao:
fixed a small typo too. PTAL.


kenrussell:

PR #1064, opened 2015-06-24
Added Negative, Draw, Multisample and TextureShadow tests.

jasomobica:


kenrussell:
Excellent work. Thank you for being so thorough with this port. Looks perfect as far as I can tell. Merging now so we can begin running these tests.

PR #1065, opened 2015-06-24
Added basic WebGL 2 query object management tests

toji:
Not the full set of tests needed by any means, but verifies baseline query object behavior.


kenrussell:
Looks good. Thanks for this contribution. Has this been tested on any WebGL 2.0 implementations?

toji:
Ours: https://codereview.chromium.org/1210673002/

kenrussell:
Great, thanks.

PR #1067, opened 2015-06-25
Improve and fix invalid enum tests

Oletus:


zhenyao:
Looks good.  Thanks for fixing them.

PR #1070, opened 2015-06-25
Added tests for the remaining WebGL 1 removed enums

matthewn4444:
These removed enums are specified in constants-and-properties.html.
This change adds tests for the remaining enums that did not have
their own test already.


zhenyao:
Looks good

PR #1071, opened 2015-06-26
Move js files from resources/ to js/

zhenyao:
Here we copied over the js files instead of moving, so we can fix paths in various tests in multiple CLs.

Once all paths are fixed, the js files in resources/ can be removed.


zhenyao:
I plan to get this migration done this weekend, so I'll land them unreviewed. It's basically file move and path fixes.

PR #1084, opened 2015-06-30
Last CL for resources/ to js/ migration (I hope)

zhenyao:
Deleted the js files in resources/, which exposed a bunch of overlooked files.

Also fixing them in this CL.


kenrussell:
Very nice. Thank you for pushing through this cleanup.

PR #1085, opened 2015-06-30
Don't remove non-whitespace characters when stripping indentations

Oletus:
Some dEQP test shaders are indented in an inconsistent way: for example,
some lines might be indented with \<tab\>\<tab\> whereas others have
\<tab\>\<space\>\<tab\>. These two indentations may look the same in an editor,
but when the framework indentation removal function looks at the
indentations, it counts <tab> as indentation by 4 and <space> as
indentation by one. The first line is used as reference and the rest of
the lines get the same amount of indentation removed. With inconsistent
indentation in the source, this could remove some meaningful characters,
like turning "void main()" into "oid main()" unless it is separately
checked that no non-whitespace characters are removed.

This fixes several shader tests, particularly in
deqp/data/gles2/shaders/constants.html.

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsShaderLibrary.js
+++ b/sdk/tests/deqp/modules/shared/glsShaderLibrary.js
@@ -100,7 +100,8 @@ var gluShaderUtil = framework.opengl.gluShaderUtil;
             for (var i = 0; i < arr.length; ++i) {
             /** @type {number} */ var removed = 0;
             /** @type {number} */ var j;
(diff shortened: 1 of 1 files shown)


Oletus:
FYI @jasomobica - might be nicer to actually fix the indentations in the test files - for example convert them to use purely spaces, but this is a faster way to get the tests working.

kenrussell:
Yikes. Thank you for catching this serious logic error. I agree this is the most expedient and robust fix.

PR #1086, opened 2015-07-01
Move conformance2/textures tests into conformance2/textures/misc

zhenyao:
So we can put tests for canvas/image/video into sub folders.

Also, refactor tex-image-and-sub-image-2d-with-canvas to accordomate ES3 needs.


toji:
LGTM

PR #1087, opened 2015-07-02
Replace deprecated JavaScript media types with application/javascript

alexhenrie:
application/x-javascript and text/javascript were both deprecated in
April 2006 by RFC 4329. WebGL does not require them, and the tests
should still pass even if a web browser drops support for the deprecated
names.

### DIFF
--- a/sdk/demos/google/high-dpi/index.html
+++ b/sdk/demos/google/high-dpi/index.html
@@ -32,12 +32,12 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 -->
     <title>Teapot Test</title>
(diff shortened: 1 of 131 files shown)


kenrussell:
Thanks for this contribution. What's the motivation for making this change now?

Regardless, I'd like to ask you to please update this pull request to limit your changes to the sdk/ and specs/ directories, and undo your changes to the files under conformance-suites/. The conformance-suites/ directories are snapshots of the top-of-tree tests under sdk/tests/, and our working group prefers to limit changes to those tests to serious bug fixes; this doesn't qualify. If you could please update this pull request and squash commits so the history of those files is untouched, we'll merge it. Thanks.

alexhenrie:
Done. I'm bringing this issue to your attention because I want to propose removing some of Firefox's support for application/x-javascript, and I don't want anyone complaining "but the WebGL tests use it!"

kenrussell:
Understood. Thanks for this contribution. If we need to consider modifying more files (such as the most recent conformance suite snapshot) we can do that later.

alexhenrie:
Thank you!

PR #1090, opened 2015-07-02
add conformance test to readBuffer for WebGL 2

Richard-Yunchao:
PTAL when you have time. Thanks!

### DIFF
--- a/sdk/tests/conformance2/renderbuffers/readbuffer.html
+++ b/sdk/tests/conformance2/renderbuffers/readbuffer.html
@@ -1,0 +1,184 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Test readBuffer Against WebGL 2</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<canvas id="canvas" width="20" height="20"> </canvas>
+<script>
+"use strict";
+description("This tests reading from fbo");
+
+var clearDrawingbuffer = function(color) {
+  gl.clearColor(color[0] / 255, color[1] / 255, color[2] / 255, color[3] / 255);
+  gl.clear(gl.COLOR_BUFFER_BIT);
+}
+
+var validateReadingFromFramebuffer = function(color, expected, msg) {
+  var pixels = new Uint8Array(1 * 1 * 4);
+  gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, pixels);
+  wtu.glErrorShouldBe(gl, expected, msg);
+  if (expected == gl.NO_ERROR)
+    wtu.checkCanvasRect(gl, 0, 0, canvas.width, canvas.height, color,
+    "the color should be [" + color + "]");
+}
+
+var setupRenderbuffer = function(attachment) {
+  var renderbuffer = gl.createRenderbuffer();
+  gl.bindRenderbuffer(gl.RENDERBUFFER, renderbuffer);
+  gl.framebufferRenderbuffer(gl.FRAMEBUFFER, attachment, gl.RENDERBUFFER, renderbuffer);
+  gl.renderbufferStorage(gl.RENDERBUFFER, gl.RGBA8, canvas.width, canvas.height);
+  return renderbuffer;
+}
+
+var testReadBufferOnDefaultFB = function() {
+  gl.readBuffer(gl.NONE);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+      "calling readBuffer with GL_NONE on the default framebuffer should succeed.");
+  gl.readBuffer(gl.BACK);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+      "calling readBuffer with GL_BACK on the default framebuffer should succeed.");
+
+  gl.readBuffer(gl.COLOR_ATTACHMENT0);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+      "calling readBuffer with GL_COLOR_ATTACHMENT0 on the default framebuffer should generate INVALID_OPERATION.");
+}
+
+var testReadbufferOnFBO = function() {
+  gl.readBuffer(gl.BACK);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+      "calling readBuffer with GL_BACK on fbo should generate INVALID_OPERATION.");
+
+  gl.readBuffer(gl.NONE);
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/renderbuffers/readbuffer.html
```diff
@@ -0,0 +1,192 @@
...
+  shouldBe('gl.getParameter(gl.READ_BUFFER)', 'gl.COLOR_ATTACHMENT1');
+
+  gl.copyTexImage2D(gl.TEXTURE_2D, 0, gl.RGBA, 0, 0, canvas.width, canvas.height, 0);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+      "should generate INVALID_OPERATION when reading from GL_COLOR_ATTACHMENT1 but this attachment has no image currently.");
+
+  var colorbuffer1 = gl.createRenderbuffer();
+  gl.bindRenderbuffer(gl.RENDERBUFFER, colorbuffer1);
+  gl.renderbufferStorage(gl.RENDERBUFFER, gl.RGBA8, canvas.width, canvas.height);
+  gl.framebufferRenderbuffer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT1, gl.RENDERBUFFER, colorbuffer1);
```
Richard-Yunchao: the status of framebuffer is incomplete, so several tests below will fail in Chromium with my patch. Is there anything wrong in this test?

sdk/tests/conformance2/renderbuffers/readbuffer.html
```diff
@@ -0,0 +1,192 @@
...
+  shouldBe('gl.getParameter(gl.READ_BUFFER)', 'gl.BACK');
+
+  var tex = gl.createTexture();
+  gl.bindTexture(gl.TEXTURE_2D, tex);
+  gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, canvas.width, canvas.height, 0,
+                      gl.RGBA, gl.UNSIGNED_BYTE, null);
+  gl.copyTexImage2D(gl.TEXTURE_2D, 0, gl.RGBA, 0, 0, canvas.width, canvas.height, 0);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+      "should be no errors when reading from GL_BACK on the default framebuffer.");
+  testPixels(0, 0, backColor);
```
kenrussell: testPixels is redundant -- there is already a similar function in WebGLTestUtils.checkCanvasRect. Please use that instead.

sdk/tests/conformance2/renderbuffers/readbuffer.html
```diff
@@ -0,0 +1,192 @@
...
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+      "calling readBuffer with GL_COLOR_ATTACHMENT0 on the default framebuffer should generate INVALID_OPERATION.");
+
+  shouldBe('gl.getParameter(gl.READ_BUFFER)', 'gl.BACK');
+
+  var tex = gl.createTexture();
+  gl.bindTexture(gl.TEXTURE_2D, tex);
+  gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, canvas.width, canvas.height, 0,
+                      gl.RGBA, gl.UNSIGNED_BYTE, null);
+  gl.copyTexImage2D(gl.TEXTURE_2D, 0, gl.RGBA, 0, 0, canvas.width, canvas.height, 0);
```
kenrussell: The results of the copyTexImage2D calls here and throughout the test are being thrown away -- at no point are the contents of the texture being validated. Please either remove the copyTexImage2D calls, or verify their results by binding the texture to a framebuffer object and using WebGLTestUtils.checkCanvasRect.

sdk/tests/conformance2/renderbuffers/readbuffer.html
```diff
@@ -0,0 +1,192 @@
...
+  var maxColorAttachments = gl.getParameter(gl.MAX_COLOR_ATTACHMENTS);
+  gl.readBuffer(gl.COLOR_ATTACHMENT0 + maxColorAttachments);
+  wtu.glErrorShouldBe(gl, gl.INVALID_ENUM,
+      "calling readBuffer with GL_COLOR_ATTACHMENTi that exceeds MAX_COLOR_ATTACHMENT on fbo should generate INVALID_ENUM.");
+  gl.readBuffer(gl.COLOR_ATTACHMENT1);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+      "calling readBuffer with GL_COLOR_ATTACHMENT1 on the fbo should succeed.");
+  shouldBe('gl.getParameter(gl.READ_BUFFER)', 'gl.COLOR_ATTACHMENT1');
+
+  gl.copyTexImage2D(gl.TEXTURE_2D, 0, gl.RGBA, 0, 0, canvas.width, canvas.height, 0);
```
kenrussell: Note that this could just as well be a readPixels operation.

sdk/tests/conformance2/renderbuffers/readbuffer.html
kenrussell: I can't see anything wrong with it, but several parts of this test are failing in both Chromium and Firefox, so it's hard to tell whether the WebGL 2.0 implementations are broken (likely) or the test has bugs.

sdk/tests/conformance2/renderbuffers/readbuffer.html
```diff
@@ -0,0 +1,192 @@
...
+  gl.bindRenderbuffer(gl.RENDERBUFFER, colorbuffer1);
+  gl.renderbufferStorage(gl.RENDERBUFFER, gl.RGBA8, canvas.width, canvas.height);
+  gl.framebufferRenderbuffer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT1, gl.RENDERBUFFER, colorbuffer1);
+  var fboColor1 = [0, 0, 255, 255];
+  gl.clearColor(fboColor1[0] / 255, fboColor1[1] / 255, fboColor1[2] / 255, fboColor[3] / 255);
+  gl.drawBuffers([gl.NONE, gl.COLOR_ATTACHMENT1]);
+  gl.clear(gl.COLOR_BUFFER_BIT);
+  gl.copyTexImage2D(gl.TEXTURE_2D, 0, gl.RGBA, 0, 0, canvas.width, canvas.height, 0);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+      "should be no errors when reading from GL_COLOR_ATTACHMENT1 on fbo.");
```
kenrussell: It's very confusing that the gl.readBuffer(gl.COLOR_ATTACHMENT1) call was made 20 lines above from this call. Please try to keep the tests simple, and minimize the state dependencies between different portions of the test. I note that there is redundant FBO and renderbuffer setup in multiple places in this test which could be factored out into helper functions.

sdk/tests/conformance2/renderbuffers/readbuffer.html
```diff
@@ -0,0 +1,192 @@
...
+  // switch to another fbo, read buffer is GL_COLOR_ATTACHMENT0, not GL_COLOR_ATTACHMENT1
+  var fb1 = gl.createFramebuffer();
+  gl.bindFramebuffer(gl.FRAMEBUFFER, fb1);
+  var buffer = gl.createRenderbuffer();
+  gl.bindRenderbuffer(gl.RENDERBUFFER, buffer);
+  gl.framebufferRenderbuffer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.RENDERBUFFER, buffer);
+  gl.renderbufferStorage(gl.RENDERBUFFER, gl.RGBA8, canvas.width, canvas.height);
+
+  shouldBe('gl.getParameter(gl.READ_BUFFER)', 'gl.COLOR_ATTACHMENT0');
+  var fbo1Color = [255, 255, 255, 255];
```
kenrussell: There are variables named "fboColor1" and "fbo1Color" in this test. This is extremely confusing and will make this test hard to maintain. Please use reasonably named variables. In this case, you could use "blue", "white", etc.

sdk/tests/conformance2/renderbuffers/readbuffer.html
```diff
@@ -0,0 +1,192 @@
...
+  gl.deleteFramebuffer(fb);
+  gl.deleteRenderbuffer(colorbuffer);
+  gl.deleteRenderbuffer(colorbuffer1);
+  gl.deleteFramebuffer(fb1);
+  gl.deleteRenderbuffer(buffer);
+  gl.deleteTexture(tex);
+  shouldBe('gl.getParameter(gl.READ_BUFFER)', 'gl.BACK');
+  gl.copyTexImage2D(gl.TEXTURE_2D, 0, gl.RGBA, 0, 0, canvas.width, canvas.height, 0);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+      "should be no errors when reading from GL_BACK after switch to default fb from fbo.");
```
kenrussell: The texture which was the target of the copyTexImage2D call was deleted four lines above, and was bound to the TEXTURE_2D target, so it's as though bindTexture(TEXTURE_2D, null) was called, and WebGL doesn't allow copyTexImage2D to work without a texture bound to the target texture unit. Therefore this call will produce an INVALID_OPERATION error. Per discussion above, I think it should be removed, and readPixels calls used everywhere in this test.

kenrussell:
Thanks for this contribution. I strongly feel that the code should be refactored and simplified to be made more clear. Right now the entire test has to be read from top to bottom in order to understand how the test works. Instead of this, it should be separated into logically separate portions, helper functions should be written for setup of objects like renderbuffers and framebuffers, and more clear variable names should be used.

Richard-Yunchao:
@kenrussell , thanks for your careful reviewing. I have refactored the code, please take a look when you have time. Thanks a lot! 

The problem of framebuffer incomplete still exists, which lead to some failures. but I think it is not caused by the test cases.

kenrussell:
Thanks, this looks much better. Please continue to keep in mind the design goal to minimize dependencies between different parts of the tests. Merging now.

kenrussell:
BTW, I agree that it's likely the incomplete framebuffer problem is a bug in implementations.

PR #1095, opened 2015-07-06
Loosen global variable initialization spec to meet implementations

Oletus:
Historically WebGL implementations have not posed many restrictions to
operations or variables in global variable initializers. This is a
violation of the OpenGL ES Shading Language spec.

Specify what is allowed in global variable initializers in WebGL.
The new spec is halfway between the actual ESSL spec and the old
de-facto functionality. It is intended to avoid widespread compatibility
regressions in browsers while bringing WebGL closer to the restrictions
in the ESSL spec.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -28,7 +28,7 @@
     <!--end-logo-->
 
     <h1>WebGL Specification</h1>
-    <h2 class="no-toc">Editor's Draft 9 June 2015</h2>
+    <h2 class="no-toc">Editor's Draft 7 July 2015</h2>
     <dl>
         <dt>This version:
             <dd>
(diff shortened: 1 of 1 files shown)


specs/latest/1.0/index.html
```diff
@@ -3788,6 +3788,49 @@ <h2 class="no-toc">Table of contents</h2>
...
+    The GLSL ES <a href="#refsGLES20GLSL">[GLES20GLSL]</a> spec restricts global variable
+    initializers to be constant expressions. In the WebGL API, it is allowed to use other global
+    variables not qualified with the <code>const</code> qualifier and uniform values in global
+    variable initializers. Global variable initializers must be global initializer expressions,
+    which are defined as one of:
+</p>
+<ul>
+<li>a literal value</li>
+<li>a global variable</li>
+<li>an uniform</li>
```
kenrussell: As strange as it sounds, I think this should be "a uniform".

kenrussell:
Thanks Olli for so thoroughly investigating and clarifying the rules here.

What do you think should be done for WebGL 2.0? Should these restrictions be enforced for ESSL 3.00 shaders? If so, does the WebGL 2.0 spec need to be updated to effectively remove this new subsection?

Do the conformance tests yet test all of these behaviors? If not, could you file an issue about adding new tests to do so?

@grorg or @jdashg should review and merge this, but it LGTM as is.

Oletus:
I amended the commit to fix typos and clarified that it applies to ESSL 1.00 shaders. I think ESSL 3.00 shaders in WebGL 2.0 should just work according to the normal ESSL 3.00 spec - meaning that the constant expression restriction would be strictly enforced.

I can also work on conformance tests for this.

kenrussell:
Thanks for clarifying that this applies only to ESSL 1.00 shaders. LGTM again.

@grorg @jdashg any comments?

kenrussell:
Merging this now. If there are any objections from the spec editors later, it can be revisited.

PR #1097, opened 2015-07-06
Switch to use texture image related test generator in WebGL 1 conform…

zhenyao:
…ance.

Also, fix a few minor bugs.


zhenyao:
This CL is a bit hard to review, but it's quite simple.

1) fix the RGB9_E5 bug from previous CL (kbr pointed out after it's merged)
2) restructure the conformance/textures tests the same as conformance2/textures, i.e., put most tests under misc/, and then canvas/ image/ image_data/ svn_image/ video/ webgl_canvas/ folders where tests are generated by the python script
3) remove a few unused args in the tex test scripts, so the generator can handle various tests in the same way
4) fix the image-data html (in py) because it requires a second canvas and different canvas size

zhenyao:
webgl 1 conformance tested with this CL.  textures/ all passing

kenrussell:
Sorry for the delay reviewing. Looks excellent. Thanks for this significant improvement!

PR #1098, opened 2015-07-07
Fix uniformBlockShader.vert

Oletus:
The shader is supposed to be in ESSL 3.00, with "in" instead of
"attribute".

Also, it is somewhat unclear in the spec whether a semicolon is required
after a uniform block declaration (it's not present in all the examples),
but the grammar seems to suggest so.


kenrussell:
Thanks for the fix. I assume the harness injects the #version directive at the beginning?

Oletus:
The file actually has the version directive before the copyright comment.

kenrussell:
Oops, I didn't click the "expand" button enough times. Thanks for pointing that out.

PR #1100, opened 2015-07-07
Add rationale for forbidding a buffer to be bound to transformfeedbac…

zhenyao:
…k and another binding point.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -2630,6 +2630,12 @@ match the <em>type</em> according to the <a href="#TEXTURE_PIXELS_TYPE_TABLE">ab
         restrictions in the WebGL 1 specification</a>.
     </p>
 
+    <div class="note rationale">
+        This restriction has been added to prevent writing to index buffers on the GPU, which would make
+        doing any CPU-side checks on index data prohibitively expensive. Handling index buffers as different
+        from other buffer data also maps better to the Direct3D API.
+    </div>
+
     <p>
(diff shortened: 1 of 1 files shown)


specs/latest/2.0/index.html
```diff
@@ -2637,9 +2643,7 @@ <h2 class="no-toc">Table of contents</h2>
     </p>
 
     <div class="note rationale">
-        This restriction has been added to prevent writing to index buffers on the GPU, which would make
-        doing any CPU-side checks on index data prohibitively expensive. Handling index buffers as different
-        from other buffer data also maps better to the Direct3D API.
+        This restriction has been added to prevent undefined values when reading from or writing to a buffer object that is simultanenously bound to a transform feedback buffer binding point and elsewhere in the GL. An example of such bindings is <code>ReadPixels</code> to a pixel buffer object binding point.
```
kenrussell: Could you please list the section in the ES 3.0.3 spec that results in undefined behavior here? We dug it up this morning.

Also, "an example of such bindings" is a little confusing -- did you mean "an example of such a situation"?

Finally, could you please update the last modified date of the spec?

kenrussell:
Thanks for taking care of this. Couple of minor comments.

zhenyao:
Updated, please take another look.

kenrussell:
Looks good to me. If no comments from @grorg or @jdashg by tomorrow, let's merge it then.

kenrussell:
Merging now.

PR #1102, opened 2015-07-07
Test max token size of 1024 in WebGL 2.

zhenyao:
Also, mark the 257 tests as WebGL 1 only.


toji:
Who would DO that? Seriously. o_O
LGMT.

zhenyao:
I don't think anyone would write such tokens manually, but auto-generated code could reach the limit.

kenrussell:
Sorry -- not LGTM. These rules should apply only to ESSL 3.00 shaders, and not affect the behavior of ESSL 1.00 shaders. Correct? WebGL 1.0 contexts should not change behavior because of the presence of a WebGL 2.0 context. If this is correct, please revert this change, modify it, and re-upload it. Thanks.

zhenyao:
Per our offline discussion, in theory we can make the token size per ESSL version, or per WebGL version. However, it seems cleaner that we have one max token size per WebGL version, so we don't need to query the shader version in functions like getUniformLocation() because we need to generate an INVALID_OPERATION if the arg name is longer than max token size.

I will send in other pull request to update the spec to make this rule cleaner.

zhenyao:
@Oletus @jdashg Do you have an opinion on this?

kenrussell:
Thanks. After our offline discussion I can definitely see the argument for raising the token size limit for all shaders in WebGL 2.0, as opposed to just the ESSL 3.00 shaders.

PR #1104, opened 2015-07-08
add test cases for reading from framebuffer and readBuffer is GL_NONE

Richard-Yunchao:
PTAL. Thanks a lot!


kenrussell:
Thanks for this improvement.

PR #1106, opened 2015-07-08
Fix a bug in texture tests in extensions/

zhenyao:
This is caused by texture-image auto-generation refactoring.


kenrussell:
Thanks for fixing this.

Issue #1099, opened 2015-07-07
Write conformance tests for global variable initializers


Oletus:
Tests need to be written for rules specified in #1095

PR #1108, opened 2015-07-09
Add a test for global variable initializer restrictions

Oletus:
For issue #1099


zhenyao:
Thanks for the thorough test cases.  LGTM.

PR #1110, opened 2015-07-09
Add a test for short-circuit behavior in a loop condition

matthewn4444:


sdk/tests/conformance2/glsl3/short-circuiting-in-loop-condition.html
```diff
@@ -0,0 +1,91 @@
...
+  description("Test behavior of a short-circuiting operator in a loop condition using a function call with side effects");
+  debug("");
+
+  var wtu = WebGLTestUtils;
+  var gl = wtu.create3DContext(undefined, undefined, 2);
+  if (!gl) {
+    testFailed("context does not exist");
+  } else {
+    var program = wtu.setupProgram(gl, ["vertex-shader", "fshader"]);
+    gl.uniform1i(gl.getUniformLocation(program, "u"), 1);
```
zhenyao: Why don't you also test the other case, setting u to 0, and test that foo() isn't executed?

matthewn4444:
I updated with a false condition in the loop.

sdk/tests/conformance2/glsl3/short-circuiting-in-loop-condition.html
```diff
@@ -0,0 +1,99 @@
...
+  int iterations = 0;
+
+  while(u && foo()) {
+    if (iterations > 10) {
+      break;
+    }
+    ++iterations;
+  }
+
+  result = sideEffectCounter == 12 ? vec4(0, 1.0, 0, 1.0) : vec4(0, 1.0, 0, 0);
```
zhenyao: I think here you also want to test sideEffectCount == 0, because if u set to false, then if foo() is not evaluated (which is correct behavior), then sideEffectCount == 0. Otherwise sideEffectCount == 1.  The current code won't be able to verify this.

matthewn4444:
Added the condition for u is false and side effect counter is 0. Now both are green if test works.

zhenyao:
Great!  Thank you.

PR #1111, opened 2015-07-09
Added simple sampler object management tests

toji:


sdk/tests/conformance2/sampler/sampler.html
```diff
@@ -0,0 +1,138 @@
...
+    
+    shouldBe("gl.SAMPLER_BINDING", "0x8919");
+    
+    gl.getParameter(gl.SAMPLER_BINDING);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "SAMPLER_BINDING query should succeed");
+    
+    // Default value is null
+    shouldBe("gl.getParameter(gl.SAMPLER_BINDING)", "null");
+    
+    debug("Testing binding a Transform Feedback object");
```
kenrussell: The test description's incorrect.

sdk/tests/conformance2/00_test_list.txt
```diff
@@ -6,6 +6,7 @@ misc/00_test_list.txt
 query/00_test_list.txt
 renderbuffers/00_test_list.txt
 rendering/00_test_list.txt
+sampler/00_test_list.txt
```
zhenyao: nit: samplers, to be consistent with textures/renderbuffers

sdk/tests/conformance2/sampler/sampler.html
```diff
@@ -0,0 +1,138 @@
...
+function runBindingTest() {
+    debug("Testing binding enum");
+    
+    shouldBe("gl.SAMPLER_BINDING", "0x8919");
+    
+    gl.getParameter(gl.SAMPLER_BINDING);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "SAMPLER_BINDING query should succeed");
+    
+    // Default value is null
+    shouldBe("gl.getParameter(gl.SAMPLER_BINDING)", "null");
```
zhenyao: You can avoid the previous getParameter call and move the glError after this one

zhenyao: Also, you can use shouldBeNull() here

sdk/tests/conformance2/sampler/sampler.html
```diff
@@ -0,0 +1,138 @@
...
+    
+    debug("Testing binding a Transform Feedback object");
+    s1 = gl.createSampler();
+    s2 = gl.createSampler();
+    shouldBeNull("gl.getParameter(gl.SAMPLER_BINDING)");
```
zhenyao: This one is unnecessary, just tested above

sdk/tests/conformance2/sampler/sampler.html
```diff
@@ -0,0 +1,138 @@
...
+}
+
+function runObjectTest() {
+    debug("Testing object creation");
+    
+    s1 = gl.createSampler();
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "createSampler should not set an error");
+    shouldBeNonNull("s1");
+    
+    // Expect true, event if never bound
```
kenrussell: Typo: event -> even

kenrussell:
This is a superb test. Clear, concise, and easy to read. Thank you for putting it together. Please merge once the above minor typos have been addressed.

@Richard-Yunchao : I'd like to suggest that you study this test as an example of how to keep the OpenGL state changes self-contained and easy-to-understand in a WebGL conformance test.

toji:
Thanks for the quick turnaround on the reviews!

Richard-Yunchao:
@kenrussell , yeah, this test is clear and easy to read. It is a very good sample. I will try to make my code to be self-contained and easy-to-understand. Thank you, Ken.

Richard-Yunchao:
From the perspective of Blink/Chromium, we need to add conformance tests against WebGL 2.0. Most of the code for this purpose will be added into conformance2/ filefolder. I wonder what about other tasks in this project? Are there some tasks about ESSL 300 and/or deqp? And/or any other tasks? @kenrussell , @zhenyao , @toji , could you tell me some tasks, I can help on this.

kenrussell:
@Richard-Yunchao let's discuss that via email -- this pull request is not the best place to have a discussion about follow-on work. In general though the goal should be to add only minimal tests to conformance2/, as the drawElements tests that Mobica is porting in sdk/tests/deqp/ are intended to be the majority of the WebGL 2.0 conformance suite. Contributions toward passing those tests would be appreciated.

PR #1112, opened 2015-07-09
Add a wtu function to query the default context version.

zhenyao:
Hook it up with gl-drawelements.html so it runs correctly in both WebGL 1/2.

Also, mark a few tests as WebGL 1 only.


toji:
LGTM

PR #1113, opened 2015-07-10
Small spec updates

zhenyao:
1) allow DEPTH_STENCIL in renderbufferStorage
2) max token size 1024 for both ESSL1 and ESSL3 shaders.


kenrussell:
LGTM too FWIW.

PR #1114, opened 2015-07-10
Fixed failing program creation in gl-enum-tests

toji:
The shader this was trying to load previously was `#version 300 es`, but this js file was used by both WebGL 1 and WebGL 2 tests. Switched it to use something that works for both.


zhenyao:
LGTM.  Since it's passing currently with WebGL 1, does that mean we have a bug in chrome impl?

PR #1115, opened 2015-07-10
Add WebGL 2 version of framebuffer-object-attachment.html

zhenyao:
Note that I didn't add tests for various combinations because deqp already has
a very thorough tests for that.

### DIFF
--- a/sdk/tests/conformance2/renderbuffers/framebuffer-object-attachment.html
+++ b/sdk/tests/conformance2/renderbuffers/framebuffer-object-attachment.html
@@ -1,0 +1,421 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+
+<script>
+"use strict";
+var wtu = WebGLTestUtils;
+var gl;
+
+function checkFramebuffer(expected) {
+    var actual = gl.checkFramebufferStatus(gl.FRAMEBUFFER);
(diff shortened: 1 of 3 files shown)


sdk/tests/conformance2/renderbuffers/framebuffer-object-attachment.html
```diff
@@ -0,0 +1,450 @@
...
+    gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+    var colorBuffer = gl.createRenderbuffer();
+    gl.bindRenderbuffer(gl.RENDERBUFFER, colorBuffer);
+    gl.framebufferRenderbuffer(
+        gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.RENDERBUFFER, colorBuffer);
+    gl.renderbufferStorage(gl.RENDERBUFFER, gl.RGBA4, 16, 16);
+    checkFramebuffer([gl.FRAMEBUFFER_COMPLETE]);
+
+    debug("");
+    debug("Wrong storage type for type of attachment be FRAMEBUFFER_INCOMPLETE_ATTACHMENT");
```
kenrussell: typo: be -> should be

sdk/tests/conformance2/renderbuffers/framebuffer-object-attachment.html
```diff
@@ -0,0 +1,450 @@
...
+    checkFramebuffer([gl.FRAMEBUFFER_COMPLETE]);
+    checkBufferBits();
+
+    debug("STENCIL_ATTACHMENT overwrites stencil set by DEPTH_STENCIL_ATTACHMENT")
+    gl.framebufferRenderbuffer(
+        gl.FRAMEBUFFER, gl.DEPTH_STENCIL_ATTACHMENT, gl.RENDERBUFFER, depthStencilBuffer);
+    gl.framebufferRenderbuffer(
+        gl.FRAMEBUFFER, gl.STENCIL_ATTACHMENT, gl.RENDERBUFFER, null);
+    checkFramebuffer([gl.FRAMEBUFFER_COMPLETE]);
+    checkBufferBits(gl.DEPTH_ATTACHMENT);
```
kenrussell: Very nice test here. Subtle.

sdk/tests/conformance2/renderbuffers/framebuffer-object-attachment.html
```diff
@@ -0,0 +1,450 @@
...
+    checkFramebuffer([gl.FRAMEBUFFER_INCOMPLETE_MISSING_ATTACHMENT]);
+
+    gl.deleteRenderbuffer(colorBuffer);
+    gl.deleteFramebuffer(fbo);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+}
+
+function testFramebufferIncompleteDimensions() {
+    debug("");
+    debug("Attachments of different sizes should be FRAMEBUFFER_INCOMPLETE_DIMENSIONS");
```
kenrussell: The ES 2.0 restrictions in this area no longer apply to ES 3.0. Please see "Whole Framebuffer Completeness" in section 4.4.4 "Framebuffer Completeness" in the ES 3.0 spec. Are you sure this block of tests is correct? It might be asserting old WebGL 1.0 behaviors.

sdk/tests/conformance2/renderbuffers/framebuffer-object-attachment.html
```diff
@@ -0,0 +1,450 @@
...
+    gl.bindRenderbuffer(gl.RENDERBUFFER, depthBuffer);
+    gl.framebufferRenderbuffer(
+        gl.FRAMEBUFFER, gl.DEPTH_ATTACHMENT, gl.RENDERBUFFER, depthBuffer);
+    gl.renderbufferStorage(gl.RENDERBUFFER, gl.DEPTH_COMPONENT16, 16, 16);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+    checkFramebuffer([gl.FRAMEBUFFER_COMPLETE]);
+
+    // We pick this combination because it works on desktop OpenGL but should not work on OpenGL ES 2.0
+    gl.renderbufferStorage(gl.RENDERBUFFER, gl.DEPTH_COMPONENT16, 32, 16);
+    checkFramebuffer([gl.FRAMEBUFFER_INCOMPLETE_DIMENSIONS]);
```
kenrussell: Is this correct? See above.

kenrussell:
Thanks for this test. It looks very good overall -- in particular the tests of depth/stencil attachment behaviors -- but I think there may be some bugs around framebuffer attachments of different sizes. Could you please re-check?

sdk/tests/conformance2/renderbuffers/framebuffer-object-attachment.html
zhenyao: You are right.  Chrome implementation generates this error, so we need to update the chrome implementation and this test.

zhenyao:
Revised.  Please take another look.

kenrussell:
Looks excellent. Thanks for being so thorough with this test.

PR #1117, opened 2015-07-15
Added test for bug in bound attribute locations in program binaries o…

kenrussell:
…n some GPUs.


toji:
Test LGTM. Behavior under test makes me want to cry. :(

PR #1119, opened 2015-07-16
Move test to function to remove 'use strict' parsing failure.

### LINKED ISSUE #1118: WebGL 2 conformance - renderbuffer-test.html fails with strict mode
In Firefox, renderbuffer/framebuffer-test.html fails with following error:
`SyntaxError: in strict mode code, functions may be declared only at top level or immediately within another function framebuffer-test.html:186:11`

djg:
Reorganized the JS code to fix the strict mode failure to resolve #1118.


kenrussell:
Very nice. Thanks for this fix.

PR #1121, opened 2015-07-17
Fix getVertexAttribOffset return type.

zhenyao:
Used to be GLsizeiptr, should be GLintptr.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -28,7 +28,7 @@
     <!--end-logo-->
 
     <h1>WebGL Specification</h1>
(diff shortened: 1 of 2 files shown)


toji:
I don't disagree with the space formatting changes, but they are unrelated and it does make it harder to spot the relevant change. I'd prefer that they were in at least a separate commit, if not a separate pull request. Sorry!

zhenyao:
The whitespace removal is automatic.  I landed another CL to just fix the whitespace removal and rebase this one.  So now it's only meaningful change.

toji:
Thanks!

PR #1122, opened 2015-07-17
Fix webgl.idl unnecessary whitespaces.

zhenyao:
Just run make once, it's removing all of them automatically.


zhenyao:
No meaningful change, just run make to get rid of whitespaces.  Merging myself.

PR #1123, opened 2015-07-17
fix a small bug in tex-image-and-sub-image-2d-with-canvas.js

zhenyao:
When testing with R or RG formats, we won't get the channels that's not there,
so need to filter the expectations of these channels to be 0.


toji:
LGTM

PR #1124, opened 2015-07-20
Fix format and type validation for TexImage2D and TexSubImage2D

qkmiao:
GL_INVALID_ENUM is generated if internalformat, format, or type is not
an accepted constant. GL_INVALID_OPERATION should be generated if
combination of internalformat, format and type is not valid.

### DIFF
--- a/sdk/tests/js/tests/tex-input-validation.js
+++ b/sdk/tests/js/tests/tex-input-validation.js
@@ -309,15 +309,29 @@ var testCases = [
     internalFormat: gl.RGBA,
     border: 0,
     format: gl.RGBA,
-    type: gl.BYTE,
-    expectedError: gl.INVALID_ENUM },
-  { target: gl.TEXTURE_2D,
-    internalFormat: gl.RGBA,
-    border: 0,
-    format: gl.RGBA,
     type: gl.UNSIGNED_BYTE,
     expectedError: gl.NO_ERROR }
 ];
+
+if (contextVersion < 2) {
+  testCases = testCases.concat([
+    { target: gl.TEXTURE_2D,
+      internalFormat: gl.RGBA,
+      border: 0,
+      format: gl.RGBA,
+      type: gl.BYTE,
+      expectedError: gl.INVALID_ENUM }
+  ]);
+} else {
+  testCases = testCases.concat([
+    { target: gl.TEXTURE_2D,
+      internalFormat: gl.RGBA,
+      border: 0,
+      format: gl.RGBA,
+      type: gl.BYTE,
+      expectedError: gl.INVALID_OPERATION }
+  ]);
+}
 
 for (var ii = 0; ii < testCases.length; ++ii) {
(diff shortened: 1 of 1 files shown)


sdk/tests/js/tests/tex-input-validation.js
```diff
@@ -310,7 +310,7 @@ var testCases = [
     border: 0,
     format: gl.RGBA,
     type: gl.BYTE,
-    expectedError: gl.INVALID_ENUM },
+    expectedError: gl.INVALID_OPERATION},
```
zhenyao: I think INVALID_ENUM is correct for ES2, where BYTE is not a valid enum for type.

sdk/tests/js/tests/tex-input-validation.js
```diff
@@ -330,11 +330,11 @@ testCases = [
   { target: gl.TEXTURE_2D,
     format: 0x1903, // GL_RED
     type: gl.UNSIGNED_BYTE,
-    expectedError: gl.INVALID_ENUM },
+    expectedError: gl.INVALID_OPERATION },
```
zhenyao: Same here, in ES2, RED is not a valid enum for format, so INAVLID_ENUM is correct.

sdk/tests/js/tests/tex-input-validation.js
```diff
@@ -330,11 +330,11 @@ testCases = [
...
   { target: gl.TEXTURE_2D,
     format: gl.RGBA,
     type: gl.BYTE,
-    expectedError: gl.INVALID_ENUM },
+    expectedError: gl.INVALID_OPERATION },
```
zhenyao: Same here, BYTE is not a valid enum for type in ES2, so INVALD_ENUM is correct.

zhenyao:
I think we should modify the test a little bit.  For these cases that generate different error in ES2 and ES3, we should check the error isn't NO_ERROR instead.

qkmiao:
Updated pull request to align with current approach to handle ES2 and ES3.

sdk/tests/js/tests/tex-input-validation.js
```diff
@@ -328,19 +342,36 @@ debug("Checking TexSubImage2D: a set of inputs that are valid in GL but invalid
...
+      expectedError: gl.INVALID_ENUM },
+    { target: gl.TEXTURE_2D,
+      format: gl.RGBA,
+      type: gl.BYTE,
+      expectedError: gl.INVALID_ENUM }
+    ]);
+} else {
+  testCases = testCases.concat([
+    { target: gl.TEXTURE_2D,
+      format: 0x1903, // GL_RED
```
zhenyao: Can you use gl.RED here? it should be defined in WebGL2. Then I can merge this fix.  Thank you.

qkmiao:
Thanks for review. Updated pull request.

zhenyao:
LGTM

PR #1125, opened 2015-07-20
Added Rasterizer Discard, Shader Indexing, Shader Loop and Shader Struct

marcorperez:
Added Rasterizer Discard, Shader Indexing, Shader Loop and Shader Struct Tests for WebGL 2


sdk/tests/deqp/framework/opengl/gluShaderUtil.js
```diff
@@ -178,6 +178,18 @@ gluShaderUtil.DataType = {
 };
 
 /**
+ * @enum {number}
+ */
+gluShaderUtil.ShaderType = {
+    VERTEX: 0,
+	FRAGMENT: 1,
```
kenrussell: Strange indentation.

sdk/tests/deqp/functional/gles3/es3fShaderLoopTests.js
```diff
@@ -0,0 +1,1224 @@
...
+	LOOPCASE_PRE_INCREMENT: 16,									// for (...; ++i) { <body>; }
+	LOOPCASE_POST_INCREMENT: 17,								// for (...; i++) { <body>; }
+	LOOPCASE_MIXED_BREAK_CONTINUE: 18,
+	LOOPCASE_VECTOR_COUNTER: 19,								// for (ivec3 ndx = ...; ndx.x < ndx.y; ndx.x += ndx.z) { ... }
+	LOOPCASE_101_ITERATIONS: 20,								// loop for 101 iterations
+	LOOPCASE_SEQUENCE: 21,										// two loops in sequence
+	LOOPCASE_NESTED: 22,										// two nested loops
+	LOOPCASE_NESTED_SEQUENCE: 23,								// two loops in sequence nested inside a third
+	LOOPCASE_NESTED_TRICKY_DATAFLOW_1: 24,						// nested loops with tricky data flow
+	LOOPCASE_NESTED_TRICKY_DATAFLOW_2: 25						// nested loops with tricky data flow
```
kenrussell: Are all of these loop variants expected to pass? How does the native test harness do on these tests? Have you noticed any loop constructs in the tests whose behavior is outside the scope of the GLSL ES 3.00 spec?

sdk/tests/deqp/functional/gles3/es3fShaderStructTests.js
```diff
@@ -0,0 +1,1951 @@
...
+			'	int				c;\n' +
+			'};\n' +
+			'\n' +
+			'void main (void)\n' +
+			'{\n' +
+			'	S s = S(${COORDS}.x, vec3(0.0), ui_one);\n' +
+			'	s.b = ${COORDS}.yzw;\n' +
+			'	${DST} = vec4(s.a, s.b.x, s.b.y, s.c);\n' +
+			'	${ASSIGN_POS}\n' +
+			'}\n',
```
kenrussell: I hope these shaders were a simple copy-paste and reformat from es3fShaderStructTests.cpp! :)

sdk/tests/deqp/functional/gles3/rasterizerdiscard.html
```diff
@@ -0,0 +1,38 @@
...
+<script>
+testName = 'lifetime';
+setCurrentTestName(testName);
+description("Functional test: " + testName + ".");
+
+var wtu = WebGLTestUtils;
+/** @type {WebGL2RenderingContext} */ var gl = wtu.create3DContextWithWrapperThatThrowsOnGLError('canvas', {preserveDrawingBuffer: true}, 2);
+
+    try{
+        /* TODO: remove next line */
```
kenrussell: This sounds bad; is the error coming from pure context initialization?

It would be better to remove the workaround and let the test fail, so implementations see any errors caused during context creation. @zhenyao @toji

kenrussell:
Thank you, these new tests look great. I only had time to skim through them but they look very good overall. Merging now so that we can start running them; we can iterate on any issues with them later.

sdk/tests/deqp/functional/gles3/rasterizerdiscard.html
jasomobica: Yes, I should've removed that workaround (as TODO suggests).
Currently chrome, at least on my computer, creates the context successfully but it sets the GL error, so the first gl.getError() call returns a non-zero error code.

sdk/tests/deqp/functional/gles3/es3fShaderLoopTests.js
dangro: As far as I can tell, all the tests should pass. All of the native tests pass but I'm observing different behavior on different browsers (~90% of the shaders do not compile on Cr and ~10% of the tests pass; FF is passing on ~100% of the tests with no shader compile errors).

So far, the shaders I've seen don't seem to be non-conformant. I'll consult with the team regarding some questions I have about the spec though.

kenrussell: That's interesting. I wonder whether Firefox is using the (still somewhat incomplete) ANGLE shader translator in their WebGL 2.0 prototype, or bypassing it. Until it's known to be ES 3.0 compliant, it's probably expected that it will reject valid shaders. CC'ing @jamiemadill @vonture @zhenyao .

kenrussell: Also @jdashg .

PR #1126, opened 2015-07-20
Fix the wrong format RGBA4444 (should be RGBA4) in conformance2/textures

zhenyao:


zhenyao:
This is a straightforward bug fix (replace the wrong enum with the correct one).

I'll merge myself.

kenrussell:
Good catch. LGTM after the fact.

PR #1131, opened 2015-07-24
Improve read-pixels-from-fbo-test.html

zhenyao:
Add all the sized ES3 texture formats/types as fbo's color image.

### DIFF
--- a/sdk/tests/conformance2/reading/read-pixels-from-fbo-test.html
+++ b/sdk/tests/conformance2/reading/read-pixels-from-fbo-test.html
@@ -44,54 +44,362 @@ description("Checks that ReadPixels from a fbo works as expected.");
 var wtu = WebGLTestUtils;
 var gl = wtu.create3DContext();
 
-function rgba8_f(buf, index) {
-  return [buf[index]/255.0, buf[index + 1]/255.0, buf[index + 2]/255.0, buf[index + 3]/255.0];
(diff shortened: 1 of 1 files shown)


zhenyao:
This is still working progress.  I'd like to add impl dependent format/type tests; read from renderbuffer tests.

However, I think this is good as is, so please take a look and please merge if good.  I will append/improve the test later.

sdk/tests/conformance2/reading/read-pixels-from-fbo-test.html
```diff
@@ -44,54 +44,362 @@
 var wtu = WebGLTestUtils;
 var gl = wtu.create3DContext();
 
-function rgba8_f(buf, index) {
-  return [buf[index]/255.0, buf[index + 1]/255.0, buf[index + 2]/255.0, buf[index + 3]/255.0];
+function rgba8(buf, index, expectedColor) {
+  var scale = 1.0 / 0xFF;
+  var color = [buf[index] * scale, buf[index + 1] * scale, buf[index + 2] * scale, buf[index + 3] * scale];
+  var tol = 2.0 * scale;
```
kenrussell: Please think a little about how the tolerances might be more easily maintained if it turns out that some GPUs need them increased.

kenrussell:
Very nice. Yes, let's check in these tests now and continue iterating on them. LGTM

PR #1132, opened 2015-07-27
Fixup logic error in drawing to MRTs.

djg:
The test was assuming that the first half color attachments would all be red. Update the test to clear all color attachments to red to satisfy this expectation.


kenrussell:
Thanks for this fix. It looks good as far as I can follow the logic. The assumptions about the state are confusing to follow. Any contributions to simplify it would be appreciated.

PR #1137, opened 2015-07-31
Added query arg to change the test root folder.

toji:
Useful for debugging a subset of the tests. So, for example: If you only wanted to see the WebGL2 portion of the conformance test you could use

`?version=2.0.0&root=conformance2`

Or maybe you're only testing WebGL1 extensions?

`?root=conformance/extensions`


kenrussell:
Nice. LGTM

PR #1138, opened 2015-08-01
Added conformance test for shaders that crash some Qualcomm drivers.

toji:
Based on http://crbug.com/498947

### DIFF
--- a/sdk/tests/conformance/glsl/bugs/00_test_list.txt
+++ b/sdk/tests/conformance/glsl/bugs/00_test_list.txt
@@ -24,3 +24,4 @@
 --min-version 1.0.4 sampler-struct-function-arg.html
 --min-version 1.0.4 sequence-operator-evaluation-order.html
 --min-version 1.0.4 logic-inside-block-without-braces.html
+--min-version 1.0.4 qualcomm-crash.html
--- a/sdk/tests/conformance/glsl/bugs/qualcomm-crash.html
+++ b/sdk/tests/conformance/glsl/bugs/qualcomm-crash.html
@@ -1,0 +1,157 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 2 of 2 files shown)


sdk/tests/conformance/misc/qualcomm-crash.html
```diff
@@ -0,0 +1,157 @@
...
+</script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<script>
+"use strict";
+description("This test checks a known bug in some Qualcomm drivers which causes crashes when linking certain shaders.");
+
+debug("");
```
kenrussell: Please add a link to the Chromium bug here; see conformance/attribs/gl-bindAttribLocation-repeated.html as an example.

sdk/tests/conformance/misc/qualcomm-crash.html
```diff
@@ -0,0 +1,157 @@
...
+
+        var program2 = wtu.setupProgram(gl, ['vshader2', 'fshader2']);
+        if (!gl.getProgramParameter(program2, gl.LINK_STATUS)) {
+            testFailed("Program failed to link");
+        }
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no errors");
+    }
+}
+
+// Cycle through a rAF once to give any webglcontextlost events a chance to propegate
```
kenrussell: typo: propagate

sdk/tests/conformance/misc/qualcomm-crash.html
```diff
@@ -0,0 +1,157 @@
...
+window.requestAnimationFrame(function() { finishTest(); });
```
kenrussell: After our offline discussion, I think this is fine. I had been thinking of the delay in conformance/glsl/misc/large-loop-compile.html , but I don't think it's necessary any more, since the synchronous getProgramParameter call is made above.

sdk/tests/conformance/misc/00_test_list.txt
```diff
@@ -9,6 +9,7 @@ invalid-passed-params.html
 is-object.html
 null-object-behaviour.html
 object-deletion-behaviour.html
+qualcomm-crash.html
```
kenrussell: Could you please put this in conformance/glsl/bugs/ ?

kenrussell:
Thank you for putting together these tests. Please merge yourself after addressing the above comments. Thanks.

PR #1139, opened 2015-08-03
Update new formats and types test for texture

qkmiao:


sdk/tests/conformance2/textures/misc/tex-new-formats.html
```diff
@@ -305,97 +261,104 @@
...
         },
         {
-            sizedformat: "RGB10_A2UI",
+            sizedformat: "RGBA32I",
             unsizedformat: "RGBA_INTEGER",
-            type: "UNSIGNED_INT_2_10_10_10_REV",
+            type: "INT",
         },
         {
-            // No ALPHA8 constant in the spec. This format cannot be specified as a sized format.
```
kenrussell: This pull request seems to lose useful comments such as this one.

kenrussell:
I'm sorry, but this pull request is messy -- it makes dozens of changes to the existing tests with no good justification, and removes useful comments.

If you intend to reorganize the test for better readability as well as add new test cases, then please do these in separate steps -- one which makes large changes but does not change the functionality in any way, and another which adds new tests.

Adding @zhenyao and @toji for other reviews.

qkmiao:
Thanks for review. I organized the test cases in order described in http://docs.gl/es3/glTexImage2D. Unsized internal format first, then the sized internal format. Some of previous test cases doesn't align with the current spec, for example texImage2D doesn't accept all unsizedformt.

jdashg:
The order (and contents) should be based on the GLES3 spec.
From GLES 3.0.4 p111-113:
"Table 3.2: Valid combinations of `format`, `type`, and sized `internalformat`."
and
"Table 3.3: Valid combinations of `format`, `type`, and unsized `internalformat`."

jdashg:
You are correct about the current test containing invalid combinations. For example, (ALPHA, ALPHA, FLOAT) is not a valid (internalFormat, unpackFormat, unpackType) tuple.

qkmiao:
@jdashg @kenrussell , I updated this pull request. I put test cases following the order in GLES3 spec 3.0.4 Table 3.2 and 3.3 . Please take another look.

kenrussell:
Thanks for the clarification and the reordering. I verified your test against the tables in the OpenGL ES 3.0.4 spec; looks good.

PR #1141, opened 2015-08-05
getFramebufferAttachmentParameter gets different value under WebGL 1 and WebGL 2 rendering context

Richard-Yunchao:
the test case conformance/misc/object-deletion-behaviour.html failed under WebGL 2 rendering context in Chromium. Because getFramebufferAttachmentParameter gets different value under WebGL 1 and WebGL 2 rendering context. 

GLES 2.0 spec: https://www.khronos.org/opengles/sdk/docs/man/xhtml/glGetFramebufferAttachmentParameteriv.xml, error section: GL_INVALID_ENUM is generated if there is no attached object at the named attachment point and pname is not GL_FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE.
GLES 3.0 spec: https://www.khronos.org/opengles/sdk/docs/man3/html/glGetFramebufferAttachmentParameteriv.xhtml, 
If the value of GL_FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE is GL_NONE, no framebuffer is bound to target. In this case querying pname GL_FRAMEBUFFER_ATTACHMENT_OBJECT_NAME will return zero (not INVALID_ENUM). 

PTAL when you have time.

### DIFF
--- a/sdk/tests/conformance/misc/object-deletion-behaviour.html
+++ b/sdk/tests/conformance/misc/object-deletion-behaviour.html
@@ -44,6 +44,7 @@ description("Tests deletion behavior for texture, renderbuffer, shader, and prog
 var wtu = WebGLTestUtils;
 var gl = wtu.create3DContext();
 var shouldGenerateGLError = wtu.shouldGenerateGLError;
+var contextVersion = wtu.getDefault3DContextVersion();
 
 debug("");
 debug("shader and program deletion");
@@ -95,7 +96,11 @@ shouldGenerateGLError(gl, gl.NO_ERROR, "gl.deleteTexture(tex)");
 // Deleting a texture bound to the currently-bound fbo is the same as
 // detaching the textue from fbo first, then delete the texture.
 shouldBe("gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE)", "gl.NONE");
-shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)");
+if (contextVersion > 1) {
+  shouldBeNull("gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)");
+} else {
+  shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)");
+}
 shouldBeFalse("gl.isTexture(tex)");
 shouldBeNull("gl.getParameter(gl.TEXTURE_BINDING_2D)");
 shouldGenerateGLError(gl, gl.NO_ERROR, "gl.bindTexture(gl.TEXTURE_2D, tex)");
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/misc/object-deletion-behaviour.html
```diff
@@ -147,7 +152,11 @@
 // Deleting a renderbuffer bound to the currently-bound fbo is the same as
 // detaching the renderbuffer from fbo first, then delete the renderbuffer.
 shouldBe("gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE)", "gl.NONE");
-shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)");
+if (contextVersion > 1) {
+  shouldBe("gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)", "0");
```
zhenyao: "null" is better than "0", as in WebGL we return the object, not the GL name.

sdk/tests/conformance/misc/object-deletion-behaviour.html
```diff
@@ -281,7 +290,11 @@
   shouldBe("gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)", "rbo");
 
   shouldGenerateGLError(gl, gl.NO_ERROR, "gl.bindFramebuffer(gl.FRAMEBUFFER, fbo)");
-  shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)");
+  if (contextVersion > 1) {
+    shouldBe("gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)", "0");
```
zhenyao: Same here, "0" -> "null"

sdk/tests/conformance/misc/object-deletion-behaviour.html
```diff
@@ -318,7 +331,11 @@
   shouldBe("gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)", "tex");
 
   shouldGenerateGLError(gl, gl.NO_ERROR, "gl.bindFramebuffer(gl.FRAMEBUFFER, fbo)");
-  shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)");
+  if (contextVersion > 1) {
+    shouldBe("gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)", "0");
```
zhenyao: Same here, "0" -> "null"

Richard-Yunchao:
@zhenyao , agree with you. In OpenGL ES, return 0 is OK. In WebGL, it is better to return "NULL" instead. In this situation, WebGL spec differs from OpenGL ES spec, then we need to revise the WebGL 2 spec (and implementation in Blink). WDYT?

zhenyao:
It's already in the WebGL spec to return WebGLRenderbuffer or WebGLTexture object in this situation.  So naturally null when nothing is attached.

Richard-Yunchao:
Yeah, we should return null for object type in WebGL, just like what we do in getParameter. Although it returns 0 ( a number type) in ES 3.0 spec

Code has been updated, PTAL.

Richard-Yunchao:
Please also take a look at this small fix in Blink: https://codereview.chromium.org/1281443003/, thanks!

zhenyao:
LGTM

PR #1144, opened 2015-08-05
Added builtin precision, fbo render, texture filtering, shader switch…

jasomobica:
… and shaderderivative tests.


sdk/tests/deqp/framework/opengl/simplereference/sglrReferenceContext.js
```diff
@@ -171,7 +175,7 @@ goog.scope(function() {
 
         // \todo [2013-08-01 pyry] Do we want to make these conditional based on renderCtx?
         addExtension("gl.EXT_color_buffer_half_float");
-        addExtension("gl.EXT_color_buffer_float");
+        addExtension("gl.WEBGL_color_buffer_float");
```
kenrussell: The planned direction is to expose EXT_color_buffer_float on WebGL 2.0 contexts, and to not move forward with the WebGL 1.0 extensions for this functionality (WEBGL_color_buffer_float, EXT_color_buffer_half_float).

kenrussell:
Superb work. Thank you for this tremendous contribution. I only had time to skim through it. Let's merge it now so we can start uncovering any bugs both in the WebGL 2.0 implementations and the tests.

PR #1145, opened 2015-08-06
fix the error in attrib-location-length-limit.html

Richard-Yunchao:
According to WebGL 2 spec: WebGL 1.0 supports tokens up to 256 characters in length. WebGL 2.0 follows The OpenGL ES Shading Language, Version 3.00 and allows tokens up to 1024 characters in length for both ESSL 1 and ESSL 3 shaders. 

So attrib-location-length-limit.html failed because the max identifier length is not 256 in WebGL 2. 

However, both conformance and conformance2 have specific test cases for identifier length, see shader-with-256-character-identifier.html, shader-with-256-character-identifier.html in conformance/, and shader-with-1024-character-identifier,frag.html and shader-with-1025-character-identifier,frag.html in conformance2/. and attrib-location-length-limit.html is used to test bindAttribLocation and getAttribLocation, so I removed the part that tests identifier max length. 

PTAL when you have time. Thanks.


kenrussell:
Not LGTM. It is critically important to test these kinds of boundary conditions. Even if the other tests cover maximum identifier lengths in shaders, there could be something about the handling of attribute names specifically that should be verified separately.

The test should instead be updated to use a different shader with a longer identifier for WebGL 2.0. You can duplicate the test into conformance2/glsl3/ , update it there, and set a max-version 1.9.9 for the version in conformance/glsl/misc/.

Richard-Yunchao:
OK. I updated the code accordingly. Chromium can pass the original tests and this new tests too. PTAL.

sdk/tests/conformance2/glsl3/attrib-location-length-limits.html
```diff
@@ -0,0 +1,117 @@
...
+} else {
+    testPassed("attrib location should not be -1");
+}
+wtu.glErrorShouldBe(gl, gl.NONE);
+
+debug("The attrib length is valid to compile and link in a shader, but attrib location is over the length limit");
+gl.bindAttribLocation(program, 0, "vPosition01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567");
+wtu.glErrorShouldBe(gl, gl.INVALID_VALUE);
+shouldBe('gl.getAttribLocation(program, "vPosition01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567")', '-1');
+wtu.glErrorShouldBe(gl, gl.INVALID_VALUE);
```
kenrussell: These two are oversights. The maximum length of names passed to bindAttribLocation, getAttribLocation and getUniformLocation should be raised to 1024 from 256 in the WebGL 2.0 spec. Other program-related entry points taking strings like getFragDataLocation, transformFeedbackVaryings, getUniformIndices, and getUniformBlockIndex should also have a note indicating that the maximum length of these names is 1024. Could you please make this change and update this test, even though Chromium will fail it for the moment?

sdk/tests/conformance2/glsl3/attrib-location-length-limits.html
```diff
@@ -0,0 +1,117 @@
...
+<script src="../../js/webgl-test-utils.js"> </script>
+</head>
+<body>
+<canvas id="example" width="50" height="50">
+There is supposed to be an example drawing here, but it's not important.
+</canvas>
+<div id="description">Verify limits on the lengths of attribute locations per WebGL spec "Maximum Uniform and Attribute Location Lengths" and WebGL 2 spec "Maximum GLSL Token Size"</div>
+<div id="console"></div>
+<script id="goodVertexShader" type="x-shader/x-vertex">
+// A vertex shader where the needed attrib locations are exactly 256 characters, 257 characters and exactly 1024 characters.
```
kenrussell: This can be updated to just use one attrib location of length 1024.

sdk/tests/conformance2/glsl3/attrib-location-length-limits.html
```diff
@@ -0,0 +1,117 @@
...
+
+var wtu = WebGLTestUtils;
+var gl = wtu.create3DContext("example");
+
+var program = wtu.loadProgramFromScript(gl, "goodVertexShader", "fragmentShader");
+shouldBe('gl.getProgramParameter(program, gl.LINK_STATUS)', 'true');
+debug("Test attrib location underneath the length limit");
+gl.bindAttribLocation(program, 0, "vPosition0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456");
+wtu.glErrorShouldBe(gl, gl.NONE);
+var attribLoc = gl.getAttribLocation(program, "vPosition0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456");
```
kenrussell: This and the above one should be updated to use the attrib location with 1024 characters. Please use variables to hold the 1024- and 1025-length attribute names.

sdk/tests/conformance2/glsl3/attrib-location-length-limits.html
```diff
@@ -0,0 +1,117 @@
...
+wtu.glErrorShouldBe(gl, gl.INVALID_VALUE);
+
+debug("Test attrib length over the length limit");
+debug("Shader compilation or link should fail");
+shouldBe('wtu.loadProgramFromScriptExpectError(gl, "badVertexShader", "fragmentShader")', 'null');
+wtu.glErrorShouldBe(gl, gl.NONE);
+
+debug("Attempt to bind too-long attrib location should produce error");
+program = gl.createProgram();
+gl.bindAttribLocation(program, 0, "vPosition01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567");
```
kenrussell: Should use the 1025-length attrib location.

sdk/tests/conformance2/glsl3/attrib-location-length-limits.html
```diff
@@ -0,0 +1,117 @@
...
+wtu.glErrorShouldBe(gl, gl.INVALID_VALUE);
+
+debug("Attempt to fetch too-long attrib location should produce error");
+program = wtu.loadStandardProgram(gl);
+shouldBe('gl.getAttribLocation(program, "vPosition01234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567")', '-1');
```
kenrussell: Same here.

kenrussell:
Thanks for writing this test. It uncovered some oversights in the spec. The spec and test should be updated further.

Richard-Yunchao:
@kenrussell , got it. The max length of the attrib location should be also 1024 in WebGL 2, the same as the max length of any identifiers in shader. I will update the spec and the test soon.

Richard-Yunchao:
@kenrussell , the code is updated. PTAL.

A small change is also submitted into Blink, please have a look when you have time: https://codereview.chromium.org/1281733003/

sdk/tests/conformance2/glsl3/attrib-location-length-limits.html
```diff
@@ -0,0 +1,108 @@
...
+shouldBe('gl.getProgramParameter(program, gl.LINK_STATUS)', 'true');
+gl.bindAttribLocation(program, 0, "vPosition1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345");
```
kenrussell: Could you please define a variable like:

var attrib1024Name = "vPosition1234..."

and use it here and below?

Same for "attrib1025Name".

kenrussell:
Thanks, this looks better. LGTM. Could you please fix up the merge conflicts and address the one comment in the new test? Thanks.

Richard-Yunchao:
@kenrussell , code has been updated accordingly. PTAL. Thanks a lot!

kenrussell:
Thank you. Looks good.

Richard-Yunchao:
Sorry, a little late to merge https://github.com/KhronosGroup/WebGL/pull/1147 into this. PTAL.

PR #1146, opened 2015-08-08
Added profile string support. Renamed constants back to _KHR.

kenrussell:


zhenyao:
I think kbr clearly explained the rationales. Merging now

PR #1147, opened 2015-08-10
fix bug in uniform-location-length-limits.html.

Richard-Yunchao:
uniform location max length is 1024 in WebGL 2.


Richard-Yunchao:
PTAL, thanks a lot!

kenrussell:

PR #1149, opened 2015-08-10
Moved WEBGL_compressed_texture_astc to draft status.

kenrussell:


kenrussell:
Per discussion on public_webgl list, there were no objections to moving this to draft status; doing so.

PR #1150, opened 2015-08-11
fix bugs in read-pixels-test.html in WebGL 2

Richard-Yunchao:
And this CL divides the original test into two parts:
1. report INVALID_ENUM for invalid format or type
2. report INVALID_OPERATION for invalid combination of format/type,
but standalone formats or types are valid enums.

### DIFF
--- a/sdk/tests/conformance/reading/read-pixels-test.html
+++ b/sdk/tests/conformance/reading/read-pixels-test.html
@@ -31,6 +31,7 @@
 <meta charset="utf-8">
 <title>WebGL ReadPixels conformance test.</title>
 <link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/desktop-gl-constants.js"></script>
 <script src="../../js/js-test-pre.js"></script>
 <script src="../../js/webgl-test-utils.js"> </script>
 </head>
@@ -45,6 +46,7 @@ description("Checks that ReadPixels works as expected.");
 var wtu = WebGLTestUtils;
 var canvas = document.getElementById("example");
 var gl = wtu.create3DContext(canvas);
+var contextVersion = wtu.getDefault3DContextVersion();
 
 var actual;
 var expected;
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:

sdk/tests/conformance/reading/read-pixels-test.html
```diff
@@ -124,80 +126,128 @@
...
+    var info = invalidTypeInfo[tt];
+    var type = info.type;
+    var dest = info.dest;
+    gl.readPixels(0, 0, 1, 1, gl.RGBA, type, dest);
+    wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "Should not be able to read as " + wtu.glEnumToString(gl, type));
+  }
+
+  var badCombinations = [
     {
       format: gl.RGB,
```
zhenyao: You haven't checked the GL_IMPLEMENTATION_COLOR_READ_FORMAT and GL_IMPLEMENTATION_COLOR_READ_TYPE yet, how can you be sure this is invalid combination? The same question for a few other combinations below.

Richard-Yunchao:
The original code also do it in the same way. I thought that the impl format/type are valid only for reading from fbo, in which we set the format/type in a texture/renderbuffer and attached the texture/renderbuffer... Is that wrong? 

The bugs for this test is that GL_LUMINANCE/GL_LUMINANCE_ALPHA is valid as a format standalone in WebGL 2, it will not generate GL_INVALID_ENUM, but the original combination is wrong, so it generates GL_INVALID_OPERATION in WebGL 2. But in WebGL 1, it generates GL_INVALID_ENUM.

Richard-Yunchao:
per readPixels from fbo, we have a test already.

zhenyao:
Then the original code is wrong (I think).  It just happens to be that these combinations are not supported through GL_IMPLEMENTATION_COLOR_READ_FORMAT and GL_IMPLEMENTATION_COLOR_READ_TYPE, so the test passes. My understanding is even for the default backbuffer, this second pair of combination is always supported.

Richard-Yunchao:
These updated format/type combinations are not supported through GL_IMPLEMENTATION_COLOR_READ_FORMAT and GL_IMPLEMENTATION_COLOR_READ_TYPE on my desktop too. So the code passed in both WebGL 1 and WebGL 2 rendering context in Chromium...

OK, I will update this patch tomorrow.

zhenyao:
It's because Chrome is limiting it through emulating GL_IMPLEMENTATION_COLOR_READ_FORMAT and GL_IMPLEMENTATION_COLOR_READ_TYPE on desktop GL.  Otherwise a few will be supported on Desktop GL as they don't have the two combination only limits there. Thank you for fixing this.

Richard-Yunchao:
@zhenyao , the code has been updated, please take a look when you have time. Thanks a lot!

sdk/tests/conformance/reading/read-pixels-test.html
```diff
@@ -124,80 +126,147 @@
...
+  if (contextVersion < 2) {
+  // They are valid in WebGL 2 or higher
+    invalidFormat = invalidFormat.concat([gl.LUMINANCE, gl.LUMINANCE_ALPHA, desktopGL.RED, desktopGL.RG_INTEGER, desktopGL.RGBA_INTEGER]);
+  }
+
+  var invalidTypeInfo = [
+    {type: gl.UNSIGNED_SHORT,                 dest: new Uint16Array(4)},
+    {type: gl.SHORT,                          dest: new Int16Array(4)},
+    {type: desktopGL.UNSIGNED_INT_24_8,       dest: new Uint32Array(4)}
+  ];
```
zhenyao: I think this is incorrect, as these types still can be the implementation selected type.

zhenyao:
Mostly looks good except the invalid type part.

Richard-Yunchao:
@zhenyao, yes, you are right. These two types(SHORT, UNSIGNED_SHORT) may be supported in WebGL 2 (but Blink didn't added them into WebGL 2, I will update this). 

I have updated the code, PTAL when you have time. Thanks!

zhenyao:
lgtm

Richard-Yunchao:
@zhenyao, I also updated the code in blink side: add SHORT and UNSIGNED_SHORT into types in WebGL 2 rendering context. PTAL at this small change: https://codereview.chromium.org/1288303002/.

PR #1151, opened 2015-08-13
Add test cases in sampler for WebGL 2

legendlee1314:
Test targets are getSamplerParameter and samplerParameter[if].
Checking the pnames and if the default and setting params are correct.
Will add params checking cases later.

### DIFF
--- a/sdk/tests/conformance2/samplers/samplers.html
+++ b/sdk/tests/conformance2/samplers/samplers.html
@@ -47,8 +47,10 @@ debug("");
 var wtu = WebGLTestUtils;
 var canvas = document.getElementById("canvas");
 var gl = wtu.create3DContext(canvas, null, 2);
+var s = null;
 var s1 = null;
 var s2 = null;
+var testCases = null;
 
 if (!gl) {
     testFailed("WebGL context does not exist");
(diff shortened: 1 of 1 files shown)


legendlee1314:
@kenrussell , @zhenyao , @toji , could you please review this? Thanks!

sdk/tests/conformance2/samplers/samplers.html
```diff
@@ -126,6 +133,72 @@
...
+    testCases = [
+      { pname: gl.TEXTURE_WRAP_S,
+        param: gl.REPEAT,
+        expectedError: gl.NO_ERROR },
+      { pname: gl.TEXTURE_MIN_FILTER,
+        param: gl.LINEAR,
+        expectedError: gl.NO_ERROR },
+      { pname: gl.TEXTURE_COMPARE_MODE,
+        param: gl.NONE,
+        expectedError: gl.NO_ERROR },
```
zhenyao: Could you please also test valid pname TEXTURE_WRAP_R, TEXTURE_WRAP_T, TEXTURE_MAG_FILTER, TEXTURE_COMPARE_FUNC? The list is short, so it's good to verify them all.

legendlee1314: Yes, updated to have all valid pnames. Moreover, add tests for the default value of them.

zhenyao:
Great!

PR #1152, opened 2015-08-18
Supplement WebGLSampler part in WebGL 2 spec

legendlee1314:
1. Add unit and sampler states of bindSampler.
2. Add pname states of samplerParameter[if].
3. bindSampler accepts null object but others not.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -27,7 +27,7 @@
     <!--end-logo-->
 
     <h1>WebGL 2 Specification</h1>
-    <h2 class="no-toc">Editor's Draft 10 August 2015</h2>
+    <h2 class="no-toc">Editor's Draft 18 August 2015</h2>
     <dl>
         <dt>This version:
             <dd>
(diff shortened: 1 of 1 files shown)


legendlee1314:
PTAL. Thank you!

specs/latest/2.0/index.html
```diff
@@ -2121,7 +2141,7 @@ <h2 class="no-toc">Table of contents</h2>
               <tr><td>TEXTURE_WRAP_T</td><td>GLenum</td></tr>
           </table>
           <p>If <em>pname</em> is not in the table above, generates an <code>INVALID_ENUM</code> error.</p>
-          <p>If an OpenGL error is generated, returns null.</p>
+          <p>If <em>sampler</em> is not a valid sampler object or an OpenGL error is generated, returns null.</p>
```
zhenyao: This is confusing, because if |sampler| is not a valid sampler object, an INVALID_OPERATION will be generated, so it is not an "or" condition, but already covered by the original condition.

A better way is to add the INVALID_OPERATION error for invalid |sampler|.

legendlee1314: Updated. Please review. :)

specs/latest/2.0/index.html
```diff
@@ -2120,8 +2140,9 @@ <h2 class="no-toc">Table of contents</h2>
               <tr><td>TEXTURE_WRAP_S</td><td>GLenum</td></tr>
               <tr><td>TEXTURE_WRAP_T</td><td>GLenum</td></tr>
           </table>
-          <p>If <em>pname</em> is not in the table above, generates an <code>INVALID_ENUM</code> error.</p>
-          <p>If an OpenGL error is generated, returns null.</p>
+          <p>If <em>pname</em> is not in the table above, generates an <code>INVALID_ENUM</code> error and returns null.</p>
```
zhenyao: You don't have to mention returns null here and the next line, because the third item covers both cases already.

legendlee1314: Agreed. Just referred other `get*` API like `getSyncParameter` and `getQueryParameter`. Maybe we can clean up them in another PR for consistency?

zhenyao:
LGTM.  Thank you.

PR #1153, opened 2015-08-18
Removed unnecessary shader logging

toji:
This was creating a massive amount of log spew on Chrome's build bots. Left it in when shader compile fails for easy debugging.


zhenyao:
Sorry I didn't see jasonmobica@ comment before merging. toji@, can you address this request?

toji:
Yeah, I was thinking about doing just that a week back. I'll try and plumb a query arg through in the next day or so.

PR #1155, opened 2015-08-19
Fix minor issues in short-circuiting in loop condition test

Oletus:
Fix path to webgl-test-utils.js, add call to finish test and add shader
logging for easier debugging.

### DIFF
--- a/sdk/tests/conformance2/glsl3/short-circuiting-in-loop-condition.html
+++ b/sdk/tests/conformance2/glsl3/short-circuiting-in-loop-condition.html
@@ -32,7 +32,7 @@
 <title>Short circuit in loop condition test</title>
 <link rel="stylesheet" href="../../resources/js-test-style.css"/>
 <script src="../../js/js-test-pre.js"></script>
-<script src="../../resources/webgl-test-utils.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
 </head>
 <body>
 <div id="description"></div>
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/glsl3/short-circuiting-in-loop-condition.html
```diff
@@ -81,7 +81,7 @@
   if (!gl) {
     testFailed("context does not exist");
   } else {
-    var program = wtu.setupProgram(gl, ["vertex-shader", "fshader"]);
+    var program = wtu.setupProgram(gl, ["vertex-shader", "fshader"], ['aPosition'], undefined, true);
```
zhenyao: toji@ can we add a general logShaders flag to webgl conformance suite, so we can turn it on for debugging and off when running on bots?

kenrussell:
This LGTM modulo @zhenyao 's question. Deferring review and merge to him.

Oletus:
What is the issue with the logging in setupProgram? The shaders are not logged to the console, only to the web page. I can imagine it adds a small performance cost, but is that significant enough to be an issue?

kenrussell:
Some of the ported dEQP tests were logging a significant amount of output (mainly shaders) to the console, and this output was not only making the logs huge, but also presumably slowing down test execution. We'd like to avoid this as much as possible. If the amount of logging your test does is small, that's OK, but in general we'd like to be able to turn off debug logging on our automated test infrastructure.

Oletus:
Most of the existing shader tests do the same thing as this patch does, they add the shaders being tested and their translations to the displayed HTML. This output doesn't go to the console. Of course you might still want to disable this output on the bots if there's some significant perf gain, but I think this output should be available by default.

zhenyao:
OK, let's deal with disabling logging shaders to the page separately.

Oletus:
Cool, thanks for reviewing. If you don't actually need the HTML output on the bots, it might be possible to do an even more radical change and add an option to eliminate most of the other messages written by the tests as well. Most of the messages go through only a few function calls.

PR #1156, opened 2015-08-19
Add test for dynamic indexing of matrices and vectors

Oletus:
This test goes through cases where bugs might slip into certain type of
code transformations done to replace dynamic indexing with static
indexing as is required on a HLSL backend of WebGL.

### DIFF
--- a/sdk/tests/conformance2/glsl3/vector-dynamic-indexing.html
+++ b/sdk/tests/conformance2/glsl3/vector-dynamic-indexing.html
@@ -1,0 +1,340 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/glsl3/vector-dynamic-indexing.html
```diff
@@ -0,0 +1,340 @@
...
+<script id="fshaderIndexMatrixTwice" type="x-shader/x-fragment">#version 300 es
+precision mediump float;
+
+out vec4 my_FragColor;
+
+uniform int u_zero;
+
+void main() {
+    mat2 m = mat2(0.0, 0.0, 0.0, 1.0);
+    float f = m[u_zero + 1][u_zero + 1];
```
zhenyao: This got me thinking, what is the behavior if dynamic index is out of range?  For static index, it's an error at compile time.

zhenyao:
The test itself LGTM

kenrussell:
Thanks for the thorough test Olli and thanks Mo for reviewing it. I didn't have time to go through it in detail but at a high level it LGTM too.

sdk/tests/conformance2/glsl3/vector-dynamic-indexing.html
Oletus: Same answer as on ANGLE code review: The base spec says that this is undefined behavior. KHR_robust_buffer_access_behavior says that either zero or any value from the active program's memory may be returned. We can consider specifying this more tightly than what KHR_robust_buffer_access_behavior does, but possible costs need to be considered carefully. ANGLE already clamps indices, but that might not be strictly necessary on all backends if the goal is only memory security.

zhenyao: Thanks for clarifying this.

PR #1158, opened 2015-08-20
Use clearBufferiv and clearBufferuiv for INT and UINT buffers.

zhenyao:
This may not be the last fix for this test, as Chrome doesn't work in this area yet.


kenrussell:
Very nice. Good catch. LGTM

PR #1159, opened 2015-08-21
Add params checking for sampler parameter test

legendlee1314:
Refactor the test steps:
    default value -> valid value set/get -> invalid value.


legendlee1314:
@kenrussell @zhenyao , PTAL. Thanks!

zhenyao:
LGTM

PR #1161, opened 2015-08-24
Implemented quiet mode

toji:
This mode drastically decreases the amount of DOM output the tests do, which can lead to decent speed-ups on automated testing systems which don't care about this output anyway.

### DIFF
--- a/sdk/tests/conformance/more/unit.js
+++ b/sdk/tests/conformance/more/unit.js
@@ -71,6 +71,46 @@ var __testFailCount__ = 0;
 var __testLog__;
 var __backlog__ = [];
 
+var getUrlOptions = (function() {
+  var _urlOptionsParsed = false;
+  var _urlOptions = {};
+  return function() {
+    if (!_urlOptionsParsed) {
+      var s = window.location.href;
+      var q = s.indexOf("?");
+      var e = s.indexOf("#");
+      if (e < 0) {
+        e = s.length;
+      }
+      var query = s.substring(q + 1, e);
+      var pairs = query.split("&");
+      for (var ii = 0; ii < pairs.length; ++ii) {
+        var keyValue = pairs[ii].split("=");
+        var key = keyValue[0];
+        var value = decodeURIComponent(keyValue[1]);
+        _urlOptions[key] = value;
+      }
+      _urlOptionsParsed = true;
+    }
+
+    return _urlOptions;
+  }
+})();
+
+if (typeof quietMode == 'undefined') {
+  var quietMode = (function() {
+    var _quietModeChecked = false;
+    var _isQuiet = false;
+    return function() {
+      if (!_quietModeChecked) {
+        _isQuiet = (getUrlOptions().quiet == 1);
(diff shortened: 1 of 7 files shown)


sdk/tests/conformance/more/unit.js
```diff
@@ -71,6 +71,44 @@ var __testFailCount__ = 0;
 var __testLog__;
 var __backlog__ = [];
 
+var getUrlOptions = (function() {
```
kenrussell: Would it be possible to refactor these more/ tests to at least share this code with js-test-pre.js instead of introducing two copies into the source base?

sdk/tests/js/webgl-test-harness.js
```diff
@@ -238,21 +238,16 @@ var getMajorVersion = function(versionString) {
 
 /**
  * @param {string} url Base URL of the test.
- * @param {number} webglVersion Integer containing the WebGL major version.
- * @param {boolean} dumpShaders add dumpShader query parameter if true.
+ * @param {map} options Map of options to append to the URL's query string.
```
kenrussell: Could you please document the options that are commonly expected?

kenrussell: Rather, the ones that are supported. It's clear that the options map is explicitly queried inside this function.

kenrussell: Apologies, it looks like I was wrong, and confused the setup of the options below with the options received here.

sdk/tests/conformance/more/unit.js
toji: I would absolutely like to see the two testing frameworks consolidated. That felt like a bigger task, though, an more appropriate for a separate patch.

toji: Filed Issue #1162 to do exactly that.

sdk/tests/webgl-conformance-tests.html
```diff
@@ -257,7 +258,10 @@
     }
     div.appendChild(button);
     var a = reporter.localDoc.createElement('a');
-    a.href = WebGLTestHarnessModule.getURLWithOptions(url, reporter.selectedWebGLVersion);
+    a.href = WebGLTestHarnessModule.getURLWithOptions(url, {
+      webglVersion: reporter.selectedWebGLVersion,
+      quiet: OPTIONS.quiet
```
kenrussell: Before committing this, could you please go through the exercise of wiring this up to Chromium's webgl_conformance.py harness and making sure that the quiet mode can be easily specified? I think using query arguments in the URL may cause problems -- possibly with the test expectations mechanism. Note that webgl_conformance.py overrides portions of the code in sdk/tests/js/webgl-test-harness.js , specifically reportResults -- if it's possible to inject the "quiet" variable with code, that may be a more robust solution.

kenrussell:
Thanks for working on this important speedup. One high-level comment about making sure it can be easily hooked up before merging this pull request. Please merge yourself once that concern is addressed. Thanks.

PR #1164, opened 2015-08-24
Replaced performance.now with Date.now for broader compatibility

toji:
It was pointed out that Safari on Mavericks and iOS8 don't have performance.now yet. This is a less accurate replacement, but I don't think the precision is needed in this case.


kenrussell:
LGTM. Thanks for taking care of this.

PR #1166, opened 2015-08-26
add test cases for reading pixels to pixel pack buffer

Richard-Yunchao:
PTAL. Thanks.

### DIFF
--- a/sdk/tests/conformance2/reading/00_test_list.txt
+++ b/sdk/tests/conformance2/reading/00_test_list.txt
@@ -1,1 +1,2 @@
 read-pixels-from-fbo-test.html
+read-pixels-into-pixel-pack-buffer.html
--- a/sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
+++ b/sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
@@ -1,0 +1,246 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL 2 Conformance Test: ReadPixels Into Pixel Pack Buffer.</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<canvas id="example" width="4" height="4"></canvas>
+<div id="description"></div>
+<div id="console"></div>
+<script>
+"use strict";
+
+function checkFormatAndType()
+{
+    debug("");
+    debug("check format / type");
+    var invalidFormat = [gl.DEPTH_COMPONENT, gl.DEPTH_STENCIL, gl.R8, gl.RGBA4];
+    var invalidType = [gl.UNSIGNED_INT_24_8];
+    for (var ff = 0; ff < invalidFormat.length; ++ff) {
+        var format = invalidFormat[ff];
+        gl.readPixels(0, 0, 1, 1, format, gl.UNSIGNED_BYTE, 0);
+        wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "Format should not be able to read as " + wtu.glEnumToString(gl, format));
+    }
+    for (var tt = 0; tt < invalidType.length; ++tt) {
+        var type = invalidType[tt];
+        gl.readPixels(0, 0, 1, 1, gl.RGBA, type, 0);
+        wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "Type should not be able to read as " + wtu.glEnumToString(gl, type));
+    }
+
+    debug("");
+    debug("check combinations of format and type");
+    var combinations = [
+        {format: gl.RGBA,         type: gl.UNSIGNED_BYTE},
+        {format: gl.RGB,          type: gl.UNSIGNED_BYTE},
+        {format: gl.RGB,          type: gl.UNSIGNED_SHORT_5_6_5},
+        {format: gl.RGBA,         type: gl.UNSIGNED_SHORT_5_5_5_1},
+        {format: gl.RGBA,         type: gl.UNSIGNED_SHORT_4_4_4_4},
+        {format: gl.ALPHA,        type: gl.UNSIGNED_BYTE},
+        {format: gl.LUMINANCE,    type: gl.UNSIGNED_BYTE},
+        {format: gl.RED,          type: gl.UNSIGNED_BYTE},
+        {format: gl.RGBA_INTEGER, type: gl.UNSIGNED_INT},
+        {format: gl.RGBA_INTEGER, type: gl.INT}
+    ];
+
+    var implFormat = gl.getParameter(gl.IMPLEMENTATION_COLOR_READ_FORMAT);
+    var implType = gl.getParameter(gl.IMPLEMENTATION_COLOR_READ_TYPE);
+    for (var tt = 0; tt < combinations.length; ++ tt) {
+        var info = combinations[tt];
+        var format = info.format;
+        var type = info.type;
+        gl.readPixels(0, 0, 1, 1, format, type, 0);
+        // Only two format/type parameter pairs are accepted. GL_RGBA/GL_UNSIGNED_BYTE is always
+        // accepted on default readbuffer. The other acceptable pair can be discovered by querying
+        // GL_IMPLEMENTATION_COLOR_READ_FORMAT and GL_IMPLEMENTATION_COLOR_READ_TYPE.
+        if ((format == gl.RGBA && type == gl.UNSIGNED_BYTE) || (format == implFormat && type == implType)) {
+            wtu.glErrorShouldBe(gl, gl.NO_ERROR, "The combination of format/type should be able to read as " +
+                                wtu.glEnumToString(gl, format) + " / " + wtu.glEnumToString(gl, type));
+        } else {
+            wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "The combination of format/type should not be able to read as " +
+                                wtu.glEnumToString(gl, format) + " / " + wtu.glEnumToString(gl, type));
+        }
+    }
+}
+
+function validatePixelPackBufferAndParameters(canvasWidth, canvasHeight)
+{
+    debug("");
+    debug("Validate PIXEL_PACK buffer and readPixels' parameters");
+    gl.clearColor(0, 0, 0, 1);
+    gl.clear(gl.COLOR_BUFFER_BIT);
+    gl.pixelStorei(gl.PACK_ALIGNMENT, 1);
+
+    var size = canvasWidth * canvasHeight * 4;
+    var buffer = gl.createBuffer();
+    gl.bindBuffer(gl.PIXEL_PACK_BUFFER, buffer);
+    gl.bufferData(gl.PIXEL_PACK_BUFFER, size, gl.STATIC_DRAW);
+    var array = new Uint8Array(size);
+
+    debug("");
+    debug("PIXEL_PACK buffer is bound");
+    gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, array);
+    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "should generate INVALID_OPERATION if pixel pack buffer is bound");
+    gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, 0);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+    debug("");
+    debug("Validate the offset of PIXEL_PACK buffer and buffer size");
+    gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, -1);
+    wtu.glErrorShouldBe(gl, gl.INVALID_VALUE, "offset < 0");
+    gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, size);
+    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "offset > buffer size");
+    gl.readPixels(0, 0, canvasWidth + 1, canvasHeight, gl.RGBA, gl.UNSIGNED_BYTE, 0);
+    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "pixel pack buffer is not large enough");
+
+    debug("");
+    debug("Validate the reading area of framebuffer");
+    gl.readPixels(-1, -2, canvasWidth, canvasHeight, gl.RGBA, gl.UNSIGNED_BYTE, 0);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "reading pixels outside of the framebuffer should succeed.");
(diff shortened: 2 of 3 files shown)


sdk/tests/conformance2/reading/00_test_list.txt
```diff
@@ -1 +1,2 @@
 read-pixels-from-fbo-test.html
+read-pixels-upon-pixel-pack-buffer.html
```
kenrussell: "upon" should really be "into" here.

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
```diff
@@ -0,0 +1,219 @@
...
+    gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, 0);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+    debug("Validate the offset of PIXEL_PACK buffer and buffer size");
+    gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, -1);
+    wtu.glErrorShouldBe(gl, gl.INVALID_VALUE, "offset < 0");
+    gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, size);
+    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "offset > buffer size");
+    gl.readPixels(0, 0, 2, 1, gl.RGBA, gl.UNSIGNED_BYTE, 0);
+    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "pixel pack buffer is not large enough");
```
zhenyao: Why this is not enough?  You have a buffer size of width * height * 4, and here you only read back 2 * 1 * 4.

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
```diff
@@ -0,0 +1,219 @@
...
+debug("");
+debug("Canvas.getContext");
+
+var wtu = WebGLTestUtils;
+var pixel = [ 0, 0, 0, 0 ];
+var expectedColor = [ 0, 0, 0, 0 ];
+
+var canvas = document.getElementById("example");
+var gl = wtu.create3DContext(canvas);
+var width = 2;
```
zhenyao: You should get rid of the global varibale width and height. Some functions are calling with parameter width/height, and some are directly using the global ones.  It's really confusing.

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
```diff
@@ -0,0 +1,219 @@
...
+{
+    debug("");
+    debug("Verify that reading pixels to PIXEL_PACK buffer works fine with various PACK_ALIGNMENT values.");
+    shouldBeNonNull("program = wtu.setupProgram(gl, ['vshader', 'fshader'], ['pos', 'colorIn'])");
+
+    var format = gl.RGBA;
+    var type = gl.UNSIGNED_BYTE;
+
+    debug("format: RGBA, type: UNSIGNED_BYTE");
+    runTestIteration(format, type, 1, 1, 2);
```
zhenyao: Why do you pass format/type to runTestIteration()?  Since they are constant in this test, you should just set them inside runTestIteration().

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
```diff
@@ -0,0 +1,219 @@
...
+    runTestIteration(format, type, 8, 1, 2);
+    runTestIteration(format, type, 1, 0, 0);
+    runTestIteration(format, type, 1, 0, -1);
+    runTestIteration(format, type, 1, -1, 0);
+    runTestIteration(format, type, 1, -1, -1);
```
zhenyao: You already tested negative height/width in validatePixelPackBufferAndParameters(). More interesting test cases would be various pack alignment, since this function name is testPackAlignment() anyway.

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
```diff
@@ -0,0 +1,219 @@
...
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL 2 ReadPixels Test Upon Pixel Pack Buffer.</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+<script id="vshader" type="x-shader/x-vertex">
```
zhenyao: The shaders are not used in this test at all.

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
```diff
@@ -0,0 +1,219 @@
...
+    gl.getBufferSubData(gl.PIXEL_PACK_BUFFER, 0, arrBuffer);
+    var array = new Uint8Array(arrBuffer);
+
+    // Check the last pixel of the last row.
+    var bytesPerRow = width * bytesPerPixel + padding;
+    var pos = bytesPerRow * (height - 1) + (width - 1) * bytesPerPixel;
+    var numComponents = bytesPerPixel;
+    for (var i = 0; i < numComponents; ++i)
+        pixel[i] = array[pos + i];
+    for (var i = numComponents; i < 4; ++i)
```
zhenyao: This is unnecessary (and confusing) since bytesPerPixel is a constant 4.

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
```diff
@@ -0,0 +1,219 @@
...
+        pixel[i] = 0;
+    expectedColor = [255, 102, 0, 255];
+    shouldBeNonNull("expectedColor");
```
zhenyao: Why do we need to test shouldBeNonNull() here?

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
```diff
@@ -0,0 +1,219 @@
...
+
+    var size = width * height * 4;
+    var buffer = gl.createBuffer();
+    gl.bindBuffer(gl.PIXEL_PACK_BUFFER, buffer);
+    gl.bufferData(gl.PIXEL_PACK_BUFFER, size, gl.STATIC_DRAW);
+    var array = new Uint8Array(size);
+
+    debug("PIXEL_PACK buffer is bound");
+    gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, array);
+    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "pixel pack buffer should not be bound");
```
kenrussell: This error message is a little confusing. I'd suggest instead "should fail if pixel pack buffer is bound".

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
```diff
@@ -0,0 +1,219 @@
...
+    var padding = 0;
+    switch (packAlignment) {
+    case 1:
+    case 2:
+    case 4:
+    case 8:
+        padding = (bytesPerPixel * width) % packAlignment;
+        if (padding > 0)
+            padding = packAlignment - padding;
+        break;
```
kenrussell: Why not just return "padding" here, rather than breaking out and returning it at the end of the function?

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
```diff
@@ -0,0 +1,219 @@
...
+    default:
+        return -1;
```
kenrussell: What about throwing 'Illegal packAlignment value' or similar here to make errors more obvious?

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
```diff
@@ -0,0 +1,219 @@
...
+}
+
+function runTestIteration(format, type, packAlignment, width, height)
+{
+    debug("Testing PACK_ALIGNMENT = " + packAlignment + ", width = " + width + ", height = " + height);
+    gl.clearColor(1, 0.4, 0, 1);
+    gl.clear(gl.COLOR_BUFFER_BIT);
+    gl.pixelStorei(gl.PACK_ALIGNMENT, packAlignment);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+    var bytesPerPixel = 4;
```
kenrussell: This is dependent on the format and type. Per @zhenyao 's comment below, given the current state of this test, format and type should be made constant, and defined alongside bytesPerPixel.

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
```diff
@@ -0,0 +1,219 @@
...
+    default:
+        return -1;
+    }
+    return padding;
+}
+
+function runTestIteration(format, type, packAlignment, width, height)
+{
+    debug("Testing PACK_ALIGNMENT = " + packAlignment + ", width = " + width + ", height = " + height);
+    gl.clearColor(1, 0.4, 0, 1);
```
kenrussell: Please define this in terms of the expected color; i.e., gl.clearColor(expectedColor[0] / 255.0, ...).

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
kenrussell: Echoing this comment -- this is confusing.

kenrussell:
@zhenyao thank you for doing such a careful review. @Richard-Yunchao thanks for the contribution; please address zhenyao's comments. I defer review to him.

sdk/tests/conformance2/reading/read-pixels-upon-pixel-pack-buffer.html
Richard-Yunchao: The caller will handle this.

Richard-Yunchao:
Thanks for your review, @zhenyao and @kenrussell . The code has been updated accordingly. A function to check format/type of readPixels is also added into the test. 

Generally speaking, the test cases are similar to those cases under conformance/reading/. In order to make the code clear, I would like to create this standalone file under conformance2/reading/. Otherwise, there would be many if-else to distinguish different rendering contexts, different readPixels APIs, and different resources(ArrayBuffer or PIXEL_PACK buffer), etc.

specs/latest/2.0/index.html
```diff
@@ -1869,7 +1869,7 @@ <h2 class="no-toc">Table of contents</h2>
         </dt>
         <dd>
           <p>If no WebGLBuffer is bound to the <code>PIXEL_PACK_BUFFER</code> target, generates an <code>INVALID_OPERATION</code> error.</p>
-          <p><em>offset</em> is the byte offset into the WebGLBuffer's data store; generates an <code>INVALID_VALUE</code> error if it's less than 0.</p>
+          <p><em>offset</em> is the byte offset into the WebGLBuffer's data store; generates an <code>INVALID_VALUE</code> error if it's less than 0. If <em>offset</em> passes the end of the pixel pack buffer, or the pixel pack buffer is not large enough to store all pixels, generates an <code>INVALID_OPERATION</code>.</p>
```
Richard-Yunchao: If the offset passes the end of the buffer, which error does the browser should generate, INVALID_ENUM or INVALID_OPERATION? 

bufferSubData generates INVALID_ENUM in this situation, but other APIs, say vertexAttribPointer, would generate INVALID_OPERATION. 

Personally, I prefer to generate INVALID_ENUM, WDYT?

sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
```diff
@@ -0,0 +1,269 @@
...
+    for (var ff = 0; ff < invalidFormat.length; ++ff) {
+        var format = invalidFormat[ff];
+        gl.readPixels(0, 0, 1, 1, format, gl.UNSIGNED_BYTE, 0);
+        wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "Format should not be able to read as " + wtu.glEnumToString(gl, format));
+    }
+    for (var tt = 0; tt < invalidType.length; ++tt) {
+        var type = invalidType[tt];
+        gl.readPixels(0, 0, 1, 1, gl.RGBA, type, 0);
+        wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "Type should not be able to read as " + wtu.glEnumToString(gl, type));
+    }
```
zhenyao: This is incorrect.  ReadPixels only supports one basic format/type, one implementation specific format/type.  Only with RGB10_A2, a third format/type is allowed.  Your test is saying all these formats and types have to be allowed.

sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
```diff
@@ -0,0 +1,269 @@
...
+<div id="console"></div>
+<script>
+"use strict";
+
+function checkFormatAndType()
+{
+    debug("");
+    debug("check format / type");
+    var validFormat = [gl.LUMINANCE, gl.RED, gl.RG_INTEGER, gl.RGBA_INTEGER];
+    var validType = [gl.UNSIGNED_SHORT, gl.BYTE, gl.INT, gl.UNSIGNED_INT_2_10_10_10_REV];
```
zhenyao: Why testsing these formats and types?  They are not complete list, and seem random to me.

sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
```diff
@@ -0,0 +1,269 @@
...
+    runTestIteration(8, 1, 2);
+    runTestIteration(4, 2, 2);
+    runTestIteration(8, 2, 2);
+    runTestIteration(4, 3, 2);
+    runTestIteration(8, 3, 2);
+    runTestIteration(1, 0, 0);
+    runTestIteration(4, 1, 1);
+    runTestIteration(8, 1, 1);
+    runTestIteration(3, 1, 2);
+    runTestIteration(16, 1, 2);
```
zhenyao: You tested with width = 3, but your canvas is 2 x 2.  I didn't see code that resize the canvas.

sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
```diff
@@ -0,0 +1,269 @@
...
+                                wtu.glEnumToString(gl, format) + " / " + wtu.glEnumToString(gl, type));
+        }
+    }
+}
+
+function validatePixelPackBufferAndParameters()
+{
+    debug("");
+    debug("Validate PIXEL_PACK buffer and readPixels' parameters");
+    var canvasWidth = 2, canvasHeight = 2;
```
zhenyao: It's better for width/height to be passed in as function parameters, to be consistent with other functions in this file.

specs/latest/2.0/index.html
kenrussell: bufferSubData generates INVALID_VALUE (not INVALID_ENUM or INVALID_OPERATION) if there is an attempt to write data past the end of the buffer's data store.

I think readPixels should also generate INVALID_VALUE if there is an attempt to read back pixel data past the end of the bound pixel pack buffer.

sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
Richard-Yunchao: Yeah. I made a mistake here. validFormat and validType should be removed.

sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
Richard-Yunchao: Sorry, this code snippet is not correct.

specs/latest/2.0/index.html
Richard-Yunchao: I thought about this again. The other readPixels API would generate INVALID_ENUM if the pixels(ArrayBuffer) is null, it would generate INVALID_OPERATION if the ArrayBuffer is not large enough. So I think this readPixels API should generate INVALID_ENUM if offset < 0, and generate INVALID_OPERATION if the remainder of the PIXEL_PACK buffer is not large enough. Then these two readPixels APIs are aligned with each other to generate appropriate errors in similar situations.

Richard-Yunchao:
Thanks for your careful review, @kenrussell and @zhenyao . Code has been updated, PTAL.

Richard-Yunchao:
After an interactive rebase to squash all commits and force update this pull request, some comments went away... 

Code has been updated, PTAL.

sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
```diff
@@ -0,0 +1,250 @@
...
+
+    var size = bytesPerPixel * width * height + padding * (height - 1);
+    if (size < 0)
+        size = 0;
+    var buffer = gl.createBuffer();
+    gl.bindBuffer(gl.PIXEL_PACK_BUFFER, buffer);
+    gl.bufferData(gl.PIXEL_PACK_BUFFER, size, gl.STATIC_DRAW);
+    var offset = 0;
+    gl.readPixels(0, 0, width, height, gl.RGBA, gl.UNSIGNED_BYTE, offset);
+    if (padding < 0) {
```
zhenyao: This should go away

sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
```diff
@@ -0,0 +1,250 @@
...
+    case 1:
+    case 2:
+    case 4:
+    case 8:
+        padding = (bytesPerPixel * width) % packAlignment;
+        if (padding > 0)
+            padding = packAlignment - padding;
+        return padding;
+    default:
+        testFail("should not reach here");
```
zhenyao: testFail -> testFailed

zhenyao:
Two tiny fixes, LGTM after that

Richard-Yunchao:
Thanks for your review, @zhenyao . Code has been updated. PTAL.

zhenyao:
LGTM.  Thanks.

PR #1167, opened 2015-08-26
check packing more toroughly

gero3:
Based on mrdoob/three.js#7037, we decided to check if  packing is implemented based on the packing algoritm in https://www.khronos.org/registry/gles/specs/2.0/GLSL_ES_Specification_1.0.17.pdf Appendix A section 7.

### DIFF
--- a/sdk/tests/conformance/glsl/misc/shader-uniform-packing-restrictions.html
+++ b/sdk/tests/conformance/glsl/misc/shader-uniform-packing-restrictions.html
@@ -262,6 +262,78 @@ for (var ss = 0; ss < shaderTypes.length; ++ss) {
       linkSuccess: true,
       passMsg: shaderType.type + " shader with " + minVars + " uniforms of " + info.type + " (the minimum required) should succeed",
     });
+    
+    if (maxPerRow > 1) {
+      
+      var generateCodeArray = function() {
+        var uniforms = [];
+        var sumTerms = [];
+        var itemcount = Math.max(Math.floor(shaderType.maxVectors / maxPerRow) + 1, minVars ); 
+        for (var uu = 0; uu < maxPerRow; ++uu) {
+          uniforms.push("    uniform " + info.type + " u_uniform" + uu + "[" + itemcount + "];");
+          sumTerms.push(wtu.replaceParams(info.uniToF, {id: uu, index: "[" + (itemcount - 1) + "]"}));
+        }
+        return {
+          uniforms: uniforms.join("\n"),
+          code: info.fType + " sum = " + sumTerms.join(" + \n            ") + ";",
+          result: wtu.replaceParams(info.fToVec4, {f: 'sum'})
+        };
+      };
+      
+      // Test required uniforms of type.
+      tests.push({
+        vShaderSource: wtu.replaceParams(shaderType.vertUniformTest, generateCodeArray(), info),
+        vShaderSuccess: true,
+        fShaderSource: wtu.replaceParams(shaderType.fragUniformTest, generateCodeArray(), info),
+        fShaderSuccess: true,
+        linkSuccess: true,
+        passMsg: shaderType.type + " shader with " + maxPerRow + "uniform arrays of " + info.type + " with " + (Math.floor(shaderType.maxVectors / maxPerRow) + 1) + " elements should succeed",
+      });
+      
+      for (var ut = ii + 1 ; ut < uniformTypes.length; ++ut) { 
+        var info2 = uniformTypes[ut];
+        if (info.componentsPerRow + info2.componentsPerRow <= 4){
+          // Compute the minimum required to work in a single array.
+          var minVars2 = Math.floor(shaderType.minVectors / info2.rows);
+          // Compute the maximum allowed as single elements
+          var maxPerRow2 = Math.floor(4 / info.componentsPerRow);
+          
+          var itemcount = Math.max(Math.floor(shaderType.maxVectors / info.rows / 2) + 1, minVars ); 
+          var itemcount2 = Math.max(Math.floor(shaderType.maxVectors / info2.rows / 2) + 1, minVars ); 
+
+          var generateCodeDifferentTypes= function() {
+            var uniforms = [];
+            var sumTerms = [];
+            
+            uniforms.push("    uniform " + info.type + " u_uniform1[" + itemcount + "];");
+            uniforms.push("    uniform " + info2.type + " u_uniform2[" + itemcount2 + "];");
+            
+            
+            sumTerms.push(wtu.replaceParams(info.fToVec4, {f: wtu.replaceParams(info.uniToF, {id: 1, index: "[" + (itemcount - 1 ) + "]"})}));
+            sumTerms.push(wtu.replaceParams(info2.fToVec4, {f: wtu.replaceParams(info2.uniToF, {id: 2, index: "[" + (itemcount2 - 1 ) + "]"})}));
+            
+            return {
+              uniforms: uniforms.join("\n"),
+              code: "vec4 sum = " + sumTerms.join(" + \n            ") + ";",
+              result: 'sum'
+            };
+          };
+          
+          // Test required uniforms of type.
+          tests.push({
+            vShaderSource: wtu.replaceParams(shaderType.vertUniformTest, generateCodeDifferentTypes(), info),
+            vShaderSuccess: true,
+            fShaderSource: wtu.replaceParams(shaderType.fragUniformTest, generateCodeDifferentTypes(), info),
+            fShaderSuccess: true,
+            linkSuccess: true,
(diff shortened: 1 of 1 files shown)


kenrussell:
Thanks for submitting this pull request.

It strongly looks like you've uncovered a bug in various WebGL implementations' packing algorithms for uniform vectors; some of the shaders your test is generating should come nowhere near the limits, but are being rejected (as I'm sure you noticed). On the other hand, reducing the size of some of your arrays by just 1 element causes them to pass the packing algorithm.

We're still investigating what is going on and will update this pull request with more information.

The main difficulty in proceeding with this pull request as is is that there aren't strong guarantees about the amount of storage that is actually available; at the end of Appendix A section 7 there is a sentence "Part of the storage may be reserved by an implementation for its own use". It's for this reason that the other tests only make two assertions:

1) Packing the minimum number of supported vertex or fragment uniform vectors must succeed.
2) Packing one more than the maximum number of supported vertex or fragment uniform vectors must fail.

We need to carefully consider whether it's feasible to test shaders that lie somewhere in between these two boundaries. The bug you've identified looks like it's preventing valid shaders from working which are using just over 1/4 of the space that's reported to be available.

sdk/tests/conformance/glsl/misc/shader-uniform-packing-restrictions.html
```diff
@@ -262,6 +262,78 @@
       linkSuccess: true,
       passMsg: shaderType.type + " shader with " + minVars + " uniforms of " + info.type + " (the minimum required) should succeed",
     });
+    
+    if (maxPerRow > 1) {
+      
+      var generateCodeArray = function() {
+        var uniforms = [];
+        var sumTerms = [];
+        var itemcount = Math.max(Math.floor(shaderType.maxVectors / maxPerRow) + 1, minVars );
```
kenrussell: "itemcount" should be computed once and its value shared when computing the message string below.

sdk/tests/conformance/glsl/misc/shader-uniform-packing-restrictions.html
```diff
@@ -262,6 +262,78 @@
...
+      };
+      
+      // Test required uniforms of type.
+      tests.push({
+        vShaderSource: wtu.replaceParams(shaderType.vertUniformTest, generateCodeArray(), info),
+        vShaderSuccess: true,
+        fShaderSource: wtu.replaceParams(shaderType.fragUniformTest, generateCodeArray(), info),
+        fShaderSuccess: true,
+        linkSuccess: true,
+        passMsg: shaderType.type + " shader with " + maxPerRow + "uniform arrays of " + info.type + " with " + (Math.floor(shaderType.maxVectors / maxPerRow) + 1) + " elements should succeed",
```
kenrussell: The Math.floor expression here looks like it's supposed to be identical to "itemcount" but the code's duplicated. Please share it.

sdk/tests/conformance/glsl/misc/shader-uniform-packing-restrictions.html
```diff
@@ -262,6 +262,78 @@
...
+          sumTerms.push(wtu.replaceParams(info.uniToF, {id: uu, index: "[" + (itemcount - 1) + "]"}));
+        }
+        return {
+          uniforms: uniforms.join("\n"),
+          code: info.fType + " sum = " + sumTerms.join(" + \n            ") + ";",
+          result: wtu.replaceParams(info.fToVec4, {f: 'sum'})
+        };
+      };
+      
+      // Test required uniforms of type.
```
kenrussell: Please update this comment with a clearer description of what's being tested. In this particular case, it looks like it's that multiple arrays which should pack into the same row must do so.

sdk/tests/conformance/glsl/misc/shader-uniform-packing-restrictions.html
```diff
@@ -262,6 +262,78 @@
...
+        passMsg: shaderType.type + " shader with " + maxPerRow + "uniform arrays of " + info.type + " with " + (Math.floor(shaderType.maxVectors / maxPerRow) + 1) + " elements should succeed",
+      });
+      
+      for (var ut = ii + 1 ; ut < uniformTypes.length; ++ut) { 
+        var info2 = uniformTypes[ut];
+        if (info.componentsPerRow + info2.componentsPerRow <= 4){
+          // Compute the minimum required to work in a single array.
+          var minVars2 = Math.floor(shaderType.minVectors / info2.rows);
+          // Compute the maximum allowed as single elements
+          var maxPerRow2 = Math.floor(4 / info.componentsPerRow);
```
kenrussell: minVars2 and maxPerRow2 are defined but never used.

gero3:
> It strongly looks like you've uncovered a bug in various WebGL implementations' packing algorithms for uniform vectors; some of the shaders your test is generating should come nowhere near the limits, but are being rejected (as I'm sure you noticed). On the other hand, reducing the size of some of your arrays by just 1 element causes them to pass the packing algorithm.

I try to keep the necessary uniforms to the absolute minimum that is needed to enforce packing multiple uniforms in one row which is why even reduing the size by 1 is enough to make it pass.

> The main difficulty in proceeding with this pull request as is is that there aren't strong guarantees about the amount of storage that is actually available; at the end of Appendix A section 7 there is a sentence "Part of the storage may be reserved by an implementation for its own use". It's for this reason that the other tests only make two assertions:
> 
> 1) Packing the minimum number of supported vertex or fragment uniform vectors must succeed.
> 2) Packing one more than the maximum number of supported vertex or fragment uniform vectors must fail.
> 
> We need to carefully consider whether it's feasible to test shaders that lie somewhere in between these two boundaries. The bug you've identified looks like it's preventing valid shaders from working which are using just over 1/4 of the space that's reported to be available.

This is the reason why i decided to only test the absolute minimum that is needed to enforce packing multiple uniforms in one row which is 1 more then half of what is reported to be available.  I think that is a reasonable assumption.

kenrussell:
> I try to keep the necessary uniforms to the absolute minimum that is needed to enforce packing
> multiple uniforms in one row which is why even reduing the size by 1 is enough to make it pass.

Ah, thanks, that's subtle. Yes, I see that packing 4 boolean vector arrays of length 257 when the max vertex uniform vectors is 1024 will require at least one of the arrays to be packed on the same row as another one.

We're continuing to investigate this ( @zhenyao is driving the investigation ) and we'll keep this pull request updated.

zhenyao:
It turns out the failures we saw on Linux with NVIDIA GPU are caused by the NVIDIA driver, not by Chrome.  I filed an ANGLE bug to track this: https://code.google.com/p/angleproject/issues/detail?id=1142

zhenyao:
To confirm, the uniform packing algorithm in Chrome/ANGLE didn't raise an error in the above mentioned  cases (by kbr).

kenrussell:
Further testing indicates that many desktop OpenGL implementations are expanding out arrays of scalar values (uniform bool[], uniform int[], uniform float[] to 4-vectors. No packing of scalar arrays into columns, or arrays of multiple types into columns, is occurring. The same behavior is not observed on some true OpenGL ES implementations.

gero3:
That means updating the specification to make this clear??

kenrussell:
It's undesirable to add this information to the WebGL specification. Unfortunately, it's legal for GLSL and GLSL ES shaders to fail compilation for basically any reason, if they use anywhere more than the minimum required number of uniform vectors (vertex or fragment). The WebGL tests assert that the minimum number of supported uniforms must compile, and one over the maximum number must fail to compile; but anything in the middle is a gray area.

We have observed that OpenGL ES drivers seem to do a better job of packing these uniforms than some desktop OpenGL drivers. We'd need to do more investigation to understand whether there are compatibility issues that cause the inefficient packing in OpenGL drivers. Since we've observed the poor behavior on multiple desktop platforms (Windows, Mac and Linux OpenGL drivers, and from multiple GPU vendors), enforcing new rules and getting them rolled out across the board is not a simple task, and would take a long time.

In the short term I recommend you work around the issue in Three.js by packing these uniform arrays yourself into 4-vectors. We'll continue to investigate this more, and if you have more information, please feel free to post it here.

gero3:
Actually I was more thinking of adding an amendum to https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.24 which would state that they webgl consumers itself shouldn't rely on the packing algoritm in https://www.khronos.org/registry/gles/specs/2.0/GLSL_ES_Specification_1.0.17.pdf
instead of what us writtennow in the specification:

> The OpenGL ES Shading Language, Version 1.00 [GLES20GLSL], Appendix A, Section 7 "Counting of Varyings and Uniforms" defines a conservative algorithm for computing the storage required for all of the uniform and varying variables in a shader. The GLSL ES specification requires that if the packing algorithm defined in Appendix A succeeds, then the shader must succeed compilation on the target platform.

kenrussell:
@gero3 Thanks for the suggestion. I've uploaded pull request https://github.com/KhronosGroup/WebGL/pull/1191 to provide some guidance. Please provide your comments.

PR #1168, opened 2015-08-27
Fixed Khronos bug 14551: WebGL 1.0.3: ternary-operators-in-global-ini…

kenrussell:
…tializers

Added "const" to the declarations of the variables on the two sides of the ternary operator.

### DIFF
--- a/conformance-suites/1.0.3/conformance/glsl/misc/ternary-operators-in-global-initializers.html
+++ b/conformance-suites/1.0.3/conformance/glsl/misc/ternary-operators-in-global-initializers.html
@@ -38,8 +38,8 @@
 <div id="console"></div>
 <script id="fragmentShader" type="text/something-not-javascript">
 precision mediump float;
-$(type) green = $(green);
-$(type) black = $(black);
+const $(type) green = $(green);
+const $(type) black = $(black);
 $(type) var = (true) ? green : black;
 void main() {
     gl_FragColor = $(asVec4);
--- a/sdk/tests/conformance/glsl/misc/ternary-operators-in-global-initializers.html
+++ b/sdk/tests/conformance/glsl/misc/ternary-operators-in-global-initializers.html
@@ -38,8 +38,8 @@
 <div id="console"></div>
 <script id="fragmentShader" type="text/something-not-javascript">
 precision mediump float;
-$(type) green = $(green);
-$(type) black = $(black);
+const $(type) green = $(green);
+const $(type) black = $(black);
 $(type) var = (true) ? green : black;
 void main() {
     gl_FragColor = $(asVec4);


kenrussell:
@zhenyao @RafaelCintron please review.

zhenyao:
LGTM

RafaelCintron:
LTGM after the fact.

Oletus:
A while back it was decided that WebGL should allow some types of non-constants in global initializers, since this had been the behavior historically. If that causes issues on the drivers of a specific vendor, a workaround should be added to shader translators such as ANGLE so that non-constant global initializers get put in a function. I don't think the test should have been changed, can we revert this?

https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.30

zhenyao:
What about we add a specific test for global initializers with non-constants and remove that from other tests?  This way testing the behavior is isolated and if we ever encounter a driver that's spec compliant, running conformance tests and failing that specific test will make it clear what the issue is.

kenrussell:
Mo, that's a good idea. Olli, this test was added specifically to test ternary operators in initializers, and the spec's clear that the two arms of the ternary operator must be constant expressions, so this strongly seemed like a bug in the test. Let's add other tests verifying the behavior of other non-constants in global initializers.

Oletus:
Actually, now that I think of it I don't think that the test works as intended after this commit - now the ternary operator will just end up being constant folded, when the problem that this patch exposed was that the non-constant ternary operator was being converted to if statements in ANGLE, which wasn't appropriate in the global scope.

There's also already a specific test for global initializers with non-constants, but only in the 1.0.4 suite, not in the 1.0.3 suite.

kenrussell:
That might be true with ANGLE as the shader translator but isn't necessarily how Microsoft's WebGL implementation works; the test was originally contributed by Microsoft and this change was LGTM'd by them above.

Let's add more tests verifying the behavior of global initializers with non-constants as you see necessary. This one previously violated the spec, and I think the complaint that was raised by one vendor was valid.

Oletus:
I wasn't aware that Microsoft contributed the test, but it did reveal a bug in ANGLE, and now that coverage is thrown away. I can't say I agree with that.

Essentially, to restore coverage, I'd just want to add this same test back to the test suite. But would that be accepted, if the complaint about the test is seen as valid? I can see some reason for removing the test from the 1.0.3 suite, as it wasn't correct according to the WebGL spec at the time when 1.0.3 was frozen, but I don't understand why we shouldn't have it in the TOT conformance suite, unless we want to revisit the spec as well.

kenrussell:
Apologies, I didn't realize this test was essentially a regression test for an ANGLE bug.

Could you please add another case to sdk/tests/conformance/glsl/misc/global-variable-init.html (or some other place) in the top of tree conformance suite testing the specific condition you're interested in? Let's please also make sure that the associated section in the WebGL specification https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.30 ensures that this is required to work.

PR #1170, opened 2015-08-28
WebGL2 - Improve tex image and sub image tests

djg:
I fixed up the texImage2D and texSubImage2D tests that work with various DOM elements to work with GL_RED as well as unsigned integer textures.

I made some improvements to the tests taking `image data` to exercise that pre-multiplied alpha was applied.

PTAL

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-utils.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-utils.js
@@ -1,0 +1,249 @@
+/*
+** Copyright (c) 2012 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+var TexImageUtils = (function() {
+
+  "use strict";
+
+  var wtu = WebGLTestUtils;
+
+  /**
+   * A vertex shader for a single texture.
+   * @type {string}
+   */
+  var simpleTextureVertexShaderES3 = [
+    '#version 300 es',
+    'in vec4 vPosition;',
+    'in vec2 texCoord0;',
+    'out vec2 texCoord;',
+    'void main() {',
+    '    gl_Position = vPosition;',
+    '    texCoord = texCoord0;',
+    '}'].join('\n');
+
+  /**
+   * A fragment shader for a single integer texture.
+   * @type {string}
+   */
+  var simpleUintTextureFragmentShaderES3 = [
+    '#version 300 es',
+    'precision mediump float;',
+    'uniform usampler2D tex;',
+    'in vec2 texCoord;',
+    'out vec4 fragData;',
+    'void main() {',
+    '    fragData.rgb = texture(tex, texCoord).rgb / 255.0;',
+    '    fragData.a = 1;',
+    '}'].join('\n');
+
+  /**
+   * A fragment shader for a single cube map integer texture.
+   * @type {string}
+   */
+    // A fragment shader for a single integer cube map texture.
+  var simpleCubeMapUintTextureFragmentShaderES3 = [
+    '#version 300 es',
+    'precision mediump float;',
+    'uniform usamplerCube tex;',
+    'uniform int face;',
+    'in vec2 texCoord;',
+    'out vec4 fragData;',
+    'void main() {',
+    // Transform [0, 1] -> [-1, 1]
+    '    vec2 texC2 = (texCoord * 2.) - 1.;',
+    // Transform 2d tex coord. to each face of TEXTURE_CUBE_MAP coord.
+    '    vec3 texCube = vec3(0., 0., 0.);',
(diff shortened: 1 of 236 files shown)


kenrussell:
Thanks Dan. This LGTM overall -- no significant comments -- but @zhenyao should review and merge.

djg:
@zhenyao - Feel free to flatten into one commit when merging. (Not sure if that is possible)

sdk/tests/js/tests/tex-image-and-sub-image-2d-utils.js
```diff
@@ -0,0 +1,249 @@
+/*
+** Copyright (c) 2012 The Khronos Group Inc.
```
zhenyao: 2012 -> 2015

sdk/tests/js/tests/tex-image-and-sub-image-2d-utils.js
```diff
@@ -0,0 +1,249 @@
...
+   * Does the GL internal format represent an unsigned integer format
+   * texture?
+   * @return {boolean}
+   */
+  var isUintFormat = function(internalFormat)
+  {
+    return (internalFormat == "R8UI" ||
+            internalFormat == "RG8UI" ||
+            internalFormat == "RGB8UI" ||
+            internalFormat == "RGBA8UI");
```
zhenyao: RGB10_A2UI

zhenyao: What about 16UI and 32UI?

sdk/tests/js/tests/tex-image-and-sub-image-2d-utils.js
```diff
@@ -0,0 +1,249 @@
...
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+var TexImageUtils = (function() {
```
zhenyao: I feel this should be part of the WTU, so it can be used in tests other than texture ones.  But let's merge with the current design and we can iterate later.

sdk/tests/js/tests/tex-image-and-sub-image-2d-utils.js
djg: Those formats are not allowed from DOM elements, which is why I didn't add the code to WebGLTestUtils.

`When the data source is a DOM element (Image, Canvas, or Video), commonly each channel's representation is an unsigned integer type of at least 8 bits. Converting such representation to signed integers or unsigned integers with more bits is not clearly defined. For example, when converting RGBA8 to RGBA16UI, it is unclear whether or not the intention is to scale up values to the full range of a 16-bit unsigned integer. Therefore, only converting to unsigned integer of at most 8 bits, half float, or float is allowed.`

sdk/tests/js/tests/tex-image-and-sub-image-2d-utils.js
djg: It could be, but the fragment shaders are specialized for the tests in conformance2/textures/{canvas,image,image_data,svg_image,video,webgl_canvas} folders. For example, unsigned integer sampling is scaled by 1/255 to bring 0-255 into 0-1 range for final colour output.

This is why I didn't add the code to WebGLTestUtils and created a new support file.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js
```diff
@@ -118,18 +144,20 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
```
zhenyao: This is problematic.  In WebGL, we generate an INVALID_OPERATION if clear* function type and the fbo color image type mismatch.  For example, if you use RED_INTEGER tex, you should use clearBufferiv() instead of clear().

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js
```diff
@@ -118,18 +144,20 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
 
             // Check the top pixel and bottom pixel and make sure they have
             // the right color.
-            debug("Checking bottom pixel");
-            wtu.checkCanvasRect(gl, 0, 0, 1, 1, bottomColor, "shouldBe " + bottomColor);
-            debug("Checking top pixel");
-            wtu.checkCanvasRect(gl, 0, 1, 1, 1, topColor, "shouldBe " + topColor);
+            debug("Checking " + (flipY ? "top" : "bottom"));
+            wtu.checkCanvasRect(gl, 0, bottom, halfWidth, halfHeight, tl, "shouldBe " + tl);
```
zhenyao: This is also problematic.  We can't use RGBA/UNSIGNED_BYTE for readPixels() if it's INTEGER type color image.  We will have to use RGBA/INTEGER.

zhenyao: The same comments apply to all other tests as well.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js
djg: No, the test isn't drawing into a RED_INTEGER texture, it's sampling from it. The framebuffer is RGBA8 in all the tests.

zhenyao:
Thanks for working on these tests.  There are some extra work necessary to make it totally correct, but this is a nice improvement toward the right direction.  I am OK merging it right now, and iterate to improve in followup CLs.  Let me know.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js
zhenyao: Ah I see.  Then your code is correct.

zhenyao:
Thanks for the contribution.  More work is needed in these tests, for example, handling signed integer tests, etc.  Let's get this right collectively.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js
djg: But the tests are all against RGBA8 framebuffer. These suite of tests only test loading image data from the DOM elements (<CANVAS>, , <SVG>, <VIDEO>, etc) into a texture that is then rendered into the default framebuffer attached to another <CANVAS> element, which is RGBA8.

zhenyao:
The method you used here is very smart, walking around the clear/readback issues I pointed out above (incorrectly).

However, I think rendering into integer/unsigned integer and readback from them should be part of these tests, as they are very comprehensive (testing over all renderable formats).  We probably don't want to generate another patch of tests testing these function over all formats.

PR #1171, opened 2015-09-01
Fixed bug where transform feedback varyings were not being added for …

kenrussell:
…those tests.

Deleted obsolete comment.


kenrussell:
@zhenyao please review.

@jasomobica , @toji : FYI.

This doesn't make the test pass yet even with Chromium-side bug fixes. Still debugging.

PR #1174, opened 2015-09-02
Clean up GLSL constructor corner case test

Oletus:
Make it use the GLSL test framework instead of having custom testing
code, and add mention of the relevant spec section.


kenrussell:
Nice cleanup. LGTM

PR #1176, opened 2015-09-02
Modified EXTDisjointTimerQuery conformance tests to match the spec.

DavidYen:
The isQueryEXT() call should return false until beginQueryEXT() is
called. This should match the OpenGL ES 3.0 behavior outlined in
section 2.13:
  These names are marked as used, for the purposes of GenQueries
  only, but no object is associated with them until the first time
  they are used by BeginQuery.
This same behavior is outlined in the original OpenGL ES
EXT_Disjoint_Timer_Query spec.

The failure values tested for beginQueryEXT and getQueryObjectEXT
should both be INVALID_OPERATION as outlined in the error section
of the EXT_disjoint_timer_query spec.


kenrussell:
LGTM after the fact.

PR #1177, opened 2015-09-02
Test generateMipmap will fail when base level texture is not specified

qkmiao:


sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
```diff
@@ -122,6 +122,22 @@
...
+  wtu.fillTexture(gl, tex, 4, 4, [255, 0, 0, 255], 3, gl.RGBA, gl.UNSIGNED_BYTE);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "fillTexture(4,4,3) should succeed");
+  wtu.fillTexture(gl, tex, 2, 2, [0, 0, 255, 255], 4, gl.RGBA, gl.UNSIGNED_BYTE);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "fillTexture(2,2,4) should succeed");
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_BASE_LEVEL, 4);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texParameter(TEXTURE_BASE_LEVEL) should succeed");
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAX_LEVEL, 2);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texParameter(TEXTURE_MAX_LEVEL) should succeed");
+  gl.generateMipmap(gl.TEXTURE_2D);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "generateMipmap should fail");
```
zhenyao: I don't think this is the correct behavior.  Looking at ES spec 3.0.4, page 157, under Mipmapping, the level_max is clamped to the reasonable range [level_base, max], so there should be no error here, just an no op.

What would be interesting to test is the base level texture is not specified and an INVALID_OPERATION is generated.

kenrussell:
The behavior verified by this test doesn't seem to be defined in the OpenGL ES 3.0.4 spec nor in the manual page for glGenerateMipmap. Section 3.8.13 "Texture Completeness" requires level_base <= level_max for mipmap completeness, but other than that it doesn't seem to be required that GenerateMipmap produce an error if level_max < level_base. What's the motivation for adding this test? If a change is going to be made in this area, the specification should be updated.

qkmiao:
Thanks @kenrussell @zhenyao . I updated the patch to check if base level texture is not specified INVALID_OPERATION  is generated.

sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
```diff
@@ -122,6 +122,20 @@
...
+
+  // Test base level texture isn't specified.
+  var tex = gl.createTexture()
+  gl.bindTexture(gl.TEXTURE_2D, tex);
+  wtu.fillTexture(gl, tex, 8, 8, [255, 0, 0, 255], 2, gl.RGBA, gl.UNSIGNED_BYTE);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "fillTexture(8,8,2) should succeed");
+  wtu.fillTexture(gl, tex, 4, 4, [255, 0, 0, 255], 3, gl.RGBA, gl.UNSIGNED_BYTE);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "fillTexture(4,4,3) should succeed");
+  wtu.fillTexture(gl, tex, 2, 2, [0, 0, 255, 255], 4, gl.RGBA, gl.UNSIGNED_BYTE);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "fillTexture(2,2,4) should succeed");
```
zhenyao: The message "(2,2,4)" is a typo?

sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
```diff
@@ -122,6 +122,20 @@
...
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_BASE_LEVEL, 1);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texParameter(TEXTURE_BASE_LEVEL) should succeed");
+  gl.generateMipmap(gl.TEXTURE_2D);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "generateMipmap should fail if base level texture is not specified");
 })();
```
zhenyao: It's worth setting base level to 2 and see generateMipmap succeeds.

sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
qkmiao: Sorry, I didn't get this.

zhenyao: I don't understand what fillTexture(2,2,4) means?

kenrussell: Looking above, it seems to refer to arguments 2, 3, and 5 above. (2x2 texture at level 4.) Is that correct?

zhenyao: Ah I see.  It will be a much cleaner message if it says fillTexture(2x2, level=4) instead. here and above.

qkmiao: The message was updated.

zhenyao:
LGTM

PR #1179, opened 2015-09-02
Fix spec compliance of conformance2/buffers/buffer-type-restrictions

toji:
The behavior being changed is described by section 5.1 and 5.2 of the WebGL 2 spec


kenrussell:
This LGTM but @zhenyao should review too.

zhenyao:
LGTM

PR #1180, opened 2015-09-03
Fixed bugs in primitive type handling, queries and loop structure.

kenrussell:
These fixes make more of the tests pass but it's unclear whether all of
them will; Chromium must virtualize transform feedback handling better.


kenrussell:
@zhenyao @toji @jasomobica : apologies but I'm merging this myself to be able to better continue work on it later. Deliberately left in some commented-out debugging code. These fixes are definitely correct, but I'm not sure all the tests pass. Need to test on other platforms. This triggers problems in Chromium's virtual GL context implementation on Linux where it looks like transform feedback objects need to be virtualized better.

PR #1182, opened 2015-09-03
Added more real query tests to ext-disjoint-timer-query.html.

DavidYen:
Previously the conformance test for EXT_disjoint_timer_query
only looked for error results. I've added a simple test which
tests that the query makes sure the results become available
and give reasonable results.

### DIFF
--- a/sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
+++ b/sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
@@ -49,6 +49,7 @@ var query = null;
 var elapsed_query = null;
 var timestamp_query1 = null;
 var timestamp_query2 = null;
+var availability_retry = 16;
 
 if (!gl) {
     testFailed("WebGL context does not exist");
@@ -68,6 +69,7 @@ if (!gl) {
         runElapsedTimeTest();
         runTimeStampTest();
 
+        gl.finish();
         setTimeout(checkQueryResults, 0);
     }
 }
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
```diff
@@ -123,10 +134,65 @@
...
+    ext.queryCounterEXT(timestamp_query2, ext.TIMESTAMP_EXT);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "Timestamp queries should have no errors");
+}
+
+function checkQueryResults() {
+    debug("");
+    debug("Testing query results");
+
+    // Make sure queries are available.
+    gl.finish();
```
toji: In Chrome at least gl.finish actually executes a flush. The only way to actually block and finish it to call gl.readPixels. You could do that here or (more elegantly IMO) check for QUERY_RESULT_AVAILABLE in a rAF loop, timing out if you don't see a success in a reasonable number of iterations/milliseconds. Otherwise this test looks good!

DavidYen: I've made it retry 16 times in 1ms intervals now, PTAL!

sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
```diff
@@ -58,6 +62,15 @@
...
         runSanityTests();
+
+        // Clear disjoint value.
+        gl.getParameter(ext.GPU_DISJOINT_EXT);
+
+        runElapsedTimeTest();
+        runTimeStampTest();
+
+        gl.finish();
+        setTimeout(checkQueryResults, 0);
```
toji: This is definitely an improvement, but I'd like to request that we use `requestAnimationFrame(checkQueryResults)` here and below instead of `setTimeout`. This test will likely be a template for future query tests, and as such we should try and model it after more realistic code/best practices.

kenrussell: Also let's please retry fetching the query's availability and result for some more reasonable period like half a second (500 ms) -- 16 ms might be optimistic.

kenrussell:
I'd like to defer review to @toji . One minor comment.

sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
DavidYen: Ok, I addressed these issues in the latest patch. Also I have added a bunch of tests testing the conditions involving getting the current query.

toji:
Thanks for the updates! LGTM!

PR #1183, opened 2015-09-03
Add a conformance test for an ANGLE crash caused by the hierarchical …

Kangz:
…voronoi Shadertoy sample

Checked it crashes in Chromium on Windows and doesn't crash with a ToT ANGLE or Chromium on Linux


kenrussell:
Excellent. Thank you for tracking this down and creating this regression test.

Kangz:
No problem, thank you for merging.

PR #1185, opened 2015-09-04
Added clipping, fbo invalidate, shader api, boolean state query and s…

jasomobica:
…ync tests.

Plus cleanup of HTML files.


toji:
I'd like to requests that simple, noisy changes like the removal of the DOCTYPE here be done in their own pull request. The way that Github presents changes for review makes it hard to filter those simple changes out and focus on the newly added tests.

jasomobica:
Sorry about that. I am traditionally squashing commits before creating pull requests.

sdk/tests/deqp/modules/shared/glsStateQuery.js
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * Verify that WebGL state 'param' has the expected value
+ * @param {number} param
+ * @param {number|Array<number>} reference
+ * @return {boolean}
+ */
+glsStateQuery.verifyGreaterOrEqual = function(param, reference) {
+    var value = gl.getParameter(param);
+    if (reference instanceof Array) {
```
kenrussell: Not a big deal, but perhaps a glsStateQuery.compareImpl could be added which takes a function object as argument which is used to compare each individual array element in the case of arrays. That way this and verifyLessOrEqual could have less duplicated code.

kenrussell:
Thanks for this contribution. Going forward I agree with Brandon's assessment that large, boilerplate changes like the doctype removal here should be done in independent pull requests. Per our offline discussion I realize that this would cause a bit of a headache with this pull request, so merging this as is now. Other than that, this looks excellent.

PR #1186, opened 2015-09-04
Add new tests for ternary operators in global initializers

Oletus:
This is to restore coverage that was lost when
glsl/misc/ternary-operators-in-global-initializers.html was recently
changed.

See #1168.

### DIFF
--- a/sdk/tests/conformance/glsl/misc/global-variable-init.html
+++ b/sdk/tests/conformance/glsl/misc/global-variable-init.html
@@ -166,6 +166,26 @@ void main() {
     gl_Position = aPosition;
 }
 </script>
+<script id="globalNonConstTernary" type="x-shader/x-fragment">
+precision mediump float;
+float green = 1.0;
+float black = 0.0;
+float f = true ? green : black;
+
+void main() {
+    gl_FragColor = vec4(0.0, f, 0.0, 1.0);
+}
+</script>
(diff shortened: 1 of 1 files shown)


kenrussell:
For future reference:

The original bug (Khronos bug 14551, unfortunately not filed on the public issue tracker) complained that the ternary-operators-in-global-initializers test violated the text in the GLSL ES specification version 1.0.17, section 4.3 "Storage Qualifiers":

"Declarations of globals without a storage qualifier, or with just the const qualifier, may include initializers, in which case they will be initialized before the first line of main() is executed. Such initializers must be a constant expression."

It was found during extensive discussion in both the WebGL and OpenGL ES working groups that there are many drivers which don't enforce the rule that global variables must be initialized with constant expressions. Further, Olli did extensive testing and found that a significant amount of existing WebGL content would break if ANGLE's shader validator began enforcing these rules. For this reason these constraints were relaxed in the current version of the WebGL specification in https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.30 .

It seemed that it was only an accident that the ternary-operators-in-global-initializers test used non-constant initializer expressions, which motivated pull request #1168 . However, as Olli points out, the test coverage was both significant, and allowed by the addendum in the WebGL specification.

Merging this now. We will have to either figure out workarounds to make this test work on stricter OpenGL ES implementations, or relax those implementations to match the behavior of others.

PR #1187, opened 2015-09-04
Added vertexAttribI4i and vertexAttribI4ui variants.

jasomobica:


kenrussell:
Thanks, seems fine. I'll follow up to this to add a typedef for these, similar to UniformMatrixFVSource.

PR #1190, opened 2015-09-07
Packing restrictions

kenrussell:


kenrussell:
Did not mean to include the change to the conformance test. Closing; will open a new one.

PR #1196, opened 2015-09-09
Add sampler drawing test for WebGL 2.0

legendlee1314:


legendlee1314:
PTAL. This test case is referred from [conformance/textures/misc/texparameter-test.html](https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/conformance/textures/misc/texparameter-test.html).

legendlee1314:
@zhenyao , @kenrussell , this test case can not pass in Chrome now and I'm looking into this issue.

The root cause is that `samplerParameter` does not affect parameters of texture object cached both in `blink` and `command buffer`. So the prediction of whether the bound texture can be rendered is always false.

If you think this test case makes sense, I'd like to discuss with you about my preliminary solution. Thank you!

zhenyao:
The general idea of the test looks good to me.  However, I feel it's more complicated than necessary.  For example, do we really need the "world" matrix transformation?  Can you design the coordinates and tex coordinates so we can verify the different behaviors from three texture wrap mode, but without the transformation?

kenrussell:
Agree with Mo's review. The world transformation seems unnecessary. Please keep the logic as simple as possible to expose the specific bug in question.

Please also use WebGLTestUtils as much as possible to set up the geometric data. There are already routines like setupTexturedQuad which will generate a unit quad with texture coordinates.

legendlee1314:
Agreed. It's a little complicated when I touch that codes in texparameter-test.html.

The patch is updated. Please review again. Thanks!

sdk/tests/conformance2/samplers/sampler-drawing-test.html
```diff
@@ -0,0 +1,145 @@
...
+        for (var xx = 0; xx < 12; ++xx) {
+            var ec = [0, 0, 0, 0];
+            switch (param) {
+                case gl.REPEAT:
+                    if (xx % 2 == 1 && yy % 2 == 1) {
+                        ec = color;
+                    }
+                    break;
+                case gl.CLAMP_TO_EDGE:
+                    if (xx < 6 && yy < 6) {
```
zhenyao: I don't think this is correct.  Your tex coordinates is [-2.5, 3.5], so it should be a value less than 6.

sdk/tests/conformance2/samplers/sampler-drawing-test.html
```diff
@@ -0,0 +1,145 @@
...
+                        ec = color;
+                    }
+                    break;
+                case gl.MIRRORED_REPEAT:
+                    if (xx % 4 < 2 && yy % 4 < 2) {
```
zhenyao: I think this should be >= 2 and >= 2.

sdk/tests/conformance2/samplers/sampler-drawing-test.html
legendlee1314: I use `wtu.setupTexturedQuadWithTexCoords(gl, [-2.5, -2.5], [3.5, 3.5]);` instead of the former case in `texparameter-test.html`, which uses a y-flipped coordinates ([-2.5, 3.5], [3.5, -2.5]). According my calculation with coordinates of `canvas`, `texture` and `readPixels`, it's correct. Since this case can not run in Chrome now, you can replace `samplerParameter` to `texParameter` to check.

zhenyao:
OK, it looks like I might have misunderstood the meaning of [-2.5, -2.5], [3.5, 3.5].  Let me re-check the coordinates later.

PR #1197, opened 2015-09-09
Fix wrong level of texSubimage3D

qkmiao:
When testing dimension out of range, level should be correct. Otherwise,
test result isn't what we want.


zhenyao:
LGTM. Thanks for catching the bug.

PR #1198, opened 2015-09-09
Fix a bunch of tests timeing out

zhenyao:
Really trivia bugs


zhenyao:
Really trivia bug fix.  Merge myself.

kenrussell:
LGTM after the fact.

PR #1204, opened 2015-09-11
fix bugs in gl-object-get-calls.js

Richard-Yunchao:


sdk/tests/js/tests/gl-object-get-calls.js
```diff
@@ -600,6 +600,7 @@ if (contextVersion > 1) {
   shouldBe('gl.getUniform(samplerForWebGL2Program, s3DValLoc)', '0');
   shouldBe('gl.getUniform(samplerForWebGL2Program, s2DArrayValLoc)', '1');
 }
+wtu.glErrorShouldBe(gl, gl.NO_ERROR);
```
Richard-Yunchao: Add this line to show the error in getUniform, and prevent the errors fall to getVertexAttrib.

zhenyao:
Great!

PR #1205, opened 2015-09-11
Fix more bugs in gl-object-get-calls.js

zhenyao:
PTAL

### DIFF
--- a/sdk/tests/resources/matForWebGL2UniformShader.vert
+++ b/sdk/tests/resources/matForWebGL2UniformShader.vert
@@ -30,12 +30,14 @@ uniform mat3x4 mval3x4;
 uniform mat4x2 mval4x2;
 uniform mat4x3 mval4x3;
 
+out vec4 position;
+
 void main()
 {
-    gl_Position = vec4(mval2x3 * vec2(1.0, 2.0), 0.0, 0.0)
-            + vec4(mval2x4 * vec2(1.0, 2.0), 0.0, 0.0)
-            + vec4(mval3x2 * vec3(1.0, 2.0, 3.0), 0.0)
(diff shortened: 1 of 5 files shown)


toji:
LGTM!

sdk/tests/resources/matForWebGL2UniformShader.vert
```diff
@@ -30,12 +30,14 @@ uniform mat3x4 mval3x4;
...
+
 void main()
 {
-    gl_Position = vec4(mval2x3 * vec2(1.0, 2.0), 0.0, 0.0)
-            + vec4(mval2x4 * vec2(1.0, 2.0), 0.0, 0.0)
-            + vec4(mval3x2 * vec3(1.0, 2.0, 3.0), 0.0)
-            + vec4(mval3x4 * vec3(1.0, 2.0, 3.0), 0.0)
-            + mval4x2 * vec4(1.0, 2.0, 3.0, 4.0)
-            + mval4x3 * vec4(1.0, 2.0, 3.0, 4.0);
+  position = vec4(mval2x3 * vec2(1.0, 2.0), 0.0); +
```
kenrussell: It looks like there's an incorrect semicolon on this line.

sdk/tests/resources/matForWebGL2UniformShader.vert
```diff
@@ -30,12 +30,14 @@ uniform mat3x4 mval3x4;
...
+      mval2x4 * vec2(1.0, 2.0); +
```
kenrussell: Here too.

kenrussell:
LGTM overall after the fact, but one minor issue which I think should be cleaned up in a follow-on pull request.

PR #1206, opened 2015-09-11
Added new deqp tests and bugfixes.

jasomobica:
Added default vertex attribute, occlusion query, sync, shader texture functions and readpixels tests.
Added state query tests for: shaders, fbo, rbo, buffer objects, internal formats, string queries and indexed parameters.

Improved test results logging and fixed a possible infinite loop in draw tests.


sdk/tests/deqp/functional/gles3/es3fShaderStateQueryTests.js
```diff
@@ -0,0 +1,2241 @@
...
+
+        // test
+        if (this.check(glsStateQuery.verifyProgram(program, gl.LINK_STATUS, true), 'Program link fail' + gl.getProgramInfoLog(program))) {
+            var indices = gl.getUniformIndices(program, ['uniformValue']);
+            var info_type = gl.getActiveUniforms(program, indices, gl.UNIFORM_TYPE)[0];
+            var info_size = gl.getActiveUniforms(program, indices, gl.UNIFORM_SIZE)[0];
+            var info_is_row_major = gl.getActiveUniforms(program, indices, gl.UNIFORM_IS_ROW_MAJOR)[0];
+            this.check(glsStateQuery.compare(info_size, size));
+            this.check(glsStateQuery.compare(info_type, type));
+            this.check(glsStateQuery.compare(info_is_row_major, isRowMajor));
```
jasomobica: I have a doubt here. According to the spec gl.UNIFORM_IS_ROW_MAJOR is a GLint.
IMO the spec should better be changed and gl.UNIFORM_IS_ROW_MAJOR should be boolean. The test in its current shape expects it to be boolean.

kenrussell: Thanks for pointing this out. Submitted #1210 for that spec update.

sdk/tests/deqp/functional/gles3/es3fDefaultVertexAttributeTests.js
```diff
@@ -0,0 +1,541 @@
...
+
+/** @const */ var RENDER_SIZE = 32;
+/** @const */ var s_valueRange = 10;
+/** @const */ var s_passThroughFragmentShaderSource = '#version 300 es\n' +
+                                                        'layout(location = 0) out mediump vec4 fragColor;\n' +
+                                                        'in mediump vec4 v_color;\n' +
+                                                        'void main (void)\n' +
+                                                        '{\n' +
+                                                        ' fragColor = v_color;\n' +
+                                                        '}\n';
```
kenrussell: It's unfortunate that there's duplication like this shader in the dEQP tests. However, I don't think it would be a good idea to try to use WebGLTestUtils' shaders here, since that would introduce an undesirable dependency preventing easy tracking of the C++ code.

sdk/tests/deqp/functional/gles3/es3fDefaultVertexAttributeTests.js
```diff
@@ -0,0 +1,541 @@
...
+    this.m_program = new gluShaderProgram.ShaderProgram(gl, gluShaderProgram.makeVtxFragSources(this.genVertexSource(), s_passThroughFragmentShaderSource));
+    if (!this.m_program.isOk())
+        testFailedOptions('could not build program', true);
+
+    var fullscreenQuad = [
+         1.0, 1.0, 0.0, 1.0,
+         1.0, -1.0, 0.0, 1.0,
+        -1.0, 1.0, 0.0, 1.0,
+        -1.0, -1.0, 0.0, 1.0
+    ];
```
kenrussell: Same duplication in things like this geometric data.

sdk/tests/deqp/functional/gles3/es3fOcclusionQueryTests.js
```diff
@@ -0,0 +1,461 @@
...
+
+    gl.endQuery(this.m_queryMode);
+    gl.finish();
+    gl.disable(gl.SCISSOR_TEST);
+    gl.disable(gl.STENCIL_TEST);
+    gl.disable(gl.DEPTH_TEST);
+
+    // Check that query result is available.
+    var resultAvailable = /** @type {boolean} */ (gl.getQueryParameter(query, gl.QUERY_RESULT_AVAILABLE));
+    assertMsgOptions(resultAvailable, 'Occlusion query failed to return a result after glFinish()', false, true);
```
kenrussell: These sorts of checks in these tests -- where a query's result must be available after a glFinish -- are going to have to be restructured. See the text in this WebGL 1.0 extension:

https://www.khronos.org/registry/webgl/extensions/EXT_disjoint_timer_query/

"A query's result must not be made available until the current frame has completed rendering."

This is needed to provide uniform behavior among implementations, and prevent developers from adding busy loops to their applications.

I'm soon going to restructure Chromium's implementation of WebGL query objects to honor that requirement, at which point basically all of the ported dEQP tests that use queries are going to break. We will need to change them to issue multiple requestAnimationFrame calls, waiting for the query result to be available, and then proceed with the test.

How difficult is it going to be to add a "second part" to all of the tests which use queries? I'm thinking of something like a callback function which acts as a predicate indicating whether it can proceed, and another callback with the rest of the test. Or, multiple callbacks in the case where the test issues multiple queries.

sdk/tests/deqp/functional/gles3/es3fShaderStateQueryTests.js
```diff
@@ -0,0 +1,2241 @@
...
+        'longlongUniformBlockName.vector2',
+        'shortUniformBlockName.vector2',
+        'shortUniformBlockName.vector4'
+    ];
+
+    // test UNIFORM_BLOCK_INDEX
+
+    var uniformIndices = gl.getUniformIndices(program, uniformNames);
+
+    /* TODO: wait for spec clarification */
```
kenrussell: What spec clarification is needed here?

sdk/tests/deqp/modules/shared/glsLifetimeTests.js
```diff
@@ -745,7 +745,7 @@ glsLifetimeTests.LifeTest.testGen = function() {
     else
         assertMsgOptions(!this.m_type.exists(obj), 'create* should not have created an object, but did', false, true);
     this.m_type.release(obj);
-    testPassed(this.fullName() + ' ' + this.description);
+    testPassed('Passed');
```
kenrussell: Out of curiosity why are these messages being simplified?

kenrussell: Never mind, I see the change below in testPassed.

kenrussell: I do think that passing the string 'Passed' is redundant here.

sdk/tests/js/js-test-pre.js
```diff
@@ -214,6 +214,9 @@ var TestFailedException = function (message) {
 
 function testPassed(msg)
 {
+    if (_currentTestName)
+      msg = _currentTestName + ': ' + msg;
```
kenrussell: Perhaps you could simplify the callers by writing something like:

  if (!msg)
    msg = 'Passed'
  if (_currentTestName)
    ...

sdk/tests/js/js-test-pre.js
```diff
@@ -225,6 +228,9 @@ function testPassed(msg)
 
 function testFailed(msg)
 {
+    if (_currentTestName)
+      msg = _currentTestName + ': ' + msg;
```
kenrussell: Same here (with 'Failed')

kenrussell:
Thank you for this substantial work. It looks good overall and I'm merging it as is. A few comments; let's please work toward addressing them in future pull requests. Thanks.

sdk/tests/deqp/functional/gles3/es3fOcclusionQueryTests.js
jasomobica: If occlusion query results are not available immediately after glFinish(), then doesn't it make occlusion query useless?
My understanding is that, in real life applications, we use occlusion query with a simple bonding box to see if an object is visible. If it is visible, then we start the real rendering.
If we need to return the control to the browser to get the query results, wouldn't it be too slow to be worth the effort?

sdk/tests/deqp/functional/gles3/es3fShaderStateQueryTests.js
jasomobica: Old comment. I'll remove it in a follow up patch.

sdk/tests/deqp/functional/gles3/es3fOcclusionQueryTests.js
kenrussell: In order to get proper pipelining of occlusion query results it's necessary to wait until the next frame to retrieve them. My understanding is that most engines use the occlusion query results from the previous frame. Yes, this means that the occlusion results are a frame behind. It doesn't make them useless.

PR #1208, opened 2015-09-14
Simplify test case of texparameter-test.html

legendlee1314:
1. Remove world transformation.
2. Use WebGLTestUtils to setup texture quad.

### DIFF
--- a/sdk/tests/conformance/textures/misc/texparameter-test.html
+++ b/sdk/tests/conformance/textures/misc/texparameter-test.html
@@ -35,167 +35,112 @@
 <script src="../../../js/webgl-test-utils.js"></script>
 </head>
 <body>
-<canvas id="example" width="24" height="24"></canvas>
-<canvas id="canvas2d" width="2" height="2"></canvas>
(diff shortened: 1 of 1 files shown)


legendlee1314:
PTAL. Thanks!

kenrussell:
This pull request does more than just this -- it also removes the array of textures set up at the beginning. While creating those textures one at a time may be functionally equivalent, I'm concerned that we may be losing key aspects of this longstanding test.

We'll need to carefully review these changes together.

zhenyao:
Can you address kbr's concern and put back the texture array, and modify runDrawingTest(param) to also take a texture parameter.

legendlee1314:
Put the texture array back. Please review.

zhenyao:
LGTM
@kenrussell Ken, do you have more concerns?

kenrussell:
The revised test looks OK. My main concern is around changing the logic of these tests too much, and accidentally losing test coverage; but after looking through the verification code, it seems it's pretty similar. Thanks for this simplification.

PR #1210, opened 2015-09-14
Changed getActiveUniforms to return sequence of GLboolean for UNIFORM…

kenrussell:
…_IS_ROW_MAJOR.

This change is based on feedback from Janusz Sobczak from Mobica.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -27,7 +27,7 @@
     <!--end-logo-->
 
     <h1>WebGL 2 Specification</h1>
-    <h2 class="no-toc">Editor's Draft 5 September 2015</h2>
+    <h2 class="no-toc">Editor's Draft 14 September 2015</h2>
     <dl>
         <dt>This version:
             <dd>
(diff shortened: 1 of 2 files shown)


kenrussell:

@jasomobica FYI.

Note that UNIFORM_NAME_LENGTH is elided from this table, per the section on String Length Queries.

specs/latest/2.0/index.html
```diff
@@ -2391,12 +2391,24 @@ <h2 class="no-toc">Table of contents</h2>
...
             <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.3.pdf#nameddest=section-2.11.6">OpenGL ES 3.0.3 &sect;2.11.6</a>,
             <a class="nonnormative" href="https://www.khronos.org/opengles/sdk/docs/man3/html/glGetActiveUniformsiv.xhtml">man page</a>)</span>
         <dd>
             Queries the value of the parameter named <code class="param">pname</code> for each of the uniforms within <code class="param">program</code> whose indices
-            are specified in the array of <code class="param">uniformIndices</code>.
+            are specified in the array of <code class="param">uniformIndices</code>. The type returned is the natural type for the requested
+            <code class="param">pname</code>, as given in the following table:
+            <table class="foo">
+                <tr><th>pname</th><th>returned type</th></tr>
+                <tr><td>UNIFORM_TYPE</td><td>sequence&lt;GLint&gt;</td></tr>
```
zhenyao: Should we change this to sequence<GLenum>?

kenrussell: Yes, thanks, good point. Will change this.

kenrussell:
@zhenyao addressed your review feedback.

specs/latest/2.0/index.html
jasomobica: I second that.

specs/latest/2.0/index.html
```diff
@@ -2391,12 +2391,24 @@ <h2 class="no-toc">Table of contents</h2>
...
+                <tr><td>UNIFORM_SIZE</td><td>sequence&lt;GLint&gt;</td></tr>
```
jasomobica: And this could GLuint because negative values are not allowed for UNIFORM_SIZE

kenrussell: Will be changed in the next version.

kenrussell:
Thanks for the feedback. Any further comments?

zhenyao:
LGTM

PR #1189, opened 2015-09-05
Added union type typedefs for vertexAttribI4iv and vertexAttribI4uiv.

kenrussell:

PR #1212, opened 2015-09-15
Add union type typedefs for vertexAttrib*fv, and make argument names …

kenrussell:
…consistent.

Follow-up to #1189.


kenrussell:

@jasomobica FYI.

zhenyao:
Great!

PR #1213, opened 2015-09-15
update bufferData and bufferSubData test

Richard-Yunchao:
1) code refactoring
2) add bufferData test without ArrayBuffer
3) add cases for bufferSubData and bufferData when size/offset is zero


sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
```diff
@@ -0,0 +1,163 @@
...
+    gl.bufferSubData(gl.ARRAY_BUFFER, 0, array);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+    gl.bufferSubData(gl.ARRAY_BUFFER, 0, new Float32Array(0));
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+    gl.bindBuffer(gl.ARRAY_BUFFER, null);
+}
+
+finishTest();
```
zhenyao: No need for finishTest() here.  This test is static.

sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
```diff
@@ -0,0 +1,163 @@
...
+
+    gl.bindBuffer(gl.ARRAY_BUFFER, buf);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+    gl.bufferData(gl.ARRAY_BUFFER, null, gl.STATIC_DRAW);
+    wtu.glErrorShouldBe(gl, gl.INVALID_VALUE);
+
+    // This should not crash, but the selection of the overload is ambiguous per Web IDL.
+    gl.bufferData(gl.ARRAY_BUFFER, null, gl.STATIC_DRAW);
+    gl.getError();
```
zhenyao: You can get rid of this case.  The spec has been updated since to return INVALID_VALUE, and it's tested right above.

sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
```diff
@@ -0,0 +1,163 @@
...
+    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION);
+
+    gl.bindBuffer(gl.ARRAY_BUFFER, buf);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+    gl.bufferData(gl.ARRAY_BUFFER, -4, gl.STATIC_DRAW);
+    wtu.glErrorShouldBe(gl, gl.INVALID_VALUE);
+
+    gl.bufferData(gl.ARRAY_BUFFER, 0, gl.STATIC_DRAW);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
```
kenrussell: Per https://codereview.chromium.org/1339273002/ this change does not look good to me. There is no good reason to support allocation of zero-sized buffers, and the checks in Chromium's implementation probably date back to workarounds for bugs in graphics drivers.

sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
```diff
@@ -0,0 +1,163 @@
...
+    gl.bufferData(gl.ARRAY_BUFFER, null, gl.STATIC_DRAW);
+    gl.getError();
+
+    gl.bufferData(gl.ARRAY_BUFFER, array, gl.STATIC_DRAW);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+    arrayZero = new ArrayBuffer(0);
+    shouldBeNonNull("arrayZero");
+    gl.bufferData(gl.ARRAY_BUFFER, arrayZero, gl.STATIC_DRAW);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
```
kenrussell: What do the other WebGL implementations (Firefox's, Apple's, Microsoft's) do in this case? Since passing a null ArrayBuffer is disallowed, and Chrome's implementation has historically not allowed 0-sized buffers to be allocated, I think this should generate an INVALID_VALUE error, and be specified to do so.

Richard-Yunchao: @kenrussell, Firefox can support both 0-sized buffers and 0-sized ArrayBuffer. It can pass this test cases.

sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
kenrussell: See follow-up below.

sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
kenrussell: I see. That wasn't made clear in your CL https://codereview.chromium.org/1339273002/ nor in the description of this new test. Please, make it clear what the motivation is for making changes like this.

Given that, this overall change LGTM, but please address @zhenyao 's comments.

RafaelCintron: IE and Edge disallow null from being passed to bufferData and bufferSubdata.  We did this because section "5.14.5 Buffer Objects" of the WebGL spec, it states that an INVALID_VALUE should be flagged if null is passed to either bufferData or bufferSubData.  The OpenGL spec states that passing null to either of these functions has undefined behavior.  

I would prefer that the behavior continue to be defined as it is in the spec, unless I am missing something.

RafaelCintron: IE and Edge allow 0 to be passed to bufferData and bufferSubData. At render time, we ensure that zero is ends up in the vertex shader and that drawing is skipped if the empty buffer is used as an element array buffer.  Same goes for the 'new ArrayBuffer(0)' case.

RafaelCintron:
@Richard-Yunchao , I encourage you to add a test that passes 'undefined' as the second parameter to bufferData and bufferSubData.  While the spec does not specifically say what happens in this case, I believe the result should be the same as if you pass null.

sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
kenrussell: The behavior of passing null to bufferData is unchanged. "new ArrayBuffer(0)" returns a non-null instance. Passing null is supposed to call the bufferData overload taking "GLenum target, BufferDataSource? data, GLenum usage". Passing 0 is supposed to call the overload taking "GLenum target, GLsizeiptr size, GLenum usage".

RafaelCintron: @kenrussell The see that on line 103 of the test, it verfies that passing null to bufferData flags INVALID_VALUE. Good so far. However, that is not consistent with the bufferSubData test (line 144) where passing null is verified to flag no error.

sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
```diff
@@ -0,0 +1,163 @@
...
+    wtu.glErrorShouldBe(gl, gl.INVALID_VALUE, "buffer overflow");
+
+    gl.bufferSubData(gl.ARRAY_BUFFER, -10, new Float32Array(8));
+    wtu.glErrorShouldBe(gl, gl.INVALID_VALUE);
+
+    gl.bufferSubData(gl.ARRAY_BUFFER, 10, array);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+    gl.bufferSubData(gl.ARRAY_BUFFER, 10, null);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
```
kenrussell: This should generate INVALID_VALUE per https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.14.5 . Please update the expectations, and any buggy WebGL implementations will be fixed.

sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
kenrussell: Thanks @RafaelCintron for pointing this out. It looks like a deviation from the spec per the bufferSubData definition at https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.14.5 .

Richard-Yunchao:
Thanks for you all, @RafaelCintron , @kenrussell , @zhenyao , code has been updated. As @RafaelCintron suggested, I add test cases for bufferData and bufferSubData when the ArrayBufferSource is undefined. I also think these case should generate GL_INVALID_VALUE.

PTAL.

RafaelCintron:
The newest change LGTM from a correctness perspective. A couple of minor things I would like to see added to the test.
* At least for the tests you added, please make more liberal use of the 'opt_msg' parameter of glErrorShouldBe. For debugging, having this makes it much easier to track down the source of the failure.
* Add a test to ensure that calling `bufferData(gl.ARRAY_BUFFER, [1,2,3], gl.STATIC_DRAW)` and `bufferSubData(gl.ARRAY_BUFFER, 0. [1,2,3])` also fail with INVALID_VALUE.  Sequences of numbers are allowed for things like uniforms but not for the buffer functions.

RafaelCintron:
I previously had a comment where I suggested that we have undefined and [1,2,3] flag INVALID_VALUE but, thinking things through some more, I think we should have these two continue to be accepted without error and create a buffer with 0 size. This is the current behavior of other Webgl functions that accept numbers. What do @kenrussell and @jdashg think?

RafaelCintron:
@Richard-Yunchao , another improvement you can make to the test if you wish is to call getBufferParameter(gl.ARRAY_BUFFER, gl.BUFFER_SIZE) and ensure that the returned size is what you expect.

kenrussell:
We should not add any special behavior for cases like "undefined" and "[1,2,3]"; the coercion of values to particular types is well defined in the Web IDL editor's draft at https://heycam.github.io/webidl/ and that's the behavior that should be enforced. Firefox's been doing a great job conforming to Web IDL and Chrome's made dramatic recent improvements in conformance.

Richard-Yunchao:
Thanks for your suggestions, @kenrussell  and @RafaelCintron , I added information for opt_msg parameter in glErrorShouldBe(). The "undefined" test cases are removed as kbr suggested. For getBufferParameter, I suppose that gl-object-get-call.js can will cover it. 

PTAL. Thanks a lot!

kenrussell:
@Richard-Yunchao I didn't mean to remove the tests where "undefined" and "[1, 2, 3]" were removed. I only meant that the behavior should be defined by Web IDL, and that no special cases should be added to the definitions of bufferData and bufferSubData.

I've compared the old and new tests by hand to verify that no test coverage was lost -- a little difficult since the file was moved as well as the tests changed. @Richard-Yunchao I would encourage you to please submit two different pull requests in the future if you are moving files, one which moves the file without changing its contents, and the second which updates the tests in the file.

I'd like to merge this now while the state of the test is well known, and add more tests in subsequent pull requests. Thanks for the contribution.

Richard-Yunchao:
@kenrussell , oh, I misunderstood you... And thanks for your suggestion of pull requests that move files.

PR #1217, opened 2015-09-15
Small fix in gl-object-get-calls.js

zhenyao:
(I think this is the last fix for this test for now!)


kenrussell:
Good catch. Looks good.

PR #1218, opened 2015-09-16
Update samplerParameter[if] part in WebGL 2 spec

legendlee1314:
Supplement the effect clarification compared to function `texParameter`.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -27,7 +27,7 @@
     <!--end-logo-->
 
     <h1>WebGL 2 Specification</h1>
-    <h2 class="no-toc">Editor's Draft 14 September 2015</h2>
+    <h2 class="no-toc">Editor's Draft 16 September 2015</h2>
     <dl>
         <dt>This version:
             <dd>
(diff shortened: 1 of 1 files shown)


legendlee1314:
PTAL. This clarification is referred from [GLES3.0 spec](https://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.3.pdf).

specs/latest/2.0/index.html
```diff
@@ -2123,6 +2123,9 @@ <h2 class="no-toc">Table of contents</h2>
         </table>
         <p>If <em>pname</em> is not in the table above, generates an <code>INVALID_ENUM</code> error.</p>
         <p>If <em>sampler</em> is not a valid sampler object, generates an <code>INVALID_OPERATION</code> error.</p>
+        <p>Modifying a parameter of a sampler object affects all texture units to which that sampler object is bound.
+        Calling <code>TexParameter</code> has no effect on the sampler object bound to the active texture unit.
+        It will modify the parameters of the texture object bound to that unit.</p>
```
legendlee1314: Personally, it's still a little confused from this sentence. As I see, there're two parts of parameters for a texture unit. One is sampling related like `wrap_mode` and `filter`, another is not. While a sampler is bound to a texture unit, the texture unit will ignore its settings of sampling part and use equivalent parameters of this sampler. If my understanding is correct, then it's better not to use `texParameter` after binding a sampler unless modifying the parameters except sampling parts.
So I suggest to change this sentence to `It will modify the parameters except those in the table above of the texture object bound to that unit.` What do you think about this?

zhenyao: I am not sure if this text adds value to the spec.  WebGL 2 spec is based on top of ES 3.0 spec, and this text is something fundamental about how sampler objects work, and it should be very clear in the ES 3.0 spec.  I see no reason to repeat that in the WebGL 2 spec.

kenrussell: I agree with Mo. It isn't necessary to duplicate the spec text from the ES 3.0 spec. Additionally, that proposed sentence is incorrect. TexParameter modifies the parameters of a texture object, regardless of whether a sampler is bound. If a sampler's currently bound (overriding the parameters of the texture object), the texture's parameters are modified, and the sampler is unbound, then the texture's newly modified parameters take effect.

kenrussell:
Sorry, but I don't think this change is warranted.

legendlee1314:
I understood it. :)

PR #1220, opened 2015-09-16
Test generateMipmap for 3D target with zero size texture at base level

qkmiao:


sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
```diff
@@ -139,6 +139,13 @@
   gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_BASE_LEVEL, 2);
   gl.generateMipmap(gl.TEXTURE_2D);
   wtu.glErrorShouldBe(gl, gl.NO_ERROR, "generateMipmap should succeed");
+
+  // Test 3D texture.
+  var tex3d = gl.createTexture();
+  gl.bindTexture(gl.TEXTURE_3D, tex3d);
+  gl.texImage3D( gl.TEXTURE_3D, 0, gl.RGBA, 0, 0, 0, 0, gl.RGBA, gl.UNSIGNED_BYTE, new Uint8Array(4));
```
zhenyao: I think you mean width, height, depth to be 1, not 0?

kenrussell: Does this new test catch the off-by-one error in the earlier Chromium commit in this area?

kenrussell: I see from the comment on https://codereview.chromium.org/1343053002/ that it does. LGTM

PR #1221, opened 2015-09-17
Fix a bug in js/tests/tex-image-and-sub-image-2d-utils.js

zhenyao:
A minor fix to get integer format texture tests to work.
Otherwise these two shaders won't compile.


kenrussell:
That seems a little strange, but LGTM.

PR #1226, opened 2015-09-18
Added state query tests and bugfixes.

jasomobica:
New tests for: sampler state, texture state, integer and floating point state parameters.

Fixed querying result in occlusion query test.
Added stencil context creation parameter where tests are using stencil buffer.
Fixed Fbo render test to skip over unsupported formats instead of reporting errors.
Fixed TF test to skip unsupported cases.
Fixed framework code for getting test results in transform feedback buffers.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fOcclusionQueryTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fOcclusionQueryTests.js
@@ -63,6 +63,18 @@ var setParentClass = function(child, parent) {
 /** @const */ var OCCLUDER_STENCIL_CLEAR = (1 << 4);
 
 /**
+ * @enum
+ */
+es3fOcclusionQueryTests.State = {
+    DRAW: 0,
+    VERIFY: 1,
+    FINISH: 2
+};
+
+/* Maximum time to wait for query result (in seconds) */
+/** @const */ var MAX_VERIFY_WAIT = 5;
+
+/**
  * @constructor
  * @extends {tcuTestCase.DeqpTest}
  */
@@ -79,6 +91,8 @@ es3fOcclusionQueryTests.OcclusionQueryCase = function(name, description, numOccl
     this.m_program = null;
     this.m_iterNdx = 0;
     this.m_rnd = new deRandom.Random(deString.deStringHash(name));
+    this.m_state = es3fOcclusionQueryTests.State.DRAW;
+    /** @type {WebGLQuery} */ this.m_query;
 };
 
 setParentClass(es3fOcclusionQueryTests.OcclusionQueryCase, tcuTestCase.DeqpTest);
@@ -140,7 +154,7 @@ es3fOcclusionQueryTests.OcclusionQueryCase.prototype.init = function() {
     gl.vertexAttribPointer(0, ELEMENTS_PER_VERTEX, gl.FLOAT, false, 0, 0);
 };
 
-es3fOcclusionQueryTests.OcclusionQueryCase.prototype.iterate = function() {
+es3fOcclusionQueryTests.OcclusionQueryCase.prototype.draw = function() {
     var colorUnif = gl.getUniformLocation(this.m_program.getProgram(), 'u_color');
 
     var targetW = gl.drawingBufferWidth;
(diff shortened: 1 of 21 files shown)


sdk/tests/deqp/functional/gles3/es3fTransformFeedbackTests.js
```diff
@@ -895,9 +895,13 @@ goog.scope(function() {
 
             if (linkFail) {
                 if (!es3fTransformFeedbackTests.isProgramSupported(this.m_progSpec, this.m_bufferMode)) {
-                    throw new Error('Not Supported. Implementation limits exceeded.');
+                    var msg = 'Not Supported. Implementation limits exceeded.';
+                    checkMessage(false, msg);
+                    throw new TestFailedException(msg);
```
kenrussell: Does this cause the test runner to report a test failure? It shouldn't -- the C++ version of the test harness considers this case as being "not supported" rather than a failure. In the WebGL version I'm not sure it's necessary to distinguish between "passing" and "not supported", but "not supported" should not be equivalent to a failure.

sdk/tests/deqp/functional/gles3/es3fOcclusionQueryTests.js
```diff
@@ -283,20 +297,34 @@ es3fOcclusionQueryTests.OcclusionQueryCase.prototype.iterate = function() {
...
     gl.disable(gl.STENCIL_TEST);
     gl.disable(gl.DEPTH_TEST);
+    this.m_state = es3fOcclusionQueryTests.State.VERIFY;
+};
 
+es3fOcclusionQueryTests.OcclusionQueryCase.prototype.verify = function() {
     // Check that query result is available.
-    var resultAvailable = /** @type {boolean} */ (gl.getQueryParameter(query, gl.QUERY_RESULT_AVAILABLE));
-    assertMsgOptions(resultAvailable, 'Occlusion query failed to return a result after glFinish()', false, true);
+    var resultAvailable = /** @type {boolean} */ (gl.getQueryParameter(this.m_query, gl.QUERY_RESULT_AVAILABLE));
```
kenrussell: The fact that verification is now asynchronous is very nice. We should do the same restructuring of the transform feedback tests, which rely on the correctness of the TRANSFORM_FEEDBACK_PRIMITIVES_WRITTEN query.

sdk/tests/deqp/modules/shared/glsShaderExecUtil.js
```diff
@@ -472,21 +472,25 @@ goog.scope(function() {
...
+                        dstPtr[scalarSize * 4 * ndx + j] = readPtr[ndx * outputBufferStride + j];
+                    }
+                outputs[outputNdx] = dstPtr;
             }
-        }
-        outputs[outputNdx] = dstPtr;
-
-              curOffset += scalarSize * 4;
+            curOffset += scalarSize * 4;
           }
```
kenrussell: Indentation looks off here.

kenrussell:
Thank you, these look excellent. Merging now. Let's handle any follow-ups in subsequent pull requests.

sdk/tests/deqp/functional/gles3/es3fTransformFeedbackTests.js
jasomobica: No, it doesn't report a test failure. I admit that TestFailedException name can be confusing, though.
When a test fails, we usually report test failure to the harness and then throw TestFailedException. This causes the test runner to move on to the next test.
In this case, we're printing a warning message and terminating the test.
All exceptions other than TestFailedException are caught by the test runner and result in test failure.

sdk/tests/deqp/functional/gles3/es3fOcclusionQueryTests.js
jasomobica: Are there any functions other than getQueryParameter() with occlusion and transform feedback queries that are affected?

kenrussell: In the core ES 3.0 spec those are the two query types. The EXT_disjoint_timer_query extension defines another one which will be affected by this change too (they're not used by the dEQP tests, though).

PR #1227, opened 2015-09-21
Add a test that indexes a vector using a uint

Oletus:
This makes sure that indexing workarounds work also for uints.


kenrussell:
Looks good. Thanks for the targeted test case.

PR #1230, opened 2015-09-22
Remove webgl_preprocessor_reserved tests that contradict dEQP

Oletus:
Native dEQP tests require that preprocessor macro names containing
double underscores are allowed, based on Khronos discussions which have
not yet been reflected in all the published specs. The updated language
concerning double underscores can be found from the GLSL 3.10 spec.


Oletus:
Thoughts on this? It's tricky to deal with this grey area where the dEQP tests and the published version of the spec disagree, but something needs to be done.

kenrussell:
What's the expected behavior for ES 2.0 and ES 3.0? It sounds like the behavior was first updated in ES 3.1. Which dEQP tests for which ES versions require support for macros containing double underscores?

Since it's likely the WebGL implementation will have to preprocess the incoming shaders anyway, isn't it likely that these spec differences can be handled there?

What would happen if (hypothetically speaking) these tests are removed, we start passing shaders down to the driver containing double underscores, and the driver rejects them?

I think it'd be possible to still enforce this rule for WebGL 1.0 contexts (as has been done for a while) even if the behavior's relaxed for WebGL 2.0 contexts (for example).

@RafaelCintron @grorg @jdashg @zhenyao @toji thoughts?

Oletus:
These are the dEQP tests that conflict with the published spec:

dEQP-GLES2.functional.shaders.preprocessor.basic.identifier_with_double_underscore*
dEQP-GLES3.functional.shaders.preprocessor.basic.identifier_with_double_underscore*

They're not in the WebGL conformance port of dEQP right now. These tests were changed from negative to positive in November 2014. This was done based on the apparent intention of Khronos to allow double underscores in all versions of the spec. I reopened a Khronos bug about this to remind the group about updating the older specs, but it hasn't resulted in changes yet.

It's true that the WebGL implementation will preprocess the incoming shaders anyway, so any double underscores in macro names shouldn't reach the driver level. There aren't any really tricky corner cases as far as I can see - redefining a built-in macro like __LINE__ will result in an error at the WebGL implementation level, but using a macro like LINE__NUMBER shouldn't cause any problems. We could still mandate a warning in this case.

If double underscores in macro names were passed down to the driver and the driver rejected them, the shader wouldn't compile and run. But by porting the dEQP tests in question we could test to a reasonable degree that this doesn't happen. What we should be worried about is the chance that the driver thinks some macro names have special meaning that changes shader behavior, but I think that making sure that these macro names aren't passed to the driver can be left to the preprocessor in the WebGL implementation.

kenrussell:
Thanks for the thought on this.

Is bringing in this dEQP test as simple as pulling the associated test from https://android.googlesource.com/platform/external/deqp/+/master/data/gles3/shaders/preprocessor.test (and the ES2 version) into the preprocessor.test files already in the WebGL repo?

jasomobica:
Yes, that's all you need.

If you wanted to copy a new .test file, then you would also need to create a corresponding html page.

Oletus:
As a general reminder, you should be careful and check the git history when pulling test files from a different version of dEQP, since some of them have had WebGL specific changes. However, right now this only applies to functions.test and scoping.test, so updating preprocessor.test should be fine.

Is it decided that we agree with and will apply the dEQP interpretation of the spec?

kenrussell:
I'd like to reach consensus with the other implementers ( @grorg @jdashg @RafaelCintron ), but my opinion is that WebGL should follow the direction defined by the ES specs and in this case driven by the dEQP tests. Therefore I think that if we reverse the behavior in this area we should remove the old tests but also add the corresponding preprocessor test to the ported dEQP tests. Do you think you could add that to your pull request? Or would you rather that be kept separate from this one?

RafaelCintron:
Macro names is not the only place where double underscores are prohibited.  They are also prohibited in function names, type names and variable names.  Are we meant to relax the rules for these places as well?  If not, what is the rationale for allowing double underscores in macro names but not the other places?

grorg:
After discussion on our most recent teleconference, we agreed to merge this change.

jdashg:
Following ES is LGTM.

Oletus:
@RafaelCintron Double underscores will still be prohibited in other places. It was a pragmatic decision on the part of the native OpenGL group to allow them specifically in macro names, since the bug had slipped into several drivers and content was already using macros with double underscores.

RafaelCintron:
@Oletus , thank you for the clarification. Looks like the ship has sailed on allowing double underscores in macro names.

Issue #1228, opened 2015-09-21
Clarify interaction between ANGLE_instanced_arrays and OES_vertex_array_object


toji:
Sean Lilley pointed out on the WebGL dev mailing list that these two extensions behave inconsistently between both the Chrome and Firefox ANGLE and OpenGL backends. With native OpenGL the vertex divisor is tracked as part of the VAO, but under ANGLE it is not. In Chrome this is because the ANGLE backend uses emulated VAOs which don't account for the instancing extension.

Unless there is concern about cross-platform compatibility I suggest that we add text to the ANGLE_instanced_arrays spec to explicitly state that if OES_vertex_array_object is also enabled the vertex divisor also be tracked as part of the VAO, along with a test to verify that behavior.

jdashg:
I agree that adding clarification would be nice.

PR #1231, opened 2015-09-23
Extend angle-instanced-arrays test to check VAO interaction

toji:
Problems with interaction between these two extensions was noted in
Issue #1228. Test currently fails on Chrome when using ANGLE, but
succeeds with --use-gl=desktop. Firefox also handles the new test
correctly.


kenrussell:
Very nice test. Thanks for reducing this.

PR #1233, opened 2015-09-24
Remove dEQP #extension directive test to keep backwards compatibility

Oletus:
When ANGLE implemented a fix to this test in May 2015, significant issues
with backwards compatibility were detected. It was decided that the ESSL1
compiler should issue a warning rather than an error in this case, though
the WebGL spec was not updated.

The spec is still strictly enforced in the case of ESSL3 shaders and
WebGL 2 with a similar gles3 dEQP test.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -27,7 +27,7 @@
     <!--end-logo-->
 
     <h1>WebGL 2 Specification</h1>
-    <h2 class="no-toc">Editor's Draft 02 October 2015</h2>
+    <h2 class="no-toc">Editor's Draft 05 October 2015</h2>
     <dl>
         <dt>This version:
             <dd>
@@ -2994,6 +2994,23 @@ match the <em>type</em> according to the <a href="#TEXTURE_PIXELS_TYPE_TABLE">ab
         specification of <a href="#GET_QUERY_PARAMETER">getQueryParameter</a> for discussion and
         rationale.
     </p>
+
+    <h3>GLSL ES 3.00 #extension directive location</h3>
+
+    <p>
+        The WebGL 1.0 specification section <a href="../1.0/index.html#EXTENSION_DIRECTIVE_LOCATION">
+        GLSL ES #extension directive location</a> only applies to OpenGL ES Shading Language 1.00
+        shaders. It does not apply to shaders that are written in OpenGL ES Shading Language 3.00.
+        In shaders written in OpenGL ES Shading Language 3.00, <code>#extension</code> directives
+        must occur before non-preprocessor tokens regardless of what is written in the extension
+        specification.
+    </p>
+
+    <div class="note rationale">
+        This is done to make WebGL 2.0 to follow the written GLSL ES 3.00 specification more
+        closely. Enforcing the restriction more strictly than native GLES drivers makes the behavior
+        well-defined.
+    </div>
 
 <!-- ======================================================================================================= -->
 
--- a/sdk/tests/deqp/data/gles2/shaders/preprocessor.test
+++ b/sdk/tests/deqp/data/gles2/shaders/preprocessor.test
@@ -2856,21 +2856,6 @@ group extensions "Extension Tests"
 			}
 		""
 	end
-
-	case after_non_preprocessing_tokens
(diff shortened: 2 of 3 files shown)


Oletus:
Also see discussion at https://code.google.com/p/angleproject/issues/detail?id=1165

Oletus:
Also, should we try to actually specify and test the backwards-compatible behavior?

kenrussell:
Thanks for putting this together. I do think it would be worth writing this up in the differences section of the WebGL 1.0 spec along with a rationale, and further, removing that difference in the WebGL 2.0 spec. It's definitely desirable to remove this legacy behavior going forward. Would you be interested in doing that, and/or should I merge this pull request now?

kenrussell:
To answer your remaining question: I'm not sure about testing the backward compatible behavior -- not sure what kinds of guarantees we would want to make, because it turns out that it's just permissive behavior from the GLSL compilers that allowed these shaders to work. Please consult with Geoff who I think did the deepest analysis of the affected shaders.

Oletus:
I'll need to consult with Geoff before I can spec or test this. It will take a while, but I can take care of it.

kenrussell:
Thanks. Again, should I merge this pull request now? Or would you prefer that it be merged along with the associated spec change and perhaps additional tests?

Oletus:
I'll edit this pull request to include at least the spec.

Oletus:
Added spec. I think this will conflict with https://github.com/KhronosGroup/WebGL/pull/1236 and I'd prefer that pull request to be merged before this one.

kenrussell:
Merged #1236 . Could you rebase this one? Thanks.

kenrussell:
Also: could you please update the last modified date, and update the WebGL 2.0 spec to specifically remove this deviation from the OpenGL ES spec?

Oletus:
Rebased, updated the last modified date, and updated WebGL 2.0 spec for GLSL ES 3.00 shaders.

kenrussell:
Thank you, this looks perfect. Any further comments @grorg @jdashg @RafaelCintron @toji @zhenyao ?

zhenyao:
LGTM

RafaelCintron:
LGTM

kenrussell:
Olli, sorry, I introduced a merge conflict with your pull request. Could you resolve it? Also, would you consider squashing commits to keep the commit history clean? See https://www.khronos.org/webgl/wiki/Using_Github_To_Contribute#Handling_requested_changes about using the --force option to push the branch again afterward. Thanks.

Oletus:
Rebased and squashed the commits.

specs/latest/2.0/index.html
```diff
@@ -2995,6 +2995,23 @@ <h2 class="no-toc">Table of contents</h2>
...
 
+    <h3>GLSL ES 3.00 #extension directive location</h3>
+
+    <p>
+        The WebGL 1.0 specification section <a href="../1.0/index.html#EXTENSION_DIRECTIVE_LOCATION">
+        GLSL ES #extension directive location</a> only applies to OpenGL ES Shading Language 1.00
+        shaders. It does not apply to shaders that are written in OpenGL ES Shading Language 3.00.
+        In shaders written in OpenGL ES Shading Language 3.00, <code>#extension</code> directives
+        must occur before non-preprocessor tokens regardless of what is written in the extension
+        specification.
```
kenrussell: Hopefully this won't cause problems with ESSL extensions.

kenrussell:
Thanks Olli for putting this together. LGTM. Merging.

grorg:
LGTM also

jdashg:
LGTM

Oletus:
Thanks for merging this. I think there was some concern about implementing this looser version of the spec on platforms where the restriction is enforced by the native GLES, so I'd like to add a note that ANGLE can parse shaders according to the looser spec and still produce shaders that meet the stricter spec with #extension directives at the top.

kenrussell:
I think a note like that would be useful. Would you mind submitting another pull request adding one?

Oletus:
I meant to just mention it in this comment thread, but I suppose it could be in the spec as well. I'll put up a pull request.

PR #1234, opened 2015-09-25
Added stress test of resource deallocation upon page reloading / navi…

kenrussell:
…gation.

Intended to ensure that Chrome's mechanism for freeing resources doesn't regress when changing how WebGL objects are tracked.


kenrussell:
Merging this myself as this test is needed on multiple machines. It's not part of the conformance suite.

toji:
LGTM after the fact.

PR #1235, opened 2015-09-25
fix the bugs in buffercopy.js

Richard-Yunchao:
For copyBufferSubData, ELEMENT_ARRAY_BUFFER are compatible with
COPY_READ_BUFFER and COPY_WRITE_BUFFER only if the COPY_READ/WRITE
buffer had been initially bound to an ELEMENT_ARRAY_BUFFER

This behavior is covered in buffer-copying-restrictions.html,
so skip these tests in buffercopy.js

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fBufferCopyTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fBufferCopyTests.js
@@ -244,16 +244,8 @@ goog.scope(function() {
                 // In WebGL 2, a copy between an ELEMENT_ARRAY_BUFFER and other data buffer
                 // (not COPY_WRITE_BUFFER nor COPY_READ_BUFFER nor ELEMENT_ARRAY_BUFFER)
                 // cannot be made, so let's skip those cases.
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fBufferCopyTests.js
```diff
@@ -241,19 +241,8 @@ goog.scope(function() {
                 if (srcTargetNdx == dstTargetNdx)
                     continue;
 
-                // In WebGL 2, a copy between an ELEMENT_ARRAY_BUFFER and other data buffer
-                // (not COPY_WRITE_BUFFER nor COPY_READ_BUFFER nor ELEMENT_ARRAY_BUFFER)
-                // cannot be made, so let's skip those cases.
```
kenrussell: This comment is still needed.

kenrussell:
Looks good except for the removal of the comment. Could you put it back?

Richard-Yunchao:
@kenrussell , done. PTAL.

PR #1236, opened 2015-09-25
Specify handling "defined" operators generated by macro replacement

Oletus:
This is already tested by the dEQP shader preprocessor tests in the
conformance suite.

Five different corner cases of this behavior were tested against Nexus 5,
Nexus 6, Nexus 9 and Nexus Player native GLES3 implementations, as well
as the recent implementation in ANGLE TOT. The existing implementations
largely agree with the spec in this commit. The only exception is Nexus
Player, where in the following kind of case:

 #define CCC 1
 #define defined BBB
 #define AAA defined
 #if AAA CCC
 #endif

"defined" in the replacement list of "AAA" is replaced by "BBB" instead
of immediately handling it as the "defined" operator. In all the other
four implementations, the defined operator is evaluated in this case and
returns "1".


kenrussell:
Thanks, this seems fine. Could you please update the last modified date of the spec as well?

Oletus:
I amended the commit to include updating the last modified date.

kenrussell:
Thanks for thinking this through so carefully. Merging.

PR #1238, opened 2015-09-27
Add bufferData and bufferSubData test cases

RafaelCintron:
bufferData has an integer overload and an ArrayBuffer overload.
Per the WebIDL spec, the overload resolution algorithm should resolve types as follows:
- If the argument is null or undefined, pick the nullable type, which is ArrayBuffer.
Per the WebGL spec, null should flag INVALID_VALUE.
- If the argument is an ArrayBuffer, then pick the ArrayBuffer overload
- Everything else should pick the numeric overload. This means things like objects, strings, floating point numbers, arrays of numbers and strings, etc should convert themselves to an integer.
    
Unlike bufferData, bufferSubData is typed to accept an ArrayBuffer. No overloads.
Per the WebIDL spec, passing undefined should convert to null.  Per the WebGL spec, null should
flag an INVALID_VALUE error.

### DIFF
--- a/sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
+++ b/sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
@@ -45,16 +45,13 @@ debug('Regression test for <a href="https://bugs.webkit.org/show_bug.cgi?id=4188
 
 var wtu = WebGLTestUtils;
 var gl = wtu.create3DContext();
-var buf = null;
-var array = null;
-var arrayZero = null;
 if (!gl) {
     testFailed("WebGL context does not exist");
 } else {
     testPassed("WebGL context exists");
 
     bufferDataTest();
-    bufferDataTestWithArrayBuffer();
+    bufferDataSizesTest();    
 
     bufferSubDataTest();
 }
@@ -63,8 +60,8 @@ function bufferDataTest() {
     debug("");
     debug("Test bufferData without ArrayBuffer input");
 
-    buf = gl.createBuffer();
-    shouldBeNonNull("buf");
+    var buf = gl.createBuffer();
+    shouldBeNonNull(buf);
 
     gl.bufferData(gl.ARRAY_BUFFER, 4, gl.STATIC_DRAW);
     wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "no buffer bound");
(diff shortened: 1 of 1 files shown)


kenrussell:
Thank you for this contribution. Have you triple-checked these new cases against the Web IDL spec?

What is the behavior of various WebGL implementations on this test? I'm particularly interested to know how Firefox performs, as it's known to have one of the best-conforming Web IDL implementations, and uncovered many bugs in the WebGL spec's IDL itself.

RafaelCintron:
I've run my code past @travisleithead and it turns out there are a couple of errors in the test with its handling of undefined. In particular, passing undefined to bufferData should pick the ArrayBuffer overload and flag INVALID_VALUE, per the WebGL spec. Similarly, passing undefined to bufferSubData should do the same thing.  I'm going to push a improved revision.  Stay tuned.

RafaelCintron:
I've updated the pull request.  Please re-review.
@kenrussell , you asked about other implemenations
- Firefox passes all of the tests except for the ones that pass null and undefined to bufferSubData. I believe this is in error because the WebGL spec says null should flag INVALID_VALUE.
- Edge passes all tests except for bufferSubData being passed undefined where it throws an exception.
- Chrome has the same failures as FireFox.  In addition, it (incorrectly) flags an error when bufferData is called with an array or object. It also flags INVALID_VALUE when passed 0 to bufferData

sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
```diff
@@ -76,59 +73,80 @@
     wtu.glErrorShouldBe(gl, gl.INVALID_VALUE,
         "calling bufferData when buffer size is negative should generate INVALID_VALUE");
 
-    gl.bufferData(gl.ARRAY_BUFFER, 0, gl.STATIC_DRAW);
```
kenrussell: The removal of this block of code and bufferDataTestWithArrayBuffer() looks like a bad merge. Could you please double-check?

RafaelCintron: @kenrussell , it is not a bad merge.  The tests that used to be in bufferDataTestWithArrayBuffer are now part of bufferDataSizesTest. In bufferDataSizesTest, we test that, among other things, ArrayBuffers and 0 can be passed to bufferData as well as verify that the created buffer is the size we expect.

kenrussell: I see. Sorry, I'd missed the test cases with new ArrayBuffer(0) and new ArrayBuffer(4).

kenrussell:
Thanks for your careful analysis of these cases. Let's merge this now, though it will cause some difficulty until all of the browsers are updated to provide consistent behavior.

sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
RafaelCintron: @kenrussell No worries.  Thank you for the review.

PR #1239, opened 2015-09-28
Add getQueryParameter to gl-object-get-call.js

renchenglei:
Please help review the code, thanks!

### DIFF
--- a/sdk/tests/js/tests/gl-object-get-calls.js
+++ b/sdk/tests/js/tests/gl-object-get-calls.js
@@ -757,6 +757,51 @@ if (contextVersion > 1) {
         function(target) {
             return gl.getIndexedParameter(target, 0);
     });
+
+    debug("");
+    debug("Test getQueryParameter");
+    var buffer = gl.createBuffer();
+    gl.bindBuffer(gl.TRANSFORM_FEEDBACK_BUFFER, buffer);
+    gl.bufferData(gl.TRANSFORM_FEEDBACK_BUFFER, 64, gl.DYNAMIC_DRAW);
+    var query = gl.createQuery();
+    gl.bindBufferBase(gl.TRANSFORM_FEEDBACK_BUFFER, 0, buffer);
+    var requestAnimationFrame = window.requestAnimationFrame ||
+	                        window.mozRequestAnimationFrame ||
+			        window.webkitRequestAnimationFrame ||
+			        window.msRequestAnimationFrame;
+    var cancelRAF = window.cancelAnimationFrame ||
+	            window.mozCancelAnimationFrame ||
+		    window.webkitCancelAnimationFrame ||
+		    window.msCancelAnimationFrame;
+    var requestId;
+    function stopAnimation() {
+        cancelRAF(requestId);
+    }
+    function animate() {
+        gl.beginQuery(gl.TRANSFORM_FEEDBACK_PRIMITIVES_WRITTEN, query);
+        gl.beginTransformFeedback(gl.TRIANGLES);
+        gl.drawArrays(gl.POINTS, 0, 3);
+        gl.endTransformFeedback();
+        gl.endQuery(gl.TRANSFORM_FEEDBACK_PRIMITIVES_WRITTEN);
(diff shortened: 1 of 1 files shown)


sdk/tests/js/tests/gl-object-get-calls.js
```diff
@@ -757,6 +757,32 @@ if (contextVersion > 1) {
...
+    gl.bindBuffer(gl.TRANSFORM_FEEDBACK_BUFFER, buffer);
+    gl.bufferData(gl.TRANSFORM_FEEDBACK_BUFFER, 64, gl.DYNAMIC_DRAW);
+    var query = gl.createQuery();
+    gl.bindBufferBase(gl.TRANSFORM_FEEDBACK_BUFFER, 0, buffer);
+    gl.beginQuery(gl.TRANSFORM_FEEDBACK_PRIMITIVES_WRITTEN, query);
+    gl.beginTransformFeedback(gl.TRIANGLES);
+    gl.drawArrays(gl.POINTS, 0, 3);
+    gl.endTransformFeedback();
+    gl.endQuery(gl.TRANSFORM_FEEDBACK_PRIMITIVES_WRITTEN);
+    shouldBe('gl.getQueryParameter(query, gl.QUERY_RESULT_AVAILABLE)', 'true');
```
kenrussell: This can't be relied upon. I've been lax in putting up this specification change, but https://github.com/KhronosGroup/WebGL/pull/1240 requires that this will not actually be true in the same frame that the query was issued. Could you please adjust this test to not rely on the query's result becoming available the same frame as it was issued?

kenrussell:
Thanks for this contribution. The negative test looks OK, but the positive test needs to be updated. Could you please do so? Thanks.

sdk/tests/js/tests/gl-object-get-calls.js
renchenglei: I have checked the latest spec. My patch has returned to the user agent's main loop. Is it right now? Please help review the code, thanks a lot. 
@kenrussell

kenrussell: It has not returned to the main loop properly. It must return to the main loop (potentially multiple times, using e.g. requestAnimationFrame) after gl.endQuery() before requiring that gl.getQueryParameter(query, gl.QUERY_RESULT_AVAILABLE) return true. The WebGL 2.0 spec now requires this in https://www.khronos.org/registry/webgl/specs/latest/2.0/#GET_QUERY_PARAMETER .

renchenglei: Thanks a lot! I will modify the code.

renchenglei: I modify the code, can you help review the code? Thanks a lot!
https://github.com/KhronosGroup/WebGL/pull/1239/files
@kenrussell

sdk/tests/js/tests/gl-object-get-calls.js
```diff
@@ -757,6 +757,51 @@ if (contextVersion > 1) {
...
+        gl.drawArrays(gl.POINTS, 0, 3);
+        gl.endTransformFeedback();
+        gl.endQuery(gl.TRANSFORM_FEEDBACK_PRIMITIVES_WRITTEN);
+        if(gl.getQueryParameter(query, gl.QUERY_RESULT_AVAILABLE) == true)
+           stopAnimation();
+        else
+           requestId = requestAnimationFrame(animate);
+    }
+    animate();
+    shouldBe('gl.getQueryParameter(query, gl.QUERY_RESULT_AVAILABLE)', 'true');
```
kenrussell: This code is still incorrect. The shouldBe call will execute immediately after the call to animate(), even if a requestAnimationFrame callback is registered.

kenrussell:
@renchenglei , https://github.com/KhronosGroup/WebGL/pull/1261 shows what I'd be comfortable merging. Please review that pull request.

kenrussell:
Since #1261 has been merged I'm closing this pull request. Thanks for the contribution, but we need to turn around these pull requests more quickly.

renchenglei:
@kenrussell  I have seen your patch, it's great! Thanks a lot for your help.

PR #1240, opened 2015-09-29
A query's result must not become available the same frame it was issued.

kenrussell:
This specification language is derived from that in the EXT_disjoint_timer_query extension. It is intended to ensure best portability and performance of code using query objects. The ported dEQP tests in the WebGL 2.0 conformance suite have already been adjusted to follow this requirement. The EXT_disjoint_timer_query conformance test will be updated shortly to follow the same requirement.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -27,7 +27,7 @@
     <!--end-logo-->
 
     <h1>WebGL 2 Specification</h1>
-    <h2 class="no-toc">Editor's Draft 14 September 2015</h2>
+    <h2 class="no-toc">Editor's Draft 02 October 2015</h2>
     <dl>
         <dt>This version:
             <dd>
@@ -2007,7 +2007,7 @@ match the <em>type</em> according to the <a href="#TEXTURE_PIXELS_TYPE_TABLE">ab
           function.</p>
       </dd>
       <dt>
-        <p class="idl-code">any getQueryParameter(WebGLQuery? query, GLenum pname)
+        <p class="idl-code"><a name="GET_QUERY_PARAMETER">any getQueryParameter</a>(WebGLQuery? query, GLenum pname)
           <span class="gl-spec">
             (<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.3.pdf#nameddest=section-6.1.7">OpenGL ES 3.0.3 &sect;6.1.7</a>,
             <a href="http://www.khronos.org/opengles/sdk/docs/man3/html/glGetQueryObjectuiv.xhtml" class="nonnormative">man page</a>)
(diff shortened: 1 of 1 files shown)


kenrussell:
@grorg @jdashg @RafaelCintron @Oletus please review this specification change to make WebGL 2.0's built-in queries behave the same as those in the EXT_disjoint_timer_query extension. The ported dEQP tests have already been updated to follow this rule, and the EXT_disjoint_timer_query conformance test will be updated shortly to do so as well.

@toji @zhenyao please review as well.

toji:
LGTM, though I definitely want the spec editors to have a look before merging. Thanks!

specs/latest/2.0/index.html
```diff
@@ -2957,6 +2984,17 @@ <h2 class="no-toc">Table of contents</h2>
         floating point in the range [0, 1].
     </p>
 
+    <h3>Queries' results may not be made available in the current frame</h3>
```
Oletus: Should this be "must not be made available"?

Oletus:
It makes sense that this would work the same way in WebGL 2 as with the WebGL 1 extension, so in general LGTM.

RafaelCintron:
@kenrussell , thank you for adding the clarifications and for the note about finish.  The pull request LGTM from a spec perspective.

specs/latest/2.0/index.html
kenrussell: Yes, thanks. Corrected in the current diff.

kenrussell:
Merging now to make forward progress on this.

grorg:
LGTM too

specs/latest/2.0/index.html
```diff
@@ -2034,6 +2034,33 @@ <h2 class="no-toc">Table of contents</h2>
 
           <p>Returns null if any OpenGL errors are generated during the execution of this
           function.</p>
+
+          <p>In order to ensure consistent behavior across platforms, queries' results must only be made
+          available when the user agent's <a
+          href="http://www.whatwg.org/specs/web-apps/current-work/multipage/webappapis.html#event-loops">event
+          loop</a> is not executing a task. In other words:
+          <ul>
+            <li> A query's result must not be made available until the current frame has completed
```
jdashg: This is not a rephrasing of the above, but rather a more strict restriction. Requiring returning to the event loop is different than requiring that the current frame is complete.

I hesitantly support the former, but do not support the latter.

jdashg:
I think I have to disagree with this change. While I support pushing developers toward safer coding practices, we shouldn't forbid access to the 'sharp tools'. In particular, this seems like it could add a large barrier for porting codebases to the web.

Requiring the current frame to composite before giving access could be particularly onerous for apps which don't care about framerate (e.g. event-driven drawing apps, or GPGPU usecases), or 'ASAP'/'vsync=off' cases where more than one frame may happen per rAF. (WebGL in workers?)

I appreciate where the desire is coming from, but I believe these changes are too restrictive.

kenrussell:
@jdashg it's a fair point that it mustn't be a requirement for the frame to composite. That imposes too many restrictions, especially for canvases in workers, and offscreen canvases. I've posted #1256 to revise both this language and that in EXT_disjoint_timer_query. Please review there. Thanks.

PR #1244, opened 2015-10-01
Add tolerance to pow-with-constant-exponent test

Oletus:
Result of (x^2)^0.5 may differ from x enough to affect the result of the
test if the mediump pow function is run at the precision suggested in the
GLSL ES 3.00 spec, where it inherits its precision from exp2(y*log2(x)).

The lack of tolerance caused the test to fail at least on Nexus 5.


zhenyao:
LGTM

PR #1245, opened 2015-10-01
Expand Transform Feedback test

toji:


toji:
Hm... something went severely wrong with a rebase here. Closing so I can push a cleaner one.

PR #1249, opened 2015-10-02
Get ready for tex 3d with web elements testing.

zhenyao:
1) rename tex-image-and-sub-image-2d-utils.js to
tex-image-and-sub-image-utils.js
(this causes a lot of texture tests to be updated.
2) copy 2d*.js to 3d*.js (but no modifying the file content)


kenrussell:
LGTM. Thanks for pushing this forward.

PR #1252, opened 2015-10-03
Verify that queries' results don't become available in a spin loop.

kenrussell:
This is a follow-up to #1240 enforcing the behavior for
EXT_disjoint_timer_query. Has been tested with Chromium's
work-in-progress CL. Will revise more of the WebGL 2.0 tests that use
query objects in a follow-on pull request.


kenrussell:
@zhenyao @RafaelCintron @grorg @jdashg @DavidYen I've tested this change with Chromium before and after implementing the desired behavior, verifying both that current Chromium fails, and that the test passes after fixing the browser. Merging this myself so others can easily run the test. Please comment and I'll make revisions in a follow-on pull request.

PR #1253, opened 2015-10-03
Fixed bug in expando-loss.html test.

kenrussell:
It was attaching renderbuffers that hadn't yet been fully
initialized (never bound to the RENDERBUFFER binding point) to a
framebuffer object, causing an INVALID_OPERATION error.


kenrussell:
@RafaelCintron in particular please review. Thanks.

Tested this in Chromium with a bug fix that makes this test finally pass; it clears up the INVALID_OPERATION errors that were being generated.

zhenyao:
Looks good to me, but I'll wait Rafael to take a look.

RafaelCintron:
LGTM.  @kenrussell Apologies for the bug and thank you for the fix.

PR #1254, opened 2015-10-03
Fixed bugs in extensions' unique object tests.

kenrussell:
A few of these tests were leaving around a global reference to the
extension object, so the attempted garbage collection of the JavaScript
wrapper was having no effect. Nulling out the global reference exposes
more bugs in WebGL implementations.


kenrussell:
@RafaelCintron @grorg @jdashg @zhenyao @toji please review. Thanks.

zhenyao:
Looks good.

PR #1255, opened 2015-10-06
Verify that fetching result or availability from active query fails.

kenrussell:
This is per David Yen's review of https://codereview.chromium.org/1385793002 .


kenrussell:
@toji @zhenyao @DavidYen please review.

toji:
LGTM.

PR #1256, opened 2015-10-06
Revised language regarding queries' result availability.

kenrussell:
Based on feedback from Jeff Gilbert from Mozilla, revised language to indicate solely that control must return to the browser's main loop in order for queries' results to become available. Follow-on to #1240.


kenrussell:
@jdashg please review this change. @RafaelCintron @grorg and others, your feedback appreciated too.

RafaelCintron:
New spec language LGTM.

toji:
LGTM as well.

jdashg:
LGTM

PR #1257, opened 2015-10-06
Update dEQP shader preprocessor tests from deqp-dev

Oletus:
Pull in latest changes to the tests from deqp-dev branch.

This contains the following changes:

-Double underscores in macro names are now expected to be accepted by the
 compiler. See previous discussion in pull request #1230.
-Unrecognized pragmas are now expected to be accepted by the compiler.
 ESSL 1.00 and 3.00 specs say "If an implementation does not recognize
 the tokens following #pragma, then it will ignore that pragma".
-Non-matching varying invariance is fixed in some #pragma tests.
-Added tests for "defined" operator corner cases for ESSL 3.00 (now
 covered by WebGL 1.0 spec section 6.31)
-Added tests for backslash in a comment for ESSL 3.00

gles2/shaders/preprocessor.test is not a direct copy from dEQP, since
WebGL-specific changes apply to the behavior of extension directives.

dEQP scripts are updated with support for "build_successful" expectation.
Also, testFailed() is now called if the shader file can't be parsed to
get debugging information automatically.

### DIFF
--- a/sdk/tests/deqp/data/gles2/shaders/preprocessor.test
+++ b/sdk/tests/deqp/data/gles2/shaders/preprocessor.test
@@ -2737,7 +2737,7 @@ group pragmas "Pragma Tests"
 	case pragma_unrecognized_token
 		expect build_successful
 		both ""
-			#pragma ¤¤½
+			#pragma **%
 
 			// trailing bytes form a valid but unrecognized preprocessor token
 
@@ -2872,7 +2872,7 @@ group extensions "Extension Tests"
 	case invalid_char_in_name
 		expect compile_fail
 		both ""
-	#extension all¤ : warn
+	#extension all* : warn
 			precision mediump float;
 			void main()
 			{
@@ -2884,7 +2884,7 @@ group extensions "Extension Tests"
 	case invalid_char_in_behavior
 		expect compile_fail
 		both ""
-	#extension all : war¤n
+	#extension all : war*n
 			precision mediump float;
 			void main()
 			{
--- a/sdk/tests/deqp/data/gles2/shaders/preprocessor.test
+++ b/sdk/tests/deqp/data/gles2/shaders/preprocessor.test
@@ -112,6 +112,21 @@ group basic "Basic Tests"
 		""
 	end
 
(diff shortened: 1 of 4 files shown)


kenrussell:
Thanks Olli for this contribution. It looks good overall. @RafaelCintron @grorg @jdashg @toji in particular do you have any comments?

I'd like to ask @zhenyao to review this too. Mo, could you please also merge this once other feedback has been assembled? Thanks.

kenrussell:
Also CC'ing @jasomobica in case he has any feedback.

Oletus:
I found one issue in the shaders: they test that "#pragma ¤¤½" compiles - based on that unrecongized pragmas should be ignored. This goes against WebGL mandating an error when there are invalid characters in the shader. Pyry suggested to me that I should open a Khronos bug about this for discussion, so I did that: https://cvs.khronos.org/bugzilla/show_bug.cgi?id=14939 (private link).

kenrussell:
Thanks for finding that behavioral difference Olli. Perhaps we could merge in the majority of these tests, leaving just that one for future work. Or we could discuss this at the F2F.

kenrussell:
One other comment: WebGL implementations already strip out Unicode characters from comments before even passing the shaders to ANGLE's shader translator, to make it easier for users to add comments in their native language. It would be straightforward to change the WebGL spec to allow these sorts of pragmas as long as they can be safely stripped out of shaders.

zhenyao:
@Oletus Can you get rid of the "#pragma ¤¤½" case and merge the rest?

@kenrussell I prefer failing shaderSource(), so we don't have to perform any complicated processing in WebGLRenderingContext at that stage except for stripping out comments.

Oletus:
Removed the invalid characters from the tests. I put that in a separate commit, that way it will be easier to track which tests were intentionally changed compared to native dEQP. I noticed that the problem wasn't unique to the pragma tests either, but some extension directive tests had the same issue.

kenrussell:
LGTM. Let's let @zhenyao re-review and merge this. He's out until next Monday.

zhenyao:
LGTM

PR #1258, opened 2015-10-07
Improve mipmap-fbo test

Oletus:
Improve the existing test case in the following ways:
-Change the size of the mipmapped texture so that it is minified when
 drawing and two levels actually need to be sampled.
-Specify viewport when the framebuffer dimensions change.
-Use RGBA format. Support for RGB framebuffer attachments is not
 guaranteed by the spec. This bugfix is ported also to versions 1.0.3
 and 1.0.2 of the conformance suite.

Also add a separate subtest for using a cube map face with incomplete
mipmaps as a framebuffer attachment.

These tests already pass at least on Chrome and Firefox on Windows and
Linux with NVIDIA HW.


toji:
LGTM, thanks!

kenrussell:
Thanks for this contribution Olli. LGTM after the fact.

PR #1259, opened 2015-10-07
Require that framebuffer attachments are cube complete

Oletus:
This rule is added to improve the portability of WebGL implementations -
some underlying GL implementations enforce this restriction found in
newer OpenGL specs and working around it is complex. The backwards
compatibility impact of enforcing this rule in WebGL is expected to be
minimal.

This rule is not extended to require mipmap completeness as well, as
using non-mipmap complete attachments is already covered by WebGL tests
and working, and there might be content relying on that.

### DIFF
--- a/sdk/tests/conformance/textures/misc/cube-incomplete-fbo.html
+++ b/sdk/tests/conformance/textures/misc/cube-incomplete-fbo.html
@@ -1,0 +1,93 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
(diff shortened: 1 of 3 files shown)


zhenyao:
For the records, this came up in https://code.google.com/p/chromium/issues/detail?id=528145

@jdashg @grorg @RafaelCintron @kenrussell @toji 

I think it is a reasonable restriction. If no one raises objections in a day or two, I'll merge.

RafaelCintron:
LGTM

sdk/tests/conformance/textures/misc/cube-incomplete-fbo.html
```diff
@@ -0,0 +1,93 @@
...
+<script>
+"use strict";
+
+var wtu = WebGLTestUtils;
+description();
+
+var gl = wtu.create3DContext("example");
+
+var testIncompleteCubemapFaceInFBO = function() {
+    // Create a cube map texture that's not mipmap complete.
```
kenrussell: Should "not mipmap complete" be "not cube complete"?

specs/latest/1.0/index.html
```diff
@@ -3882,6 +3882,20 @@ <h2 class="no-toc">Table of contents</h2>
     compatible with existing content.
 </div>
 
+<h3>Completeness of Cube Map Framebuffer Attachments</h3>
```
kenrussell: Looks good. Could you please update the last modified date too?

kenrussell:
Thanks for this contribution Olli. Looks good. Two minor comments.

Oletus:
Addressed comments with the amended patch.

kenrussell:
Thanks Olli. LGTM. Let's let Mo re-review and merge.

zhenyao:
LGTM

PR #1261, opened 2015-10-14
Incorporate test from @renchenglei in #1239.

kenrussell:
Simplify the test. Verified locally. (Portions of this test are still failing.)


zhenyao:
LGTM

PR #1263, opened 2015-10-14
Generate all 3D texture tests with HTML element input.

zhenyao:
The tests might still be buggy, but let's add them first so we can easily run/test them.


toji:
Welp. Github doesn't like that at all... :) You have to look at the raw patch file to review tex_image_test_generator.py.

LGTM, and thanks!

kenrussell:
I trust that you'll follow through on these tests. LGTM

PR #1268, opened 2015-10-19
Fix OES 2D texture with image data tests.

apchhee:
The test script [sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js](../blob/master/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js) populates a 2x2 test image, but `conformance/extensions/oes-texture*-with-image-data.html` tests use 1x2 canvas, which results in no-op checks. This is fixed in the pull request.

Additionally, the test script checks all four image pixels, and it should not check out-of-canvas-bounds areas.


kenrussell:
Thank you for catching and fixing this. Per our offline discussion, these changes fix key tests which accidentally became no-ops recently.

I'd like @zhenyao to review this too but LGTM from my standpoint.

zhenyao:
Thanks for finding and fixing this bug. LGTM.

PR #1269, opened 2015-10-19
Update dEQP shader keywords tests from deqp-dev

Oletus:
This removes double underscore test cases from dEQP. After this commit,
the dEQP tests don't address double underscores in identifiers in any
way. Using double underscores in identifiers is covered elsewhere in the
WebGL conformance suite, so the intent of this patch is simply to bring
the copy of the tests closer to latest native dEQP, not to change the
expected behavior.

The other test that makes sure double underscores are not allowed in
identifiers is:

/conformance/glsl/misc/shader-with-reserved-words.html


kenrussell:
LGTM. Thanks for helping unify the behavior here.

PR #1278, opened 2015-10-22
Declare that only the std140 layout is supported for uniform blocks.

kenrussell:
This was resolved by the working group a long time ago for portability but was never incorporated into the specification. At the time, confirmed with the OpenGL ES working group that this restriction does not impose a performance penalty.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -3012,6 +3012,24 @@ match the <em>type</em> according to the <a href="#TEXTURE_PIXELS_TYPE_TABLE">ab
         well-defined.
     </div>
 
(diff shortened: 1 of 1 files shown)


kenrussell:
@grorg @jdashg Please review.

This was resolved by the working group a very long time ago but incorporating it into the spec was forgotten. @jasomobica discovered this in the ported dEQP uniform buffer object tests.

@toji @zhenyao FYI.

kenrussell:
@Oletus please take a look as well. This affects several of the ported dEQP tests per #1279 .

Oletus:
Seems like a fairly reasonable restriction. LGTM.

kenrussell:
There haven't been any objections to adding this, so merging it now to avoid conflicts with other upcoming spec changes.

PR #1279, opened 2015-10-22
Removed shared and packed layouts tests.

jasomobica:

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fUniformBlockTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fUniformBlockTests.js
@@ -393,7 +393,7 @@ goog.scope(function() {
         ];
 
         /** @type {Array<Object.<string, glsUniformBlockCase.UniformFlags>>} */
-        var layoutFlags = [{ name: 'shared', flags: glsUniformBlockCase.UniformFlags.LAYOUT_SHARED },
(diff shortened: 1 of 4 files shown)


kenrussell:
Thanks, these changes LGTM.

I think that it may be difficult long term to track the differences between the WebGL versions of these tests and the vanilla ES3 tests.

@Oletus @jdashg @toji @zhenyao do any of you have comments on these test changes?

sdk/tests/deqp/functional/gles3/es3fUniformBlockTests.js
```diff
@@ -393,7 +393,7 @@ goog.scope(function() {
         ];
 
         /** @type {Array<Object.<string, glsUniformBlockCase.UniformFlags>>} */
-        var layoutFlags = [{ name: 'shared', flags: glsUniformBlockCase.UniformFlags.LAYOUT_SHARED },
+        var layoutFlags = [ /* { name: 'shared', flags: glsUniformBlockCase.UniformFlags.LAYOUT_SHARED }, */
```
Oletus: Why is this commented out rather than just removed?

jasomobica: I left it commented out because this code could be potentially useful to somebody.

kenrussell:
Merging now. Thanks for cleaning this up.

PR #1280, opened 2015-10-23
Don't match the type for TexSubImage2D in ES3

qkmiao:
ES3 doesn't have such limitation for texSubImage2D. Chrome removed the
restrict in https://codereview.chromium.org/1344373004/.


kenrussell:

kenrussell:
FYI: see #1282 .

PR #1281, opened 2015-10-23
Fix fragment shader for 2D array target

qkmiao:


toji:
LGTM, thanks for fixing that.

PR #1282, opened 2015-10-25
Remove restriction on texture type in texSubImage2D calls.

kenrussell:
This restriction seems to have been based on a too-strict reading of the ES 2.0 spec, and is no longer a useful restriction for ES 3.0. See https://codereview.chromium.org/1344373004/ and http://crbug.com/532708 for more background.


kenrussell:
@zhenyao @toji @grorg @jdashg please review.

@qkmiao FYI.

zhenyao:
LGTM

PR #1284, opened 2015-10-27
Add a dynamic indexing test for indexing a uniform vector

Oletus:


sdk/tests/conformance2/glsl3/vector-dynamic-indexing.html
```diff
@@ -275,6 +275,21 @@
     my_FragColor = vec4(1.0 - f, f, 0.0, 1.0);
 }
 </script>
+<script id="fshaderIndexUniform" type="x-shader/x-fragment">#version 300 es
```
kenrussell: Would it be worth calling this test fShaderIndexUniformWithUniform and adding another one which indexes a global variable with a uniform? The test above indexes a local variable with a uniform, which I could imagine being handled differently.

Oletus: I don't think it's worth it to exhaustively test all possible variable types, I'd rather just have this targeted regression test.

kenrussell: OK.

kenrussell:
LGTM

PR #1287, opened 2015-10-27
Fix #1286: MIN_PROGRAM_TEXEL_OFFSET should be <= -8.

### LINKED ISSUE #1286: gl-get-calls.html tests MIN_PROGRAM_TEXEL_OFFSET wrong
It should be using shouldBeLessThanOrEqual (which doesn't exist) for a minimum instead of shouldBeGreaterThanOrEqual

kenrussell:
Add shouldBeLessThanOrEqual primitive and use it for this test.

Verified in Chrome Canary.


kenrussell:
@toji @zhenyao @jdashg please review.

toji:
LGTM, thanks!

PR #1289, opened 2015-10-28
Fix bugs in dEQP fragment output test

Oletus:
After these fixes, a lot of the tests pass in latest Chromium.

Bugs fixed:

1. Pass the whole array being indexed to readVec4 and readIVec4,
JavaScript doesn't have pointers.

2. Keep indices bound to ELEMENT_ARRAY_BUFFER when calling drawElements.

3. Remove incorrect Math.floor() calls in places where the intent is to
do calculations in floating point (see native dEQP).

4. Call setPixelInt instead of setPixel when rendering int reference.

5. Upload integer buffer data converted to Int32's instead of Float32's.

6. Fix minimum integer in integer range. Also fix similar issue in
another test.

7. Add in missing gl.readBuffer call for tests that draw into multiple
buffers.

Also fix general issues in tcuImageCompare:

1. Report failed image comparison to test harness.

2. Set error mask values directly to 255 instead of 1 in, since the
JavaScript port of dEQP uses tcuSurface instead of PixelBufferAccess to
store the error mask, and tcuSurface doesn't automatically convert stored
floats to bytes.

3. Log scale and bias in a more understandable format similar to native
dEQP.

### DIFF
--- a/sdk/tests/deqp/framework/common/tcuImageCompare.js
+++ b/sdk/tests/deqp/framework/common/tcuImageCompare.js
@@ -288,7 +288,7 @@ tcuImageCompare.floatUlpThresholdCompare = function(imageSetName, imageSetDesc, 
 
                 maxDiff = deMath.max(maxDiff, diff);
 
-                errorMask.setPixel(x, y, isOk ? [0, 1, 0, 1] : [1, 0, 0, 1]);
+                errorMask.setPixel(x, y, isOk ? [0, 255, 0, 255] : [255, 0, 0, 255]);
             }
         }
     }
@@ -361,7 +361,7 @@ tcuImageCompare.floatThresholdCompare = function(imageSetName, imageSetDesc, ref
 
                 maxDiff = deMath.max(maxDiff, diff);
 
-                errorMask.setPixel(x, y, isOk ? [0, 1, 0, 1] : [1, 0, 0, 1]);
+                errorMask.setPixel(x, y, isOk ? [0, 255, 0, 255] : [255, 0, 0, 255]);
             }
         }
     }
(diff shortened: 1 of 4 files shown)


Oletus:
Also, is it intentional that the image compare tests don't actually report testPassed when the image comparison succeeds? This hurts the ease-of-debugging of the tests quite a bit, particularly ones which do several image comparisons.

jasomobica:
The comparison functions shouldn't report test pass or fail to the framework. The reason is that we want to have just one PASS or FAIL message per test and sometimes there are many image comparisons per test.
IIRC, there are also some tests where image comparison is expected to fail.

The correct fix is to pass/fail the test based on the value of 'allLevelsOk' at the end of es3fFragmentOutputTests.FragmentOutputCase.prototype.iterate().

sdk/tests/deqp/framework/common/tcuImageCompare.js
```diff
@@ -157,7 +157,7 @@ tcuImageCompare.intThresholdCompare = function(imageSetName, imageSetDesc, refer
     var compareOk = deMath.boolAll(deMath.lessThanEqual(maxDiff, threshold));
 
     if (!compareOk) {
-        debug('Image comparison failed: max difference = ' + maxDiff + ', threshold = ' + threshold);
+        testFailed('Image comparison failed: max difference = ' + maxDiff + ', threshold = ' + threshold);
```
kenrussell: Per @jasomobica 's comments, let's revert these back to debug() statements and add at the bottom of es3fFragmentOutputTests.FragmentOutputCase.prototype.iterate:

  if (allLevelsOk) {
      testPassed('');
  } else {
      testFailed('One or more image comparisons failed');
  }

kenrussell:
Thanks @Oletus for tracking down and fixing these issues and @jasomobica for reviewing them.

I reviewed all of the source code changes, comparing to the original C++ sources, and they look good.

Olli, if you could please make the one change that Janusz suggested and which I commented on, this looks great and ready to merge.

Oletus:
Updated the commit with testPassed/testFailed being called in the right place. Also added one more fix: floating point render targets are only used when EXT_color_buffer_float is supported.

kenrussell:
Thanks Olli for these fixes. Merging now.

PR #1290, opened 2015-10-29
Add more check in framebuffer-texture-layer.html

qkmiao:


sdk/tests/conformance2/renderbuffers/framebuffer-texture-layer.html
```diff
@@ -105,6 +105,7 @@
     gl.framebufferTextureLayer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, tex3d, maxLevels - 1, 0);
     wtu.glErrorShouldBe(gl, gl.NO_ERROR,
         "calling framebufferTextureLayer with an appropriate mipmap level should succeed.");
+    checkFramebuffer([gl.FRAMEBUFFER_INCOMPLETE_ATTACHMENT]);
```
zhenyao: This should be FRAMEBUFFER_COMPELETE, I think.

kenrussell: Should it? I think it should be incomplete. The layer that was attached doesn't exist in the 3D texture.

qkmiao: The attached texture failed completeness check due to texture image doesn't exist in (maxLevels -1) level.

zhenyao:
You are right.  I thought the texture was mipmaped already.

PR #1291, opened 2015-10-30
Fix dEQP texture sampling when generating reference images

Oletus:
These bring the port in line with native dEQP, fixing the reference
rendering in most of the texture2darrayshadow tests.

1. Pass entire offset instead of just the first value when calling
sampleNearest2D() in sample2DCompare().

2. Remove extraneous Math.round from sampleNearest*().


Oletus:
I'm not entirely sure if all shadertexturefunction*.html reference renderings are still correct even after this patch, but this should move us closer at least.

jasomobica:
LGTM.

zhenyao:
LGTM

PR #1292, opened 2015-11-02
Improve short circuiting in loop condition test

Oletus:
Test having a short circuiting operator in the for loop expression, in
do-while loop condition, and inside a sequence operator in a loop
condition.

Also make the while loop test a bit more intuitive. Shuffling around
a few operations makes the sideEffectCounter value at the end match
the number "iterations" is compared against.


kenrussell:
Thanks. These improvements to these tests LGTM.

PR #1293, opened 2015-11-03
Fix typo in uniformi.html

Kangz:
Will fix our being puzzled when looking at that failure. http://build.chromium.org/p/chromium.gpu.fyi/builders/Linux%20Release%20%28ATI%29/builds/43033

It says it runs uniformi and fails with testUniformf


toji:
Thanks, good catch.

PR #1294, opened 2015-11-05
Fix invalid operation on framebuffer in expando-loss.html

qkmiao:
In WebGL 2.0, DEPTH_STENCIL_ATTACHMENT is considered an alias for
DEPTH_ATTACHMENT + STENCIL_ATTACHMENT. Attaching a different render
buffer to STENCIL_ATTACHMENT would make different images bound to depth
and stencil attachment points.
If attachment is DEPTH_STENCIL_ATTACHMENT and different images are bound
to the depth and stencil attachment points, getFramebufferAttachmentParameter
generates an INVALID_OPERATION error. Move DEPTH_STENCIL_ATTACHMENT to
the end of the attachments array to avoid such issue.


kenrussell:
Thanks, that's a bit subtle. Good catch.

PR #1295, opened 2015-11-05
Use RGBA for renderable textures, as RGB isn't renderable.

jdashg:
The WebGL 1 spec guarantees that RGBA/UNSIGNED_BYTE textures are renderable (valid framebuffer formats), but it does not guaratee that RGB/UNSIGNED_BYTE textures are. 

(See GLES 2.0.25, p117, Table 4.5 for the list of renderable formats in GLES2)

### DIFF
--- a/sdk/tests/conformance/rendering/framebuffer-switch.html
+++ b/sdk/tests/conformance/rendering/framebuffer-switch.html
@@ -48,7 +48,7 @@ var program = wtu.setupTexturedQuad(gl);
 
 var tex1 = gl.createTexture();
 gl.bindTexture(gl.TEXTURE_2D, tex1);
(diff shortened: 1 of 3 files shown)


kenrussell:
This is a little confusing because GLES 2.0.25 table 4.5 describes renderbuffer sized internal formats, not textures' internal formats. RGB565 is listed as a renderable format in that table, but that doesn't describe the behavior when a texture is attached as a framebuffer attachment.

I think the WebGL spec text in https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.6 really defines the behavior here, since framebuffers can be reported as incomplete for any number of implementation-defined reasons.

This patch looks good; thanks for the fixes. Please submit follow-on pull requests for earlier snapshots of the conformance suite if they're needed.

PR #1296, opened 2015-11-06
multiple uniform blocks in shader should be well-supported

Richard-Yunchao:
multiple uniform blocks in shader will expose an bug in Chromium. PTAL.


kenrussell:
For the record, the associated Chromium bug is https://code.google.com/p/chromium/issues/detail?id=544853 and the patch fixing this particular issue was https://chromium-review.googlesource.com/#/c/310271/ .

kenrussell:
Thanks for this contribution.

PR #1297, opened 2015-11-09
To define the same uniform in two uniform blocks should fail

Richard-Yunchao:
PTAL.

### DIFF
--- a/sdk/tests/conformance2/buffers/uniform-buffers.html
+++ b/sdk/tests/conformance2/buffers/uniform-buffers.html
@@ -40,6 +40,27 @@ void main()
     gl_Position = vec4(p.xyz, 1.0);
 }
 </script>
+<script id='fbadshader' type='x-shader/x-fragment'>#version 300 es
+precision mediump float;
+layout(location=0) out vec4 oColor;
+
+uniform UBOData {
+    float UBORed;
+    float UBOGreen;
+    float UBOBlue;
+};
+
+uniform Color {
+    float Red;
+    float UBOGreen;
+    float Blue;
+};
+
+void main()
+{
+    oColor = vec4(UBORed * Red, UBOGreen * UBOGreen, UBOBlue * Blue, 1.0);
+}
+</script>
 <script id='fshader' type='x-shader/x-fragment'>#version 300 es
 precision mediump float;
 layout(location=0) out vec4 oColor;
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/buffers/uniform-buffers.html
```diff
@@ -127,8 +148,15 @@
 
     wtu.setupUnitQuad(gl);
 
+    var program = wtu.setupProgram(gl, ['vshader', 'fbadshader']);
+    if (!program) {
+        testPassed("To define the same uniform in two uniform blocks should fail");
+    }
+
     var program = wtu.setupProgram(gl, ['vshader', 'fshader']);
-    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be able to compile shader with uniform blocks without error");
```
Richard-Yunchao: I added the new uniform blocks in fbadshader, used glErrorShouldBe( ) to test the compile/link status, found it succeed to compile in Chromium. It was a bug, I thought. However, I found that the JS code was wrong: if shaders fail to compile or link, it will not generate glError. As a result, using wtu.glErrorShouldBe( ) to detect setupProgram( ) is not appropriate. 

In fact, setupProgram will detect the COMPILE_STATUS and LINK_STATUS and return NULL if the shaders are failed to compile/link.

sdk/tests/conformance2/buffers/uniform-buffers.html
```diff
@@ -127,8 +148,15 @@
 
     wtu.setupUnitQuad(gl);
 
+    var testProgram = wtu.setupProgram(gl, ['vshader', 'fbadshader']);
+    if (!testProgram) {
+        testPassed("To define the same uniform in two uniform blocks should fail");
+    }
+
     var program = wtu.setupProgram(gl, ['vshader', 'fshader']);
-    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be able to compile shader with uniform blocks without error");
```
Richard-Yunchao: I added the new uniform blocks in fbadshader, used glErrorShouldBe( ) to test the compile/link status, found it succeed to compile in Chromium. It was a bug, I thought. However, I found that the JS code was wrong: if shaders fail to compile or link, it will not generate glError. As a result, using wtu.glErrorShouldBe( ) to detect setupProgram( ) is not appropriate.

In fact, setupProgram will detect the COMPILE_STATUS and LINK_STATUS and return NULL if the shaders are failed to compile/link.

zhenyao:
Looks good.

PR #1299, opened 2015-11-10
Test texImage3D and texSubImage3D with not enough pixel data

qkmiao:


zhenyao:
LGTM

kenrussell:
Thanks for this contribution. LGTM after the fact.

By the way, have you checked to see whether these error cases are already covered by the recently-ported dEQP tests? I think we should avoid creating unnecessary duplicated tests.

qkmiao:
@kenrussell  Thanks for reminder. I checked dEQP tests covers part of  tests added in my patch: negative width, height and depth, max width, height and depth for level 0 but for other levels. The dEQP tests doesn't cover pixel data size validation either.  Should I remove the duplicate one?

kenrussell:
Going forward, let's avoid adding too many duplicate tests, but in this case, it seems fine to have your tests be more complete and self-contained.

PR #1300, opened 2015-11-10
Add a simple ESSL3 test for constant array initialization

Oletus:
Constant array initialization in the global scope can be problematic,
for the combination of these reasons:

1. It's often undesirable to constant fold array constructors.

2. Constant array constructors that are not folded may need to be
   implemented using function calls on HLSL.

3. Function calls are not allowed in the global scope.

So a WebGL test for this is necessary.

### DIFF
--- a/sdk/tests/conformance2/glsl3/00_test_list.txt
+++ b/sdk/tests/conformance2/glsl3/00_test_list.txt
@@ -7,6 +7,7 @@ array-in-complex-expression.html
 array-length-side-effects.html
 attrib-location-length-limits.html
 compare-structs-containing-arrays.html
+const-array-init.html
 frag-depth.html
 invalid-default-precision.html
 misplaced-version-directive.html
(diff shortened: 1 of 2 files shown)


Oletus:
ANGLE currently passes the first included test but fails the second.

kenrussell:
Thanks for this contribution.

Is this test going to be feasible to pass, or is this a case that we're likely going to have to disallow in WebGL 2.0?

Oletus:
This is feasible to pass. Constant folding arrays should be entirely possible to implement in a smart way inside ANGLE so that it will only be done when it makes sense (so that there would not be a copy of the array when it is dynamically indexed for example). Constant folding of all constant expressions is necessary for validating case labels, array sizes and array indexing, so I think it's worthwhile to still do that, even if the simpler use cases work even without constant folding arrays.

Oletus:
Updated the pull request with one more test.

kenrussell:
Apologies for the long delay re-reviewing this. Looks good. Thanks for this contribution and your analysis.

PR #1301, opened 2015-11-10
Add a test for ANGLE constructor bugs

Oletus:
Passing invalid parameters to constructors can pass shader validation
with the current version of ANGLE. Most driver compilers should fail
compilation and the test will still pass, but running the test does cause
issues at least on OSS NVIDIA Linux driver stack.

### DIFF
--- a/sdk/tests/conformance/glsl/bugs/angle-constructor-invalid-parameters.html
+++ b/sdk/tests/conformance/glsl/bugs/angle-constructor-invalid-parameters.html
@@ -1,0 +1,77 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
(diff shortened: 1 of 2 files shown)


Oletus:
These same tests were added to ANGLE unittests lately. I'm not sure how much value we get from including them in the CTS as well, but could consider it.

sdk/tests/conformance/glsl/bugs/angle-constructor-invalid-parameters.html
```diff
@@ -0,0 +1,79 @@
...
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<script id="fshaderSamplerInConstructorArguments" type="x-shader/x-fragment">
+precision mediump float;
+uniform sampler2D s;
+void main()
+{
+    vec2 v = vec2(0.0, s);
```
kenrussell: Strictly speaking, are the temporary vec2's needed, or could the invalid argument be passed to the vec4 constructor directly?

kenrussell:
Thanks, these tests look good. Only one small question. I think we should add these to the WebGL conformance suite.

sdk/tests/conformance/glsl/bugs/angle-constructor-invalid-parameters.html
Oletus: The vec4 constructor can be used directly. I'll update the pull request.

Oletus:
The patch set has been updated.

kenrussell:
Thanks. Sorry for the delay re-reviewing. LGTM.

PR #1303, opened 2015-11-11
Fix some tests to allow ambiguous error cases.

jdashg:
There are a number of tests that assume the most obvious GL errors will be generated. However, a single function call can match multiple error generation criteria.

I add functions that take a list of possible errors, and make sure that the generated error in the given list.

### DIFF
--- a/sdk/tests/conformance/context/context-lost-restored.html
+++ b/sdk/tests/conformance/context/context-lost-restored.html
@@ -157,7 +157,7 @@ function testLosingAndRestoringContext()
       // restore the context after this event has exited.
       setTimeout(function() {
         shouldGenerateGLError(gl, gl.NO_ERROR, "WEBGL_lose_context.restoreContext()");
-        // The context should still be lost. It will not get restored until the 
+        // The context should still be lost. It will not get restored until the
         // webglrestorecontext event is fired.
         shouldBeTrue("gl.isContextLost()");
         shouldBe("gl.getError()", "gl.NO_ERROR");
@@ -248,7 +248,8 @@ function testOESTextureFloat() {
     // Extension must still be lost.
     var tex = gl.createTexture();
     gl.bindTexture(gl.TEXTURE_2D, tex);
-    shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, 1, 1, 0, gl.RGBA, gl.FLOAT, null)");
+    wtu.shouldGenerateGLErrorIn(gl, [gl.INVALID_ENUM, gl.INVALID_OPERATION],
+                                "gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, 1, 1, 0, gl.RGBA, gl.FLOAT, null)");
     // Try re-enabling extension
     OES_texture_float = reGetExtensionAndTestForProperty(gl, "OES_texture_float", false);
     shouldGenerateGLError(gl, gl.NO_ERROR, "gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, 1, 1, 0, gl.RGBA, gl.FLOAT, null)");
--- a/sdk/tests/conformance/extensions/ext-sRGB.html
+++ b/sdk/tests/conformance/extensions/ext-sRGB.html
@@ -125,9 +125,7 @@ function testInvalidFormat(fn, internalFormat, formatName, enabled) {
   var err = gl.getError();
   if (err == gl.NO_ERROR) {
     testFailed("should NOT be able to create type " + formatName);
-  } else if (err == gl.INVALID_OPERATION) {
-    testFailed("should return gl.INVALID_ENUM for type " + formatName);
-  } else if (err == gl.INVALID_ENUM) {
+  } else if (err == gl.INVALID_ENUM || err == gl.INVALID_OPERATION) {
     testPassed("not able to create invalid format: " + formatName);
   }
 }
@@ -328,7 +326,7 @@ function runFramebufferTextureConversionTest(format) {
   wtu.glErrorShouldBe(gl, gl.NO_ERROR);
 
   shouldBe('gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, ext.FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING_EXT)', 'ext.SRGB_EXT');
-  
+
   if (validFormat) {
     shouldBe("gl.checkFramebufferStatus(gl.FRAMEBUFFER)", "gl.FRAMEBUFFER_COMPLETE");
 
--- a/sdk/tests/conformance/extensions/oes-texture-float.html
+++ b/sdk/tests/conformance/extensions/oes-texture-float.html
@@ -157,7 +157,8 @@ function runTextureCreationTest(testProgram, extensionEnabled, opt_format, opt_n
     }
     gl.texImage2D(gl.TEXTURE_2D, 0, format, width, height, 0, format, gl.FLOAT, data);
     if (expectFailure) {
-        wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "floating-point texture allocation must be disallowed if OES_texture_float isn't enabled");
+        wtu.glErrorShouldBe(gl, [gl.INVALID_ENUM, gl.INVALID_OPERATION],
+                            "floating-point texture allocation must be disallowed if OES_texture_float isn't enabled");
         return;
     } else {
         wtu.glErrorShouldBe(gl, gl.NO_ERROR, "floating-point texture allocation should succeed if OES_texture_float is enabled");
@@ -181,10 +182,10 @@ function arrayToString(arr, size) {
         mySize = size;
     var out = "[";
     for (var ii = 0; ii < mySize; ++ii) {
-	if (ii > 0) {
-	    out += ", ";
-	}
-	out += arr[ii];
+    if (ii > 0) {
+        out += ", ";
+    }
+    out += arr[ii];
     }
     return out + "]";
 }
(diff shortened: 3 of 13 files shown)


zhenyao:
This looks good to me.  I agree the conformance tests shouldn't over-testing the error cases.

sdk/tests/conformance/context/context-lost-restored.html
```diff
@@ -248,7 +248,8 @@
     // Extension must still be lost.
     var tex = gl.createTexture();
     gl.bindTexture(gl.TEXTURE_2D, tex);
-    shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, 1, 1, 0, gl.RGBA, gl.FLOAT, null)");
+    wtu.shouldGenerateGLErrorIn(gl, [gl.INVALID_ENUM, gl.INVALID_OPERATION],
```
kenrussell: This change looks incorrect. The only error this call is supposed to generate is INVALID_ENUM because of the use of the gl.FLOAT type, since the OES_texture_float extension is lost when the context is restored.

sdk/tests/conformance/extensions/ext-sRGB.html
```diff
@@ -125,9 +125,7 @@
   var err = gl.getError();
   if (err == gl.NO_ERROR) {
     testFailed("should NOT be able to create type " + formatName);
-  } else if (err == gl.INVALID_OPERATION) {
-    testFailed("should return gl.INVALID_ENUM for type " + formatName);
-  } else if (err == gl.INVALID_ENUM) {
+  } else if (err == gl.INVALID_ENUM || err == gl.INVALID_OPERATION) {
```
kenrussell: This change also looks incorrect. Per https://www.khronos.org/opengles/sdk/docs/man/xhtml/glTexImage2D.xml the only error condition we should be hitting is that for INVALID_ENUM, since the format isn't an accepted value.

sdk/tests/conformance/extensions/oes-texture-float.html
```diff
@@ -157,7 +157,8 @@
     }
     gl.texImage2D(gl.TEXTURE_2D, 0, format, width, height, 0, format, gl.FLOAT, data);
     if (expectFailure) {
-        wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "floating-point texture allocation must be disallowed if OES_texture_float isn't enabled");
+        wtu.glErrorShouldBe(gl, [gl.INVALID_ENUM, gl.INVALID_OPERATION],
```
kenrussell: As above, the only error that should be possible from the texImage2D call above is INVALID_ENUM, since gl.FLOAT isn't supported because the OES_texture_float extension hasn't been enabled.

sdk/tests/conformance/extensions/ext-sRGB.html
kenrussell: Looking more closely, it looks like INVALID_VALUE could be generated here since internalformat isn't an accepted format, but INVALID_OPERATION should not be able to be generated.

sdk/tests/conformance/extensions/oes-texture-half-float.html
```diff
@@ -232,19 +232,20 @@
...
-          ". Expect " + (expectSuccess ? "Success" : "Failure"));   
+          ". Expect " + (expectSuccess ? "Success" : "Failure"));
 
     var texture = allocateTexture();
     var width = 2;
     var height = 2;
     gl.texImage2D(gl.TEXTURE_2D, 0, format, width, height, 0, format, halfFloatOESEnum, data);
     if(!extensionEnabled) {
-        wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "Half floating point texture must be disallowed if OES_texture_half_float isn't enabled");
+        wtu.glErrorShouldBeIn(gl, [gl.INVALID_ENUM, gl.INVALID_OPERATION],
```
kenrussell: INVALID_ENUM is the only error that should be generated here.

sdk/tests/conformance/extensions/webgl-depth-texture.html
```diff
@@ -116,8 +116,10 @@
 
     var tex = gl.createTexture();
     gl.bindTexture(gl.TEXTURE_2D, tex);
-    wtu.shouldGenerateGLError(gl, gl.INVALID_ENUM, 'gl.texImage2D(gl.TEXTURE_2D, 0, gl.DEPTH_COMPONENT, 1, 1, 0, gl.DEPTH_COMPONENT, gl.UNSIGNED_SHORT, null)');
-    wtu.shouldGenerateGLError(gl, gl.INVALID_ENUM, 'gl.texImage2D(gl.TEXTURE_2D, 0, gl.DEPTH_COMPONENT, 1, 1, 0, gl.DEPTH_COMPONENT, gl.UNSIGNED_INT, null)');
+    wtu.shouldGenerateGLErrorIn(gl, [gl.INVALID_ENUM, gl.INVALID_OPERATION],
+                                'gl.texImage2D(gl.TEXTURE_2D, 0, gl.DEPTH_COMPONENT, 1, 1, 0, gl.DEPTH_COMPONENT, gl.UNSIGNED_SHORT, null)');
+    wtu.shouldGenerateGLErrorIn(gl, [gl.INVALID_ENUM, gl.INVALID_OPERATION],
```
kenrussell: For these two calls, it looks like INVALID_ENUM and INVALID_VALUE are acceptable errors, but not INVALID_OPERATION.

sdk/tests/conformance/extensions/webgl-draw-buffers.html
```diff
@@ -228,8 +228,8 @@
 
   debug("Testing drawBuffersWEBGL with default drawing buffer");
   shouldBe("gl.getParameter(ext.DRAW_BUFFER0_WEBGL)", "gl.BACK");
-  wtu.shouldGenerateGLError(gl, gl.INVALID_VALUE, "ext.drawBuffersWEBGL([])");
-  wtu.shouldGenerateGLError(gl, gl.INVALID_VALUE, "ext.drawBuffersWEBGL([gl.NONE, gl.NONE])");
+  wtu.shouldGenerateGLErrorIn(gl, [gl.INVALID_VALUE, gl.INVALID_OPERATION], "ext.drawBuffersWEBGL([])");
```
kenrussell: It looks to me like this test has always been buggy, and that it should have produced INVALID_OPERATION, not INVALID_VALUE, since the length of the incoming sequence is 0 (corresponding to the n==0 case in the EXT_draw_buffers spec). Do you agree?

sdk/tests/conformance/extensions/webgl-draw-buffers.html
```diff
@@ -228,8 +228,8 @@
...
+  wtu.shouldGenerateGLErrorIn(gl, [gl.INVALID_VALUE, gl.INVALID_OPERATION], "ext.drawBuffersWEBGL([gl.NONE, gl.NONE])");
```
kenrussell: Similarly, it looks like this one should always have been producing INVALID_OPERATION, not INVALID_VALUE, since n > 1 and the default back buffer is bound. Do you agree?

sdk/tests/conformance/textures/misc/texture-formats-test.html
```diff
@@ -83,9 +83,7 @@
       var err = gl.getError();
       if (err == gl.NO_ERROR) {
           testFailed("should NOT be able to create texture of type " + formatName);
-      } else if (err == gl.INVALID_OPERATION) {
-          testFailed("should return gl.INVALID_ENUM for type " + formatName);
-      } else if (err == gl.INVALID_ENUM) {
+      } else if (err == gl.INVALID_ENUM || err == gl.INVALID_OPERATION) {
```
kenrussell: These should only generate INVALID_ENUM or INVALID_VALUE because neither format or internalformat is legal, though they're both the same, per https://www.khronos.org/opengles/sdk/docs/man/xhtml/glTexImage2D.xml .

kenrussell:
@jdashg could you please review the feedback above? Thanks.

sdk/tests/conformance/context/context-lost-restored.html
jdashg: Rereading the specs, I agree. (It actually must be INVALID_OPERATION for WebGL 2)

sdk/tests/conformance/extensions/ext-sRGB.html
jdashg: Looks like it for GLES2. GLES3 is ENUM or OPERATION.

sdk/tests/conformance/extensions/oes-texture-float.html
jdashg: Rereading the spec for GLES2, I think you are correct.

sdk/tests/conformance/extensions/webgl-draw-buffers.html
jdashg: Yeah, I don't see a way for this to be INVALID_VALUE.

sdk/tests/conformance/more/functions/texImage2DBadArgs.html
```diff
@@ -61,7 +61,8 @@
     assertGLError(gl, gl.INVALID_ENUM, "bad target", function(){
         gl.texImage2D(gl.FLOAT, 0, gl.RGBA, 1,1,0,gl.RGBA,gl.UNSIGNED_BYTE, null);
     });
-    assertGLError(gl, gl.INVALID_ENUM, "bad internal format/format", function(){
+    assertGLErrorIn(gl, [gl.INVALID_ENUM, gl.INVALID_OPERATION],
```
jdashg: I agree.

sdk/tests/conformance/more/functions/texSubImage2DBadArgs.html
```diff
@@ -91,7 +91,7 @@
     assertGLError(gl, gl.INVALID_ENUM, "bad format", function(){
         gl.texSubImage2D(gl.TEXTURE_2D, 0, 0,0, 1,1,gl.FLOAT,gl.UNSIGNED_BYTE, new Uint8Array([0,0,0,0]));
     });
-    assertGLError(gl, gl.INVALID_ENUM, "bad type", function(){
+    assertGLErrorIn(gl, [gl.INVALID_ENUM, gl.INVALID_OPERATION], "bad type", function(){
```
jdashg: I think I agree with this now.

sdk/tests/conformance/renderbuffers/framebuffer-test.html
```diff
@@ -156,16 +156,18 @@
   wtu.glErrorShouldBe(gl, gl.INVALID_ENUM,
             "calling framebufferRenderbuffer with attachment = COLOR_ATTACHMENT1 should generate INVALID_ENUM.");
 
+  gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, fbtex, 0);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "attaching a texture to a framebuffer should succeed.");
+
+  // This must come after the successful attachment, otherwise it could also
+  // generate INVALID_OPERATION.
```
jdashg: Correct, I was reading GetRenderbufferParameteriv.

sdk/tests/conformance/extensions/oes-texture-half-float.html
jdashg: Yep.

sdk/tests/conformance/extensions/webgl-depth-texture.html
jdashg: Yep.

sdk/tests/conformance/extensions/ext-sRGB.html
kenrussell: Note that this extension is promoted to core and not available in ES3/WebGL 2.0, so the ES3 rules don't apply to this error check.

sdk/tests/conformance/extensions/webgl-draw-buffers.html
kenrussell: I've split off this fix into #1318 . The rest will come in another pull request.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -1688,7 +1668,35 @@ var glErrorShouldBe = function(gl, glErrors, opt_msg) {
   glErrorShouldBeImpl(gl, glErrors, true, opt_msg);
 };
 
+/**
+ * Tests that the first error GL returns is in the specified error list.
+ * @param {!WebGLContext} gl The WebGLContext to use.
+ * @param {!Array.<number>} expectedErrorList The list of expected gl errors.
+ * @param {string} opt_msg
+ */
+var glErrorShouldBeIn = function(gl, expectedErrorList, opt_msg) {
```
kenrussell: Why was this needed? It looks like it has the same basic functionality as glErrorShouldBeImpl, below.

PR #1305, opened 2015-11-12
Fix 2D array uint texture fragment shader

qkmiao:


sdk/tests/js/tests/tex-image-and-sub-image-utils.js
```diff
@@ -163,7 +163,7 @@ var TexImageUtils = (function() {
     'in vec2 texCoord;',
     'out vec4 fragData;',
     'void main() {',
-    '    vec4 data = vec4(texture(tex, texCoord, 0).rgb 1.0);',
+    '    vec4 data = vec4(texture(tex, vec3(texCoord, 0.0)).rgb, 1.0);',
```
zhenyao: This is incorrect.  For Sampler3dArray, it is texture(tex, xy, z(integer)), not texture(tex, xyz)

qkmiao: GLSL spec 3.00.3 defines "gvec4 texture (gsampler2DArray sampler, vec3 P [, float bias] )". https://www.khronos.org/registry/gles/specs/3.0/GLSL_ES_Specification_3.00.3.pdf

zhenyao:
Ah, you are right.  I don't know where I got the wrong idea. :(

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
kenrussell: Would it not make more sense to define:

  uvec4 data = texture(tex, vec3(texCoord, 0.0));
  fragData = vec4(float(data[0])/255.0, float(data[1])/255.0, float(data[2])/255.0, 1.0);

The cases to float below aren't currently necessary given that data is a float vector.

kenrussell:
LGTM overall after the fact, with one comment.

PR #1306, opened 2015-11-12
Fix bug in invalidate-framebuffer.html.

Richard-Yunchao:
Samplers for all attachments of the fbo should be the same.
Otherwise, it will lead to FRAMEBUFFER_INCOMPLETE


Richard-Yunchao:
Together with https://codereview.chromium.org/1435183002/. The failed cases in invalidate-framebuffer.html will be fixed in Chromium. PTAL.

zhenyao:
LGTM

kenrussell:
Thanks, this fix LGTM after the fact.

PR #1307, opened 2015-11-12
Make extension status in WebGL 2.0 clearer on extension pages

Oletus:
The extension registry front page already mentions that extensions
promoted to core are no longer available in the version where they were
promoted, but it's nicer if that is also clearly stated on each
individual extension's page.


toji:
LGTM

kenrussell:
Nice, thanks. LGTM after the fact.

Issue #1031, opened 2015-06-03
Clarify behavior of ternary operator with arrays


kenrussell:
In #1024 @Oletus pointed out that the ESSL 3.00.4 spec is ambiguous in its definition of the ternary operator's behavior. In Section 5.9 "Expressions", the restriction from ESSL 1.0.17, where the ternary operator may not operate on arrays, was removed. However, in section 5.7 "Structure and Array Operations" the table defining which operators may operate on structures and arrays as whole entities was not updated. Clarification is needed in the WebGL spec and conformance suite.

Oletus:
Khronos discussion on this issue in the ESSL specs: https://cvs.khronos.org/bugzilla/show_bug.cgi?id=13837

Oletus:
Ran into this old bug that should have been closed ages ago. Good to sometimes decrease the open count a bit. :)

kenrussell:
Thanks @Oletus for helping clean things up. :)

PR #1308, opened 2015-11-12
Add disallowed variants of ESSL3 operators to spec

Oletus:
Supporting ternary and sequence operators on void and arrays is likely to
be made optional in GLSL ES 3.00 according to current draft for a future
version of the spec. For WebGL, it's the best to disallow any features
that may not be implemented on all reasonable target platforms. These
restrictions on target platforms could be worked around, but that adds an
extra burden on implementations for very little gain in corner case
features in this case.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -27,7 +27,7 @@
     <!--end-logo-->
 
     <h1>WebGL 2 Specification</h1>
-    <h2 class="no-toc">Editor's Draft 24 October 2015</h2>
+    <h2 class="no-toc">Editor's Draft 18 November 2015</h2>
     <dl>
         <dt>This version:
             <dd>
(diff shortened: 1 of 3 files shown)


Oletus:
This addresses #1031

specs/latest/2.0/index.html
```diff
@@ -3038,6 +3038,22 @@ <h2 class="no-toc">Table of contents</h2>
         layouts that are specific to one vendor's GPUs.
     </div>
 
+    <h3>Disallowed variants of GLSL ES 3.00 operators</h3>
```
kenrussell: Could you please update the last modified date?

kenrussell:
Thanks, this LGTM and sorry for the delay reviewing it. Could you please rebase it and update the last modified date in the spec? Thanks. Feel free to merge yourself afterward. (Let me know if you don't have privileges to do this after you accept the invitation to the WebGL Team associated with this repository.)

Oletus:
Rebased and updated last modified date

kenrussell:
Thanks. LGTM

PR #1311, opened 2015-11-13
Add test for extensions that have been promoted to core

Oletus:
Test that trying to enable promoted extensions in shaders results in
compilation failure, and that the compiler flags for the shaders are
not defined.

### DIFF
--- a/sdk/tests/conformance2/extensions/00_test_list.txt
+++ b/sdk/tests/conformance2/extensions/00_test_list.txt
@@ -1,1 +1,2 @@
 ext-color-buffer-float.html
+promoted-extensions-in-shaders.html
--- a/sdk/tests/conformance2/extensions/promoted-extensions-in-shaders.html
+++ b/sdk/tests/conformance2/extensions/promoted-extensions-in-shaders.html
@@ -1,0 +1,136 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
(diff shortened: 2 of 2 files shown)


Oletus:
The EXT_draw_buffers and EXT_frag_depth tests fail in current Chrome Canary.

kenrussell:
Thanks for catching this.

This is a little concerning because it means that any WebGL 1.0 shader that uses extensions like OES_standard_derivatives won't work in WebGL 2.0 unless the shader is upgraded to ESSL 3.00. I suspect many fragment shaders are using this particular extension. The intention was that most ESSL 1.00 shaders from WebGL 1.0 were expected to work with the WebGL 2.0 context.

I wonder whether the behavior in these situations should be defined differently -- namely, that any ESSL 1.00 shaders using extensions which were promoted to core will have that extension directive defined in WebGL 2.0, and it can be required. Then the browser can translate these shaders to ESSL 3.00 behind the scenes to pick up the built-in functionality. (We're considering doing this in Chrome's WebGL 1.0 implementation to be able to provide EXT_shader_texture_lod on ES 3.0 devices that don't support the extension in ES 2.0.)

Thoughts? + @jdashg @zhenyao @toji @RafaelCintron @grorg

Oletus:
At least in the case of EXT_draw_buffers, if we'd want to enable the functionality for ESSL 1.00 with WebGL 2.0, we run into conflicts between the different specs. ESSL 3.00 spec explicitly forbids indexing gl_FragData with anything other than 0 in ESSL 1.00. If we'd want to be less strict in WebGL, we'd need to override the decision that was made in the native spec. We would also have to decide how the API user enables the functionality - do they need to get the extension from the API? Do they need to use the extension directive in the shader? And when we have everything specified, we have a lot more things to test. We might also run into more cases where native dEQP conflicts with WebGL.

Overall it seems like we'd be signing up for a lot of extra work if we decide to enable extension features in ESSL 1.00 shaders in WebGL 2.0. Is this valuable enough that it justifies the delay caused by implementing it? Would postponing this to "WebGL 2.0.1" be an option? If this is really wanted, making the spec less strict after the initial release shouldn't break any content.

zhenyao:
I think the concerning one is OES_standard_derivatives.  I can go either way, but if it's put to vote, my vote would be to stay the same as the ES spec.

Oletus:
Rebased the patch.

zhenyao, are you concerned about not supporting OES_standard_derivatives or the difficulty of supporting it?

It does add convenience for developers if they're able to use ESSL1+extension shaders in WebGL 2.0, but I think we should only enable it case-by-case basis with careful consideration if at all. Maybe with OES_standard_derivatives it's trivial enough.

"Keeping it the same as ES spec" is unfortunately not very well-defined in this case, because with native ES it's not always clear how an old extension and a new spec are supposed to interact, and since WebGL's extension model differs from native ES. It will need new additions to the WebGL 2 spec if we want to make the intended functionality clear.

zhenyao:
Sorry for being unclear. What I mean is, I prefer to let developers explicitly upgrade shaders to ES3 semantic where OES_standard_derivatives is concerned. However, I won't object to the other way since it won't be hard to translate it in our implementation.

kenrussell:
OK, let's go with this rule for the time being and potentially loosen things up later. Thanks for finding and testing this corner case.

PR #1314, opened 2015-11-16
Added test that checks the behavior of gl.lineWidth

toji:
Test written in response to http://crbug.com/551081

The behavior here is very non-obvious, (storing and returning out of range values without erroring) but it's what the spec says should happen.


kenrussell:
Thanks for putting together this test. LGTM

PR #1315, opened 2015-11-16
WebGL 2.0 supports INT/UINT/FIXED, so we don't want to run these test…

bkotzz:
WebGL 2.0 supports INT/UINT/FIXED, so we don't want to run these tests when the GL version is 2.


toji:
Thanks. LGTM!

PR #1316, opened 2015-11-19
Added conformance test for struct constructor bug discovered by Three…

kenrussell:
….js developers in https://github.com/mrdoob/three.js/pull/7556 .


kenrussell:
Merging myself as I've verified that this test catches the bug on a Nexus 5.

PR #1317, opened 2015-11-19
Mandate texture offset validation at compile time

Oletus:
Texture offset is required to be constant, so validating it against the
correct range can be done at compile time. Add a test for this and edit
the WebGL 2.0 spec.


Oletus:
Surprisingly enough Chrome Canary on Windows already seems to pass this test, even though validation hasn't been added to ANGLE yet. Most likely explanation is that the DirectX compiler already checks that the offset is within the valid range, though I couldn't find documentation mentioning this.

kenrussell:
Thanks for noticing this and for the improved tests. Looks good.

PR #1318, opened 2015-11-19
Fixed expected errors in WEBGL_draw_buffers conformance test.

kenrussell:
When the default framebuffer is bound, EXT_draw_buffers defines certain
error cases as generating INVALID_OPERATION, not INVALID_VALUE. Thanks
to @jdashg for discovering this issue in #1303. This test fix has been
backported to 1.0.3, where the test was introduced. It causes Chrome and
Firefox to begin to fail the test, but the browsers must be fixed.


kenrussell:
@jdashg @toji @zhenyao please review.

toji:
LGTM.

PR #1320, opened 2015-11-20
Integrate tests from web-platform-tests.

jdashg:
There are a couple tests in https://github.com/w3c/web-platform-tests/tree/master/webgl that we should match.

### DIFF
--- a/sdk/tests/conformance/more/functions/bufferSubDataBadArgs.html
+++ b/sdk/tests/conformance/more/functions/bufferSubDataBadArgs.html
@@ -67,8 +67,14 @@ Tests.testBufferData = function(gl) {
         function(){gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, 19, new Uint16Array([1,2,3]));});
     assertFail("data too large",
         function(){gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, 0, new Uint16Array(data.concat([1])));});
-    assertOk("offset + data too large",
+    assertOk("offset + data.length = end",
         function(){gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, 18, new Uint16Array([1,2,3]));});
+    assertOk("offset Infinity",
+        function(){gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, Infinity, new Uint16Array([1,2,3]));});
+    assertOk("offset -Infinity",
+        function(){gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, -Infinity, new Uint16Array([1,2,3]));});
+    assertOk("offset NaN",
+        function(){gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, NaN, new Uint16Array([1,2,3]));});
     assertOk("good args",
         function(){gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, 0, new Uint16Array([1,2,3]));});
     throwError(gl, 'bufferData0');
--- a/sdk/tests/conformance/textures/misc/compressed-tex-image.html
+++ b/sdk/tests/conformance/textures/misc/compressed-tex-image.html
@@ -58,6 +58,8 @@ if (!gl) {
 
   var tex = gl.createTexture();
   gl.bindTexture(gl.TEXTURE_2D, tex);
+  wtu.shouldGenerateGLError(gl, gl.INVALID_ENUM,
+                            "gl.compressedTexSubImage2D(gl.TEXTURE_2D, 0, 0, 0, 10, 10, COMPRESSED_RGB_PVRTC_4BPPV1_IMG, new Uint8Array(8));");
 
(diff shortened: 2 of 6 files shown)


zhenyao:
LGTM

zhenyao:
Please rebase and merge yourself.

sdk/tests/conformance/more/functions/bufferSubDataBadArgs.html
```diff
@@ -67,8 +67,14 @@
...
         function(){gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, 0, new Uint16Array(data.concat([1])));});
-    assertOk("offset + data too large",
+    assertOk("offset + data.length = end",
         function(){gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, 18, new Uint16Array([1,2,3]));});
+    assertOk("offset Infinity",
+        function(){gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, Infinity, new Uint16Array([1,2,3]));});
+    assertOk("offset -Infinity",
+        function(){gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, -Infinity, new Uint16Array([1,2,3]));});
+    assertOk("offset NaN",
+        function(){gl.bufferSubData(gl.ELEMENT_ARRAY_BUFFER, NaN, new Uint16Array([1,2,3]));});
```
kenrussell: What are the rules for these three cases? Do the invalid inputs get converted to 0?

sdk/tests/js/webgl-test-utils.js
```diff
@@ -1767,6 +1767,45 @@ var shouldGenerateGLErrorIn = function(gl, expectedErrorList, evalStr) {
 };
 
 /**
+ * Tests that a function throws or not.
+ * @param {!WebGLContext} gl The WebGLContext to use.
+ * @param throwType Type of thrown error (e.g. TypeError), or false.
+ * @param {string} info Info on what's being tested
+ * @param {function} func The func to test.
+ */
+var shouldThrow = function(gl, throwType, info, func) {
```
kenrussell: It's confusing that "shouldThrow" handles assertions both that the callee throws and doesn't. What about adding two functions "shouldThrow" and "shouldNotThrow" with different signatures and making this one an implementation function instead?

kenrussell:
Looks fine overall; couple of minor comments. Agree with Mo's review.

sdk/tests/conformance/more/functions/bufferSubDataBadArgs.html
jdashg: Yes.

sdk/tests/js/webgl-test-utils.js
jdashg: I guess.

sdk/tests/conformance/more/functions/bufferSubDataBadArgs.html
RafaelCintron: Consider adding these tests to buffers/buffer-data-and-buffer-sub-data.html and verify that the value returned from getBufferParameter(X, BUFFER_SIZE) is what you expect, zero in this case.

kenrussell: Agree with merging these into the other test, which is more complete -- these more/ tests haven't been updated in a while.

In this case, I think getBufferParameter(..., BUFFER_SIZE) will return 12 -- the original size of the array. If we want to validate the result of the call, we should add this to the WebGL 2.0 conformance suite and use getBufferSubData to ensure that (1, 2, 3) were written to the beginning of the ELEMENT_ARRAY_BUFFER.

RafaelCintron: Apologies, I was getting the parameter and original size confused.  @kenrussell is correct, BUFFER_SIZE should return the original buffer size. However, the original buffer has 12 elements and is being filled with a Uint16Array, which has 2 bytes per element.  Since BUFFER_SIZE is in bytes, the total should be 24 bytes, 12. No?

kenrussell: You're right. Thanks for the correction.

sdk/tests/conformance/textures/misc/compressed-tex-image.html
```diff
@@ -62,6 +62,8 @@
 
   var tex = gl.createTexture();
   gl.bindTexture(gl.TEXTURE_2D, tex);
+  wtu.shouldGenerateGLError(gl, [gl.INVALID_ENUM, gl.INVALID_OPERATION],
```
jdm: Why is INVALID_OPERATION allowed here?

kenrussell: I'm not sure why this was allowed when this test was integrated, but this code doesn't exist any more in the top-of-tree conformance suite.
https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/conformance/textures/misc/compressed-tex-image.html

Even before it was deleted, this had been changed at some point to only allow INVALID_ENUM.
https://github.com/KhronosGroup/WebGL/blob/75650c75640cf3dbda29c8a57b6e4ba7edc4f176/sdk/tests/conformance/textures/misc/compressed-tex-image.html

PR #1321, opened 2015-11-20
Add missing precision qualifiers for samplers in shaders

Oletus:
New sampler types introduced in ESSL 3.00 don't have a default precision
qualifier according to ESSL 3.00.4 section 4.5.4.


zhenyao:
Thanks for catching the bug.

PR #1323, opened 2015-11-23
Check framebuffer texture attachments are mipmap complete

qkmiao:


qkmiao:
ES3 spec states:
"If the value of FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE is TEXTURE and the value of FRAMEBUFFER_ATTACHMENT_OBJECT_NAME does not name an immutable-format texture and the value of FRAMEBUFFER_ATTACHMENT_TEXTURE_LEVEL is not levelbase, then the texture must be mipmap complete, and if FRAMEBUFFER_ATTACHMENT_OBJECT_NAME names a cubemap texture, the texture must also be cube complete."

sdk/tests/conformance/textures/misc/mipmap-fbo.html
```diff
@@ -117,9 +118,28 @@
                         "Framebuffer with a non-mipmap complete cube map attachment should be green");
 }
 
+var testIncompleteMipmapInFBO = function () {
```
kenrussell: After discussion with @zhenyao and @toji we agree that tests of new WebGL 2.0 functionality should not be added to the conformance/ directory using tests of getDefault3DContextVersion. Tests of new functionality should be put into conformance2/ .

In order to save time I'm going to merge this as is, but am going to refactor it immediately afterward to move the new test. Please follow this example in the future. Thanks.

PR #1325, opened 2015-11-23
Add test verifying resizing of canvas's back buffer with preserveDraw…

kenrussell:
…ingBuffer=true.

The preserveDrawingBuffer=false case is exercised by several other tests. This test doesn't catch any new bugs, but is intended as verification for http://crbug.com/557848 .

### DIFF
--- a/sdk/tests/conformance/canvas/canvas-preservedrawingbuffer-resize-test.html
+++ b/sdk/tests/conformance/canvas/canvas-preservedrawingbuffer-resize-test.html
@@ -1,0 +1,134 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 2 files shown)


kenrussell:
Please review. Verified with Chrome Canary and Firefox Nightly on Mac OS X.

kenrussell:
(CC: @zhenyao @toji ; FYI @ds-hwang )

RafaelCintron:
I verified in Edge and IE11.  LGTM.

sdk/tests/conformance/canvas/canvas-preservedrawingbuffer-resize-test.html
```diff
@@ -0,0 +1,127 @@
...
+
+  wtu.setupUnitQuad(gl);
+
+  wtu.requestAnimFrame(render);
+}
+
+function render() {
+  var resultGreen = false;
+
+  if (currentSize == largeSize) {
```
toji: If the point of this test is to catch cases in Chrome where an immutable texture is accidentally recycled and we attempt to resize it, I don't know that this ping-ponging between a single large an small size will predictably catch it. Depending on the recycling behavior you may just happen to click into a stride that recycles the large and small buffers at the right time by accident. I think the test would be more effective if there were more variants or a more difficult-to-hit pattern involved.

kenrussell: Thanks, will update this to alternate sizes every few frames instead.

kenrussell:
@toji Please re-review. Thanks.

toji:
LGTM

ds-hwang:
Thank you for adding this test! I verified this in Ubuntu with and without http://crrev.com/1449043005
LGTM with comments

```The preserveDrawingBuffer=false case is exercised by several other tests. ```
WebglConformance.conformance_textures_webgl_canvas_tex_image_and_sub_image_2d_with_webgl_canvas covers ```preserveDrawingBuffer=false``` case, but it's not intended. It's just by change. IMO, it's better for this test to cover both ```preserveDrawingBuffer=true``` and ```preserveDrawingBuffer=false``` cases

kenrussell:
@ds-hwang without the style: attribute the text jumps all over the place, which is jarring.

When generalizing this test it really became apparent it's a duplicate of the existing rapid-resizing.html test so I'm going to submit a pull request that deletes this new test and generalizes the previous one.

PR #1326, opened 2015-11-23
Update to disallow attached images for fbo to have different sizes in…

zhenyao:
… WebGL2.


zhenyao:
@RafaelCintron @jdashg @grorg Do you have any objections for this?

RafaelCintron:
No objections here.  LGTM.

specs/latest/2.0/index.html
```diff
@@ -3076,6 +3076,11 @@ <h2 class="no-toc">Table of contents</h2>
         This restriction ensures easy portability across OpenGL ES 3.0 supporting devices.
     </div>
 
+    <h3><a name="INCOMPLETE_DIMENSIONS">checkFramebufferStatus may return FRAMEBUFFER_INCOMPLETE_DIMENSIONS</a></h3>
```
kenrussell: Please also update the last modified date at the top of the spec.

kenrussell:
LGTM. Given some of the other restrictions that have been enforced in WebGL 2.0 compared to ES 3.0, I think this is better than an inefficient implementation of this functionality.

jdashg:
Thanks for providing rationale.

PR #1327, opened 2015-11-24
Run EXT_disjoint_timer_query test against WebGL 2.0 contexts.

kenrussell:
This extension is exposed against the WebGL 2.0 context as well as the 1.0 context and the test should be run there as well.


kenrussell:
@toji @zhenyao @jdashg please review.

Tested against Chrome's WebGL 2.0 implementation (with an incoming patch to expose the extension in the WebGL 2.0 context).

toji:
LGTM

PR #1328, opened 2015-11-25
Generalized existing rapid-resizing.html test and removed new test fr…

kenrussell:
…om #1325.

The rapid-resizing.html test basically covered this scenario already. Made the test run with preserveDrawingBuffer set to both true and false.


sdk/tests/conformance/canvas/rapid-resizing.html
```diff
@@ -58,80 +59,97 @@
...
-
-    8, 9, 10,
-    8, 10, 11,
-
-    12, 13, 14,
-    12, 14, 15
-  ]), gl.STATIC_DRAW);
-
-  wtu.requestAnimFrame(render);
+function nextTest() {
```
kenrussell: This is largely a refactoring to let the main loop of the test run twice, once with preserveDrawingBuffer=true and once with it false. Unfortunately it's not easy to see that in the code review; the refactoring is going to be hard to review anyway, so I think we should just go ahead with the change.

kenrussell:
@toji @zhenyao @RafaelCintron @jdashg @grorg please review.

@ds-hwang FYI.

ds-hwang:
Thank you for improving tests. LGTM

toji:
LGTM, but it may be nice to have at least one other browser look at it.

RafaelCintron:
I didn't go over the code with a fine tooth comb but I tested in IE, Edge, Chrome and Firefox and verified the test passed.  LGTM.

kenrussell:
Thanks for testing @RafaelCintron . Merging now.

PR #1329, opened 2015-11-25
Add a test for sampler variables with no precision

Oletus:
Sampler types introduced in ESSL3 don't have predefined default
precision. Test that trying to declare samplers with no precision
qualifier fails compilation.

### DIFF
--- a/sdk/tests/conformance2/glsl3/sampler-no-precision.html
+++ b/sdk/tests/conformance2/glsl3/sampler-no-precision.html
@@ -1,0 +1,109 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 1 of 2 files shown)


Oletus:
Test should pass with TOT ANGLE, is failing with current Canary.

zhenyao:
Great!

sdk/tests/conformance2/glsl3/sampler-no-precision.html
```diff
@@ -0,0 +1,109 @@
...
+<script src="../../js/webgl-test-utils.js"></script>
+<script src="../../js/glsl-conformance-test.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<script id="vshaderSamplerNoPrecision" type="x-shader/x-vertex">#version 300 es
+precision mediump float;
+
+uniform $(samplerType) u_sampler;
```
kenrussell: Is it worth referencing the uniform in this and the fragment shader?

kenrussell:
Looks good. One comment.

sdk/tests/conformance2/glsl3/sampler-no-precision.html
Oletus: Shouldn't matter, as it's a negative test. It's fine to not to use an uniform.

PR #1334, opened 2015-11-26
Added test case from Sketchfab team failing on Nexus 5 with Android 6.0.

kenrussell:
Unclear at this time whether Nexus 4 (Adreno 320) is also affected.
Adreno 400 series GPUs appear to be unaffected. Tracked under
http://crbug.com/551937 .


kenrussell:
Merging this myself in order to be able to reference it when filing a bug with Qualcomm.

PR #1336, opened 2015-11-30
Fix dEQP test compressedteximage2d_invalid_size

qkmiao:
Don't need to test "-1" image size for compressedTexImage2D in WebGL due
to no such parameter for WebGL api.


zhenyao:
I don't think I understand why removing the test cases.  They look valid to me.

jasomobica:
I agree with @zhenyao 
There is no imageSize parameter but there's ArrayBufferView and WebGL can verify if it has a valid size.

qkmiao:
Thanks for review @zhenyao @jasomobica . The two test cases should be valid other than negative cases. I updated the pull request.

zhenyao:
Looks correct.  https://www.khronos.org/registry/webgl/sdk/tests/conformance/more/functions/texImage2DBadArgs.html tests texImage2D with 0,0 size with no GL error.  compressedTexImage* should be the same.

PR #1337, opened 2015-11-30
Added new compressed texture formats for WebGL 2.0

qkmiao:
This is copied from conformance/textures/misc/compressed-tex-image.html.
Modified number of supported compressed texture format from 0 to 10. ES3
adds 10 compressed texture formats for CompressedTex{Sub}Image{2D|3D}.

### DIFF
--- a/sdk/tests/conformance2/textures/misc/compressed-tex-image.html
+++ b/sdk/tests/conformance2/textures/misc/compressed-tex-image.html
@@ -1,0 +1,85 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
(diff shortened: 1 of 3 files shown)


sdk/tests/conformance2/textures/misc/compressed-tex-image.html
```diff
@@ -0,0 +1,85 @@
...
+
+var formats = null;
+
+if (!gl) {
+  testFailed("context does not exist");
+} else {
+  testPassed("context exists");
+
+  var tex = gl.createTexture();
+  gl.bindTexture(gl.TEXTURE_2D, tex);
```
zhenyao: Should you test 3D also?  Maybe in a follow up CL?

PR #1338, opened 2015-12-01
Update compressed texture extension tests for WebGL 2.0

qkmiao:
WebGL 2.0 supports 10 new compressed texture formats.


kenrussell:
Thanks for fixing these. It would be ideal if we could share more code among these tests but this looks OK to me. Any comments @zhenyao @toji @jdashg ? Anyone else, please feel free to merge this one.

zhenyao:
LGTM

PR #1340, opened 2015-12-02
Fix wrong parameter passed to glsTextureTestUtil.sampleTexture2DArray

qkmiao:
glsTextureTestUtil.sampleTexture2DArray requires a
tcuTexture.Texture2DArrayView object but a tcuTexture.Texture2DArray
object is passed in. In the native dEQP test, there is an implicit type
conversion from Texture2DArray to Texture2DArrayView. This patch does
explict type conversion in js.


kenrussell:
Thanks for catching and fixing this.

I'm surprised that this isn't also necessary for the other tests in this file, but let's merge this now and fix other instances as they're found.

kenrussell:
@jasomobica FYI.

qkmiao:
Texture2D, Texture3D, and TextureCube have getSubview method to get Texture2DView, Texture3DView, and TextureCubeView object. So other tests won't fail. 
@kenrussell @jasomobica, how are the native dEQP tests translated into js code, by tool or manually? If it's translated by a tool, I wonder could we make the tool more smarter to handle such issue?

jasomobica:
The code was translated manually.
2 gru 2015 18:54 "Qiankun Miao" <notifications@github.com> napisał(a):

PR #1343, opened 2015-12-03
Improve zero size compressed texture validation

qkmiao:


zhenyao:
Agree this is closer to the original deqp c++ tests.

PR #1344, opened 2015-12-03
dEQP: avoid duplicate elements in set emulated with array type

qkmiao:
In dEQP test, set is emulated with array type. We should avoid duplicate
elements in such set. "conditions" in the native implementation is a
set, so check if element is already in it before inserting.


kenrussell:
Thanks, this looks good. Compared to the original C++ source at https://android.googlesource.com/platform/external/deqp/+/master/modules/glshared/glsAttributeLocationTests.cpp#129 .

PR #1345, opened 2015-12-03
Add a js test for texImage2D&&texSubImage2D(ImageBitmap)

xidachen:
This test verifies texImage2D and texSubImage2D code paths taking ImageBitmap.


kenrussell:
Thanks, this looks like a good start. Could you please create a new folder conformance/textures/image_bitmap and add at least one manually written test that uses this? We can modify sdk/tests/py/tex_image_test_generator.py later to auto-generate the ImageBitmap tests.

Thanks. I'll merge this now; please file a follow-on pull request adding the HTML file that's the actual test. (Don't add it to 00_test_list.txt yet.)

PR #1346, opened 2015-12-03
Copy buffer sub data

Richard-Yunchao:
One of the CL to fix http://code.google.com/p/chromium/issues/detail?id=565347 

PTAL. Thanks a lot.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
@@ -481,16 +481,15 @@ goog.scope(function() {
         testGroup.addChild(new es3fApiCase.ApiCaseCallback(
             'copy_buffer_sub_data', 'Invalid gl.copyBufferSubData() usage', gl,
             function() {
-                var data = new Float32Array(32 * 32);
                 var buf = {
                     r: gl.createBuffer(),
                     w: gl.createBuffer()
                 };
 
                 gl.bindBuffer(gl.COPY_READ_BUFFER, buf.r);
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
```diff
@@ -481,12 +481,13 @@ goog.scope(function() {
         testGroup.addChild(new es3fApiCase.ApiCaseCallback(
             'copy_buffer_sub_data', 'Invalid gl.copyBufferSubData() usage', gl,
             function() {
-                var data = new Float32Array(32 * 32);
+                var data = new Float32Array(32);
```
zhenyao: By reading the original deqp test (original test is correct, except it allocate more data than necessary), I think the bug here is we only need to supply 32 bytes to the bufferData call.

So the simplest fix is to change the array size to 32/Float32Array.BYTES_PER_ELEMENT.

kenrussell: Agree. This is the only change that should be made. All of the changes below deviate from the original C++ test and should be removed.

Alternatively, you can just remove "data" and pass the number 32 explicitly to the gl.bufferData calls. The original test initializes the buffer objects to 32 bytes in size and that's what this test should do. WebGL guarantees that the buffer's contents will be zeroed out if you just pass the size.

Richard-Yunchao:
Thanks for your suggestions, Zhenyao and Ken. Considering that variable 'data' is not used afterwards, I would like to assign a constant size to bufferData directly, as Ken suggested.

PTAL.

zhenyao:
Looks good.

kenrussell:
Thanks. LGTM too after the fact.

PR #1347, opened 2015-12-03
fix the errors when calling drawBuffers in negativebufferapi

Richard-Yunchao:
Fix the bug of http://code.google.com/p/chromium/issues/detail?id=565347. 

PTAL. Thanks a lot.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
@@ -565,21 +565,22 @@ goog.scope(function() {
                 this.expectError(gl.NO_ERROR);
 
                 bufferedLogToConsole('gl.INVALID_ENUM is generated if one of the values in bufs is not an accepted value.');
-                gl.drawBuffers(values.splice(2, 2));
+                gl.drawBuffers(values.slice(2, 4));
                 this.expectError(gl.INVALID_ENUM);
 
-                bufferedLogToConsole('gl.INVALID_OPERATION is generated if the GL is bound to the default framebuffer and n is not 1.');
-                gl.drawBuffers(values.splice(0, 2));
+                bufferedLogToConsole('gl.INVALID_OPERATION is generated if the GL is bound to the default framebuffer and the number of queried buffers is not 1.');
+                gl.bindFramebuffer(gl.FRAMEBUFFER, null);
+                gl.drawBuffers(values.slice(0, 2));
                 this.expectError(gl.INVALID_OPERATION);
 
                 bufferedLogToConsole('gl.INVALID_OPERATION is generated if the GL is bound to the default framebuffer and the value in bufs is one of the gl.COLOR_ATTACHMENTn tokens.');
                 gl.bindFramebuffer(gl.FRAMEBUFFER, null);
-                gl.drawBuffers(values.splice(1, 2));
+                gl.drawBuffers(values.slice(2, 3));
                 this.expectError(gl.INVALID_OPERATION);
 
                 bufferedLogToConsole('gl.INVALID_OPERATION is generated if the GL is bound to a framebuffer object and the ith buffer listed in bufs is anything other than gl.NONE or gl.COLOR_ATTACHMENTSi.');
                 gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
-                gl.drawBuffers(values.splice(1, 1));
+                gl.drawBuffers(values.slice(1, 2));
                 this.expectError(gl.INVALID_OPERATION);
 
                 gl.deleteTexture(texture);


sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
```diff
@@ -564,22 +564,23 @@ goog.scope(function() {
                 gl.checkFramebufferStatus(gl.FRAMEBUFFER);
                 this.expectError(gl.NO_ERROR);
 
-                bufferedLogToConsole('gl.INVALID_ENUM is generated if one of the values in bufs is not an accepted value.');
-                gl.drawBuffers(values.splice(2, 2));
-                this.expectError(gl.INVALID_ENUM);
+                bufferedLogToConsole('gl.INVALID_OPERATION is generated if the GL is bound to framebuffer object and one of the values in bufs is not an accepted value.');
+                gl.drawBuffers(values.slice(0, 2));
+                this.expectError(gl.INVALID_OPERATION);
```
Richard-Yunchao: GLES3 man page says that there should be an INVALID_ENUM (see https://www.khronos.org/opengles/sdk/docs/man3/html/glDrawBuffers.xhtml), but GLES 3.0.4 spec says that there should be an INVALID_OPERATION. see https://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-4.2.1 (p185 - p186)

zhenyao: No, the original test is correct.  gl.DEPTH is not an accepted value, and INVALID_ENUM should be generated.

sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
```diff
@@ -564,22 +564,23 @@ goog.scope(function() {
...
-                gl.drawBuffers(values.splice(0, 2));
+                bufferedLogToConsole('gl.INVALID_OPERATION is generated if the GL is bound to the default framebuffer and the number of queried buffers is not 1.');
+                gl.bindFramebuffer(gl.FRAMEBUFFER, null);
+                gl.drawBuffers(values.slice(0, 2));
                 this.expectError(gl.INVALID_OPERATION);
 
                 bufferedLogToConsole('gl.INVALID_OPERATION is generated if the GL is bound to the default framebuffer and the value in bufs is one of the gl.COLOR_ATTACHMENTn tokens.');
                 gl.bindFramebuffer(gl.FRAMEBUFFER, null);
-                gl.drawBuffers(values.splice(1, 2));
+                gl.drawBuffers(values.slice(2, 3));
```
zhenyao: Why change?  slice(1, 2) include a COLOR_ATTACHMENTi, which serves the purpose here.

sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
```diff
@@ -564,22 +564,23 @@ goog.scope(function() {
...
                 this.expectError(gl.INVALID_OPERATION);
 
                 bufferedLogToConsole('gl.INVALID_OPERATION is generated if the GL is bound to a framebuffer object and the ith buffer listed in bufs is anything other than gl.NONE or gl.COLOR_ATTACHMENTSi.');
                 gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
-                gl.drawBuffers(values.splice(1, 1));
+                gl.drawBuffers([gl.NONE, gl.COLOR_ATTACHMENT0]);
```
zhenyao: This is incorrect.  You will need a gl.BACK. The original code is correct.

sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
kenrussell: Agree that the ES 3.0.4 spec is unclear for this error case, but the manual page is pretty clear that INVALID_ENUM should be generated here. The reason it's generated is that DEPTH_ATTACHMENT is not a valid buffer to name in the drawBuffers call.

There is a bug here though. It should be calling:
  gl.drawBuffers(values.slice(2, 4))
not "splice".

sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
```diff
@@ -564,22 +564,23 @@ goog.scope(function() {
...
-                bufferedLogToConsole('gl.INVALID_ENUM is generated if one of the values in bufs is not an accepted value.');
-                gl.drawBuffers(values.splice(2, 2));
-                this.expectError(gl.INVALID_ENUM);
+                bufferedLogToConsole('gl.INVALID_OPERATION is generated if the GL is bound to framebuffer object and one of the values in bufs is not an accepted value.');
+                gl.drawBuffers(values.slice(0, 2));
+                this.expectError(gl.INVALID_OPERATION);
 
-                bufferedLogToConsole('gl.INVALID_OPERATION is generated if the GL is bound to the default framebuffer and n is not 1.');
-                gl.drawBuffers(values.splice(0, 2));
+                bufferedLogToConsole('gl.INVALID_OPERATION is generated if the GL is bound to the default framebuffer and the number of queried buffers is not 1.');
```
kenrussell: The original bufferedLogToConsole call was correct. Please leave it as is.

sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
```diff
@@ -564,22 +564,23 @@ goog.scope(function() {
...
+                gl.bindFramebuffer(gl.FRAMEBUFFER, null);
+                gl.drawBuffers(values.slice(0, 2));
```
kenrussell: Good catch.

sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
kenrussell: "splice" needs to be changed to "slice", but otherwise, the arguments should not be changed.

sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
kenrussell: Agreed. It should be "slice" not "splice", but the arguments should be left unchanged.

sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
Richard-Yunchao: Ken, the drawBuffers() API in WebGL has no argument 'n', this is a little different from the native one. So I revised the log.

sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
Richard-Yunchao: After double check with the function 'slice(startIndex, endIndex)', I think the code is correct.

sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
Richard-Yunchao: Agree. It is my fault... I should keep the buffers the same if it is not wrong.

Richard-Yunchao:
Zhenyao and Ken, thanks for your careful review. The code has been updated, PTAL. @zhenyao , @kenrussell

sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
Richard-Yunchao: Thank you, Zhenyao and Ken. The Blink code was updated to generate INVALID_ENUM per your feedback. PTAL at https://codereview.chromium.org/1495023002/

zhenyao:
I misunderstood slice semantic.  Your CL is in alignment with the original deqp test. LGTM.

sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
kenrussell: Agreed, your new code is correct. The original code is:
https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/gles3/functional/es3fNegativeBufferApiTests.cpp#667

glDrawBuffers(1, &values[2]);

so slice(2, 3) provides the same behavior.

kenrussell:
LGTM too after the fact. Thanks for catching and fixing these bugs.

PR #1348, opened 2015-12-03
fix the bug when calling renderbufferStorageMultisample in negativebufferapi

Richard-Yunchao:
Fix the bug of http://code.google.com/p/chromium/issues/detail?id=565347. The Blink side CL is at https://codereview.chromium.org/1495963002. 

PTAL. Thanks a lot!


zhenyao:
LGTM

kenrussell:
LGTM overall but see also #1350 .

PR #1349, opened 2015-12-03
Clarify TexImage and TexSubImage errors.

kenrussell:
A too-small ArrayBufferView should generate INVALID_OPERATION, not INVALID_VALUE, because under different circumstances the same ArrayBufferView argument could be valid. Clarified this in both the 1.0 and 2.0 specs. This is already tested by conformance/textures/misc/tex-image-with-invalid-data.html . Recent changes to the 2.0 spec were incorrect.

Passing an ImageData whose data attribute has been neutered should generate INVALID_VALUE, since such an ImageData object is never a valid argument. This will be tested by a forthcoming pull request.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -28,7 +28,7 @@
     <!--end-logo-->
 
     <h1>WebGL Specification</h1>
-    <h2 class="no-toc">Editor's Draft 12 October 2015</h2>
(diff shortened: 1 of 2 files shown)


kenrussell:
Please review @grorg @jdashg @RafaelCintron @toji @zhenyao .

zhenyao:
LGTM

jdashg:
This seems reasonable.

PR #1350, opened 2015-12-03
Add back in type annotation lost in #1348 (fixing it at the same time).

kenrussell:


kenrussell:
@Richard-Yunchao @zhenyao FYI.

I need to get the type checking script fully working again so we can check our pull requests and avoid regressing them again. Will work on this.

zhenyao:
Ah I didn't realize these annotation are required. Thanks.

Richard-Yunchao:
Ken, Thanks for your reminder. I didn't catch this.

PR #1351, opened 2015-12-04
Manually adding new test

xidachen:
Adding conformance/textures/image_bitmap/ folder and manually created several tests without creating 00_test_list.txt.


kenrussell:
Thanks, looks fine. Please help us keep track of adding these to the test generator, and turning them on by default. Since ImageBitmap isn't supported everywhere yet, we may need to make the JS test return early if window.createImageBitmap doesn't exist.

PR #1352, opened 2015-12-04
Logical error when compute scalars array index, and the array size is…

xhcao:
Logical error when compute scalars array index, and the array size and values are all not expected.

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsShaderLibraryCase.js
+++ b/sdk/tests/deqp/modules/shared/glsShaderLibraryCase.js
@@ -993,7 +993,7 @@ glsShaderLibraryCase.execute = function() {
                         for (var repNdx = 0; repNdx < numVerticesPerDraw; repNdx++)
                             for (var i = 0; i < numCols; i++)
                                 for (var j = 0; j < numRows; j++)
-                                    scalars[repNdx * numCols + i * numRows * numVerticesPerDraw + j] = val.elements[arrayNdx * scalarSize + i * numRows + j];
+                                    scalars[repNdx * numRows + i * numRows * numVerticesPerDraw + j] = val.elements[arrayNdx * scalarSize + i * numRows + j];
                     } else
                         for (var repNdx = 0; repNdx < numVerticesPerDraw; repNdx++)
                             for (var ndx = 0; ndx < scalarSize; ndx++)


kenrussell:
Thanks for this contribution, but which test or tests (under deqp/data/gles3/shaders/, I assume) is this intended to fix?

The change looks incorrect to me -- it looks like this mirrors code at https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/glshared/glsShaderLibraryCase.cpp#1184 , and after your proposed change, "numCols" is no longer referenced in the array indexing expression.

@jasomobica do you have any comments on this?

jasomobica:
IIRC, the code that this patch modifies has been verified, so the logic should already be correct.
Does this patch fix any of the shader tests?

xhcao:
Hi, kenrussell & jasombica, 
I found this error when debugging /deqp/data/gles3/shaders/conversions.html, if I modified it, the case would pass.

xhcao:
@Richard-Yunchao pls help check it, thank you.

kenrussell:
I highly doubt that this patch is correct. It is probably causing the test to verify only a subset of the input data. I'm closing this pull request because it shouldn't be merged. If you would like to continue debugging this then please build the native dEQP tests, stop in the debugger, print the data the native tests computes, and compare to the data this test produces.

qkmiao:
I just noted the JS logic is not the same as C++ code. It seems the fix is correct. @xhcao can you explain it more clearly why your change is correct.  @jasomobica  can you have a double check about this change? Thanks!

xhcao:
@kenrussell & @jasomobica, pls help have a double to check the patch.

 First, I would explain why the logical is error.
Assue matrix is 3*2 and num vertices per Draw is 4, so numVerticesPerDraw = 4, numCols = 3, numRows = 2. The expected array size of scalars is 4*3*2 = 24, the assigned code execute 24 times in the 3 loops.
there are two errors:
1. the max value of repNdx * numCols + i * numRows * numVerticesPerDraw + j is 3 * 3 + 2 * 2 * 4 + 1 = 26, so current the size of scalars is 27.
2. some elements of scalars array would be assigned twice, for example
   index 9 = 0 * 3 + 1 * 2 * 4 + 1 = 3 * 3 + 0 * 2 * 4 + 0;
   index 17 = 0 * 3 + 2 * 2 * 4 + 1 =  3 * 3 + 1 * 2 * 4 + 0;
and some elements would not be assigned.

Second, the intention of the 3 loops is 
if matrix is 3*2 = {1 2 3 4 5 6}, scalars array should be {1 2 1 2 1 2 1 2 3 4 3 4 3 4 3 4 5 6 5 6 5 6 5 6},
I think my modification could implement it.

Richard-Yunchao:
@kenrussell and @jasomobica , could you have a look?

kenrussell:
First, thanks for this pull request and I'm sorry for taking so long to get back to it.

After instrumenting both the native dEQP harness and the ported JavaScript harness, it appears that the data layout for these test cases is quite different between the two harnesses. Fundamentally, it looks like the JavaScript harness doesn't use the stride in the gluDrawUtil.VertexArrayBinding that's used to send down the input data for these vertex attributes.

The native harness uses client-side data for these vertex attributes, and that's probably where the design difference comes from. In hindsight, I would have advised to upload all of the data into one buffer object, and use offsets and strides into that buffer object to keep the code more similar.

@xhcao , I can confirm that your change is correct. With your change, the values in the scalars array are assigned as the current code expects, though that differs from how the native harness does it. The input values are being validated and that validation is succeeding. Reopening this pull request and merging it now.

PR #1353, opened 2015-12-04
Upgrade ReadPixels to take into consideration ES3 pack parameters.

zhenyao:
This is the ReadPixels with a bound PIXEL_PACK_BUFFER case.

I'll add another test for ReadPixels with no bound PIXEL_PACK_BUFFER in a followup CL.

### DIFF
--- a/sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
+++ b/sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
@@ -164,31 +164,48 @@ function calculatePaddingBytes(bytesPerPixel, packAlignment, width)
     }
 }
 
-function runTestIteration(packAlignment, width, height)
+function runTestIteration(
+    width, height, packAlignment, packRowLength, packSkipPixels, packSkipRows)
 {
-    debug("Testing PACK_ALIGNMENT = " + packAlignment + ", width = " + width + ", height = " + height);
+    debug("Testing width = " + width + ", height = " + height +
+          ", PACK_ALIGNMENT = " + packAlignment + ", ROW_LENGTH = " + packRowLength +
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/reading/read-pixels-into-pixel-pack-buffer.html
```diff
@@ -202,19 +219,34 @@
...
-    runTestIteration(8, 1, 2);
-    runTestIteration(4, 2, 2);
-    runTestIteration(8, 2, 2);
-    runTestIteration(4, 3, 2);
-    runTestIteration(8, 3, 2);
-    runTestIteration(1, 0, 0);
-    runTestIteration(4, 1, 1);
-    runTestIteration(8, 1, 1);
+    debug("Verify that reading pixels to PIXEL_PACK buffer works fine with various pack alignments.");
+    runTestIteration(1, 3, 1, 0, 0, 0);
```
toji: Not going to block landing this, but a suggestion here: The wall of numbers here is getting a little hard to read, and in a lot of cases you're just passing 0 anyway. Maybe make the pack parameters a dictionary? So the calls could look more like this:

```
runTestIteration(1, 3, { packAlignment: 2 });
runTestIteration(3, 3, { packAlignment: 8, packRowLength: 2 });
runTestIteration(3, 3, { packAlignment: 8, packSkipPixels: 2 });
```

toji: One suggestion, but otherwise LGTM. Feel free to merge when you're ready.

zhenyao: Good suggestion.  I will revise in a followup CL, so it's clearer what's been done.

PR #1354, opened 2015-12-07
test cases of bufSize < 0 for getActiveUniform and getActiveAttrib are not necessary

Richard-Yunchao:
getActiveUniform and getActiveAttrib have no bufSize argument at all. This feature is different from that in native API. I suppose the implementation in browser always allocate memory and set a correct bufSize to call into native API, no glError will be generated. 

This patch deleted these 2 code snippets to fix failures in negativestateapi.html in WebGL dEQP conformance test. Here is the crbug: http://code.google.com/p/chromium/issues/detail?id=565347

PTAL. Thanks a lot!


Richard-Yunchao:
see the native API below:
getActiveUniform: https://www.khronos.org/opengles/sdk/docs/man/xhtml/glGetActiveUniform.xml 
getActiveAttrib: https://www.khronos.org/opengles/sdk/docs/man/xhtml/glGetActiveAttrib.xml

zhenyao:
Looks good.

PR #1355, opened 2015-12-07
Refactor: using object instead for function parameters

zhenyao:
in conformance2/reading/read-pixels-into-pixel-pack-buffer.html


toji:
LGTM! Thanks for making it more readable!

PR #1356, opened 2015-12-07
Testing pack parameters for ReadPixels into client array buffer.

zhenyao:
Currently passing on Chrome.

### DIFF
--- a/sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
+++ b/sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
@@ -1,0 +1,170 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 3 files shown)


sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
```diff
@@ -0,0 +1,170 @@
...
+    gl.pixelStorei(gl.PACK_SKIP_PIXELS, packParams.skipPixels);
+    gl.pixelStorei(gl.PACK_SKIP_ROWS, packParams.skipRows);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+    gl.clearColor(expectedColor[0] / 255.0, expectedColor[1] / 255.0, expectedColor[2] / 255.0, expectedColor[3] / 255.0);
+    gl.clear(gl.COLOR_BUFFER_BIT);
+
+    var actualWidth = packParams.rowLength > 0 ? packParams.rowLength : width;
+
+    var bytesPerPixel = 4; // see readPixels' parameters below, the format is gl.RED, type is gl.UNSIGNED_BYTE
```
toji: Shouldn't this be gl.RGBA like above?

toji:
One comment, but otherwise it LGTM. It would be nice to try and take some of the duplicated code in this new test and break it out into common utility functions, but I'm not sure where the best place is to do that. I'll leave it as an optional follow up patch.

Merge when you're ready.

zhenyao:
Thanks.  I'll do some further refactoring to share the code in the next CL.

sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
```diff
@@ -0,0 +1,170 @@
...
+        // Revisit this if spec goes the other way.
+        wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION);
+        return;
+    }
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+    if (!size)
+        return;
+
+    // Check the last pixel of the last row.
```
kenrussell: We may want to consider checking more pixels in these pack alignment tests. It'd be useful to check the first and last pixels of multiple rows.

kenrussell:
Looks good after the fact.

sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
zhenyao: Agree.  I will enhance the tests in followup CLs.

PR #1357, opened 2015-12-07
Move tests around to reduce code duplication.

zhenyao:
Now read-pixels-pack-parameters.html tests both using and not using PIXEL_PACK_BUFFER.


toji:
LGTM, thanks!

PR #1358, opened 2015-12-08
fix the ambiguous test case of getActiveUniformBlockParameter

Richard-Yunchao:
The original test case use a standard program, which have no uniformBlock at all. When call into
getActiveUniformBlockParameter(program, 0, desktopGL['UNIFORM_BLOCK_NAME_LENGTH'])
, both the index and pname are invalid. We should generate INVALID_VALUE and INVALID_ENUM for invalid index and invalid pname respectively. But the original test case expect an INVALID_ENUM. This lead to my CL(https://chromiumcodereview.appspot.com/1495113004/) failed.

See the [Errors] section in gles3 man page for invalid index and invalid pname: https://www.khronos.org/opengles/sdk/docs/man3/html/glGetActiveUniformBlockiv.xhtml. 
"GL_INVALID_VALUE is generated if uniformBlockIndex is greater than or equal to the value of GL_ACTIVE_UNIFORM_BLOCKS or is not the index of an active uniform block in program"

This change tests getActiveUniformBlockParameter against a uniformBlockProgram. So the uniformBlockIndex is correct, this test is specific for 'pname' of this function. 

PTAL. Thanks a lot!


zhenyao:
Thanks for finding the issue.  This is a more meaningful test than before.

PR #1360, opened 2015-12-08
Fix wrong cube map texture target

qkmiao:


zhenyao:
Thanks for catching this.  Now it's the same as the original DEQP test.

PR #1361, opened 2015-12-08
Fix wrong destination framebuffer in es3fNegativeBufferApiTests.js

qkmiao:
Native implementation is here: https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fNegativeBufferApiTests.cpp#1198


zhenyao:
Again, thanks for catching this.  Now it's aligned with the original deqp test.

PR #1362, opened 2015-12-09
Remove unsupported internalformats testing from es3fInternalFormatQueryTests.js

qkmiao:
These unsized internalformats are invalid formats for renderbuffer
objects, while only sized internalformats that is color-, depth-, or
stencil-renderable are valid.


kenrussell:
Sorry, simply removing these tests is not the right thing to do. As the test states, even though these unsized internal formats aren't useful, it's still legal to call GetInternalformativ (or WebGL's getInternalFormatParameter) against them.

Either the test should be changed to expect an error when passing these enums to getInternalformatParameter (and the WebGL 2.0 spec changed to mark this as a difference), or the call should succeed for these two internalformats, and return an array containing the value 0 for the SAMPLES parameter query.

PR #1364, opened 2015-12-09
Test more pixels to make it more thorough.

zhenyao:
@kenrussell per your suggestion.  please review.

### DIFF
--- a/sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
+++ b/sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
@@ -115,12 +115,22 @@ function runTestIteration(width, height, packParams, usePixelPackBuffer)
         var array = new Uint8Array(arrBuffer);
     }
 
-    // Check the last pixel of the last row.
-    var bytesPerRow = actualWidth * bytesPerPixel + padding;
-    var pos = skipSize + bytesPerRow * (height - 1) + (width - 1) * bytesPerPixel;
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
```diff
@@ -115,12 +115,22 @@
...
-    shouldBe("pixel", "expectedColor");
+    // Check the first and last pixels of each row.
+    // TODO(zmo): Revisit this if row_length < width cases are allowed.
+    for (var row = 0; row < height; ++row) {
+        var bytesPerRow = actualWidth * bytesPerPixel + padding;
+        // First pixel
+        var pos = skipSize + bytesPerRow * row;
+        for (var i = 0; i < bytesPerPixel; ++i)
+            pixel[i] = array[pos + i];
+        shouldBe("pixel", "expectedColor");
```
kenrussell: Can we make the rows different colors? That'll make the test more meaningful.

kenrussell:
Thanks for the update. One question.

sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
zhenyao: Sure.  Currently we use clear() to set the fbo color, we just need to switch to shaders and draw()

zhenyao:
Closing this one.

PR #1365, opened 2015-12-10
Fix one test case in constant-precision-qualifier.html

Oletus:
All of the cases in this test try to assign highp 4096.5 to "c" and then
check that the precision qualification of that variable affects the
result of a subsequent operation. One of the cases had the error that
"4096.5 + uTest" could be evaluated at mediump due to uTest being
mediump, even if the target of the assignment was highp. Fix this by
adding a separate highp uniform variable to the test.

The earlier version of this test in 1.0.3 doesn't have this bug.


zhenyao:
What are the rules with an operator with two numbers of different precision? It's not defined?

Oletus:
The resulting expression gets the higher precision.

zhenyao:
Thanks. LGTM

kenrussell:
Thanks for this fix. LGTM too after the fact.

PR #1367, opened 2015-12-11
Remove duplicate test case for constants in loop conditions

Oletus:
The same identical test should only be run once.


zhenyao:
Thanks for cleaning this up. I don't recall a reason for this duplication either.

PR #1368, opened 2015-12-11
use the max samples supported for the specific internalformat

Richard-Yunchao:
Max samples supported for a specific internalformat may differs from max samples for multisampling. The former should be less than or equal to the latter. 

This is the root cause of the failures in the WebGL 2 CTS on OSX. See this reverted CL: https://codereview.chromium.org/1495963002/.

PTAL, thanks a lot!

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fRboStateQueryTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fRboStateQueryTests.js
@@ -245,7 +245,9 @@ es3fRboStateQueryTests.RboSamplesCase.prototype.test = function() {
 
     this.check(glsStateQuery.verifyRenderbuffer(gl.RENDERBUFFER_SAMPLES, 0));
 
-    var max_samples = /** @type {number} */ (gl.getParameter(gl.MAX_SAMPLES));
+    /** @type {Int32Array} */ var samplesSupportedRGBA8 = /** @type {Int32Array} */ gl.getInternalformatParameter(gl.RENDERBUFFER, gl.RGBA8, gl.SAMPLES);
+    // supported samples are written in descending numeric order, so the first one is the max samples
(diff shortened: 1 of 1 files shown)


toji:
LGTM, thanks.

sdk/tests/deqp/functional/gles3/es3fRboStateQueryTests.js
```diff
@@ -245,7 +245,9 @@ es3fRboStateQueryTests.RboSamplesCase.prototype.test = function() {
 
     this.check(glsStateQuery.verifyRenderbuffer(gl.RENDERBUFFER_SAMPLES, 0));
 
-    var max_samples = /** @type {number} */ (gl.getParameter(gl.MAX_SAMPLES));
+    /** @type {Int32Array} */ var samplesSupportedRGBA8 = /** @type {Int32Array} */ gl.getInternalformatParameter(gl.RENDERBUFFER, gl.RGBA8, gl.SAMPLES);
+    // supported samples are written in descending numeric order, so the first one is the max samples
+    var max_samples = samplesSupportedRGBA8[0];
```
zhenyao: samplesSupportedRGBA8[0] is at least gl.MAX_SAMPLES, so the original code is not wrong.  Can you verify if it's a macosx driver bug?

toji:
Sorry, @zhenyao pointed out that the spec does say that the internalformat max samples should always be _at least_ MAX_SAMPLES, so if this is failing it's a driver bug and should be handled differently.

Apologies for the merge/revert.

Richard-Yunchao:
@toji @zhenyao , thanks for your review. I have no Mac, I could not debug the bug on Mac. However, the original test case indeed passes on Linux. So I would like to skip rbostatequery.html on Mac. WDYT? Please take a look at https://codereview.chromium.org/1495963002/. Thanks a lot!

PR #1371, opened 2015-12-12
fix the ambiguous test cases for instanced drawing in NegativeVertexArrayAPI

Richard-Yunchao:
The original code uses 'null' program for instanced drawing. It always generate INVALID_OPERATION. This is ambiguous for many other case to generate appropriate glError, except the one that need to test this feature. 

When count or primcount are negative in drawArraysInstanced/drawElementsInstanced,  Chromium should generates INVALID_VALUE, but it generated INVALID_OPERATION because the original code uses 'null' program for all cases. Now it is fixed. 

I think the original deqp code need to be revised too: https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/gles3/functional/es3fNegativeVertexArrayApiTests.cpp

PTAL. Thanks.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fNegativeVertexArrayApiTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fNegativeVertexArrayApiTests.js
@@ -525,7 +525,6 @@ goog.scope(function() {
 
         testGroup.addChild(new es3fApiCase.ApiCaseCallback('draw_arrays_instanced_invalid_program', 'Invalid gl.drawArraysInstanced() usage', gl, function() {
             gl.useProgram(null);
-            /** @type{WebGLFramebuffer} */ var fbo;
 
             /** @type{WebGLBuffer} */ var bufElements = gl.createBuffer();
             gl.bindBuffer(gl.ARRAY_BUFFER, bufElements);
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fNegativeVertexArrayApiTests.js
```diff
@@ -544,10 +545,12 @@ goog.scope(function() {
             this.expectError(gl.INVALID_VALUE);
```
Richard-Yunchao: After the revision, all these cases above in draw_arrays_instanced_invalid_program has been covered in draw_arrays_instanced. So I would like to remove these cases directly. 

So does many other cases in draw_arrays_instanced_invalid_program and draw_elements_instanced_invalid_program.

zhenyao:
No the original DEQP test is correct.  With useProgram(0), the rendering results are undefined but this is not an error.

However, in WebGL, we don't allow undefined results, so the WebGL spec demands an INVALID_OPERATION for drawArrays and drawElements. I think we need to specify that to cover draw*Instanced().

zhenyao:
@kenrussell I tend to agree to remove these redundant test cases.  What do you think?

kenrussell:
@zhenyao agree to remove these test cases, which generate ambiguous errors under the WebGL spec. LGTM

PR #1372, opened 2015-12-14
Test null pixel data for CompressedTexSubImage3D

qkmiao:


toji:
null should not be allowed by this function. See comments on https://codereview.chromium.org/1525583002/

PR #1374, opened 2015-12-14
Throw exception if pixels data is null for CompressedTex{Sub}Image

qkmiao:
https://codereview.chromium.org/1527723003/ fixes this test.


toji:
Thanks for the test. This looks good but I'm going to make one minor request: shouldThrowException seems like a useful utility function that multiple tests may want to use. Could you put it in webgl_test_utils.js instead?

kenrussell:
Actually, there's already a "shouldThrow" in sdk/tests/js/js-test-pre.js . Let's re-use that.

qkmiao:
Thanks! The existed shouldThrow is used now.

toji:
Thanks for the update! Merging.

PR #1375, opened 2015-12-14
Test case for shader compiler bug on Adreno 300 GPUs.

kenrussell:


kenrussell:
Merging myself. Have received permission from Qualcomm to incorporate this test into the WebGL conformance suite.

PR #1376, opened 2015-12-14
Check more pixels in read-pixels-pack-parameters.html

zhenyao:
@kenrussell This is per your request.  PTAL.

### DIFF
--- a/sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
+++ b/sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
@@ -35,14 +35,13 @@
 </head>
 <body>
 <canvas id="example" width="4" height="4"></canvas>
-<canvas id="canvas2d" width="4" height="4"></canvas>
 <div id="description"></div>
 <div id="console"></div>
 <script>
 "use strict"
 
 var wtu = WebGLTestUtils;
-var expectedColor = [[255, 102, 0, 255],
+var expectedColor = [[249, 102, 0, 255],
                      [2, 200, 102, 255],
                      [134, 87, 234, 255],
                      [99, 5, 76, 255]];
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
```diff
@@ -35,9 +35,18 @@
...
 <div id="description"></div>
 <div id="console"></div>
 <script>
+"use strict"
+
+var wtu = WebGLTestUtils;
+var expectedColor = [[255, 102, 0, 255],
+                     [2, 200, 102, 255],
+                     [134, 87, 234, 255],
+                     [99, 5, 76, 255]];
```
kenrussell: How did you choose these colors? Would it be better to just use more readily visible ones (red, green, blue, yellow) for example?

sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
```diff
@@ -56,6 +65,45 @@
...
 
+function paintWebGLCanvas(gl)
+{
+    var canvas2d = document.getElementById("canvas2d");
+    var ctx2d = canvas2d.getContext("2d");
+    for (var row = 0; row < 4; ++row) {
+        ctx2d.fillStyle = "rgba(" + expectedColor[row][0] + ", " +
+            expectedColor[row][1] + ", " + expectedColor[row][2] + ", " +
+            expectedColor[row][3] + ")";
+        ctx2d.fillRect(0, row, 4, 1);
```
kenrussell: This way of setting up this 2D texture seems to test more code paths than desired. We've been discussing the exact color spaces that are used for 2D canvas, WebGL, etc., and they might not match. A better way would just be to create a Uint8Array and fill it with the data. Just allocating a 1x4 texture should be sufficient. Could you do that instead?

sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
```diff
@@ -56,6 +65,45 @@
...
+    }
+
+    var program = wtu.setupTexturedQuad(gl);
+    gl.disable(gl.DEPTH_TEST);
+    gl.disable(gl.BLEND);
+
+    var tex = gl.createTexture();
+    gl.bindTexture(gl.TEXTURE_2D, tex);
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, canvas2d);
+    gl.generateMipmap(gl.TEXTURE_2D);
```
kenrussell: Do we really need mipmaps? This texture shouldn't be down- or upsampled; it should use NEAREST filtering.

sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
```diff
@@ -142,8 +187,7 @@
 
     debug("");
     debug("Verify that reading pixels to " + destText + " works fine with pack row length.");
-    // PACK_ROW_LENGTH < width case, pending working group decision.
-    // runTestIteration(3, 3, {alignment:8, rowLength:2}, usePixelPackBuffer);  // PACK_ROW_LENTH < width
+    runTestIteration(3, 3, {alignment:8, rowLength:2}, usePixelPackBuffer);  // PACK_ROW_LENTH < width
```
kenrussell: Typo in comments: PACK_ROW_LENTH -> PACK_ROW_LENGTH

sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
```diff
@@ -161,12 +205,8 @@
 debug("");
 debug("Canvas.getContext");
 
-var wtu = WebGLTestUtils;
-var pixel = [0, 0, 0, 0];
-var expectedColor = [255, 102, 0, 255];
-
 var canvas = document.getElementById("example");
-var gl = wtu.create3DContext(canvas);
+var gl = wtu.create3DContext(canvas, {preserveDrawingBuffer: true});
```
kenrussell: Do we really need preserveDrawingBuffer:true? If not let's not add it.

kenrussell:
Thanks, this is great overall. A few comments. Would you mind updating the pull request? If it's too much then I can make these changes afterward.

sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
zhenyao: The reason I don't use the standard colors is that the bugs are likely 1 or more bytes shifted due to incorrect computation.  So with random numbers instead 255 and 0, it's easier to understand what the bug could be.

zhenyao:
@kenrussell All your comments addressed.  Please take another look.

kenrussell:
Very good. Thanks for expanding the test. LGTM

PR #1378, opened 2015-12-15
bind buffer to ARRAY_BUFFER target for vertexAttribPointer

Richard-Yunchao:
According to the WebGL spec, If no buffer bound to ARRAY_BUFFER object, INVALID_OPERATION should be generated against vertexAttribPointer. Then no tests in vertex_attrib_pointer will succeed. 

This small change create a buffer and bind it to ARRAY_BUFFER. 

PTAL. Thanks.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fNegativeVertexArrayApiTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fNegativeVertexArrayApiTests.js
@@ -122,6 +122,9 @@ goog.scope(function() {
         }));
 
         testGroup.addChild(new es3fApiCase.ApiCaseCallback('vertex_attrib_pointer', 'Invalid gl.vertexAttribPointer() usage', gl, function() {
+            /** @type{WebGLBuffer} */ var buffer = gl.createBuffer();
+            gl.bindBuffer(gl.ARRAY_BUFFER, buffer);
+
             bufferedLogToConsole('gl.INVALID_ENUM is generated if type is not an accepted value.');
             gl.vertexAttribPointer(0, 1, 0, true, 0, 0);
             this.expectError(gl.INVALID_ENUM);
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fNegativeVertexArrayApiTests.js
```diff
@@ -166,7 +169,7 @@ goog.scope(function() {
 
         }));
 
-        testGroup.addChild(new es3fApiCase.ApiCaseCallback('vertex_attrib_i_pointer', 'Invalid gl.vertexAttribPointer() usage', gl, function() {
+        testGroup.addChild(new es3fApiCase.ApiCaseCallback('vertex_attrib_i_pointer', 'Invalid gl.vertexAttribIPointer() usage', gl, function() {
```
kenrussell: Good catch.

sdk/tests/deqp/functional/gles3/es3fNegativeVertexArrayApiTests.js
```diff
@@ -166,7 +169,7 @@ goog.scope(function() {
...
             bufferedLogToConsole('gl.INVALID_ENUM is generated if type is not an accepted value.');
             gl.vertexAttribIPointer(0, 1, 0, 0, 0);
```
kenrussell: Did you want to bind a small buffer for this case too? The same issue applies.

kenrussell:
Thanks, this looks good.

sdk/tests/deqp/functional/gles3/es3fNegativeVertexArrayApiTests.js
Richard-Yunchao: @kenrussell , see https://github.com/KhronosGroup/WebGL/pull/1380. Although test cases for vertexAttribIPointer are all passed in Chromium, this is a potential issue for other browser vendors.

kenrussell: They're just bugs -- all browser vendors eventually need to pass all these tests.

PR #1379, opened 2015-12-15
Generate NO_ERROR for deleting null texture

qkmiao:


kenrussell:
Thanks, this looks good. While this is definitely tested elsewhere in the tests, it doesn't seem to be tested anywhere else in the ported dEQP tests.

PR #1380, opened 2015-12-16
bind buffer to ARRAY_BUFFER target for vertexAttribIPointer

Richard-Yunchao:
According to the WebGL spec, If no buffer bound to ARRAY_BUFFER object, INVALID_OPERATION should be generated against vertexAttribIPointer. This is ambiguous to test other invalid operations. 

See the similar one https://github.com/KhronosGroup/WebGL/pull/1378. 

PTAL. Thanks!


kenrussell:
Thanks. LGTM

PR #1381, opened 2015-12-16
Validate sized internal format for generateMipmap

qkmiao:


qkmiao:
See GLES spec 3.0.4 Page 160 for reference:
https://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf.

"If the levelbase array was not specified with an unsized internal format from table
3.3 or a sized internal format that is both color-renderable and texture-filterable
according to table 3.13, an INVALID_OPERATION error is generated."

sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
```diff
@@ -146,6 +146,19 @@
...
+  gl.bindTexture(gl.TEXTURE_2D, tex);
+  gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA8, 0, 0, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+  gl.generateMipmap(gl.TEXTURE_2D);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "generateMipmap should succeed");
+  gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA8UI, 0, 0, 0, gl.RGBA_INTEGER, gl.UNSIGNED_BYTE, null);
+  gl.generateMipmap(gl.TEXTURE_2D);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "generateMipmap should fail for non-texture-filterable format");
+
+  gl.deleteTexture(tex);
+  gl.deleteTexture(tex3d);
```
zhenyao: If you want to delete textures, you probably should do it for each of the above test block, just to be consistent.

qkmiao:
Thanks @zhenyao , I updated this pull request. PTAL.

zhenyao:
lgtm

PR #1382, opened 2015-12-16
Large canvas should success for Tex{Sub}Image2D

qkmiao:


qkmiao:
This conformance test exposes a chromium bug: https://code.google.com/p/chromium/issues/detail?id=568645.

sdk/tests/conformance/textures/misc/tex-sub-image-2d-bad-args.html
```diff
@@ -107,6 +107,15 @@
     wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "type not same as original RGBA 4_4_4_4");
 }
 
+var largeCanvas = document.createElement("canvas");
```
zhenyao: The test itself looks fine, but it doesn't fit in this tex-sub-image-2d-bad-args.html as all args are good here.

qkmiao: I updated this pull request. Now it does the same checks as above code for small canvas.

sdk/tests/conformance/textures/misc/tex-sub-image-2d-bad-args.html
```diff
@@ -76,34 +76,49 @@
...
+wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "format not same as original RGBA 4_4_4_4");
+if (contextVersion < 2) {
+    gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGBA, gl.UNSIGNED_BYTE, c);
+    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "type not same as original RGBA 4_4_4_4");
+}
+
+var largeCanvas = document.createElement("canvas");
+largeCanvas.width = 257;
+largeCanvas.height = 257;
+var largeCanvasContext = largeCanvas.getContext("2d");
```
zhenyao: Could you add a comment here to explain we test with largeCanvas to trigger gpu-to-gpu fast path testing

sdk/tests/conformance/textures/misc/tex-sub-image-2d-bad-args.html
```diff
@@ -76,34 +76,49 @@
...
+gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_SHORT_4_4_4_4, largeCanvas);
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "make texture RGBA 4_4_4_4");
+gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGBA, gl.UNSIGNED_SHORT_4_4_4_4, c);
```
zhenyao: Is the intention here to use largeCanvas instead of c? and also below.

qkmiao: I intend to use c here. largeCanvas is used in texImage2D which trigger GPU fast path, while small canvas 'c' is used in texSubImage2D which trigger normal path. It can trigger the bug by doing this.

zhenyao: I see.  Then can you add another texSubImage2D with largeCanvas, and make sure that path is also correct?

kenrussell: We should keep this test targeted at "tex-sub-image-2d-bad-args". Another test might be appropriate for that case.

sdk/tests/conformance/textures/misc/tex-sub-image-2d-bad-args.html
qkmiao: Done.

sdk/tests/conformance/textures/misc/tex-sub-image-2d-bad-args.html
qkmiao: @zhenyao fixed you comments.
@kenrussell  should I move the positive tests to "tex-sub-image-2d.html"?

kenrussell: @qkmiao this test currently looks fine. But if we're going to add another texSubImage2D with largeCanvas, let's put that in tex-sub-image-2d.html.

kenrussell:
Thanks for these updates. LGTM from my standpoint.

kenrussell:
Sorry for forgetting about this. Merging now, but again, let's please keep the bad-args tests focused on bad arguments, and add positive tests to other tests. Thanks.

PR #1384, opened 2015-12-18
Change spec for readPixels/copySubImage2D out of bounds read.

zhenyao:
Before we set these pixels to (0, 0, 0, 0), now the working group
decided these pixels should remain untouched.

Note that I only changed the readPixels behavior in the test.  I will change the copyTexSubImage2D behavior in the test(s) in a separate CL.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -2381,9 +2381,8 @@ WebGLRenderingContext implements WebGLRenderingContextBase;
             If an attempt is made to call this function with no WebGLTexture bound (see above), an
             <code>INVALID_OPERATION</code> error is generated. <br><br>
 
-            For any pixel lying outside the frame buffer, all channels of the associated texel are
-            initialized to 0; see <a href="#READS_OUTSIDE_FRAMEBUFFER">Reading Pixels Outside the
-            Framebuffer</a>. <br><br>
+            For any pixel lying outside the frame buffer, the corresponding destination pixel remains untouched;
+            see <a href="#READS_OUTSIDE_FRAMEBUFFER">Reading Pixels Outside the Framebuffer</a>. <br><br>
 
             If this function attempts to read from a complete framebuffer with a missing attachment,
             an <code>INVALID_OPERATION</code> error is generated
@@ -3004,8 +3003,9 @@ WebGLRenderingContext implements WebGLRenderingContextBase;
             error is generated.
             <br><br>
 
-            For any pixel lying outside the frame buffer, the value read contains 0 in all channels;
-            see <a href="#READS_OUTSIDE_FRAMEBUFFER">Reading Pixels Outside the Framebuffer</a>.
+            For any pixel lying outside the frame buffer, the corresponding destination buffer range
+            remains untouched; see <a href="#READS_OUTSIDE_FRAMEBUFFER">Reading Pixels Outside the
+            Framebuffer</a>.
             <br><br>
 
             If this function attempts to read from a complete framebuffer with a missing color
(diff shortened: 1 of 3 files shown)


specs/latest/1.0/index.html
```diff
@@ -3565,10 +3565,7 @@ <h2 class="no-toc">Table of contents</h2>
 <h3><a name="READS_OUTSIDE_FRAMEBUFFER">Reading Pixels Outside the Framebuffer</a></h3>
 
 <p>
-In the WebGL API, functions which read the framebuffer
-(<code>copyTexImage2D</code>, <code>copyTexSubImage2D</code>, and <code>readPixels</code>) are
-defined to generate the RGBA value (0, 0, 0, 0) for any pixel which is outside of the bound
-framebuffer.
+In the WebGL API, three functions read the framebuffer: <code>copyTexImage2D</code>, <code>copyTexSubImage2D</code>, and <code>readPixels</code>. <code>copyTexImage2D</code> is defined to generate the RGBA value (0, 0, 0, 0) for any pixel outside of the bound framebuffer. <code>copyTexSubImage2D</code> and <code>readPixels</code> are defined not to touch the corresponding destination range for any pixel outside the bound framebuffer.
```
jdashg: CopyTexImage should have the same OOB behavior.

The guarantee should be that CopyTexImage2D functions as TexImage2D(null)+CopyTexSubImage2D. That is, if the texture data is previously undefined, we define it when we write to a subset of the pixels.

zhenyao: I don't think I fully understand what you mean.  If CopyTexImage2D behaves as TexImage2D(null) + CpyTexSubImage2D, then it is the same as saying we initialize the out-of-bound pixels to (0, 0, 0, 0) in CopyTexImage2D.

RafaelCintron: @zhenyao Please keep the wrap the text to no more than 100 columns.  It's painful to scroll these long sentences.
I would also like to see a corresponding copyTexSubImage2D test that reflects the new behavior.

zhenyao:
OK, I reformatted the spec change into multiple lines as Rafael suggested.

Also, made changes for testing behaviors on copyTexSubImage2D.

(I am not sure if we have more tests that may be affected by this spec change though.  I'll have to modify chrome behavior and run the entire suites to see who may fail).

Please review.

specs/latest/1.0/index.html
```diff
@@ -28,7 +28,7 @@
     <!--end-logo-->
 
     <h1>WebGL Specification</h1>
-    <h2 class="no-toc">Editor's Draft 14 December 2015</h2>
+    <h2 class="no-toc">Editor's Draft 17 December 2015</h2>
```
kenrussell: Could you revise the last modified date?

specs/latest/1.0/index.html
```diff
@@ -2381,9 +2381,8 @@ <h2 class="no-toc">Table of contents</h2>
             If an attempt is made to call this function with no WebGLTexture bound (see above), an
             <code>INVALID_OPERATION</code> error is generated. <br><br>
 
-            For any pixel lying outside the frame buffer, all channels of the associated texel are
-            initialized to 0; see <a href="#READS_OUTSIDE_FRAMEBUFFER">Reading Pixels Outside the
-            Framebuffer</a>. <br><br>
+            For any pixel lying outside the frame buffer, the corresponding destination pixel remain untouched;
```
kenrussell: typo: remain -> remains

specs/latest/1.0/index.html
```diff
@@ -3001,8 +3000,9 @@ <h2 class="no-toc">Table of contents</h2>
             error is generated.
             <br><br>
 
-            For any pixel lying outside the frame buffer, the value read contains 0 in all channels;
-            see <a href="#READS_OUTSIDE_FRAMEBUFFER">Reading Pixels Outside the Framebuffer</a>.
+            For any pixel lying outside the frame buffer, the corresponding destination buffer range
+            remain untouched; see <a href="#READS_OUTSIDE_FRAMEBUFFER">Reading Pixels Outside the
```
kenrussell: remain -> remains

kenrussell:
This spec change looks correct to me. copyTexImage2D must continue to initialize out-of-range texels (those not covered by the framebuffer) to (0, 0, 0, 0) -- that's the behavior that would happen with texImage2D + copyTexSubImage2D. The changes to readPixels and copyTexSubImage2D look correct.

Not sure about the remaining conformance tests -- agree we'll need to do a full conformance run with the changes in place to find all of them.

I advised Mo in an offline discussion to update the snapshotted conformance tests in a follow-on pull request. Ideally they'll be updated to allow the behavior both before and after this spec change.

zhenyao:
date updated (so no change since the current date is today) and typos fixed.

kenrussell:
LGTM.

zhenyao:
Let me merge this now to unblock myself.  We can revise if there are different opinions on wording.

PR #1385, opened 2015-12-18
Add getUniformBlockIndex to gl.object-get-call.js

renchenglei:

### DIFF
--- a/sdk/tests/js/tests/gl-object-get-calls.js
+++ b/sdk/tests/js/tests/gl-object-get-calls.js
@@ -918,7 +918,7 @@ if (contextVersion > 1) {
     shouldBe('gl.getError()', 'gl.NO_ERROR');
 
     var numActiveUniforms = gl.getProgramParameter(program, gl.ACTIVE_UNIFORMS);
-    var blockIndex = gl.getUniformBlockIndex(program, "Transform")
+    var blockIndex = gl.getUniformBlockIndex(program, "Transform");
     var uniformIndices = [];
     for (var i = 0; i < numActiveUniforms; i++)
       uniformIndices.push(i);
@@ -964,6 +964,21 @@ if (contextVersion > 1) {
 	    return gl.getActiveUniforms(program, uniformIndices, pname);
         }
     );
+
+    debug("");
+    debug("Test getUniformBlockIndex");
+    var program = wtu.loadUniformBlockProgram(gl);
+    gl.linkProgram(program);
+    shouldBeTrue('gl.getProgramParameter(program, gl.LINK_STATUS)');
+    shouldBe('gl.getUniformBlockIndex(program, "Transform")', '0');
+    shouldBe('gl.getUniformBlockIndex(program, "u_modelViewMatrix")', 'gl.INVALID_INDEX');
+    shouldBe('gl.getUniformBlockIndex(program, "normal")', 'gl.INVALID_INDEX');
+    shouldBe('gl.getUniformBlockIndex(program, "u_normal")', 'gl.INVALID_INDEX');
+    var noUniformProgram = wtu.loadStandardProgram(gl);
+    gl.linkProgram(noUniformProgram);
+    shouldBeTrue('gl.getProgramParameter(noUniformProgram, gl.LINK_STATUS)');
+    shouldBe('gl.getUniformBlockIndex(noUniformProgram, "u_modelViewProjMatrix")', 'gl.INVALID_INDEX');
+    shouldBe('gl.getUniformBlockIndex(noUniformProgram, "u_normal")', 'gl.INVALID_INDEX');
 }
 
 wtu.glErrorShouldBe(gl, gl.NO_ERROR);


sdk/tests/js/tests/gl-object-get-calls.js
```diff
@@ -964,6 +964,16 @@ if (contextVersion > 1) {
 	    return gl.getActiveUniforms(program, uniformIndices, pname);
         }
     );
+
+    debug("");
+    debug("Test getUniformBlockIndex");
+    var program = wtu.loadUniformBlockProgram(gl);
+    gl.linkProgram(program);
+    shouldBe('gl.getProgramParameter(program, gl.LINK_STATUS)', 'true');
```
zhenyao: nit: you can use shouldBeTrue

sdk/tests/js/tests/gl-object-get-calls.js
```diff
@@ -964,6 +964,16 @@ if (contextVersion > 1) {
...
+    shouldBeNull('gl.getUniformBlockIndex(program, "Transform")');
+    var blockIndex = gl.getUniformBlockIndex(program, "Transform");
```
zhenyao: The above two statements contradict each other.  Did you run this locally?

renchenglei:
Thanks for your suggestion, I have modified the code. Please help review the code, thanks a lot~
@zhenyao

sdk/tests/js/tests/gl-object-get-calls.js
```diff
@@ -964,6 +964,16 @@ if (contextVersion > 1) {
 	    return gl.getActiveUniforms(program, uniformIndices, pname);
         }
     );
+
+    debug("");
+    debug("Test getUniformBlockIndex");
+    var program = wtu.loadUniformBlockProgram(gl);
+    gl.linkProgram(program);
+    shouldBeTrue('gl.getProgramParameter(program, gl.LINK_STATUS)');
+    shouldBeNonNull('gl.getUniformBlockIndex(program, "Transform")');
```
zhenyao: Why not do shouldBeNonNull(blockIndex) instead?

sdk/tests/js/tests/gl-object-get-calls.js
```diff
@@ -918,7 +918,7 @@ if (contextVersion > 1) {
     shouldBe('gl.getError()', 'gl.NO_ERROR');
 
     var numActiveUniforms = gl.getProgramParameter(program, gl.ACTIVE_UNIFORMS);
-    var blockIndex = gl.getUniformBlockIndex(program, "Transform")
+    var blockIndex = gl.getUniformBlockIndex(program, "Transform");
```
zhenyao: This statement should be removed here.

sdk/tests/js/tests/gl-object-get-calls.js
```diff
@@ -964,6 +964,16 @@ if (contextVersion > 1) {
...
+
+    debug("");
+    debug("Test getUniformBlockIndex");
+    var program = wtu.loadUniformBlockProgram(gl);
+    gl.linkProgram(program);
+    shouldBeTrue('gl.getProgramParameter(program, gl.LINK_STATUS)');
+    shouldBeNonNull('gl.getUniformBlockIndex(program, "Transform")');
+    var blockIndex = gl.getUniformBlockIndex(program, "Transform");
+    gl.uniformBlockBinding(program, blockIndex, 1);
+    shouldBe('gl.getUniformBlockIndex(program, "Transform")', '1');
```
zhenyao: Are you trying to verify the binding?  Then you should use getActiveUniformBlockParameter() instead.  Besides, it is another get function and should not be put here.

sdk/tests/js/tests/gl-object-get-calls.js
kenrussell: Actually, it's referenced in this block of tests below.

sdk/tests/js/tests/gl-object-get-calls.js
kenrussell: Agree with Mo's review. I confirmed that this could be written as:

shouldBe('gl.getActiveUniformBlockParameter(program, ' + blockIndex + ', gl.UNIFORM_BLOCK_BINDING)', '1');

Since this requires some setup it seems acceptable to change this section's name to "Test getUniformBlockIndex and getActiveUniformBlockParameter". I looked through the dEQP tests and this test doesn't seem to be duplicated, so it seems fine to add it here.

sdk/tests/js/tests/gl-object-get-calls.js
Richard-Yunchao: Yeah, It is used on line 937 in this file.

sdk/tests/js/tests/gl-object-get-calls.js
Richard-Yunchao: Zhenyao and Ken, you are correct. We should use getActiveUniformBlockParameter to check that it has been bound to a correct uniform buffer binding point. However, I advice @renchenglei to submit another patch to add the tests you suggested against getActiveUniformBlockParameter. For this one, let's keep it simple. 

I would like to add some invalid test cases against getUniformBlockIndex: 
1) test a uniform name (not uniform block name) by this API. 
2) test a non-existed uniform block name.
3) test a program without any uniform blocks at all (but may have uniforms). 
@renchenglei , could you add these test cases?

renchenglei:
I have addressed your suggestion. Please help review the code, thanks a lot~
@kenrussell  @zhenyao  @Richard-Yunchao

sdk/tests/js/tests/gl-object-get-calls.js
```diff
@@ -964,6 +964,21 @@ if (contextVersion > 1) {
 	    return gl.getActiveUniforms(program, uniformIndices, pname);
         }
     );
+
+    debug("");
+    debug("Test getUniformBlockIndex");
+    var program = wtu.loadUniformBlockProgram(gl);
+    gl.linkProgram(program);
+    shouldBeTrue('gl.getProgramParameter(program, gl.LINK_STATUS)');
+    shouldBe('gl.getUniformBlockIndex(program, "Transform")', '0');
```
renchenglei: "Indices of the active uniform blocks of a program are assigned in consecutive order, beginning with zero. "
https://www.khronos.org/opengles/sdk/docs/man3/html/glGetUniformBlockIndex.xhtml
There is only one uniform block in 'program', so the index should be '0'.

kenrussell:
Thanks, this looks fine now.

PR #1386, opened 2015-12-18
add variables into tfVaryings to test transformFeedbackVaryings

Richard-Yunchao:
A small change to fix the bug. 

We need to add variables into 'tfVaryings' to make the array length exceed the max number of supported transformFeedbackVarying separate attribs.


zhenyao:
You are correct.  This is a bug introduced in the c++ -> js translation.

PR #1388, opened 2015-12-21
vector's length should be conform to WebGL spec in uniform*fv

Richard-Yunchao:
According to the WebGL spec: If the array passed to any of the vector forms (those ending in v) has an invalid length, an INVALID_VALUE error will be generated. The length is invalid if it is too short for or is not an integer multiple of the assigned type. 

So the original tests are ambiguous, which lead to failures. The change set appropriate vectors to those uniform*fv APIs, in order to get the only expected error.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
@@ -453,13 +453,13 @@ goog.scope(function() {
 
                 bufferedLogToConsole('gl.INVALID_OPERATION is generated if the size of the uniform variable declared in the shader does not match the size indicated by the glUniform command.');
                 gl.useProgram(program.getProgram());
-                gl.uniform1fv(vec4_v, data);
+                gl.uniform1fv(vec4_v, new Float32Array(1));
                 this.expectError(gl.INVALID_OPERATION);
-                gl.uniform2fv(vec4_v, data);
+                gl.uniform2fv(vec4_v, new Float32Array(2));
                 this.expectError(gl.INVALID_OPERATION);
-                gl.uniform3fv(vec4_v, data);
+                gl.uniform3fv(vec4_v, new Float32Array(3));
                 this.expectError(gl.INVALID_OPERATION);
-                gl.uniform4fv(vec4_v, data);
+                gl.uniform4fv(vec4_v, new Float32Array(4));
                 this.expectError(gl.NO_ERROR);
 
                 bufferedLogToConsole('gl.INVALID_OPERATION is generated if glUniform{1234}fv is used to load a uniform variable of type /** @type {number} */ var , ivec2, ivec3, ivec4, unsigned /** @type {number} */ var , uvec2, uvec3, uvec4.');
@@ -471,7 +471,7 @@ goog.scope(function() {
 
                 bufferedLogToConsole('gl.INVALID_OPERATION is generated if a sampler is loaded using a command other than glUniform1i and glUniform1iv.');
                 gl.useProgram(program.getProgram());
-                gl.uniform1fv(sampler_f, data);
+                gl.uniform1fv(sampler_f, new Float32Array(1));
                 this.expectError(gl.INVALID_OPERATION);
 
                 gl.useProgram(null);
@@ -489,7 +489,7 @@ goog.scope(function() {
                 if (vec4_v == null)
                     assertMsgOptions(false, 'Failed to retrieve uniform location', false, true);
 
-                /** @type {Float32Array} */ var data = new Float32Array(8);
+                /** @type {Float32Array} */ var data = new Float32Array(12);
 
                 bufferedLogToConsole('gl.INVALID_OPERATION is generated if count is greater than 1 and the indicated uniform variable is not an array variable.');
                 gl.useProgram(program.getProgram());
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
```diff
@@ -521,7 +521,7 @@ goog.scope(function() {
                 if (vec4_v == null)
                     assertMsgOptions(false, 'Failed to retrieve uniform location', false, true);
 
-                /** @type {Float32Array} */ var data = new Float32Array(8);
+                /** @type {Float32Array} */ var data = new Float32Array(12);
```
Richard-Yunchao: If the length of the vector is 8. It is not an integer multiple of the assigned type for uniform3fv() below. Then an INVALID_VALUE may generated according to WebGL spec. 

if the vector length is 12, then it is an integer multiple of the assigned type for all APIs below.

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
```diff
@@ -653,7 +656,7 @@ goog.scope(function() {
                 if (ivec4_f == null)
                     assertMsgOptions(false, 'Failed to retrieve uniform location', false, true);
 
-                /** @type {Int32Array} */ var data = new Int32Array(8);
+                /** @type {Int32Array} */ var data = new Int32Array(12);
```
Richard-Yunchao: The same reason I have stated above.

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
```diff
@@ -795,7 +801,7 @@ goog.scope(function() {
                 if (uvec4_f == null)
                     assertMsgOptions(false, 'Failed to retrieve uniform location', false, true);
 
-                /** @type {Uint32Array} */ var data = new Uint32Array(4);
+                /** @type {Uint32Array} */ var data = new Uint32Array(12);
```
Richard-Yunchao: The same reason.

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
```diff
@@ -885,7 +896,7 @@ goog.scope(function() {
                 if (mat4_v == null)
                     assertMsgOptions(false, 'Failed to retrieve uniform location', false, true);
 
-                /** @type {Float32Array} */ var data = new Float32Array(32);
+                /** @type {Float32Array} */ var data = new Float32Array(144);
```
Richard-Yunchao: The same reason.

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
```diff
@@ -485,11 +485,11 @@ goog.scope(function() {
 
                 bufferedLogToConsole('gl.INVALID_OPERATION is generated if the size of the uniform variable declared in the shader does not match the size indicated by the glUniform command.');
                 gl.useProgram(program.getProgram());
-                gl.uniform1fv(vec4_v, data);
+                gl.uniform1fv(vec4_v, new Float32Array(1));
```
zhenyao: You can just switch data to the size of 12, like what you did below.

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
```diff
@@ -603,39 +603,42 @@ goog.scope(function() {
                 if (vec4_v == null || ivec4_f == null || uvec4_f == null || sampler_f == null)
                     assertMsgOptions(false, 'Failed to retrieve uniform location', false, true);
 
-                /** @type {Int32Array} */ var data = new Int32Array(4);
+                /** @type {Int32Array} */ var data1 = new Int32Array(1);
```
zhenyao: Same here, why not just set data size to 12?

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
```diff
@@ -740,44 +743,47 @@ goog.scope(function() {
                 if (vec4_v == null || ivec4_f == null || uvec4_f == null || sampler_f == null)
                     assertMsgOptions(false, 'Failed to retrieve uniform location', false, true);
 
-                /** @type {Uint32Array} */ var data = new Uint32Array(4);
+                /** @type {Uint32Array} */ var data1 = new Uint32Array(1);
```
zhenyao: Same here, why not set data to size 12?

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
```diff
@@ -824,50 +830,55 @@ goog.scope(function() {
                 if (mat4_v == null || sampler_f == null)
                     assertMsgOptions(false, 'Failed to retrieve uniform location', false, true);
 
-                /** @type {Float32Array} */ var data = new Float32Array(16);
+                /** @type {Float32Array} */ var data4 = new Float32Array(4);
```
zhenyao: I think size 72 will do here?

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
Richard-Yunchao: Zhenyao, if we set the size to 12, then there are two errors: 
1) the size of the uniform variable declared in the shader (with 4 elements) does not match the size indicated by the glUniform command (uniform1fv, means 1 element). 
2) count is greater than 1 (12 / 1 = 12), but the variable declared in the shader is not an array. 

If the data is Float32Array(1) and Float32Array(2), etc as I set here, the count will be 1, there would be no 2) error.

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
Richard-Yunchao: The same reason as I explained.

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
Richard-Yunchao: The same reason.

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
Richard-Yunchao: the same reason.

Richard-Yunchao:
@zhenyao , thanks for you review. could you have another look?

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
```diff
@@ -485,11 +485,11 @@ goog.scope(function() {
...
-                gl.uniform1fv(vec4_v, data);
+                gl.uniform1fv(vec4_v, new Float32Array(1));
                 this.expectError(gl.INVALID_OPERATION);
-                gl.uniform2fv(vec4_v, data);
+                gl.uniform2fv(vec4_v, new Float32Array(2));
                 this.expectError(gl.INVALID_OPERATION);
-                gl.uniform3fv(vec4_v, data);
+                gl.uniform3fv(vec4_v, new Float32Array(3));
                 this.expectError(gl.INVALID_OPERATION);
                 gl.uniform4fv(vec4_v, data);
```
zhenyao: It's better to be consistent and just get rid of data and use new Float*(4) here.

zhenyao:
Thanks.  You are correct.  I missed the count>1 && not array case.  One minor suggestion.   Good to merge once that's fixed.

Richard-Yunchao:
addressed your suggestion, @zhenyao .

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
```diff
@@ -453,13 +453,13 @@ goog.scope(function() {
```
zhenyao: Is |data| still used?  If not, could you please remove it.

Richard-Yunchao: Yes. |data| is still used by other test cases. @zhenyao

zhenyao:
lgtm then.

kenrussell:
LGTM too after the fact. Verified all of these changes by hand. Thanks for these fixes.

PR #1389, opened 2015-12-21
In native tests, we can set a uniform location and call uniform[1234]f.

Richard-Yunchao:
However, it is not possible to set a uniform location directly in WebGL.

This change removed uniformf_invalid_location in negativeshaderapi.html.

We can get uniform location from one program, then compile and link another program and call into uniform[1234]f to that uniform location. But this is different from the native deqp tests. And it is covered by conformance test already.  So I prefer to remove these tests directly.

WDYT?

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
@@ -435,38 +435,6 @@ goog.scope(function() {
             }
         ));
 
-        testGroup.addChild(new es3fApiCase.ApiCaseCallback('uniformf_invalid_location', 'Invalid glUniform{1234}f() usage', gl,
-            function() {
-                /** @type {gluShaderProgram.ShaderProgram} */ var program = new gluShaderProgram.ShaderProgram(gl, gluShaderProgram.makeVtxFragSources(uniformTestVertSource, uniformTestFragSource));
-
-                gl.useProgram(program.getProgram());
-                this.expectError(gl.NO_ERROR);
-
-                bufferedLogToConsole('gl.INVALID_OPERATION is generated if location is an invalid uniform location for the current program object and location is not equal to -1.');
-                gl.useProgram(program.getProgram());
-                gl.uniform1f(null, 0.0);
-                this.expectError(gl.INVALID_OPERATION);
-                gl.uniform2f(null, 0.0, 0.0);
-                this.expectError(gl.INVALID_OPERATION);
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
```diff
@@ -435,38 +435,6 @@ goog.scope(function() {
...
-        testGroup.addChild(new es3fApiCase.ApiCaseCallback('uniformf_invalid_location', 'Invalid glUniform{1234}f() usage', gl,
-            function() {
-                /** @type {gluShaderProgram.ShaderProgram} */ var program = new gluShaderProgram.ShaderProgram(gl, gluShaderProgram.makeVtxFragSources(uniformTestVertSource, uniformTestFragSource));
-
-                gl.useProgram(program.getProgram());
-                this.expectError(gl.NO_ERROR);
-
-                bufferedLogToConsole('gl.INVALID_OPERATION is generated if location is an invalid uniform location for the current program object and location is not equal to -1.');
-                gl.useProgram(program.getProgram());
-                gl.uniform1f(null, 0.0);
```
Richard-Yunchao: native deqp tests use -2 (and -1) as the uniform location for these tests, see https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/gles3/functional/es3fNegativeShaderApiTests.cpp#809. Seems that we can not wrap -2 / -1 into WebGLUniformLocation variable. 

null is thought as 0, so the tests will not generate INVALID_OPERATION.

sdk/tests/deqp/functional/gles3/es3fNegativeShaderApiTests.js
```diff
@@ -435,38 +435,6 @@ goog.scope(function() {
...
-                this.expectError(gl.INVALID_OPERATION);
-                gl.uniform2f(null, 0.0, 0.0);
-                this.expectError(gl.INVALID_OPERATION);
-                gl.uniform3f(null, 0.0, 0.0, 0.0);
-                this.expectError(gl.INVALID_OPERATION);
-                gl.uniform4f(null, 0.0, 0.0, 0.0, 0.0);
-                this.expectError(gl.INVALID_OPERATION);
-
-                gl.useProgram(program.getProgram());
-                gl.uniform1f(null, 0.0);
```
Richard-Yunchao: The tests here and the tests above are exactly the same, but they expect different errors...

zhenyao:
Thanks.  The removal is reasonable.

PR #1391, opened 2015-12-22
Remove unnecessary logging from previous debugging

zhenyao:
The bug was already fixed (where RAF can be dropped in Chrome)

This makes log much longer on our bots.  Remove it.


kenrussell:
LGTM.

PR #1392, opened 2015-12-22
Fix detailed description for getShaderPrecisionFormat.

kenrussell:


kenrussell:
The detailed description was out-of-sync compared to the IDL.

kenrussell:
@zhenyao @toji @grorg @jdashg @RafaelCintron please review this small change.

zhenyao:
lgtm

PR #1393, opened 2015-12-23
Load file asynchronously in bad-arguments-test.html

qkmiao:
Synchronous XMLHttpRequest on the main thread is deprecated.
Both chrome and firefox report error on this.

### DIFF
--- a/sdk/tests/js/webgl-test-utils.js
+++ b/sdk/tests/js/webgl-test-utils.js
@@ -2015,6 +2015,21 @@ var loadShaderFromFile = function(
       opt_logShaders, undefined, file, opt_skipCompileStatus);
 };
 
+var loadShaderFromFileAsync = function(
+    gl, file, type, opt_errorCallback, opt_logShaders, opt_skipCompileStatus, callback) {
+  loadTextFileAsync(file, function(gl, type, opt_errorCallback, opt_logShaders, file, opt_skipCompileStatus){
+      return function(success, shaderSource) {
+        if (success) {
+          var shader = loadShader(gl, shaderSource, type, opt_errorCallback,
+              opt_logShaders, undefined, file, opt_skipCompileStatus);
+          callback(true, shader);
+        } else {
+          callback(false, null);
+        }
+      }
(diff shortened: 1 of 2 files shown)


sdk/tests/js/webgl-test-utils.js
```diff
@@ -2015,6 +2015,19 @@ var loadShaderFromFile = function(
...
 };
 
+var loadShaderFromFileAsync = function(
+    gl, file, type, opt_errorCallback, opt_logShaders, opt_skipCompileStatus, callback) {
+  loadTextFileAsync(file, function(gl, type, opt_errorCallback, opt_logShaders, file, opt_skipCompileStatus){
+      return function(success, shaderSource) {
+        if (success) {
+          var shader = loadShader(gl, shaderSource, type, opt_errorCallback,
+              opt_logShaders, undefined, file, opt_skipCompileStatus);
+          callback(true, shader);
```
kenrussell: In the error case, "callback" isn't called with a false first argument. This will cause callers to hang in the case of an error. Could you please make the code more robust by calling:
  callback(false, null);
in the error case?

sdk/tests/js/webgl-test-utils.js
```diff
@@ -2073,6 +2086,24 @@ var loadStandardProgram = function(gl) {
...
+    return function(success, vs) {
+      loadStandardFragmentShaderAsync(gl, function(vs) {
+        return function(success, fs) {
+          var program = gl.createProgram();
+          gl.attachShader(program, vs);
+          gl.attachShader(program, fs);
+          gl.bindAttribLocation(program, 0, "a_vertex");
+          gl.bindAttribLocation(program, 1, "a_normal");
+          linkProgram(gl, program);
+          callback(true, program);
```
kenrussell: Same basic problem here. First, "success" is being ignored multiple times. Second, "callback" is never called with "false" as the first argument, so callers can not be robust to errors.

kenrussell:
Thanks for this contribution. It's a good direction but there are a couple of robustness issues.

qkmiao:
Thanks @kenrussell . I fixed your comments.  Please take another look.

kenrussell:
Thanks for this improvement.

PR #1394, opened 2015-12-23
Generate INVALID_ENUM if internal format is not acceptable for TexImage3D

qkmiao:


qkmiao:
Seems there are errors in the man pages: 
https://www.khronos.org/opengles/sdk/docs/man/xhtml/glTexImage2D.xml
https://www.khronos.org/opengles/sdk/docs/man3/html/glTexImage3D.xhtml

zhenyao:
Note that the internalformat is actually GLint type, not GLenum type. That indicates an INVALID_VALUE error, not an INVALID_ENUM error.

kenrussell:
Agree with Mo's assessment. Both TexImage2D and TexImage3D generate INVALID_VALUE in this case. The WebGL implementations need to be updated, not the tests.

I do see that the native test allows either GL_INVALID_VALUE or GL_INVALID_OPERATION for this case, so updating the test to allow either would be acceptable.

Closing this pull request in the meantime.

qkmiao:
Thanks @zhenyao @kenrussell . It's a little strange internalformat is GLint type for TexImage* while GLenum for other APIs (compressedTexImage*, copyTexImage*). Also my Linux desktop Nvidia driver report INVALID_ENUM for texImage*. 

However, I will follow ES spec to clean texImage* in command buffer and blink.

zhenyao:
It does seem strange to me. I hope there is a good reason for that.

qkmiao:
I found in WebGL 1.0 and 2.0 spec GLenum internalformat was accepted by TexImage2D, but TexImage3D accepted GLint internalformat. Should WebGL spec align with GLES spec? WDYT @zhenyao @kenrussell ?

zhenyao:
I think it's a mistake in WebGL spec.  It should be aligned with the ES2/ES3 spec.

kenrussell:
Yes. I vaguely recall early in WebGL 1.0 spec development that the type of this argument was changed, but agree that the WebGL spec should align with the ES spec.

PR #1395, opened 2015-12-23
fix a bug caused by incorrect javascript operator precedence

Richard-Yunchao:
and a typo: s_latersParams -> s_layersParams.

According to JavaScript operator precedence, the precedence of inequality (!=) is higher than that of bitwise-and(&). see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

This change fixed some bugs in deqp fbo related tests, for example, test cases in fobcompleteness.html. 

PTAL. Thanks!


zhenyao:
LGTM
Thanks for fixing the bugs and cleaning up the typo.

kenrussell:
Thanks for finding and fixing these bugs.

PR #1396, opened 2015-12-25
remove unsized internalformat/format

Richard-Yunchao:
unsized internalformat RGBA and GRB  are not supported by renderbufferStorageMultisample / renderbufferStorage, see page 205 in gles 3.0.4 spec: https://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-4.4.2

They would lead to INVALID_ENUM error when call into renderbufferStorage or renderbufferStorageMultisample in fbocompleteness.html

PTAL. Thanks!

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fFboCompletenessTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fFboCompletenessTests.js
@@ -32,19 +32,6 @@ goog.scope(function() {
         /**
         * @type {Array<number>}
         */
-        es3fFboCompletenessTests.s_es3UnsizedColorRenderables = [
-            // "...or if it is unsized format RGBA or RGB."
-            // See Table 3.3 in GLES3.
-            glsFboUtil.formatkey(gl.RGBA, gl.UNSIGNED_BYTE),
-            glsFboUtil.formatkey(gl.RGBA, gl.UNSIGNED_SHORT_4_4_4_4),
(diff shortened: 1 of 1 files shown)


zhenyao:
I don't see the RENDERBUFFER_VALID flag for these formats, so why should there be a failure in the tests?

kenrussell:
Sorry, but this pull request looks incorrect. These unsized internal formats are not supposed to be tested with renderbufferStorage or renderbufferStorageMultisample; that's why (as Mo already pointed out) the RENDERBUFFER_VALID flag isn't present. I just ran the same tests from the native dEQP harness and they all pass, so I suspect either some bug in the porting of the test harness, or a bug in the WebGL implementation.

I'm closing this as it definitely should not be merged as is.

jasomobica:
The bug is probably in the JS port, the framework throws an exception when a GL error happens. In the case of fbo completeness tests, some subtests are expected to cause a GL error.
I will submit a fix soon.

Richard-Yunchao:
Thanks for your review, @zhenyao, @kenrussell and @jasomobica . Could you tell me the instructions to run deqp in native environment? I have download the source code and tried. But it seems that I can not run them on my Linux desktop. Could we run these tests on Linux / Mac machine?

kenrussell:
@Richard-Yunchao I've emailed you and @qkmiao instructions on building and running the native dEQP harness on Linux and Android.

PR #1397, opened 2015-12-27
Fix version string spec.

greggman:
According to the spec the correct version string is

    WebGL<space>2.0<space>...


kenrussell:
The space and vendor-specific information should be optional, not mandatory. The 1.0 version of the test doesn't mandate the space and neither should the 2.0 test. Clarifying this in #1398 .

kenrussell:
The spec's been updated in #1398 so closing this pull request.

PR #1398, opened 2015-12-28
Made vendor-specific information optional.

kenrussell:
The space and vendor-specific information should be optional to clear up confusion in #1397 .


kenrussell:
@grorg @jdashg @toji @zhenyao @RafaelCintron please review.

PR #1399, opened 2015-12-28
Changed the logic when uploading matrix attributes in glsShaderLibrar…

jasomobica:
…yCase.

This logic matches more closely the approach in the original deqp tests.


jasomobica:
This is a follow up patch complimenting the change in https://github.com/KhronosGroup/WebGL/pull/1352

The scalar arrays, offsets and strides should be the same as in the original code.
I still have a separate buffer per column.

kenrussell:
Thanks for this cleanup Janusz.

I verified manually that sdk/tests/deqp/data/gles3/shaders/conversions.html continues to pass with this change applied. This does make the calling code more understandable. Merging now.

@Richard-Yunchao @xhcao @qkmiao FYI.

PR #1401, opened 2015-12-30
Specify texture before updating subregion of it using TexSubImage

qkmiao:


kenrussell:
LGTM. I assume the original tests are ambiguous because they might generate either an INVALID_OPERATION or INVALID_VALUE error.

PR #1403, opened 2015-12-30
Changed error checking in fbo completeness tests.

jasomobica:
Before this patch, any GL error caused a test failure.
However, some of the tests expect GL errors.

Removed throwing exceptions when a GL error occurs in fbo completeness
tests.


kenrussell:
Thanks. Looks good.

PR #1405, opened 2015-12-31
Add getActiveUniformBlockName to gl.object-get-call.js

renchenglei:


kenrussell:
Thanks for this contribution. LGTM

Let's however please continue to be careful about not duplicating existing tests. sdk/tests/deqp/functional/gles3/es3fShaderStateQueryTests.js and sdk/tests/deqp/functional/gles3/uniformbuffers.html seem to cover this functionality, though I agree your test is simple and the positive and negative cases are easy to understand.

PR #1408, opened 2016-01-04
Fix ambiguous texture input test.

demo99:
According to the spec https://www.khronos.org/opengles/sdk/docs/man/xhtml/glTexImage2D.xml. GL_INVALID_VALUE is generated if internalformat is not an accepted format and GL_INVALID_ENUM is generated if format or type is not an accepted value. So this expectedError should be INVALID_VALUE or gl.INVALID_ENUM.

### DIFF
--- a/sdk/tests/js/tests/tex-input-validation.js
+++ b/sdk/tests/js/tests/tex-input-validation.js
@@ -292,7 +292,7 @@ var testCases = [
     border: 0,
     format: 0x1903, // GL_RED
     type: gl.UNSIGNED_BYTE,
-    expectedError: gl.INVALID_ENUM },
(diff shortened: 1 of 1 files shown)


kenrussell:
Thanks. This update looks correct.

(demo99, are you one of the regular submitters? Is this a different account?)

demo99:
Hi Ken,
No, this is my first time to submit this project.

zhenyao:
We will want to update the older conformance snapshots also (1.01, 1.02, 1.03).

kenrussell:
Good point Mo. demo99, please feel free to submit a pull request updating the older test suites.

PR #1407, opened 2015-12-31
Add incomplete framebuffer test for CopyTexSubImage3D

qkmiao:

PR #1410, opened 2016-01-04
dEQP: use same framebuffer as C++ code

qkmiao:
This is a follow-up patch of https://github.com/KhronosGroup/WebGL/pull/1407. Aligned with native code.


kenrussell:
Thanks. Looks good.

PR #1411, opened 2016-01-04
dEQP: add format and internalformat checks for TexImage2D

qkmiao:
This aligns with test for TexImage3D. C++ code also misses such checks.


kenrussell:
Thanks. Looks good.

PR #1412, opened 2016-01-05
Added validation for vertexAttrib error conditions

toji:
Found that these cases weren't covered when attempting to refactor some validation within Chrome.


zhenyao:
lgtm

kenrussell:
Very nice. LGTM after the fact.

PR #1402, opened 2015-12-30
Update WebGL spec to align internalformat type of TexImage{2D|3D} with ES spec

qkmiao:
Change type of internalformat from GLenum to GLint.

PR #1413, opened 2016-01-05
Generate INVALID_VALUE for TexImage with invalid internalformat

qkmiao:
Align with spec change in https://github.com/KhronosGroup/WebGL/pull/1402.

### DIFF
--- a/sdk/tests/js/tests/tex-input-validation.js
+++ b/sdk/tests/js/tests/tex-input-validation.js
@@ -514,7 +514,7 @@ var testCases = [
     border: 0,
     format: gl.RGBA,
     type: gl.UNSIGNED_BYTE,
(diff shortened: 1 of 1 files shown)


zhenyao:
lgtm

zhenyao:
Forgot to mention: can you also merge this back to older version of conformance tests?  In older version, we will want to accept both INVALID_VALUE and INVALID_ENUM, so we don't regress already certified conformant implementations.

qkmiao:
@zhenyao it seems no TexImage3D tests in older conformance tests. I sent a pull request to fix TexImage2D at: https://github.com/KhronosGroup/WebGL/pull/1414.

zhenyao:
Thanks.

PR #1414, opened 2016-01-06
Fix ambiguous texture input test in older snapshot

qkmiao:
Backport https://github.com/KhronosGroup/WebGL/pull/1408 to older snapshot to align with spec change in https://github.com/KhronosGroup/WebGL/pull/1402.


zhenyao:
thanks. lgtm

PR #1415, opened 2016-01-06
Change numVertexAttribs to a valid value when testing invalid array lengths

bkotzz:
The second-last block of tests checks out-of-range vertexAttrib indexes, by passing in the MAX_VERTEX_ATTRIBS as the index.  The last block of tests is meant to check invalid array lengths, but it still uses numVertexAttribs as the max.  We only want the last block of tests to be testing one invalid parameter at a time, not two, so we change the numVertexAttribs to a valid value.


kenrussell:
Thanks for the fix. I'd have suggested just replacing numVertexAttribs in this block with numVertexAttribs - 1, but it seems OK as is. @toji , FYI.

PR #1417, opened 2016-01-07
Added test for using instance names with uniform blocks

toji:
Should be noted that compiling the new shader in Chrome crashes the GPU process with a call stack that bottoms out in gpu::gles2::Program::GetUniformBlockMembers<>(). Joy.

### DIFF
--- a/sdk/tests/conformance2/buffers/uniform-buffers.html
+++ b/sdk/tests/conformance2/buffers/uniform-buffers.html
@@ -82,6 +82,21 @@ void main()
     oColor = vec4(UBORed * UBOR, UBOGreen * UBOG, UBOBlue * UBOB, 1.0);
 }
 </script>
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/buffers/uniform-buffers.html
```diff
@@ -82,6 +82,27 @@
...
+    float Red;
+    float Green;
+    float Blue;
+} UBOA;
+
+uniform UBOData2 {
+    float Red;
+    float Green;
+    float Blue;
+} UBOB;
```
kenrussell: Is the second named uniform block here needed in order to provoke the crash? If not, let's consider removing it. If so, let's set it up via the API calls below.

kenrussell:
Good catch. Let's get this test in ASAP. Please address the above question and merge yourself. Thanks.

PR #1420, opened 2016-01-09
The global invariant pragma is illegal in fragment shaders.

kenrussell:
GLSL ES 3.00.4 section 4.6.1 page 59 states:
 "It is an error to use this pragma in a fragment shader."

Changed the tests which stated that the fragment shader should compile
successfully to instead state that it should fail to compile, and
change one test that stated that the link should succeed to say that
it should fail.

This causes Chrome to start failing one of these tests on NVIDIA's
drivers, but that's because the shader translator isn't properly
checking this, which needs to be fixed.


kenrussell:
@zhenyao in particular, please review. @Oletus would be good to have your feedback too.

kenrussell:
Sorry, that should be Page 53 above.

Oletus:
Why should this affect GLSL ES 1.00? If this is just because WebGL sometimes needs to convert GLSL ES 1.00 shaders to other shading languages, I think a mention in the WebGL 1.0 spec is required.

kenrussell:
@Oletus thanks for pointing that out. I didn't realize there was a difference in behavior between GLSL ES 1.00 and GLSL ES 3.00 in this area.

Please don't merge this pull request yet. It will have to be updated to have different expectations for these tests between WebGL 1.0 and 2.0.

zhenyao:
I think WebGL 1 should just forbid this pragma as well as WebGL 2.  We should update the spec and also the test.

kenrussell:
Unfortunately there's a dEQP test (pragma_fragment in sdk/tests/deqp/data/gles2/shaders/preprocessor.test) which requires that this pragma be ignored in ESSL 1.00 fragment shaders, at least for WebGL 1.0.

I don't know whether the best thing to do is change the test or relax the requirement in the WebGL 1.0 context (or ESSL 1.00 shaders in general). I'm modifying a bug fix I'm currently working on ( http://crbug.com/625363 ) to make this test pass.

kenrussell:
This was fixed better in #1904 .

PR #1421, opened 2016-01-11
fixes for gl-object-get-calls.js

howareyou322:


zhenyao:
GL_COLOR is not a valid arg when default framebuffer is bound.  GL_BACK is correct.

jdashg:
Right, COLOR is only used for ClearBuffer and InvalidateFramebuffer.

howareyou322:
Thanks for comment, I also found it from spec. Will create another pull request to enable stencil.

kenrussell:
Thanks. If you submit another pull request, please consider the bug I filed about it https://github.com/KhronosGroup/WebGL/issues/1426 . Ideally it should test both with and without a default stencil buffer, rather than only one of those cases.

howareyou322:
I'm looking into #1426 now.

PR #1422, opened 2016-01-11
Fixed verification in the transform feedback test.

kenrussell:
The verification of a given test case may need to be retried because
the query's result was not available yet. Fixed the logic in this area.

Thanks to Shuai Shao for pointing out these failures.


kenrussell:
I've tested this fix locally and it works. The failures started happening when queries' results were changed to become available asynchronously. The failures went unnoticed because the change to queries was made after this test was restructured to support their new semantics, and we're not running this test yet on our automated test systems.

Several tests still fail on my system but I believe they are all due to bugs in the driver. Will do a more comprehensive analysis after this is merged.

I'm merging this myself to make more of the tests start passing. @jasomobica @zhenyao @toji @jdashg FYI.

PR #1423, opened 2016-01-11
Add internalformat tests for CopyTexImage2D

qkmiao:


zhenyao:
not lgtm

deqp tests should track the c++ version unless there is a behavior difference between ES and WebGL.

You should add extra test cases in conformance2/textures

kenrussell:
I agree with Mo's assessment. Let's keep these in sync with the C++ versions of the tests as much as possible.

qkmiao:
Thanks @zhenyao and @kenrussell, I added the test in conformance2/textures/misc. Please refer https://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf page 139 and https://www.khronos.org/opengles/sdk/docs/man3/html/glCopyTexImage2D.xhtml for the valid internalformats for copyTexImage2D.

sdk/tests/conformance2/textures/misc/copy-texture-image.html
```diff
@@ -0,0 +1,174 @@
...
+    gl.bindTexture(gl.TEXTURE_2D, texture);
+
+    function testFormat (internalformat, srcTexFormatsTypes) {
+        var fbo = gl.createFramebuffer();
+        var srcTexture = gl.createTexture();
+        gl.bindTexture(gl.TEXTURE_2D, srcTexture);
+        gl.texImage2D(gl.TEXTURE_2D, 0, srcTexFormatsTypes.internalformat, 64, 64, 0, srcTexFormatsTypes.format, srcTexFormatsTypes.type, null);
+        gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+        gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, srcTexture, 0);
+
```
zhenyao: You should verify the fbo is complete because it is not guaranteed to be.

sdk/tests/conformance2/textures/misc/copy-texture-image.html
```diff
@@ -0,0 +1,174 @@
...
+    goodIntFormats.forEach(function(internalformat) {
+        var srcTexFormatsTypes = { internalformat: gl.RGBA32I, format: gl.RGBA_INTEGER, type: gl.INT };
+        testFormat(internalformat, srcTexFormatsTypes);
+    });
+    goodUnsingedIntFormats.forEach(function(internalformat) {
+        var srcTexFormatsTypes = { internalformat: gl.RGBA32UI, format: gl.RGBA_INTEGER, type: gl.UNSIGNED_INT };
+        testFormat(internalformat, srcTexFormatsTypes);
+    });
+    badFormats.forEach(function(internalformat) {
+        gl.copyTexImage2D(gl.TEXTURE_2D, 0, gl[internalformat], 0, 0, 64, 64, 0);
```
zhenyao: I think you should set up just like the good internalformat test for the bad ones.  Otherwise the error may be generated from somewhere else and you are not actually testing what you intend to test.

qkmiao:
@zhenyao fixed your comments, please take a look.

sdk/tests/conformance2/textures/misc/copy-texture-image.html
```diff
@@ -0,0 +1,225 @@
...
+    gl.bindTexture(gl.TEXTURE_2D, texture);
+
+    function testFormat(internalformat, srcTexFormatsTypes, fboAttachmentType, expected, msg) {
+        var fbo = gl.createFramebuffer();
+        var srcTexture = gl.createTexture();
+        gl.bindTexture(gl.TEXTURE_2D, srcTexture);
+        gl.texImage2D(gl.TEXTURE_2D, 0, srcTexFormatsTypes.internalformat, 64, 64, 0, srcTexFormatsTypes.format, srcTexFormatsTypes.type, null);
+        gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+        gl.framebufferTexture2D(gl.FRAMEBUFFER, fboAttachmentType, gl.TEXTURE_2D, srcTexture, 0);
+        checkFramebuffer([gl.FRAMEBUFFER_COMPLETE]);
```
zhenyao: Actually it is still spec compliant if the underlying driver does not support a combination of color/depth/stencil.

So what should be done here if to check if the framebuffer is complete.  If not, return early, but it's not a test failure.

zhenyao: Never mind.  That's ES2.  ES3 requires a bunch of combinations to be supported.

zhenyao:
lgtm

PR #1424, opened 2016-01-11
Add ImageBitmap to TexImageSource

junov:
This change adds ImageBitmap support to texImage2D and
texSubImage2D. The origin restrictions logic was updated to take
into account the origin-clean flag of an ImageBitmap's bitmap.
The terminology for the origin-clean flag was tweaked to match the
language used in the HTML spec: "the origin-clean flag of an
HTMLCanvasElement" becomes "the origin-clean flag of an
HTMLCanvasElement's bitmap".

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -28,7 +28,7 @@
     <!--end-logo-->
 
     <h1>WebGL Specification</h1>
-    <h2 class="no-toc">Editor's Draft 30 December 2015</h2>
+    <h2 class="no-toc">Editor's Draft 11 January 2016</h2>
     <dl>
         <dt>This version:
             <dd>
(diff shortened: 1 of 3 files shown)


zhenyao:
Looks good in general, but can you
1) change the "Editor's Draft" date in the beginning of the spec?
2) make the same change to WebGL 2 spec?

junov:
Noob question: WebGL 2 does not define TexImageSource. Is it meant to be interpreted as a diff over WebGL 1?

junov:
The WebGL 2 spec for texImage2D says: Only differences from <a href="../1.0/index.html#TEXIMAGE2D_HTML">texImage2D</a> in WebGL 1 are described here.

So I think only tex(Sub)Image3D need to be changed in the v2 spec. Does that sound right?

junov:
New patch

zhenyao:
lgtm

kenrussell:
LGTM after the fact. Thanks for the patch.

PR #1425, opened 2016-01-11
Add another ReadPixels test case

zhenyao:
i.e., if pack_row_length < width, the bytes after pack_row_length should
be pixel data until the row end of |width| pixels are reached.

### DIFF
--- a/sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
+++ b/sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
@@ -149,11 +149,10 @@ function runTestIteration(xoffset, yoffset, width, height, packParams, usePixelP
         if (size > 0) {
             gl.bufferData(gl.PIXEL_PACK_BUFFER, arrayWrongSize, gl.STATIC_DRAW);
             gl.readPixels(xoffset, yoffset, width, height, gl.RGBA, gl.UNSIGNED_BYTE, offset);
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/reading/read-pixels-pack-parameters.html
```diff
@@ -211,12 +210,25 @@
...
         samePixel(array, pos, refColor, row, "last");
 
-        // Check padding bytes or bytes beyond rowLength are unchanged.
+        // Check padding bytes are unchanged and bytes beyond rowLength set correctly.
         pos += bytesPerPixel;
         if (row + 1 < height) {
-            for (var ii = 0; ii < bytesPerRow - xSpan * bytesPerPixel; ++ii) {
+            if (packParams.rowLength > 0 && packParams.rowLength < width) {
+                var trailingBytes = Math.min((width - packParams.rowLength) * bytesPerPixel,
+                                             bytesPerRow - packParams.rowLength * bytesPerPixel);
```
kenrussell: Could you explain this math in a little more detail? I don't get it.

kenrussell: Per our offline discussion: please at least add a comment with some of the rationale behind this code. The rationale is quite complicated even if the code is fairly simple in the end.

kenrussell:
This is excellent work and I suspect some of the most comprehensive ReadPixels tests yet. LGTM once the above comment is addressed.

PR #1427, opened 2016-01-12
fix read-pixels-from-fbo-test.

mephisto41:
There are 2 problems in this test.
1. The readback value of SRGB8_ALPHA8 didn't consider sRGB conversion.
2. If type is integer or float, calling getParameter with IMPLEMENTATION_COLOR_READ_FORMAT/TYPE would cause gl error then test will fail.

### DIFF
--- a/sdk/tests/conformance2/reading/read-pixels-from-fbo-test.html
+++ b/sdk/tests/conformance2/reading/read-pixels-from-fbo-test.html
@@ -224,7 +224,7 @@ var textureTestCases = [
   {
     texInternalFormat: 'SRGB8_ALPHA8', texFormat: 'RGBA', texType: 'UNSIGNED_BYTE',
     readFormat: 'RGBA', readType: 'UNSIGNED_BYTE',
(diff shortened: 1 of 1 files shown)


zhenyao:
I don't find any spec text supporting 2.  If it's not in the spec, then it's a driver bug and we definitely should test it.

kenrussell:
Agree with Mo -- I don't see anything in the spec that would support change (2). It sounds to me like Firefox's WebGL 2.0 implementation is what's generating the error, not the OpenGL implementation. (This test currently passes on Chrome's WebGL 2.0 prototype.)

mephisto41:
Ah, I am confused with this spec. https://www.opengl.org/registry/specs/ARB/ES2_compatibility.txt

I'll check our implementation. Thanks.

kenrussell:
@mephisto41 please do open a new pull request for change (1) if you've double-checked and there is a bug in the test. Chrome currently passes the test but there could certainly still be a bug.

PR #1429, opened 2016-01-13
fix ambiguous test for blitFramebuffer

Richard-Yunchao:
This test is ambiguous:
1. the read buffer contains unsigned integer, while the draw buffer contain floating-point data. It should generate INVALID_OPERATION.
2. the linear interpolation will be applied to unsigned integer data. It should generate INVALID_OPERATION.

This case tend to test 2#, so 1# is not necessary. This small change make sure that the format of the read buffer is the same with the format of the draw buffer. 

PTAL. Thanks!

The native dEQP have the same problem too. See https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/gles3/functional/es3fNegativeBufferApiTests.cpp#1157.


zhenyao:
I agree, the original test isn't testing what it claims to test.

We probably should also fix the upstream test.

@kenrussell Do you know how we can send a patch to the upstream repo?

kenrussell:
@zhenyao @Richard-Yunchao https://source.android.com/source/life-of-a-patch.html describes the process for submitting patches to Android overall. I will also alert the dEQP maintainers about this problem. Thanks for catching it.

PR #1432, opened 2016-01-13
Fix bugs in conformance suites 1.0.1, 1.0.2, 1.0.3

zhenyao:
All of these are due to spec change after the snapshots were taken.


kenrussell:
Thanks for fixing these. LGTM

PR #1433, opened 2016-01-14
fix read-pixels-from-fbo-test. 

mephisto41:
In the chromium, IMPLEMENTATION_COLOR_READ_TYPE/FORMAT always return pre defined value if there is no extension GL_OES_read_format. (see https://chromium.googlesource.com/chromium/src/gpu/+/master/command_buffer/service/gles2_cmd_decoder.cc#5111)

In my machine, GL_OES_read_format doesn't exist. But I still able to query this parameter because I have ES2_compatibility extension or latest OpenGL(seems latest OpenGL supports this without extension). And the value I queried is different from the pre defined value created by chromium.

So I add more functions for handling the extra formats that I've found. I think those changes would break the test result of chromium because of the problem I mentioned in first paragraph.

### DIFF
--- a/sdk/tests/conformance2/reading/read-pixels-from-fbo-test.html
+++ b/sdk/tests/conformance2/reading/read-pixels-from-fbo-test.html
@@ -45,252 +45,345 @@ var wtu = WebGLTestUtils;
 var gl = wtu.create3DContext();
 gl.pixelStorei(gl.PACK_ALIGNMENT, 1);
 
-function r(buf, index, expectedColor, tol) {
-  var color = [buf[index]];
-  if (Math.abs(color[0] - expectedColor[0]) > tol[0]) {
-    testFailed("Expected red = " + expectedColor + ", was = " + color);
-    return false;
+function getChannelCount(format) {
+  switch (format) {
+    case gl.RED:
+    case gl.RED_INTEGER:
+    case gl.ALPHA:
+    case gl.LUMINANCE:
+      return 1;
+    case gl.RB:
+    case gl.RB_INTEGER:
+    case gl.LUMINANCE_ALPHA:
+      return 2;
+    case gl.RGB:
+    case gl.RGB_INTEGER:
+      return 3;
+    case gl.RGBA:
+    case gl.RGBA_INTEGER:
+      return 4;
+    default:
+      return 0;
   }
-  return true;
 }
 
-function rg(buf, index, expectedColor, tol) {
-  var color = [buf[index], buf[index + 1]];
-  if (Math.abs(color[0] - expectedColor[0]) > tol[0] ||
-      Math.abs(color[1] - expectedColor[1]) > tol[1]) {
-    testFailed("Expected rg = " + expectedColor + ", was = " + color);
-    return false;
+function getUnpackInfo(type) {
+  switch (type) {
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/reading/read-pixels-from-fbo-test.html
```diff
@@ -209,11 +226,13 @@
     texInternalFormat: 'RGB565', texFormat: 'RGB', texType: 'UNSIGNED_BYTE',
     readFormat: 'RGBA', readType: 'UNSIGNED_BYTE',
     clearColor: [0.5, 0.7, 0.2, 0], readBackColor: [127, 178, 51, 0], tol: [17, 9, 17, 0],
+    implReadBackColor: [0.5, 0.7, 0.2, 0], implTol: [0.02, 0.02, 0.02, 0.02],
```
zhenyao: I don't think this is robust.  It is very likely to break on drivers other than yours.  You assume the impl-dependent format/type is RGB and UNSIGNED_SHORT_5_6_5, thus the implReadBackColor.  But it is not guaranteed.  What if the format/type is RED/UNSIGNED_BYTE?  Then this implReadBackColor and implTol will just be wrong.

The conformance tests should work across GPUs/drivers.

mephisto41: Given this fact, should we remove this tests for IMPLEMENTATION_COLOR_READ_TYPE/FORMAT? Since the value is not consistent across drivers.

kenrussell: Absolutely not. There have been bugs in the handling of IMPLEMENTATION_COLOR_READ_FORMAT and TYPE in the past, and removing the test will allow bugs to creep in to implementations. The test should be generalized to handle all of the possible implementation color formats and types, not discarded.

zhenyao: I think what you can do is: write test cases for all these extra read format/type, and trigger the right test based on the returned impl-dependent format/type.  This requires some refactoring of the original design.

mephisto41:
I have updated the test which is generalized to handle all formats and types.

sdk/tests/conformance2/reading/read-pixels-from-fbo-test.html
```diff
@@ -369,9 +462,12 @@
     case gl.INT:
       return 'INT';
     case gl.UNSIGNED_SHORT_5_6_5:
-    case gl.UNSIGNED_SHORT_4_4_4_4:
+      return 'UNSIGNED_SHORT_5_6_5';
     case gl.UNSIGNED_SHORT_5_5_5_1:
+      return 'UNSIGNED_SHORT_5_5_5_1';
     case gl.UNSIGNED_INT_2_10_10_10_REV:
+      return 'UNSIGNED_INT_2_10_10_10_REV';
+    case gl.UNSIGNED_SHORT_4_4_4_4:
```
zhenyao: It seems you already handled UNSIGNED_SHORT_4_4_4_4, so why drop the support here?

zhenyao:
This is great improvement! LGTM with one suggestion (unsigned_short_4_4_4_4 support).  I am merging it now.  Please land another CL to fix the issue if you agree with my suggestion.

PR #1434, opened 2016-01-15
To clear stencil buffer, we should use clearBufferiv, instead of clearBufferfv

Richard-Yunchao:
To clear stencil buffer, we should use clearBufferiv, instead of clearBufferfv.  The original code will lead to INVALID_VALUE error. 

The native dEQP is correct. see https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/gles3/functional/es3fRasterizerDiscardTests.cpp#338


kenrussell:
Looks good. Verified that the native test is already doing this. Thanks for catching this.

Issue #1426, opened 2016-01-12
Bug in WebGL 2.0 getFramebufferAttachmentParameter test for STENCIL buffer


kenrussell:
The getFramebufferAttachmentParameter tests in sdk/tests/js/tests/gl-object-get-calls.js for WebGL 2.0, for the default framebuffer, assert that the FRAMEBUFFER_ATTACHMENT_STENCIL_SIZE is non-zero. This is dubious, since the stencil buffer is disabled by default in the context creation attributes. This should be fixed, either by asserting that the size is zero, or, ideally, testing two contexts, one with stencil enabled and one with it disabled.

See https://github.com/KhronosGroup/WebGL/pull/1421 which has some other problems and probably won't be committed.

PR #1435, opened 2016-01-15
test getFramebufferAttachmentParameter with stencil option enabled/di…

howareyou322:


howareyou322:
This patch is trying to fix issue #1426

sdk/tests/js/tests/gl-object-get-calls.js
```diff
@@ -107,186 +91,221 @@ testInvalidArgument(
...
+];
+
+for (var run = 0; run < testCases.length; ++run) {
+  debug("");
+  debug("Test getFramebufferAttachmentParameter with stencil " + testCases[run].contextStencil);
+
+  if (testCases[run].contextStencil) {
+    gl = wtu.create3DContext(null, {stencil: true});
+  } else {
+      gl = wtu.create3DContext(null, {stencil: false});
```
zhenyao: wrong indent

zhenyao:
Mostly looks good.  In theory you only need to test the default fbo with stencil on/off.  The client fbo isn't affected, but that's not a big deal. Merging now.  Please submit another CL to fix the minor indent issue.

sdk/tests/js/tests/gl-object-get-calls.js
howareyou322: From line 94, it is the beginning of getFramebufferAttachmentParameter test case. Therefore, there are no indent in line 94, just like other test cases.

zhenyao: I am talking about line 106

PR #1436, opened 2016-01-15
Set test time for skipped case to 0

gyagp:
The startTime of skipped case is always 0, thus the totalTime for
it would be wrongly a very large number.


gyagp:
To test this, just uncheck any test and run it, a very large number of test time would show up. For example, uncheck the first case in 1.0.4 and run it, something similar to the following would show up: conformance/attribs/gl-bindAttribLocation-aliasing.html (0 of 1 passed in 1452854701594.0 ms ).

zhenyao:
Looks good.

PR #1437, opened 2016-01-15
Add successful and total case number to folder

gyagp:
For better statistics, sometimes we need to know the number of
successful and total cases in each folder.
This PR also removes an unnecessary space for result of each page.


zhenyao:
This is great improvements!

PR #1438, opened 2016-01-15
Add detection for createImageBitmap and ImageBitmap

xidachen:
We need to add a detection in the conformance test for ImageBitmap to detect whether or not createImageBitmap and ImageBitmap is available. If not, then we should quit the test right away. Also, we add the 00_test_list.txt for the ImageBitmap test so that they can be picked up by the bot.


xidachen:
@kenrussell  Please review.

kenrussell:
Thanks. LGTM.

PR #1441, opened 2016-01-16
Temporary remove the ImageBitmap tests

xidachen:
It appears that the ImageBitmap tests are failing, I didn't catch it in the early pull request. We should disable them for them, and re-enable them once I fix the tests.

@zhenyao: please review.


xidachen:
Please do not merge, I found the solution to it.

PR #1442, opened 2016-01-16
Update ImageBitmap conformance test

xidachen:
Hi, there was a bug in the code which was causing failure. I fixed the bug and all the ImageBitmap passes locally.

@zhenyao: please review. Thank you.

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap.js
@@ -88,8 +88,9 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
         gl.texParameteri(bindingTarget, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
         gl.texParameteri(bindingTarget, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
         // Set up pixel store parameters
-        gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, flipY);
-        gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, false);
+        // In the case of ImageBitmap, we ignore these parameters.
+        // gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, flipY);
+        // gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, true);
         var targets = [gl.TEXTURE_2D];
         if (bindingTarget == gl.TEXTURE_CUBE_MAP) {
             targets = [gl.TEXTURE_CUBE_MAP_POSITIVE_X,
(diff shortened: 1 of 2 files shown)


sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap.js
```diff
@@ -89,7 +89,7 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
         gl.texParameteri(bindingTarget, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
         // Set up pixel store parameters
         gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, flipY);
-        gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, premultiplyAlpha);
+        gl.pixelStorei(gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL, false);
```
zhenyao: I disagree with such fix.  It silently changes the test parameters.

Instead, we basically should change the caller side and make sure we don't call runOneIteration(premultuplyAlpha==true), and add an assert in runOneIteration.

Or simply remove the premultiplyAlpha arg and we don't even need to call pixelStorei as its default is false.

xidachen: After taking a closer look, I think there is another bug somewhere and I don't think it would be a quick fix. Is it possible to merge the "temporary remove the imagebitmap tests" only? Or should I create a separate pull request? I will re-enable these tests once the bug is fixed. Thank you.

zhenyao:
There is no need to land anything temporary.  We don't expect every test to pass, so we have a mechanism in Chrome to set test expectations.  Just land the appropriate fix once you have it.

kenrussell:
@xidachen it occurs to me that the problem you're running into may be that we want to pick up image decoding properties like "premultiplyAlpha" or "apply browser-default colorspace conversions" from the ImageBitmap constructor options, rather than the WebGL pixelStorei parameters. We've had this conversation before in the WebGL WG. Is this the underlying issue here?

xidachen:
@kenrussell I actually don't know what the problem is here. At the moment, the created ImageBitmap always has premultiplyAlpha=true, just like a HTMLCanvasElement. But I tried to set gl.UNPACK_PREMULTIPLY_ALPHA_WEBGL to be false or true, and both cases failed the test. I would have to look into it.

kenrussell:
@xidachen that's not that surprising since options like premultiplyAlpha=false would have to be applied during the decoding of the image. That's what happens when an HTMLImageElement is uploaded to a WebGL texture, and this synchronous re-decoding is exactly what we're trying to get rid of with ImageBitmap and its constructor options.

I think that the WebGL pixelStorei settings like UNPACK_FLIP_Y_WEBGL, UNPACK_PREMULTIPLY_ALPHA_WEBGL and UNPACK_COLORSPACE_CONVERSION_WEBGL should be ignored when uploading an ImageBitmap. Instead, those should be specified via ImageBitmap constructor options. We should update https://www.khronos.org/registry/webgl/specs/latest/1.0/ to clarify this.

Does that help determine a direction for both these tests and Chromium's implementation?

PR #1443, opened 2016-01-18
fix a bug about uniform1fv

Richard-Yunchao:
This small change fixed a bug about uniform1fv. The data type of the original code is not correct, we should use an array. 

Attach the native code FYI: https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/gles3/functional/es3fFboTestUtil.cpp#821


zhenyao:
Looks good.

PR #1444, opened 2016-01-18
remove ambiguous drawing test cases against invalid program

Richard-Yunchao:
In GLES, when call drawing APIs (drawElements, drawArrays and instanced drawing) with an invalid program (for example, call useProgram(0); ), the result is undefined, but it will not generate error. 

However, in WebGL, it will generate INVALID_OPERATION in the same situation. So these test cases are ambiguous. This pr removed these ambiguous test cases against invalid program.

PTAL. Thanks!


zhenyao:
I agree not much value in these tests.

PR #1446, opened 2016-01-20
Add getActiveUniformBlockParameter to gl.object-get-call.js

renchenglei:


kenrussell:
Thanks for this contribution; it is clear. I must ask though: did you check sdk/tests/deqp/functional/gles3/es3fShaderStateQueryTests.js and sdk/tests/deqp/modules/shared/glsUniformBlockCase.js and see whether these queries were already covered?

renchenglei:
@kenrussell , I have looked into these two files. They didn't cover all the features. This one is more comprehensive, just like other getter APIs.

PR #1447, opened 2016-01-21
Simplify WebGLTestUtils shader program setup functionality

Oletus:
Instead of exposing a lot of rarely used functions that simply load a
predefined shader from source, expose WTU collection of shader source as
read-only string properties on the WTU API. These are simple to use with
loadShader() or setupProgram().

Also rely on setupProgram() more in WTU functions that load a shader
program based on predefined shader sources to reduce complexity.


kenrussell:
Thanks, this is a nice cleanup.

PR #1448, opened 2016-01-22
Conformance test for ImageBitmap

xidachen:
This is the conformance tests for texImage2D and texSubImage2D(ImageBitmap). In the case of ImageBitmap, the property of premultiplyAlpha and flipY is already encoded during its constructor time. So in the conformance test, we should ignore setting these in the gl.pixelStorei.

At this moment, I comment out the code that sets premultiplyAlpha and flipY. I am implementing the feature of setting premultiplyAlpha = false and flipY = true in the constructor of ImageBitmap, so that later on we can test these settings.

@zhenyao @toji Please review.


zhenyao:
Sorry I didn't raise this before (didn't realize it until now):

But you should modify the sdk/tests/py/tex_image_test_generator.py to auto-generate all the html files and also the 00_test_list, not hand-written.

Otherwise when you move to ES3, it's going to be a lot of formats.

xidachen:
@kenrussell @zhenyao @toji Please review my last commit that says "both 2d and 3d tests works". In there, I changed the auto generator python file, and those new .html files are generated by that script. One thing I want to mention is that in the tex_image_test_generator.py it says "Copyright (c) 2015 The Khronos Group Inc.", I tried to change it to 2016 but then all the existing files changed. So I am not sure whether I should change that or not.

xidachen:
Sorry that I forgot to commit the 3d test js file. Locally all tests passed.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap.js
```diff
@@ -170,14 +165,14 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
 
     function runTestOnBindingTarget(bindingTarget, program) {
         var cases = [
-            { sub: false, flipY: true, premultiplyAlpha: false },
+            // { sub: false, flipY: true, premultiplyAlpha: false },
```
zhenyao: You can just remove these lines instead of commenting out them

sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap.js
```diff
@@ -0,0 +1,152 @@
...
+                                      0, 255, 0, 0]),
+                                      2, 2);
+
+        createImageBitmap(imageData).then(imageBitmap => {
+            bitmap = imageBitmap;
+            runTest();
+        });
+    }
+
+    function runOneIteration(useTexSubImage2D, flipY, premultiplyAlpha, bindingTarget, program)
```
zhenyao: As I mentioned before, you can just get rid of the premultiplyAlpha param in this function.

zhenyao:
A few minor issues, but let's merge now. Please submit another patch to address them.

PR #1449, opened 2016-01-26
Fix wrong texture 2d array view

qkmiao:
deqp/functional/gles3/texturefiltering09.html complains
"src.sample is not a function".
glsTextureTestUtil.verifyTexture2DArrayResult requires
tcuTexture.Texture2DArrayView object for the second argument. This patch
fixes the bug which is similar to https://github.com/KhronosGroup/WebGL/pull/1340.


zhenyao:
lgtm

I was just about to dig into this issue.  Thanks for the fix.

PR #1450, opened 2016-01-26
Add sample function for Texture2DArray to run texturespecification25.html

qkmiao:
"Intentionally empty. Call method from child class instead" error was
reported when running deqp/functional/gles3/texturespecification25.html.
Add sample() for sglrReferenceContext.Texture2DArray. This also aligns
with C++ code: https://android.googlesource.com/platform/external/deqp/+/master/framework/opengl/simplereference/sglrReferenceContext.cpp#5133.


zhenyao:
Looks good.  Thanks for the fix.

PR #1451, opened 2016-01-26
Remove flipY and premultiplyAlpha from the test

xidachen:
@zhenyao @toji @kenrussell Please review, both parameters are now removed from the test

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap.js
@@ -73,10 +73,9 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
         });
     }
 
-    function runOneIteration(useTexSubImage2D, flipY, premultiplyAlpha, bindingTarget, program)
+    function runOneIteration(useTexSubImage2D, bindingTarget, program)
     {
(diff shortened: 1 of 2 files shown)


sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap.js
```diff
@@ -139,12 +130,11 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
 
     function runTestOnBindingTarget(bindingTarget, program) {
         var cases = [
-            { flipY: false, premultiplyAlpha: false },
+            { },
```
zhenyao: You can get rid of this cases.

sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap.js
```diff
@@ -73,10 +73,9 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
         });
     }
 
-    function runOneIteration(useTexSubImage2D, flipY, premultiplyAlpha, bindingTarget, program)
+    function runOneIteration(useTexSubImage2D, bindingTarget, program)
```
zhenyao: You can also get rid of useTexSubImage2D as it is always true and this arg is not used in the function at all.

xidachen:
Thanks for catching it. The cases parameter has been removed.

zhenyao:
lgtm

PR #1453, opened 2016-01-27
Use correct uniform2fv for vec2 uniform variable

qkmiao:
This fixes deqp/functional/gles3/shaderstruct.html which failed before
because uniform1fv was used for a vec2 uniform variable.


zhenyao:
Thanks for fixing the bug!

PR #1454, opened 2016-01-27
Transpose matrix for uniformMatrix*fv manually

qkmiao:
Transpose for uniformMatrix*fv is not supported in WebGL. Do it manully
for deqp/functional/gles3/shadermatrix.html. This fixes part of
shadermatrix.html tests.

### DIFF
--- a/sdk/tests/deqp/framework/delibs/debase/deMath.js
+++ b/sdk/tests/deqp/framework/delibs/debase/deMath.js
@@ -130,6 +130,22 @@ deMath.swizzle = function(a, indices) {
 };
 
 /**
+ *  @param {goog.NumberArray} a Source array
+ *  @param {number} row
+ *  @param {number} col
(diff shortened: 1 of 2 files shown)


zhenyao:
Sorry, but I don't remember we enforce transpose==true in WebGL, and I couldn't find any text in the spec about this.  If I miss something, please point out.

zhenyao:
OK, ES2 does forbit transpose == true, but not ES3, therefore not WebGL2.

WebGL2 should allow transpose == true.  See ES spec 3.0.4 section 2.12: "If transpose is FALSE , the matrix is specified in column major order, otherwise in row major order."

We could probably mention it in the difference between WebGL 1/2 section in the WebGL 2 spec.

qkmiao:
Yes, we should allow transpose == true for WebGL 2.

kenrussell:
Let's close this pull request so it isn't accidentally merged.

PR #1455, opened 2016-01-27
Fix uniform_value_matrix test in shaderstatequery.html

qkmiao:
Use slice to replace wrong splice function. splice will remove elements
from array while slice returns a copy of part of an array.


zhenyao:
splice -> slice looks good to me.

Not the removing the transpose part.

qkmiao:
Updated the pull request. Thanks!

PR #1456, opened 2016-01-28
Add float texture test for generateMipmap

qkmiao:
When extensions EXT_color_buffer_float and OES_texture_float_linear are
enabled, generateMipmap should success for float type texture.

### DIFF
--- a/sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
+++ b/sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
@@ -160,6 +160,11 @@ wtu.glErrorShouldBe(gl, gl.NO_ERROR, "Should be no errors from setup.");
   gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA8UI, 0, 0, 0, gl.RGBA_INTEGER, gl.UNSIGNED_BYTE, null);
   gl.generateMipmap(gl.TEXTURE_2D);
   wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "generateMipmap should fail for non-texture-filterable format");
+  if (gl.getExtension('EXT_color_buffer_float') && gl.getExtension('OES_texture_float_linear')) {
+      gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA32F, 0, 0, 0, gl.RGBA, gl.FLOAT, null);
+      gl.generateMipmap(gl.TEXTURE_2D);
+      wtu.glErrorShouldBe(gl, gl.NO_ERROR, "generateMipmap should succeed");
+  }
   gl.deleteTexture(tex);
 })();
 


sdk/tests/deqp/functional/gles3/es3fNegativeTextureApiTests.js
```diff
@@ -1147,9 +1147,11 @@ goog.scope(function() {
             gl.texImage2D(gl.TEXTURE_2D, 0, gl.R8I, 0, 0, 0, gl.RED_INTEGER, gl.BYTE, null);
             gl.generateMipmap(gl.TEXTURE_2D);
             this.expectError(gl.INVALID_OPERATION);
-            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA32F, 0, 0, 0, gl.RGBA, gl.FLOAT, null);
-            gl.generateMipmap(gl.TEXTURE_2D);
-            this.expectError(gl.INVALID_OPERATION);
+            if (!gl.getExtension('EXT_color_buffer_float') || !gl.getExtension('OES_texture_float_linear')) {
```
jasomobica: Hmm, my understanding is that extensions are not in effect unless explicitly requested with a getExtension().
So, if none of the extensions you listed has been requested, then generateMipmap() should still fail.

kenrussell: That's correct. The original code should have always generated an INVALID_OPERATION.

A new clause here could read something like:

if (gl.getExtension('EXT_color_buffer_float') && gl.getExtension('OES_texture_float_linear')) {
  gl.generateMipmap(gl.TEXTURE_2D);
  this.expectError(gl.NO_ERROR);
}

qkmiao:
Thanks for review. I updated this pull request.

sdk/tests/deqp/functional/gles3/es3fNegativeTextureApiTests.js
```diff
@@ -1150,6 +1150,11 @@ goog.scope(function() {
             gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA32F, 0, 0, 0, gl.RGBA, gl.FLOAT, null);
             gl.generateMipmap(gl.TEXTURE_2D);
             this.expectError(gl.INVALID_OPERATION);
+            if (gl.getExtension('EXT_color_buffer_float') && gl.getExtension('OES_texture_float_linear')) {
```
zhenyao: This is a negative test.  It's better if you add the new test case to a new float texture test with extensions.

qkmiao:
What about tex-mipmap-levels.html?

kenrussell:
This LGTM though it's a little strange to generate mipmaps for a 0x0 texture. Per reading of the EXT_color_buffer_float and OES_texture_float_linear specs I think this new test is correct. @zhenyao , what do you think?

zhenyao:
LGTM too.  Agree that testing 0x0 is not the best isolated testing.

Also, it's worth testing that enabling one of the extensions, generateMipmap still fails.  It has to have both extensions to succeed.

qkmiao:
@zhenyao @kenrussell  thanks for review. Could you also review https://codereview.chromium.org/1527763003/? That CL makes this test pass.

PR #1460, opened 2016-02-02
Check level size of generated mipmap texture

qkmiao:
In ES 3.0.4 spec page 158, it defines how to generate texture mipmap.
For TEXTURE_2D_ARRAY target, depth size of generated image for non-base
levels should be same as base level. This is different with TEXTURE_3D.

This patch also fixes generate mipmaps for a 0x0 texture.


zhenyao:
The improvement is great.

However, we did lose 0 size texture coverage. Checking the spec, 0-size looks legal to me.  Can we add back these coverage?

zhenyao:
Merging this for now.

PR #1461, opened 2016-02-02
Stricter spec for uniform block matching

Oletus:
This change specifies what this dEQP test already requires:
dEQP-GLES3.shaders.linkage.uniform.block.layout_qualifier_mismatch_3

It makes sure shaders stay portable across GLES 3.0 implementations.


zhenyao:
Looks good, but please update the edit date.

Oletus:
Updated pull request with the edit date.

kenrussell:
LGTM, but Mo should re-review.

PR #1464, opened 2016-02-04
3D targt should be INVALID_ENUM for 2d texture apis

qkmiao:
This patch valids TEXTURE_3D is invalid enum for 2d texture apis.
deqp/functional/gles3/negativetextureapi.html already covers validation
of TEXTURE_2D target for 3d texture apis.


zhenyao:
lgtm

qkmiao:
Thanks @zhenyao ! I have a local blink fix for these tests. Since you are doing texture validation clean-up in blink, should I discard the blink side fix and let command buffer do the validation when the cleanup is ready.

zhenyao:
Sounds good.

PR #1465, opened 2016-02-05
Fix bugs for pixelbufferobject.html in deqp

Richard-Yunchao:
This change will fix failures in pixelbufferobject.html in deqp test.

PTAL. Thanks!

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
+++ b/sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
@@ -176,7 +176,7 @@ var tcuImageCompare = framework.common.tcuImageCompare;
      */
     es3fPixelBufferObjectTest.ReadPixelsTest.prototype.renderTriangle = function(a, b, c) {
 
-        var positions = [];
+        var positions = new Float32Array(36);
 
         positions[0] = a[0];
         positions[1] = a[1];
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
```diff
@@ -214,8 +214,15 @@ var tcuImageCompare = framework.common.tcuImageCompare;
         gl.enableVertexAttribArray(coordLoc);
         assertMsgOptions(gl.getError() === gl.NO_ERROR, 'enableVertexAttribArray failed ', false, true);
 
-        gl.vertexAttribPointer(coordLoc, 3, gl.FLOAT, false, 0, positions[0]);
-        gl.vertexAttribPointer(colorLoc, 4, gl.FLOAT, false, 0, colors[0]);
```
Richard-Yunchao: It is not correct to feed data to vertex attribs from client memory in WebGL, we should use ARRAY_BUFFER. In native GLES, it is correct in this way.

sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
```diff
@@ -394,7 +401,7 @@ var tcuImageCompare = framework.common.tcuImageCompare;
 
         gl.bindBuffer(gl.PIXEL_PACK_BUFFER, pixelBuffer);
         gl.bufferData(gl.PIXEL_PACK_BUFFER, readReference.getLevel(0).getDataSize(), gl.STREAM_READ);
-        gl.readPixels(0, 0, width, height, readPixelsFormat, readPixelsType, null);
+        gl.readPixels(0, 0, width, height, readPixelsFormat, readPixelsType, 0);
```
Richard-Yunchao: The original code will call into this one: 
void readPixels(GLint x, GLint y, GLsizei width, GLsizei height, GLenum format, GLenum type, ArrayBufferView? pixels)
In this API, pixel buffer should not be bound. 

After the revision, it will call into the correct one: 
void readPixels(GLint x, GLint y, GLsizei width, GLsizei height, GLenum format, GLenum type, GLintptr offset);

zhenyao:
Great!  Thanks for fixing these bugs!

PR #1466, opened 2016-02-05
Test compressed texture rendering with both NEAREST and LINEAR filter…

kenrussell:
…ing.

This exposed a bug in Chrome after some recent changes to the texture validation.


zhenyao:
lgtm

PR #1467, opened 2016-02-08
Format GL_ALPHA generates an invalid enum for readPixels

qkmiao:
Read pixels in GL_ALPHA format should generate INVALID_ENUM error in ES3
semantic. That format is deprecated in ES3.
https://codereview.chromium.org/1658703003 fixed this issue in chrome.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fNegativeBufferApiTests.js
@@ -160,15 +160,15 @@ goog.scope(function() {
                 gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_SHORT_5_6_5, ushortData);
                 this.expectError(gl.INVALID_OPERATION);
                 gl.readPixels(0, 0, 1, 1, gl.ALPHA, gl.UNSIGNED_SHORT_5_6_5, ushortData);
-                this.expectError(gl.INVALID_OPERATION);
+                this.expectError(gl.INVALID_ENUM);
                 gl.readPixels(0, 0, 1, 1, gl.RGB, gl.UNSIGNED_SHORT_4_4_4_4, ushortData);
                 this.expectError(gl.INVALID_OPERATION);
                 gl.readPixels(0, 0, 1, 1, gl.ALPHA, gl.UNSIGNED_SHORT_4_4_4_4, ushortData);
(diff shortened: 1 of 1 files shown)


zhenyao:
I'd like a second opinion on this since its a deviation from the C++ test cases.

The spec is clear that ALPHA is never an accepted format for ReadPixels.  In that sense, by tradition this should be an INVALID_ENUM error.  The spec doesn't say anything about this error, but that should be the default behavior.  There are other functions in the spec that INVALID_ENUM isn't specified but expected by default.

The c++ test expects an INVALID_OPERATION, due to the pair isn't an accepted combination.

@kenrussell What do you think?

kenrussell:
Please point to the spec text that deprecates readbacks with GL_ALPHA. ReadPixels (ES 3.0.4, section 4.3, p. 193) refers to section 3.7.2 (p.110) which refers to Tables 3.2 and 3.3. Table 3.3 continues to have ALPHA / UNSIGNED_BYTE as an entry.

Regardless, I do not think the test should deviate from the C++ version. Let's instead fix Chrome so it produces the error the test expects.

zhenyao:
The spec (ES 3.0.4, section 4.3, p. 193) actually only refers to Table 3.2.  Table 3.3 isn't included. That's why we rejected the ALPHA in Chrome.

However, I do agree that we should minimize the deviation from C++ version unless it's very necessary.  Let's change the Chrome implementation instead.

Qiankun, does this sound good to you?

msc-:
> On Feb 9, 2016, at 3:09 PM, zhenyao <notifications@github.com> wrote:
> 
> The spec (ES 3.0.4, section 4.3, p. 193) actually only refers to Table 3.2. Table 3.3 isn't included. That's why we rejected the ALPHA in Chrome.
> 
> However, I do agree that we should minimize the deviation from C++ version unless it's very necessary. Let's change the Chrome implementation instead.
> 
> 

OpenGL ES 3.0 is backwards compatible with OpenGL ES 2 therefore all deprecated formats remain usable in ES 3.0.

Regards

    -Mark

qkmiao:
Thanks you all. I agree that we should change chrome to make it accept GL_ALPHA. I proposed a CL: https://codereview.chromium.org/1673363004. Please help to review.

qkmiao:
https://codereview.chromium.org/1673363004 landed. Close this pull request.

PR #1468, opened 2016-02-08
Fix non-color-renderable texture attachment

qkmiao:
Framebuffer attached a non-color-renderable texture is incomplete.
ClearBufferfv on incomplete framebuffer will fail. Check and enable
EXT_color_buffer_float extension before clearing a framebuffer attached
a float texture which isn't color-renderable.


zhenyao:
lgtm

PR #1469, opened 2016-02-10
Fix typos

pra85:
Remove extra `the`
`simultanenously` → `simultaneously`


kenrussell:
Thanks for the fixes. I'll fix the duplicate "the" in the 1.0 spec.

PR #1471, opened 2016-02-10
Adding error checking after instanceof calls to clear any set error flags

bkotzz:
When we load the shader source from setupProgram, we check for GL errors, assuming that they would've been set from the preceding function call.  However, a GL error that is set during one of the instanceof tests won't be caught until setting up the program, and would be incorrectly attributed to a problem with loading the shaders. To fix this, we check for errors after each GL call, if only for the purpose of clearing the flag.

### DIFF
--- a/sdk/tests/js/tests/instanceof-test.js
+++ b/sdk/tests/js/tests/instanceof-test.js
@@ -31,6 +31,13 @@ description(document.title);
 debug("Tests that instanceof works on WebGL objects.");
 debug("");
 
+function checkGLError(message) {
+  var error = gl.getError();
+  if (error != gl.NO_ERROR) {
+    wtu.error("Error: " + message + " caused " + wtu.glEnumToString(gl, error));
+  }
+}
+
 var gl = wtu.create3DContext("canvas");
 if (contextVersion === 1) {
   shouldBeTrue('gl instanceof WebGLRenderingContext');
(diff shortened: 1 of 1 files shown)


kenrussell:
This seems OK, but it's a very specialized fix. On what browser, OS and GPU combination did you see this sort of failure? If Chrome, please provide about:gpu (copy/paste is fine); if Firefox, please provide about:support. Thanks.

bkotzz:
I'm not sure how this is browser specific? The failure that I am seeing is just that a GL call is raising an error, and the error flag isn't getting checked/reset, so it's causing problems down the line.

I'm working with this on Firefox, where an error is set because Samplers haven't yet been implemented.  That error causes several unrelated tests to fail later on, because it never gets cleared.  I don't see what this has to do with a specific browser, as a GL call failing is something that can happen across the board.

zhenyao:
LGTM, but Ken should do the final review.

kenrussell:
The problem you're seeing is very much browser specific. As you point out, it's only Firefox's WebGL 2.0 implementation which is breaking in this manner. If it were complete, then this test wouldn't be failing in this way.

Again, this change looks fine; merging it now. My overall point is that it is very difficult to predict how these tests will fail under an incomplete WebGL implementation, and we don't want to make tons of changes to tons of tests in order to make the tests fail gracefully in this situation.

bkotzz:
I understand your point (and its a good one). I only proposed the fix because I wasted time debugging unexpected failures, only to find out that they were indirectly caused by an expected failure, and I just didn’t want anyone else to go through the same thing.

PR #1477, opened 2016-02-15
Fix constant values in shader

qkmiao:
Some constant values are not aligned with C++ version. These wrong
values make related cases in es3fShaderOperatorTests.html fail:
(1) float number with all zero decimal part is handled as an integer in js. It caused integer overflow errors in shader
(2) wrong values cause image comparison failure.
(3) wrong input and output type cause image comparison failure.

C++ version can be found at: https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fShaderOperatorTests.cpp.


zhenyao:
Wow.  Thanks.

I only randomly double-checked half of the changes. They are consistent with the original c++ tests.

kenrussell:
Thank you @qkmiao for tracking down these discrepancies between the original C++ tests and the JavaScript port. Great work. LGTM after the fact.

PR #1479, opened 2016-02-16
Select correct z element for vec4

qkmiao:
C++ code can be found at: https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fShaderOperatorTests.cpp#996.

Index of z element should be 2 other than 3.


zhenyao:
lgtm
thx for the fix

PR #1480, opened 2016-02-16
Use same mod algorithm as C++ version

qkmiao:
GLSLangSpec.3.30.6 defines mod(x, y) = x - y * floor(x/y), see page.84 https://www.opengl.org/registry/doc/GLSLangSpec.3.30.6.clean.pdf.  
C++ dEQP also use this define, see: https://android.googlesource.com/platform/external/deqp/+/master/framework/common/tcuVectorUtil.hpp#58.   The mod in line 57 is not used in mod cases in deqp/functional/gles3/shaderoperator.html.

In JS -1.1 % 1 = -0.1, while -1.1 - 1 * Math.floor(-1.1/1) = 0.9.

### DIFF
--- a/sdk/tests/deqp/framework/delibs/debase/deMath.js
+++ b/sdk/tests/deqp/framework/delibs/debase/deMath.js
@@ -195,6 +195,15 @@ deMath.divideScale = function(a, b) {
 };
 
 /**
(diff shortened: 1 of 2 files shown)


zhenyao:
lgtm

PR #1481, opened 2016-02-17
Resolve the big-litte endian issue and read pixel out of range issue.

xhcao:


kenrussell:
Thank you for these fixes. They look correct, so merging now.

What is the effect of these fixes? Do all of the tests pass? Are there still failures?

xhcao:
@kenrussell  Parts of tests of uniformbuffers.html will be passed. Some tests still fail, but the cause is related with parsing uniform block name wrongly on command buffer side or angle side.

PR #1482, opened 2016-02-17
Add test to verify texImage3D and texStorage3D won't leak memories.

zhenyao:
I just realized I should also test texStorage2D.  Will do it in a followup CL.


kenrussell:
Looks good.

PR #1483, opened 2016-02-17
Fix missing {} for if-else

qkmiao:
This makes deqp/framework/opengl/simplereference/referencecontext.html
pass.


zhenyao:
Looks good to me.

kenrussell:
Thanks for finding and fixing this. I should have enforced usage of { } during the port. There are some very hard to read passages of ported code.

PR #1485, opened 2016-02-18
Use correct integer type

qkmiao:
This aligns with int type for result in C++ code. Some constant values
can be negative, e.g. MIN_PROGRAM_TEXEL_OFFSET in
deqp/functional/gles3/shaderbuiltinvar.html.

C++ code: https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fShaderBuiltinVarTests.cpp#133


zhenyao:
lgtm

PR #1486, opened 2016-02-18
Correct reference number of vector

qkmiao:
The number of vectors is number of components divided by four.
getVectorsFromComps checks "value%4 === 0", but it forgets doing the
division.

C++ code for reference: https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fShaderBuiltinVarTests.cpp#71


zhenyao:
lgtm

PR #1488, opened 2016-02-19
fix bug in shaderstatequery when query vertex attribute offset

Richard-Yunchao:
The test vertex_attrib_offset in shaderstatequery.html need to get the gl.VERTEX_ATTRIB_ARRAY_POINTER info, we should use getVertexAttribOffset(), instead of getVertexAttrib(). Otherwise, the pname is invalid. 

This change fixed this bug.


kenrussell:
Thanks. Good catch.

PR #1489, opened 2016-02-19
Remove unsupported attribute aliasing tests

qkmiao:
Vertex attribute aliasing causes link error (WebGl 2.0 spec section 5.17
Vertex Attribute Aliasing). So remove related tests. Also align shader
source code generation with C++ code. 
deqp/functional/gles3/attriblocation.html passed with this patch.

C++ code: https://android.googlesource.com/platform/external/deqp/+/master/modules/glshared/glsAttributeLocationTests.cpp#254

### DIFF
--- a/sdk/tests/deqp/framework/opengl/gluShaderUtil.js
+++ b/sdk/tests/deqp/framework/opengl/gluShaderUtil.js
@@ -37,6 +37,15 @@ gluShaderUtil.GLSLVersion = {
 };
 
 /**
+ * gluShaderUtil.glslVersionUsesInOutQualifiers
+ * @param {gluShaderUtil.GLSLVersion} version
+ * @return {boolean}
+ */
+gluShaderUtil.glslVersionUsesInOutQualifiers = function(version) {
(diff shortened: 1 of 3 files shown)


sdk/tests/deqp/framework/opengl/gluShaderUtil.js
```diff
@@ -37,6 +37,15 @@ gluShaderUtil.GLSLVersion = {
 };
 
 /**
+ * gluShaderUtil.glslVersionUsesInOutQualifiers
+ * @param {gluShaderUtil.GLSLVersion} version
+ * @return {boolean}
+ */
+gluShaderUtil.glslVersionUsesInOutQualifiers = function(version) {
+    return version == gluShaderUtil.GLSLVersion.V300_ES;
```
kenrussell: It might be more future-proof to say "version != gluShaderUtil.GLSLVersion.V100_ES".

kenrussell:
Thank you for discovering and removing these invalid tests, and for finding the bugs in the WebGL 1.0 versions of the same tests. Merging now.

PR #1490, opened 2016-02-19
Vertex attribute stride should be aligned with the size of vertex attribute type in WebGL 2.

Richard-Yunchao:
This feature differs from that of native gles3. see the section <Buffer Offset and Stride Requirements> in webgl spec: https://www.khronos.org/registry/webgl/specs/latest/1.0/index.html#6.4 

Considering that the stride alignment is not covered in conformance test. I add the expected GL_ERROR into this case to test this feature, instead of deleting the cases that will generate INVALID_OPERATION error. 

PTAL. Thanks!

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fShaderStateQueryTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fShaderStateQueryTests.js
@@ -1075,12 +1075,12 @@ setParentClass(es3fShaderStateQueryTests.VertexAttributeStrideCase, es3fApiCase.
 
 es3fShaderStateQueryTests.VertexAttributeStrideCase.prototype.test = function() {
     var pointers = [
-        [1, gl.FLOAT, 0, 0],
-        [1, gl.FLOAT, 1, 0],
-        [1, gl.FLOAT, 4, 0],
-        [1, gl.HALF_FLOAT, 0, 0],
-        [1, gl.HALF_FLOAT, 1, 0],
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fShaderStateQueryTests.js
```diff
@@ -1089,25 +1089,45 @@ es3fShaderStateQueryTests.VertexAttributeStrideCase.prototype.test = function()
     // Test with default VAO
 
     for (var ndx = 0; ndx < pointers.length; ++ndx) {
-        gl.vertexAttribPointer(0, pointers[ndx][0], pointers[ndx][1], false, pointers[ndx][2], pointers[ndx][3]);
-        this.check(glsStateQuery.verifyVertexAttrib(0, gl.VERTEX_ATTRIB_ARRAY_STRIDE, pointers[ndx][2]));
+        try {
+            gl.vertexAttribPointer(0, pointers[ndx][0], pointers[ndx][1], false, pointers[ndx][2], pointers[ndx][3]);
+        } catch (err) {
```
kenrussell: This try/catch mechanism for catching the GL errors only works because sdk/tests/deqp/functional/gles3/shaderstatequery.html is using wtu.create3DContextWithWrapperThatThrowsOnGLError. It's unfortunate that this causes a significant change in the control flow, especially because at some point we'll likely want to turn off that context wrapper. I'll merge this now and look into how this can be cleaned up.

PR #1491, opened 2016-02-19
change the specs of pixel storage parameters for ImageBitmap

xidachen:
We basically add a note for the parameters of UNPACK_FLIP_Y_WEBGL, UNPACK_PREMULTIPLY_ALPHA_WEBGL and UNPACK_COLORSPACE_CONVERSION_WEBGL, stating that these three parameters will be ignored if the texImageSource is an ImageBitmap.

Please review @kenrussell @zhenyao @toji

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -2503,10 +2503,6 @@ WebGLRenderingContext implements WebGLRenderingContextBase;
             If this function is called with an <code>ImageBitmap</code> that has been neutered,
             an <code>INVALID_VALUE</code> error is generated. <br><br>
 
-            If this function is called with an <code>ImageBitmap</code>, the
-            <a href="#PIXEL_STORAGE_PARAMETERS">Pixel Storage Parameters</a> associated with this
-            function are ignored. <br><br>
-
             If this function is called with an <code>HTMLImageElement</code>
             or <code>HTMLVideoElement</code> whose origin differs from the origin of the containing
             Document, or with an <code>HTMLCanvasElement</code> or <code>ImageBitmap</code> whose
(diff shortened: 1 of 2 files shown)


specs/latest/1.0/index.html
```diff
@@ -3575,6 +3575,11 @@ <h2 class="no-toc">Table of contents</h2>
 
 </dl>
 
+<p>
+If the TexImageSource is an ImageBitmap, then these three parameters will be ignored. Instead
+the equivalent <a href="https://wiki.whatwg.org/wiki/ImageBitmap_Options">ImageBitmapOptions</a>
+should be used to create an ImageBitmap with desired format.
```
kenrussell: To avoid confusion, the entry points taking the TexImageSource union type should be modified somehow to more clearly point out that these pixel storage parameters don't apply when an ImageBitmap is being uploaded. Re-reading the docs for texImage2D, I think they could use a larger restructuring, so I'll take care of that after merging this pull request.

kenrussell:
Thanks, this looks good. Let's merge it now.

jdashg:
Did we discuss this in-meeting? I was surprised to see this, but it's years ago so it's totally possible I forgot. @kenrussell

kenrussell:
Looking back through the meeting minutes - yes, it's minuted in the meeting from 18 Feb 2016 that there was agreement to ignore the WebGL-specific pixel unpack parameters when uploading ImageBitmaps to WebGL textures. These are actually all of imageOrientation, premultiplyAlpha, and colorSpaceConversion, which were added specifically on request of the WebGL working group and correspond to UNPACK_FLIP_Y_WEBGL, UNPACK_PREMULTIPLY_ALPHA_WEBGL, and UNPACK_COLORSPACE_CONVERSION_WEBGL.

jdashg:
Do you have a link to those minutes? I can't seem to find a way to access them.

kenrussell:
Link sent offline. (While it requires authentication, seems best not to post Khronos-internal links here.)

PR #1492, opened 2016-02-19
Update docs for tex{Sub}Image[23]D taking TexImageSource.

kenrussell:
Clarify that the WebGL-specific pixel storage parameters do not apply to passed ImageBitmaps, since their parameters are already baked in when the ImageBitmap's promise is resolved. Reorganize the text to make this more clear, and to leave the description of all errors at the end of each function's documentation.


kenrussell:
@grorg @jdashg please review.

@zhenyao @toji @RafaelCintron @xidachen FYI; any feedback welcome.

zhenyao:
looks good on my behalf.  please also update the date on 1.0

kenrussell:
The 1.0 spec was already updated today, so no date modification was needed.

kenrussell:
There haven't been any objections to this documentation update so I'm merging it now to make forward progress.

PR #1493, opened 2016-02-19
Stop using create3DContextWithWrapperThatThrowsOnGLError in shadersta…

kenrussell:
…tequery.html.

Other tests which use expectError() have already stopped using this error wrapper, which was only intended for debugging the tests themselves.


kenrussell:
@zhenyao please review.

@Richard-Yunchao @qkmiao @xhcao @toji FYI.

zhenyao:
lgtm

Issue #1474, opened 2016-02-11
newFloat64 looks wrong


jrmuizel:
tcuFloat.newFloat64 = function(value) {
    return new tcuFloat.deFloat().deFloatParameters(value, tcuFloat.description64);
};

deFloatParameters takes a binary representation as the first parameter not the floating point value.

kenrussell:
sdk/tests/deqp/framework/common/tcuFloat.js:

tcuFloat.deFloat.prototype.deFloatParameters = function(jsnumber, description) { ... }

The value is the first argument.

It does look like tcuFloat.newFloat64 could use tcuFloat.newDeFloatFromParameters(), but that doesn't change the argument order.

If there's a specific test that's failing please point it out. Closing as there doesn't seem to be an error here.

jrmuizel:
newFloat64 takes a floating point value like 0.2. deFloatParameters takes the binary representation of that number as an integer. newFloat64 is just passing the floating point value in as though it was the binary representation.

kenrussell:
I investigated this in detail and the code is correct. The documentation implied that deFloatParameters took in the bitwise representation of the number but that's not correct. Updated the documentation in #1494 . Verified that for example precision.float.highp_mul_vertex in deqp/functional/gles3/shaderprecision.html is passing with the current newFloat64 code and making it less correct (for example, by calling tcuFloat.newDeFloatFromParameters, which truncates it to 32 bits) makes the test fail.

PR #1494, opened 2016-02-19
Update documentation for deFloatParameters and newDeFloatFromParameters.

kenrussell:
The code was correct but the documentation was confusing. Addresses #1474.


kenrussell:
I confirmed that this documentation update is all that is needed. The code is correct. Making any changes to how newFloat64 works (for example, calling tcuFloat.newDeFloatFromParameters) breaks precision.float.highp_mul_vertex in deqp/functional/gles3/shaderprecision.html .

I'm merging this myself to address the reported issue.

PR #1495, opened 2016-02-20
Enhance uninitialized-test-2.html

zhenyao:
1) Cover TexStorage2D
2) Cover for int and uint formats
3) Cover the partially initialized cases

### DIFF
--- a/sdk/tests/conformance2/misc/uninitialized-test-2.html
+++ b/sdk/tests/conformance2/misc/uninitialized-test-2.html
@@ -49,9 +49,21 @@ else
   testPassed("Context created.");
 
 function setupTexture(target, texWidth, texHeight, texDepth) {
+    var is3d = (target == gl.TEXTURE_3D || target == gl.TEXTURE_2D_ARRAY);
     var texture = gl.createTexture();
     gl.bindTexture(target, texture);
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/misc/uninitialized-test-2.html
```diff
@@ -72,68 +95,208 @@
...
         for (var x = 0; x < texWidth; ++x) {
             var index = (y * texWidth + x) * 4;
+            var is_exception = false;
+            for (var ii = 0; ii < exceptions.length; ++ii) {
+                if (exceptions[ii].x == x && exceptions[ii].y == y) {
+                    is_exception = true;
+                    if (data[index] != exceptions[ii].r ||
+                        data[index + 1] != exceptions[ii].g ||
+                        data[index + 2] != exceptions[ii].b ||
+                        data[index + 3] != exceptions[ii].a) {
```
kenrussell: Is a tolerance needed for this comparison? If the back buffer happens to change semantics slightly (for example, it might use an sRGB texture on some implementations) then the read-back pixel may not be exactly what was sampled in the texture.

kenrussell:
The enhancement to the test looks very good. One question. Please merge yourself after addressing.

sdk/tests/conformance2/misc/uninitialized-test-2.html
zhenyao: Very good point.  Added a tolerance now.

PR #1496, opened 2016-02-22
Rename image-bitmap to image-bitmap-from-image-data

xidachen:
An ImageBitmap can be created from multiple kinds of sources such as ImageData, HTMLImageElement, HTMLVideoElement and HTMLCanvasElement. Right now the conformance & conformance2 takes an ImageBitmap created from ImageData.

What I plan to do in the future is to change it such that the conformance & conformance2 will take ImageBitmap created from HTMLImageElement, HTMLVideoElement and HTMLCanvasElement. In that case, we have broader coverage for ImageBitmap. As the first step, I rename the file name for under image-bitmap to image-bitmap-from-image-data.


zhenyao:
lgtm

kenrussell:
LGTM after the fact. Agree that it's important to test uploading of ImageBitmaps created from all of the valid sources (ImageData, HTMLImageElement, HTMLCanvasElement, HTMLVideoElement).

Perhaps some code can be shared between the tests which create textures directly from these sources, and the tests which create textures from ImageBitmaps which were created from these sources.

xidachen:
Good point Ken, my plan is to start working on these tests without sharing any code and make sure that they work fine. In this case, if there is a bug somewhere, it is easy to debug. Once we have the test coverage for all the valid sources, we can move some common functions to some util.js. Does that make sense?

kenrussell:
Sure, that sounds fine.

PR #1497, opened 2016-02-22
Add new conformance for ImageBitmap created from an HTMLImageElement

xidachen:
The existing conformance(2) tests for ImageBitmap takes an ImageBitmap created from ImageData. Now we add tests that takes an ImageBitmap created from an HTMLImageElement. All tests pass locally on a Linux desktop.


sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap-from-image.js
```diff
@@ -0,0 +1,120 @@
...
+          default:
+            break;
+        }
+
+        gl.clearColor(0,0,0,1);
+        gl.clearDepth(1);
+
+        var image = new Image();
+        image.src = resourcePath + "red-green.png";
+        image.onload = function() {
```
kenrussell: The onload handler needs to be assigned before src, because otherwise there's a race condition where the onload handler might not fire.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image.js
```diff
@@ -0,0 +1,158 @@
...
+          default:
+            break;
+        }
+
+        gl.clearColor(0,0,0,1);
+        gl.clearDepth(1);
+
+        var image = new Image();
+        image.src = resourcePath + "red-green.png";
+        image.onload = function() {
```
kenrussell: See below; the onload handler needs to be assigned before the src attribute.

kenrussell:
Thanks for contributing this new test; it looks good overall. Could you please address the two issues above?

xidachen:
Hi Ken, thank you for the comments, I have addressed both. Please take a look.

kenrussell:
Thanks for the updates. LGTM.

PR #1498, opened 2016-02-22
Add more test cases for uninitialized-test-2.html

zhenyao:
Test with larger texture sizes.  This triggers different code paths in chrome.

### DIFF
--- a/sdk/tests/conformance2/misc/uninitialized-test-2.html
+++ b/sdk/tests/conformance2/misc/uninitialized-test-2.html
@@ -179,126 +179,389 @@ function checkNonZeroPixels(texture, target, format, type, texWidth, texHeight, 
     }
 }
 
-var width = 512;
-var height = 512;
-var depth = 2;
+function testTexImage3D() {
 
-var targets = [ "TEXTURE_2D", "TEXTURE_CUBE_MAP" ];
-for (var ii = 0; ii < targets.length; ++ii) {
-    debug("");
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/misc/uninitialized-test-2.html
```diff
@@ -179,127 +179,390 @@
...
-var read_types = [ gl.UNSIGNED_BYTE, gl.UNSIGNED_INT, gl.INT ];
+      // TEXTURE_2D_ARRAY + RGBA8
+      {
+        target: "TEXTURE_2D_ARRAY",
+        internal_format: "RGBA8",
+        format: gl.RGBA,
+        type: gl.UNSIGNED_BYTE,
+        read_type: gl.UNSIGNED_BYTE,
+        width: 2048,
+        height: 2048,
```
kenrussell: Would it be possible to reproduce the same issues with a smaller width/height for these 2D texture array cases, like 1024x1024? If so please adjust the test cases. Thanks.

kenrussell:
Looks good overall. One comment above. Please merge yourself after addressing one way or the other. Thanks.

sdk/tests/conformance2/misc/uninitialized-test-2.html
zhenyao: Yes we can.  I'll adjust to 1024.

PR #1499, opened 2016-02-22
Add a new case to uninitialized-test.html

zhenyao:
If we create texture using TexImage2D with null data, and then
initialize part of it using TexSubImage2D, and then ReadPixels() the
entire texture. The uninitialized part should be cleared to 0, and the
initialized part should stay untouched.

Chrome has a bug where the initialized part will also be cleared to 0.

This is fixed in https://codereview.chromium.org/1706773003/


kenrussell:
Very nice. LGTM

PR #1500, opened 2016-02-23
Add new conformance tests for ImageBitmap(HTMLVideoElement)

xidachen:
Following this pull request: https://github.com/KhronosGroup/WebGL/pull/1497 which adds new conformance tests for ImageBitmap created from a HTMLImageElement, we now add tests for ImageBitmap created from a HTMLVideoElement. All tests passed locally on a Linux desktop.


sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-video.js
```diff
@@ -0,0 +1,160 @@
...
+            break;
+          default:
+            break;
+        }
+
+        gl.clearColor(0,0,0,1);
+        gl.clearDepth(1);
+
+        var video = document.createElement("video");
+        video.oncanplaythrough = function() {
```
kenrussell: The use of the oncanplaythrough event is interesting. I've seen it in some other recent layout tests. The other video-related WebGL tests use a couple of other events and a bunch of heuristics to try to figure out whether the first frame of the video is being displayed. If oncanplaythrough has consistent semantics across all browsers, the other tests (rather, webgl-test-utils' startPlayingAndWaitForVideo) could be simplified. Xida, what do you think?

kenrussell:
Very nice work Xida. LGTM

xidachen:
I used the "oncanplaythrough" here because I think that it fits our need. I actually didn't test the behavior of other browsers' behavior of "oncanplaythrough". If we were to use this instead of startPlayingAndWaitForVideo, I'd wait until we see that all these tests are consistently passing on major platforms after a chromium DEPS roll.

PR #1501, opened 2016-02-24
pixels(type is ArrayBufferView) in texSubImage* can not be null in WebGL

Richard-Yunchao:
The feature differs from that of native gles, in which the data(type is void*) can be null. 

The original code make many test ambiguous. It may generate INVALID_VALUE error if pixels is null. 

PTAL. Thanks!


kenrussell:
Thanks for these fixes. Agreed that the tests are ambiguous as ported. Your changes look good.

PR #1502, opened 2016-02-24
Let's flip ImageBitmap ^o^

xidachen:
createImageBitmap(validSource, flipY=true) is fully supported for all valid sources. Let's try it. I change the test of image-bitmap-from-image-data with flipY options. All conformance(2) tests passed locally on a Linux desktop.

The future plan it to apply flipY to all sources and after that, premultiplyAlpha.


xidachen:
It is worth noting that tex(Sub)Image2D and texSubImage3D for ImageBitmap is now running without re-decoding at all.

kenrussell:
Very nice. LGTM

PR #1505, opened 2016-02-25
Add conformance tests for ImageBitmap from canvas

xidachen:
This change adds conformance(2) tests for ImageBitmap created from an HTMLCanvasElement. Also, the test cases here covers flipY.


zhenyao:
lgtm

PR #1506, opened 2016-02-26
Fix sync status for WebGL

qkmiao:
For WebGL, sync status should not be signaled in the same frame as
glFinish issued. See section 5.23 "Queries' result must not be made
available in the current frame" in WebGL 2.0 spec.

There are some discussions in https://codereview.chromium.org/1736433003/. glFinish in WebGL issues glFlush to driver other than real glFinish.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fSyncTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fSyncTests.js
@@ -206,9 +206,9 @@ goog.scope(function() {
 
         gl.finish();
 
-        if (this.m_caseOptions & es3fSyncTests.CaseOptions.FINISH_BEFORE_WAIT && waitValue != gl.ALREADY_SIGNALED) {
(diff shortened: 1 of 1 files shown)


zhenyao:
The change is conforming to WebGL 2 spec. However, we seems to lose test coverage by making this change.

@kenrussell @toji What do you think?

kenrussell:
Looking back at the WebGL 2.0 spec, I was wrong about the stated semantics of sync objects -- sorry about that.

The rules aren't laid out clearly. It's legal for a sync object to become signaled within a given frame. However, the intent is that it's not required, even if a glFinish is issued. The limitations on ClientWaitSync are intended to define this behavior, but I'll update the specification to make this more clear.

Instead of simply disabling this part of the test, we should change the iterate() function to have a few states, and re-run the "verify" state up to a certain number of times, returning control to the browser in between. es3fTransformFeedbackTests.js shows how to do this. Instead of asking you to do this restructuring, let me put up another pull request implementing it and you can review.

qkmiao:
Thank you ken. I close this pull request.

PR #1509, opened 2016-02-26
Align dEQP lifetime tests with C++ code

qkmiao:
C++ code: https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fLifetimeTests.cpp#343


zhenyao:
looks good.  thanks for fixing them.

kenrussell:
Very nice. Thank you for tracking these down. LGTM after the fact.

PR #1510, opened 2016-02-26
Fix vertex array life time test

qkmiao:
Detach array buffer from vertex attribute isn't implemented in js. This
patch gets the vertex buffer first, then removes it which makes the
vertex buffer detach from vertex attribute.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
@@ -100,8 +100,12 @@ es3fLifetimeTests.ScaleProgram.prototype.draw = function(vao, scale, tf, dst) {
 es3fLifetimeTests.ScaleProgram.prototype.setPos = function(buffer, vao) {
     gl.bindBuffer(gl.ARRAY_BUFFER, buffer);
     gl.bindVertexArray(vao);
-    if (buffer)
+    if (buffer) {
         gl.vertexAttribPointer(this.m_posLoc, NUM_COMPONENTS, gl.FLOAT, false, 0, 0);
+    } else {
+        var name = gl.getVertexAttrib(this.m_posLoc, gl.VERTEX_ATTRIB_ARRAY_BUFFER_BINDING);
+        gl.deleteBuffer(name);
+    }
     gl.bindVertexArray(null);
     gl.bindBuffer(gl.ARRAY_BUFFER, null);
 };
@@ -254,7 +258,7 @@ es3fLifetimeTests.BufferVAOAttacher.prototype.detach = function(element, target)
 es3fLifetimeTests.BufferVAOAttacher.prototype.getAttachment = function(target) {
     var vao = /** @type {WebGLVertexArrayObject} */ (target);
     gl.bindVertexArray(vao);
-    var name = gl.getVertexAttrib(0, gl.VERTEX_ATTRIB_ARRAY_BUFFER_BINDING);
+    var name = gl.getVertexAttrib(this.m_posLoc, gl.VERTEX_ATTRIB_ARRAY_BUFFER_BINDING);
     gl.bindVertexArray(null);
     return name;
 };


sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
```diff
@@ -254,7 +258,7 @@ es3fLifetimeTests.BufferVAOAttacher.prototype.detach = function(element, target)
 es3fLifetimeTests.BufferVAOAttacher.prototype.getAttachment = function(target) {
     var vao = /** @type {WebGLVertexArrayObject} */ (target);
     gl.bindVertexArray(vao);
-    var name = gl.getVertexAttrib(0, gl.VERTEX_ATTRIB_ARRAY_BUFFER_BINDING);
+    var name = gl.getVertexAttrib(this.m_posLoc, gl.VERTEX_ATTRIB_ARRAY_BUFFER_BINDING);
```
zhenyao: This deviates from the c++ code.

zhenyao:
I checked the c++ code and don't see glDeleteBuffers there either. Can you explain why it is necessary in the js?

sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
qkmiao: See setPos, array buffer is bound to vertex attributes at index this.m_posLoc. So we should get buffer name from that index. I think C++ code has the same bug. In this case this.m_posLoc is always 0, so no difference between this.m_posLoc and 0.

sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
```diff
@@ -100,8 +100,12 @@ es3fLifetimeTests.ScaleProgram.prototype.draw = function(vao, scale, tf, dst) {
 es3fLifetimeTests.ScaleProgram.prototype.setPos = function(buffer, vao) {
     gl.bindBuffer(gl.ARRAY_BUFFER, buffer);
     gl.bindVertexArray(vao);
-    if (buffer)
+    if (buffer) {
         gl.vertexAttribPointer(this.m_posLoc, NUM_COMPONENTS, gl.FLOAT, false, 0, 0);
```
qkmiao: When array buffer is null, vertexAttribPointer doesn't work in WebGL, while it works in GLES. vertexAttribPointer in GLES has an extra  'pointer' argument. When bound array buffer is 0, vertexAttribPointer will attach null pointer to vertex attribute at index.

zhenyao: Right. Instead it has an offset that points to the bound buffer.

Still, why do we need to delete the buffer?  The C++ code doesn't.  Does that cause a behavior change? I try to understand the motivation.

sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
zhenyao: Got it.

sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
qkmiao: Delete the buffer will remove unbind the buffer from vertex attribute array. Otherwise detach (called from glsLifetimeTests.AttachmentTest.testDeletedNames) does nothing. Or are there any other methods to detach array buffer?

qkmiao: It's not a good implementation. But I didn't find a better one.

zhenyao: We should probably change WebGL 2 spec, and when a vao is bound, and nothing is bound to ARRAY_BUFFER, let the vertexAttribPointer(offset == 0) to be accepted.

Let's merge this for now.

PR #1512, opened 2016-02-29
Add flipY to ImageBitmap created from image and video

xidachen:
This change adds flipY option to conformance(2) tests for ImageBitmap that is created from an HTMLImageElement or HTMLVideoElement. With this change, flipY is covered in all the current tests for ImageBitmap.

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image-data.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image-data.js
@@ -33,7 +33,7 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
 
     function init()
     {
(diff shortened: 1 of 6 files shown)


xidachen:
Actually let's hold this for now, I have an on-going CL that adds support for premultiplyAlpha=false, once that CL landed, I will add both flipY and premultiplyAlpha=false in this PR.

xidachen:
With the latest commit, flipY is covered in all cases. premultiplyAlpha=false is covered in the case of image-bitmap-from-image-data. It is a bit tricky in the case of image-bitmap-from-image/video because I would have to create an image or video that is transparent. I am working on that but would require a bit of work. So maybe we could merge this change first?

kenrussell:
Yes, that's fine. Let's merge this now. premultiplyAlpha=false can really only be tested effectively with image and ImageData sources. Canvas and video are problematic. (Most 2D canvas implementations premultiply alpha internally, so premultiplyAlpha=false is lossy; and there aren't many/any video formats that support an alpha channel.)

PR #1513, opened 2016-03-01
Check single transform feedback output

qkmiao:
This exposes a bug of bindBufferBase in chrome. Between two calls of
"gl.bindBuffer(gl.TRANSFORM_FEEDBACK_BUFFER, out_add_buffer);"
bound_transform_feedback_buffer_ doesn't change, so
GLES2Implementation::BindBufferHelper doesn't issue bindBuffer to
command buffer server side.
Please refer chrome command buffer client code: https://code.google.com/p/chromium/codesearch#chromium/src/gpu/command_buffer/client/gles2_implementation.cc&q=BindBufferHelper&sq=package:chromium&type=cs&l=3811

### DIFF
--- a/sdk/tests/conformance2/transform_feedback/transform_feedback.html
+++ b/sdk/tests/conformance2/transform_feedback/transform_feedback.html
@@ -78,7 +78,8 @@ if (!gl) {
 
     runBindingTest();
     runObjectTest();
-    runFeedbackTest();
+    runOneOutFeedbackTest();
+    runTwoOutFeedbackTest();
 }
 
 function runBindingTest() {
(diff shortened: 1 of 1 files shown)


toji:
LGTM, thanks.

kenrussell:
Thanks for uncovering this bug with this test case. LGTM after the fact. Has a bug been filed against Chrome?

qkmiao:
@kenrussell  no bug has been filed against Chrome.

kenrussell:
Please file a bug in that case pointing to this new test case. Thanks.

kenrussell:
Or, if https://codereview.chromium.org/1752703003/ fixes this, no need to file a new bug.

qkmiao:
I filed a bug. You can find it at https://bugs.chromium.org/p/chromium/issues/detail?id=591258

kenrussell:
Thanks.

PR #1514, opened 2016-03-01
Implement lifetime.delete_active.transform_feedback test

qkmiao:
Align js code with C++ code.

C++ code: https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fLifetimeTests.cpp#489

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
@@ -400,20 +400,47 @@ es3fLifetimeTests.TfDeleteActiveTest = function(name, description) {
 setParentClass(es3fLifetimeTests.TfDeleteActiveTest, tcuTestCase.DeqpTest);
 
 es3fLifetimeTests.TfDeleteActiveTest.prototype.iterate = function() {
+/** @const */ var s_xfbVertexSource =
+    '#version 300 es\n' +
+    'void main ()\n' +
+    '{\n' +
+    ' gl_Position = vec4(float(gl_VertexID) / 2.0, float(gl_VertexID % 2) / 2.0, 0.0, 1.0);\n' +
+    '}\n';
+
+/** @const */  var s_xfbFragmentSource =
+    '#version 300 es\n' +
+    'layout(location=0) out mediump vec4 dEQP_FragColor;\n' +
+    'void main (void)\n' +
+    '{\n' +
+    ' dEQP_FragColor = vec4(1.0, 1.0, 0.0, 1.0);\n' +
+    '}\n';
+    var buf = gl.createBuffer();
+
+    var sources = new gluShaderProgram.ProgramSources();
+    sources.add(new gluShaderProgram.VertexSource(s_xfbVertexSource));
+    sources.add(new gluShaderProgram.FragmentSource(s_xfbFragmentSource));
+    sources.add(new gluShaderProgram.TransformFeedbackMode(gl.SEPARATE_ATTRIBS));
+    sources.add(new gluShaderProgram.TransformFeedbackVarying('gl_Position'));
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
```diff
@@ -400,20 +400,46 @@ es3fLifetimeTests.TfDeleteActiveTest = function(name, description) {
 setParentClass(es3fLifetimeTests.TfDeleteActiveTest, tcuTestCase.DeqpTest);
 
 es3fLifetimeTests.TfDeleteActiveTest.prototype.iterate = function() {
+/** @const */ var s_xfbVertexSource =
+    '#version 300 es\n' +
+    'void main ()\n' +
+    '{\n' +
+    ' gl_Position = vec4(float(1) / 2.0, float(1 % 2) / 2.0, 0.0, 1.0);\n' +
```
qkmiao: This deviates from c++ code.  I didn't use gl_VertexID here. Chrome doesn't support gl_VertexID yet and gl_VertexID doesn't influence the test result in this case.

sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
```diff
@@ -400,20 +400,47 @@ es3fLifetimeTests.TfDeleteActiveTest = function(name, description) {
...
-    try {
-        gl.deleteTransformFeedback(tf);
-    } catch (err) {
-        errCode = err.error;
-    }
-    gl.endTransformFeedback();
     gl.deleteTransformFeedback(tf);
+    errCode = gl.getError();
     assertMsgOptions(errCode == gl.INVALID_OPERATION,
         'Deleting active transform feedback must produce INVALID_OPERATION', false, true);
```
qkmiao: Chrome will hang when running this test even with https://codereview.chromium.org/1752703003/. You may see a lot of gl errors, such as
[1291:1291:0301/225346:ERROR:gles2_cmd_decoder.cc(2101)] [.CommandBufferContext.RenderWorker-0x24b0aba1b6e0]GL ERROR :GL_INVALID_OPERATION : GLES2DecoderImpl::D
 error from previous GL command.

But after you remove above 4 lines, chrome will not hang. Similar thing can be seen for link_program case in negativeshaderapi.html. Also chrome will hang when you debug conformance2/transform_feedback/transform_feedback.html with inspector. Chrome doesn't reponse after beginTransformFeedback and drawArrays.  I guess there are still bugs in chrome for transform feedback.

sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
zhenyao: I don't feel this deviation is necessary.  Failing in Chrome due to an implementation bug is not a good reason to change the test.

sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
zhenyao: What you said seems like a good reason to keep this test case, not to delete it.

zhenyao:
Mostly looks good, but can you change back to the in consistence with the original c++ test?  It's OK we still can't run it in Chrome browser - we will fix whatever bugs and then pass the test.

qkmiao:
Thanks @zhenyao ! I updated the pull request to make js test consist with c++ test. PTAL.

zhenyao:
lgtm

kenrussell:
Mo, thanks for pointing out the rule that the WebGL conformance tests mustn't be modified to work around browser bugs. I've done this myself in the past and it was a big mistake. The tests can certainly be changed to fix bugs, but shouldn't be changed to work around bugs in WebGL implementations.

PR #1515, opened 2016-03-01
Let WebGL spec points to whatwg specs of ImageBitmapOptions

xidachen:
At this moment, the WebGL spec has a statement that is pointing to a whatwg wiki for ImageBitmapOptions, now that ImageBitmapOptions is in the whatwg specs, let's change that statement to point to the whatwg spec instead of wiki.


kenrussell:
Looks fine. Thanks for the update.

PR #1516, opened 2016-03-02
Draw refSurface before attaching buffer to TRANSFORM_FEEDBACK_BUFFER

qkmiao:
A buffer bound to TRANSFORM_FEEDBACK_BUFFER can not be bound to any
other targets. See section 5.1 Buffer Object Binding in WebGL 2.0 spec.

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsLifetimeTests.js
+++ b/sdk/tests/deqp/modules/shared/glsLifetimeTests.js
@@ -992,11 +992,11 @@ glsLifetimeTests.OutputAttachmentTest.prototype.iterate = function() {
     var element = elementType.gen();
 
     attacher.initAttachment(0, element);
+    this.m_outputAttacher.drawAttachment(element, refSurface);
     attacher.attach(element, container);
 
     // For reference purposes, make note of what refSeed looks like.
     this.m_outputAttacher.setupContainer(refSeed, container);
-    this.m_outputAttacher.drawAttachment(element, refSurface);
     elementType.release(element);
 
     bufferedLogToConsole('Writing to a container after deletion of attachment');


sdk/tests/deqp/modules/shared/glsLifetimeTests.js
```diff
@@ -992,11 +992,11 @@ glsLifetimeTests.OutputAttachmentTest.prototype.iterate = function() {
     var element = elementType.gen();
 
     attacher.initAttachment(0, element);
+    this.m_outputAttacher.drawAttachment(element, refSurface);
     attacher.attach(element, container);
```
qkmiao: in drawAttachment, buffer is bound to ARRAY_BUFFER firstly. After drawing, buffer is unbound from ARRAY_BUFFER.
In attach(elemetn, container), buffer is bound to TRANSFORM_FEEDBACK_BUFFER, but isn't unbound.
So moving drawAttachment before attach will not trigger "A buffer bound to TRANSFORM_FEEDBACK_BUFFER can not be bound to any other targets" error.

kenrussell: This fix doesn't look correct. There is more than just one pair of setupContainer / drawAttachment calls in this file, and it looks like the harness assumes these are called in this order.

I suggest you add a new virtual function to glsLifetimeTests.Attacher like "detachSecondaryBindings" or "detachAuxiliaryBindings" which is empty for most of the attachment types in es3fLifetimeTests.js, except for es3fLifetimeTests.BufferTfAttacher. Then add this call between the calls to setupContainer and drawAttachment, below. Will that work?

qkmiao: Hi ken, can you explain what will the new function "detachSecondaryBindings" do for es3fLifetimeTests.BufferTfAttacher?

kenrussell: The idea was that it would unbind the buffer from TRANSFORM_FEEDBACK_BUFFER to avoid the collision where the same buffer is bound to both the ARRAY_BUFFER and TRANSFORM_FEEDBACK_BUFFER binding points.

Maybe I'm thinking about the problem incorrectly, but rather than reordering the drawAttachment/setupContainer calls, we should add more hooks to allow the state to be compatible with WebGL's semantics.

qkmiao:
I think we cannot detach the buffer from TRANSFORM_FEEDBACK_BUFFER before the test finishes. That's part of what the case tests. The buffer should be detached at the end of the case, see attacher.detach(element, container).

kenrussell:
@qkmiao Please think about another way to fix this bug that doesn't involve switching the order of setupContainer and drawAttachment, but instead adding another WebGL-specific abstract method if necessary. If you need more help figuring out what to do here please let me know.

qkmiao:
Thanks for review. I will think about another method fo fix this test.

PR #1518, opened 2016-03-02
Add premultiplyApha=false for ImageBitmap from HTMLImageElement and Canvas

xidachen:
Using HTMLCanvasElement as example, I set canvas to be rgba(255, 0, 0, 0.5) and create an ImageBitmap from it. When premultiplyAlpha=default, the color we see should be somewhere around 128 (usually won't be exactly 128 because of rounding error). Now if we create an ImageBitmap from it with premultiplyAlpha=false, we should see solid red color.

In the case of HTMLImageElement, I use canvas.toDataURL to obtain an img element and create ImageBitmap from it.

Locally conformance tests all pass, the tests always fail for 7 format under conformance2:
r8_red_unsigned_byte
r16f_red_half_float
r16f_red_float
r32f_red_float
r8ui_red_integer_unsigned_byte
srgb8_rgb_unsigned_byte
srgb8_alpha8_rgba_unsigned_byte

I don't know exactly why these 7 format failed, but will investigate later. What I plan to do next is to add new tests for ImageBitmap created from a Blob and from an ImageBitmap. After that, we have full coverage for all valid sources.

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image.js
@@ -26,6 +26,8 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
     var tiu = TexImageUtils;
     var gl = null;
     var successfullyParsed = false;
+    var halfRedColor = [128, 0, 0];
+    var halfGreenColor = [0, 128, 0];
     var redColor = [255, 0, 0];
     var greenColor = [0, 255, 0];
     var bitmaps = [];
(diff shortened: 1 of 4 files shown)


sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image.js
```diff
@@ -58,12 +60,18 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
         gl.clearColor(0,0,0,1);
         gl.clearDepth(1);
 
+        var testCanvas = document.createElement('canvas');
+        var ctx = testCanvas.getContext("2d");
+        setCanvasToRedGreen(ctx);
```
kenrussell: For the construction of ImageBitmap from image, it would be better to create a resource .png with the required 4 colors, and load it with premultiplyAlpha={default,none} as well as colorspaceConversion=none. The reason is that every current canvas implementation uses a premultiplied alpha representation internally, which is lossy, and which is probably why the tolerances are so high.

In a follow-up pull request, could you construct such an image using the Gimp or Photoshop (let me know if you need help) and replace the canvas in both this and the 3d-with-image-bitmap-from-image test below with that? Thanks.

kenrussell:
Very nice work Xida. One comment above about the precision of the image-bitmap-from-image test, but merging now so you can move forward. LGTM

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image.js
xidachen: No problem. I tried GIMP before, that didn't work. But I managed to generate a .png file with half transparency using Matlab. I will change this in a follow-up pull request. Thank you.

PR #1519, opened 2016-03-02
Add tests for ImageBitmap created from a Blob

xidachen:
Summary of changes:
1. create new tests for ImageBitmap created from a Blob. Similar to HTMLImageElement, the tests fail on exactly the same 7 formats.
2. Add a new .png file of 2*2 pixels, pixels are: solid red, red with alpha=0.5, solid green, green with alpha=0.5.
3. Change tests for ImageBitmap created from a HTMLImageElement to take the .png file instead of using canvas.toDataURL. Unfortunately the tolerance has to be 8. I tried to set it to 5, and some tests passed, and some failed. It appears that 8 is the limit.


xidachen:
I forgot to mention, I tried to create a video of alpha=0.5, but I don't think I can do that even with Matlab.

sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap-from-image.js
```diff
@@ -138,7 +117,7 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
 
         // Check a few pixels near the top and bottom and make sure they have
         // the right color.
-        var tolerance = 10;
+        var tolerance = 8;
```
kenrussell: I suspect the reason the tolerance needs to be so high is that the ImageBitmap constructor argument colorspaceConversion='none' is not supported yet. Other than that, the tolerances should be very close to 0. Please see sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image.js. Please give yourself a TODO to look at this again once that constructor argument is supported, and further, to use it in these tests.

kenrussell:
Thanks Xida for contributing these tests. They look good with the one caveat mentioned above. Please follow up on it later. Don't worry about creating a video with an alpha channel; they don't exist in practice. Merging now so you can move forward with your work.

sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap-from-image.js
xidachen: Thank you for your comments, Ken. I didn't realize that colorspaceConversion would have impact on the result in this case. Once colorspaceConversion is supported, I will definitely add that to all these test cases.

msc-:
The MPEG-4 specification most definitely supports arbitrary shaped video (i.e. video with an alpha channel). I don't know about the video file formats used on the web though.

PR #1520, opened 2016-03-03
Deleted buffer should be invalid

qkmiao:
When a buffer is deleted, its name immediately becomes invalid. But the
continue using the object (page 315 ES spec 3.0.4).


kenrussell:
This LGTM. @zhenyao would you like to review as well?

zhenyao:
lgtm

Can we also merge back to earlier snapshots for conformance suite?

PR #1522, opened 2016-03-03
Experimental: Please do not merge

xidachen:
What I am trying:
1. Change the source image to be "red-green-semi-transparent.png", which is a 2*2 image that has solid red, red with alpha=0.5, solid green, and green with alpha=0.5.
2. Change UNPACK_PREMULTIPLY_ALPHA_WEBGL=true
3. In this case, the top-right color should be somewhere around 128 in the red channel. But what I observe is that the tolerance has to be 8. 
4. Even with tolerance=8, the following 7 formats fail under conformance2 tests:
r8_red_unsigned_byte
r16f_red_half_float
r16f_red_float
r32f_red_float
r8ui_red_integer_unsigned_byte
srgb8_rgb_unsigned_byte
srgb8_alpha8_rgba_unsigned_byte

@kenrussell @zhenyao Please take a look. I think this is the problem that I am having for image-bitmap-from-image, and image-bitmap-from-blob. In the case when I try to create an ImageBitmap from "red-green-semi-transparent.png" with premultiplyAlpha=true, the top-right needs to have a tolerance of 8.

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image.js
@@ -27,6 +27,8 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
     var gl = null;
     var successfullyParsed = false;
     var imgCanvas;
+    var halfRedColor = [128, 0, 0];
+    var halfGreenColor = [0, 128, 0];
     var redColor = [255, 0, 0];
     var greenColor = [0, 255, 0];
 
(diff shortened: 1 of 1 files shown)


kenrussell:
Thanks for investigating this. What if you leave UNPACK_PREMULTIPLY_ALPHA_WEBGL set to false and explicitly check the alpha channel as well as the color channels? Does that allow a tolerance of 0 to be used in that case?

xidachen:
Good point. I leave it as false, and found that tolerance=0 makes all cases pass including conformance2 tests.

So I also checked the image-bitmap-from-image tests, particularly the cases of createImageBitmap(image, premultiplyAlpha=false), and all the tests pass with tolerance=0 too. So it looks like the faulty tests is in the case of premultiplyAlpha=true.

I investigated a bit further. So texImage2D(ImageBitmap) calls bitmap->copyBitmapData() which returns a const unsigned char array, I print out this array in the case of createImageBitmap(image, premultiplyAlpha=true), and for the red pixel with alpha=0.5, it is actually [128, 0, 0, 128]. It basically says that the WebGLImageConversion::extractImageData() which does the format conversion seems to return a different value than [128, 0, 0, 128]. Does that make sense?

zhenyao:
Can you provide the extractImageData output for the failing formats? That will give us some clue what went wrong.

xidachen:
OK, I have a better example, will do a new pull request and close this one.

PR #1523, opened 2016-03-03
Make some uses of "defined" shader preprocessor errors

Oletus:
The GLES working group decided in recent discussions that these cases
should be errors, but left the spec undefined in older versions of the
API. WebGL should specify them to be errors to be consistent with the
direction GLES is going.

The changes to the tests will be upstreamed to native dEQP as well.


kenrussell:
Thank you. LGTM. Could you please update the last modified date and merge yourself?

You should have an invite to join the "WebGL Team" on Github already, which should allow you to merge your own pull requests.

Oletus:
Updated the date in the amended commit. Took a bit of digging to get to the invite, but looks like I'm able to merge now.

qkmiao:
@Oletus , Should these test be fixed in ANGLE? 

FYI: @kenrussell @zhenyao

Kangz:
@Oletus do you have a link to the CL proposed upstream, if any?

Oletus:
If you mean upstream dEQP, I didn't submit a patch there. The GLES standard issue is suffering from a severe case of being stuck in limbo, and I didn't think it made sense to do changes to the tests before people are sure what's the intended behavior.

There are candidate ANGLE changes here, the thing is just that they change the behavior for all GLES, not just WebGL:

https://chromium-review.googlesource.com/#/c/352470/
https://chromium-review.googlesource.com/#/c/352471

Kangz:
Ok, I think it makes sense to land these changes in ANGLE even if they break GLES compatibility: it is a corner case and I don't think GLES applications in the wild use that. Also in the case an application relies on this, we can add an optional outputVersion parameter to the preprocessor.

Oletus:
There was actually a recommendation from the GLES working group to make these cases errors in WebGL just now. So I'll do the changes in ANGLE. It's still a bit unclear what will happen in upstream dEQP.

PR #1524, opened 2016-03-03
Add conformance tests for ImageBitmap from an imagebitmap

xidachen:
To make sure we have a good coverage, this test works like this:
1. Start with a 2*2 ImageData of solid red, red with alpha=0, solid green, green with alpha=0.
2. Using the ImageData, create bitmap1 with premultiplyAlpha=true, and bitmap2 with premultiplyAlpha=false.
3. Create a bunch of bitmaps from bitmap1 and bitmap2 with various constructor options.

@kenrussell @zhenyao @toji Please review.


sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap-from-image-bitmap.js
```diff
@@ -0,0 +1,169 @@
...
+
+        var bitmap1; // bitmap1 will be in premultiplied format
+        var bitmap2; // bitmap2 will be in unpremultiplied format
+        var p1 = createImageBitmap(imageData).then(function(imageBitmap) { bitmap1 = imageBitmap });
+        var p2 = createImageBitmap(imageData, {imageOrientation: "none", premultiplyAlpha: "none"}).then(function(imageBitmap) { bitmap2 = imageBitmap });
+
+        Promise.all([p1, p2]).then(function() {
+            // Now create new ImageBitmaps from the above two ImageBitmaps.
+            var p3 = createImageBitmap(bitmap1).then(function(imageBitmap) { bitmaps.defaultOption1 = imageBitmap });
+            var p4 = createImageBitmap(bitmap1, {imageOrientation: "none", premultiplyAlpha: "default"}).then(function(imageBitmap) { bitmaps.noFlipYPremul1 = imageBitmap });
```
kenrussell: Thinking about this more, is "default" really an appropriate value to use for these tests? It's implementation-specific. These tests run on multiple browsers -- Firefox, Safari, Edge, and Chrome -- and must have no ambiguities. Should we use "premultiply" everywhere that "default" is used?

sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap-from-image-bitmap.js
```diff
@@ -0,0 +1,169 @@
...
+            var p5 = createImageBitmap(bitmap1, {imageOrientation: "none", premultiplyAlpha: "none"}).then(function(imageBitmap) { bitmaps.noFlipYUnpremul1 = imageBitmap });
```
kenrussell: So to clarify, this bitmap is specified as having "unmultiplied" alpha, because bitmap1 had premultiplied alpha? Is this really a case worth testing? It involves a divide-by-zero on the alpha channel, does it not? Premultiplying the alpha channel into the color channels is lossy and I am not sure it can be reliably tested.

sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap-from-image-bitmap.js
xidachen: That's a good point. I think to be safe, we should probably remove all the test cases of "default" and just use "premultiply".

sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap-from-image-bitmap.js
```diff
@@ -0,0 +1,169 @@
...
+            var p6 = createImageBitmap(bitmap1, {imageOrientation: "flipY", premultiplyAlpha: "default"}).then(function(imageBitmap) { bitmaps.flipYPremul1 = imageBitmap });
+            var p7 = createImageBitmap(bitmap1, {imageOrientation: "flipY", premultiplyAlpha: "none"}).then(function(imageBitmap) { bitmaps.flipYUnpremul1 = imageBitmap });
+            var p8 = createImageBitmap(bitmap2).then(function(imageBitmap) { bitmaps.defaultOption2 = imageBitmap });
```
kenrussell: Here again I am dubious of the value of testing the default options for the constructor arguments, in particular premultiplyAlpha="default", since its behavior is allowed to vary across browsers.

kenrussell:
Very nice work overall Xida. These tests are very thorough. Please see comments above. I'll be out of the office for the next few days, so please work with @zhenyao and @toji in particular to finish the review and get this merged. Comments also welcome and solicited @jdashg @grorg @RafaelCintron . These are key test cases for uploads of ImageBitmaps to WebGL textures.

sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap-from-image-bitmap.js
xidachen: Agreed, I think I will remove all "default" cases and replace them with "premultiply" to avoid any ambiguity.

sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap-from-image-bitmap.js
xidachen: IMO, there should be no divide-by-zero here. I think the way premultiplyAlpha works is redChannel*alpha/255, so there should be no divide-by-zero. Am I missing something from you comment?

bitmap1 is created with premultiplyAlpha=true, which means the entire bitmap1's alpha channel should be 255. The point that I put bitmap1 is that it is premultiplied, so its top-right and bottom-right color will always be black when do createImageBitmap(bitmap1, premultiplyAlpha) no matter whether the second parameter here is true or false. bitmap2 is different, it is created with premultiplyAlpha=false, so its alpha channel is 0 for the top-right and bottom-right pixels.

kenrussell: That doesn't sound correct. For an RGBA value (255, 0, 0, 0), the premultiplied alpha version is (0, 0, 0, 0) -- not (0, 0, 0, 255). Alpha needs to be untouched during the premultiplication process in order for blending to work correctly. See http://eelpi.gotdns.org/blog.wiki.html and search for his original post on "Premultiplied alpha" dated 18 March 2015 (created 15 July 2006). I agree that the 0 alpha zeros out the color channels, but the unmultiplication process after that is somewhat ambiguous.

xidachen:
I have made changes to address Ken's comments.
1. Change the "default" to "premultiply" in all ImageBitmap test cases to avoid any ambiguity.
2. Delete two test cases in image-bitmap-from-image-bitmap that doesn't make much sense.

Please review. @kenrussell  @zhenyao @toji

sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image-bitmap-from-image-bitmap.js
```diff
@@ -0,0 +1,165 @@
...
+        runTestOnBindingTarget(gl.TEXTURE_3D, program);
+        program = tiu.setupTexturedQuadWith2DArray(gl, internalFormat);
+        runTestOnBindingTarget(gl.TEXTURE_2D_ARRAY, program);
+
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be no errors");
+        finishTest();
+    }
+
+    function runTestOnBindingTarget(bindingTarget, program) {
+        runOneIteration(bindingTarget, program, bitmaps.defaultOption1, false, true);
```
kenrussell: Again, is there any point in testing bitmaps.defaultOption1 or defaultOption2? I don't think so, and think these should be removed.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image-bitmap.js
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    function runTestOnBindingTarget(bindingTarget, program) {
+        var cases = [
+            { sub: false },
+            { sub: true },
+        ];
+
+        for (var i in cases) {
+            runOneIteration(cases[i].sub, bindingTarget, program, bitmaps.defaultOption1, false, true);
```
kenrussell: I think defaultOption1 should be removed entirely.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image-bitmap.js
```diff
@@ -0,0 +1,198 @@
...
+        var bitmap2; // bitmap2 will be in unpremultiplied format
+        var p1 = createImageBitmap(imageData).then(function(imageBitmap) { bitmap1 = imageBitmap });
+        var p2 = createImageBitmap(imageData, {imageOrientation: "none", premultiplyAlpha: "none"}).then(function(imageBitmap) { bitmap2 = imageBitmap });
+
+        Promise.all([p1, p2]).then(function() {
+            // Now create new ImageBitmaps from the above two ImageBitmaps.
+            var p3 = createImageBitmap(bitmap1).then(function(imageBitmap) { bitmaps.defaultOption1 = imageBitmap });
+            var p4 = createImageBitmap(bitmap1, {imageOrientation: "none", premultiplyAlpha: "premultiply"}).then(function(imageBitmap) { bitmaps.noFlipYPremul1 = imageBitmap });
+            var p5 = createImageBitmap(bitmap1, {imageOrientation: "flipY", premultiplyAlpha: "premultiply"}).then(function(imageBitmap) { bitmaps.flipYPremul1 = imageBitmap });
+            var p6 = createImageBitmap(bitmap2).then(function(imageBitmap) { bitmaps.defaultOption2 = imageBitmap });
```
kenrussell: I think defaultOption1 and defaultOption2 should be removed entirely.

kenrussell:
Looking better, but a few cleanups still needed as far as I can tell.

Please work with @toji and @zhenyao for further reviews for the next few days.

xidachen:
I made some more changes, which deletes all the cases with default options that could be ambiguous. So right now the options are either "premultiply" or "none".

Please review. @zhenyao @toji

zhenyao:
lgtm

PR #1525, opened 2016-03-03
Fix wrong array type in shaderprecision.html

qkmiao:
This patch fixes two bugs exposed by deqp/functional/gles3/shaderprecision.html:
1. Use 32-bit int array for 32-bit int type.
2. Use glVertexAttribIPointer for int vertex array buffer.

C++ code: https://android.googlesource.com/platform/external/deqp/+/master/framework/opengl/gluDrawUtil.cpp#415


zhenyao:
thanks for the fix

PR #1526, opened 2016-03-03
Add tex-unpack-params.html

zhenyao:
This tests all ES3 unpack parameters.

Note that this is not an fully implemented test yet. I plan to add pixel
verification in a followup CL.

### DIFF
--- a/sdk/tests/conformance2/textures/misc/tex-unpack-params.html
+++ b/sdk/tests/conformance2/textures/misc/tex-unpack-params.html
@@ -1,0 +1,415 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/textures/misc/tex-unpack-params.html
```diff
@@ -0,0 +1,409 @@
...
+function runTests() {
+  var gl = wtu.create3DContext("example");
+  if (!gl) {
+    testFailed("Fail to get a WebGL context");
+    return;
+  }
+
+  // For 2D cases, depth, zoffset, imageHeight, skipImages are ignored.
+  var testCases = [
+    { width: 5, height: 7, depth: 3, xoffset: 2, yoffset: 3, zoffset: 1,
```
toji: This is getting a bit pedantic, but it would be nice to see some tests with different width/height/depth/offsets, just to make sure we don't happen to be hitting some really weird edge case. Specifically it would be good to do a few tests where the textures are powers of 2. That should be the "simple" case, but it's worth testing.

Also, if these are going to be the same for a large number of cases in may be better to patch them in down below rather than repeat them over and over. For example:

```
for (var ii = 0; ii < testCases.length; ++ii) {
    var testCase = testCases[ii];
    testCase.width = 5;
    testCase.height = 7;
    // Etc.
    runTestIteration2D(gl, testCase, useUnpackBuffer);
    runTestIteration3D(gl, testCase, useUnpackBuffer);

    testCase.width = 4;
    testCase.height = 8;
    // Etc.
    runTestIteration2D(gl, testCase, useUnpackBuffer);
    runTestIteration3D(gl, testCase, useUnpackBuffer);
 }
```

toji:
LGTM with one suggestion.

sdk/tests/conformance2/textures/misc/tex-unpack-params.html
zhenyao: Good suggestion.  I'll add a few more cases.

kenrussell:
Very nice test @zhenyao and good suggestion @toji . No further comments aside from those. LGTM after the fact.

PR #1527, opened 2016-03-03
Experimental: Please do not merge

xidachen:
This is a better example than the previous one.

What I did here is changing the "tex-image-and-sub-image-2d-with-image-data.js", and set two pixels in the ImageData to be alpha=128 instead of 0. In this case, when the pixel store parameter of premultiplyAlpha=true, the expected color should be very close to 128.

Here is what I get locally:
rgb_rgb_unsigned_short_5_6_5: [132, 0, 0]
rgba_rgba_unsigned_short_4_4_4_4: [136, 0, 0]
rgba_rgba_unsigned_short_5_5_5_1: [132, 0, 0]

PTAL. @kenrussell @zhenyao

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js
@@ -27,6 +27,8 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
     var gl = null;
     var successfullyParsed = false;
     var imageData = null;
+    var halfRedColor = [128, 0, 0];
+    var halfGreenColor = [0, 128, 0];
     var blackColor = [0, 0, 0];
     var redColor = [255, 0, 0];
     var greenColor = [0, 255, 0];
@@ -66,7 +68,7 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
         data[4] = 255;
         data[5] = 0;
         data[6] = 0;
-        data[7] = 0;
+        data[7] = 128;
         data[8] = 0;
         data[9] = 255;
         data[10] = 0;
@@ -74,7 +76,7 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
         data[12] = 0;
         data[13] = 255;
         data[14] = 0;
-        data[15] = 0;
+        data[15] = 128;
 
         runTest();
     }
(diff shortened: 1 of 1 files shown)


xidachen:
@zhenyao In this case, there are two input pixels of [255, 0, 0, 128] and [0, 255, 0, 128]. Here the result of WebGLImageConversion::extractImageData is:
rgb_rgb_unsigned_byte: [0, 0, 0, 255], [10, 0, 0, 0]
rgb_rgb_unsigned_short_5_6_5: [224, 7, 0, 4], [37, 43, 0, 16]
rgba_rgba_unsigned_short_4_4_4_4: [15, 15, 8, 8], [37, 43, 0, 16]
rgba_rgba_unsigned_short_5_5_5_1: [193, 7, 1, 4], [37, 43, 0, 16]

I hope that helps.

kenrussell:
I think there may be a colorspace conversion going on: I think CanvasRenderingContext2D is transforming the input image data to the color profile of the display. @junov may know more. I remember running into this before and, further, that there are some customers who put a simple gray ramp into a canvas and then read it back in order to understand the transfer function from the canvas to the display, for better color fidelity.

xidachen:
OK, thanks for investigating. I think we would at least know that this is not caused by the construction of ImageBitmap. I will change the ImageBitmap tests to take a alpha=0 image so that some of the tests won't fail.

kenrussell:
Well, I would say this certainly bears more investigation! First, it's essential to implement ImageBitmap's colorspaceConversion constructor option; and second, we should understand why it's not possible to lower the tolerances.

junov:
ImageData -> 2D canvas does not involve any color correction in blink's current implementation.  The only discrepancies I would expect to see are compounded losses of precision from the multiply by alpha and bit depth conversion round-trips.  There seem to be more than that going on.  I smell a bug.

zhenyao:
What is the input pixel data to WebGLImageConversion::packPixels? That can help us determine where the pixels get changed. The output pixels do not look reasonable to me.

kenrussell:
I'm reopening this to indicate that investigation is still ongoing.

xidachen:
@zhenyao 
The input is in rgba format and it is [255, 0, 0, 128], [0, 255, 0, 128], and the output is:
rgb_rgb_unsigned_byte: [0, 0, 0, 255], [10, 0, 0, 0]
rgb_rgb_unsigned_short_5_6_5: [224, 7, 0, 4], [37, 43, 0, 16]
rgba_rgba_unsigned_short_4_4_4_4: [15, 15, 8, 8], [37, 43, 0, 16]
rgba_rgba_unsigned_short_5_5_5_1: [193, 7, 1, 4], [37, 43, 0, 16]

zhenyao:
This is surprising.  It looks like a bug in packPixels, but it's been working so far (I think well tested) with the conformance tests.

xidachen:
After debugging with @zhenyao , found that this is not a bug, it is just loss of precision. For example, when converting 128 to unsigned_short_4_4_4_4, it is converted to 8, and when converting back, the pixel value becomes 8 / 15 * 255 = 136.

xidachen:
Closing this after the fact that this is not a bug.

PR #1529, opened 2016-03-04
Print the successful results of UniformBlockCase cases.

xhcao:


kenrussell:
Thanks. We may want to handle this more generally in the future somehow, but fixing it for these tests sounds good.

PR #1530, opened 2016-03-04
Backport: Deleted buffer should be invalid

qkmiao:
When a buffer is deleted, its name immediately becomes invalid. But the
continue using the object.


zhenyao:
lgtm. thx.

PR #1531, opened 2016-03-04
Add a helpful post-commit hook for spec editing

Oletus:
The hook notifies the user if they've changed
specs/latest/*/index.html files but haven't changed the line with the
date, identified by the "Editor's Draft" string in the script.

It's implemented as a post-commit hook so it also runs on git commit
--amend. Users should link or copy it to their .git/hooks/ directory.

### DIFF
--- a/other/post-commit
+++ b/other/post-commit
@@ -1,0 +1,38 @@
+#!/bin/sh
+#
+# Copyright (c) 2016 The Khronos Group Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a
+# copy of this software and/or associated documentation files (the
+# "Materials"), to deal in the Materials without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Materials, and to
+# permit persons to whom the Materials are furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Materials.
+#
+# THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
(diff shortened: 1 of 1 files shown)


Oletus:
Made this mistake enough times by now that it would be nice to make it less likely, and this might be helpful for others too :)

Automatically updating the date would be even nicer, but much more effort to implement.

zhenyao:
This is useful.

msc-:
> Automatically updating the date would be even nicer, but much more effort to implement.

It is pretty easy to do this but there are a couple of caveats...
1. There needs to be a marker around the date, e.g. $date$. This sort of marker will be visible in the published document. An invisible alternative would be an html tag, e,g. <date>...</date>. I don't know if a suitable tag exists though.
2. Users who wish to have the automatic updating must make a one time small edit to the .git/config file in their clone of the repo. See http://stackoverflow.com/questions/18329621/storing-git-config-as-part-of-the-repository for details.

msc-:
If the caveats I mentioned are acceptable, I can make the changes and create a PR. Note that this will not be a post-commit hook. It will use smudge and clean filters. Also note that the .git/config file on the Khronos server will need to be edited.

kenrussell:
If they implement the desired behavior, then yes, please:
1. The smudged date must be the actual last modified date -- not the current date.
2. The date has to be completely substituted -- no stuff left over after smudging.

Otherwise I think the goal should be to implement this as a post-commit hook.

Other comments?

msc-:
> If they implement the desired behavior, then yes, please:
> 1. The smudged date must be the actual last modified date -- not the current date.

Obviously. That is the intention and is not a problem. I've done it before.

> 2. The date has to be completely substituted -- no stuff left over after smudging.

I'm not 100% sure what you mean by this. Do you mean no $date ...$? If so, I'm pretty sure I can use `<span id="date">...</span>` to mark where the date should be inserted in the html. This will be invisible to readers of the document. Note that whether its a smudge or a post-commit there needs to be some way to identify where the date is/should go.

msc-:
In case you missed it, I created PR #1545 for automatically updating the date.

kenrussell:
Thank you Mark. Apologies for not having reviewed it yet but I need to test on multiple platforms, Windows in particular. Will do so after GDC.

msc-:
Not a problem. On Windows I think you will need to have the directory containing bash.exe in your %PATH% for it to work properly with the likes of TortoiseGit. From the Git for Windows command there should be no problem.

PR #1532, opened 2016-03-04
Add pixel verification to tex-unpack-params.html

zhenyao:


sdk/tests/conformance2/textures/misc/tex-unpack-params.html
```diff
@@ -42,6 +42,7 @@
 "use strict";
 
 var wtu = WebGLTestUtils;
+var __verbose__ = false;
```
toji: I feel like we already have a common flag for this somewhere? Not a big deal, though.

toji:
LGTM

PR #1536, opened 2016-03-07
unaligned stride is not allowed in WebGL

Richard-Yunchao:
This feature in WebGL differs from that in gles. See the spec in WebGL: https://www.khronos.org/registry/webgl/specs/latest/1.0/index.html#BUFFER_OFFSET_AND_STRIDE. 

This change can fix all remaining bugs in vertexarrays.html in webgl deqp. 

However, the unaligned tests can still pass in native deqp on Linux desktop. Moreover, when I remove the validation code for vertexAttribPointer in command buffer, the unaligned webgl tests can pass too. The underlying driver will not report gl error in this situation. I suppose the driver will re-calculate the stride and discard the one that user provides. 

PTAL. Thanks!


zhenyao:
lgtm

PR #1537, opened 2016-03-07
Add test for ANGLE ambiguous function call bug

Oletus:
HLSL compiler can't resolve between float4 and float2x2 parameters
when determining which overloaded function is being called. ANGLE
exposes this property to ESSL, but ESSL should be able to resolve
which function is being called in this case. Add a test for this.


Oletus:
This test currently fails on Chrome on Windows but passes on Chrome on Linux.

zhenyao:
I thought we always rewrite function name in ANGLE, appending args into the function name, so overloaded functions will have different names?

Oletus:
Nope, the names go through hashing in WebGL, but the function parameter types are not appended to the name. I'll soon submit a fix to ANGLE that does that in the required cases though.

PR #1538, opened 2016-03-08
There should be at least one enabled vertex attribute array that has a divisor of zero

Richard-Yunchao:
This feature differs from that in gles.
This limitation is added to keep compatible with D3D.
See the WebGL spec: https://www.khronos.org/registry/webgl/specs/latest/2.0/index.html#5.6. 

This change removed the test cases when no vertex divisor equals to zero during drawing. It can fix some bugs in draw.html in WebGL deqp tests. PTAL.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fDrawTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fDrawTests.js
@@ -1004,6 +1004,7 @@ goog.scope(function() {
             if (!spec.valid())
                 continue;
 
+            var hasZeroDivisor = false;
             for (var attrNdx = 0; attrNdx < attributeCount;) {
                 /** @type {boolean} */ var valid;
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fDrawTests.js
```diff
@@ -1030,12 +1031,19 @@ goog.scope(function() {
...
 
             // Do not collapse all vertex positions to a single positions
-            if (spec.primitive != glsDrawTests.DrawTestSpec.Primitive.POINTS)
+            if (spec.primitive != glsDrawTests.DrawTestSpec.Primitive.POINTS) {
                 spec.attribs[0].instanceDivisor = 0;
+                hasZeroDivisor = true;
+            }
+
+            if (hasZeroDivisor == false)
+                continue;
```
zhenyao: Can you add a comment here explaining why we skip such test cases?

Richard-Yunchao:
Thanks for your review, Zhenyao. The comment has been added. PTAL. @zhenyao

zhenyao:
Thanks.  Looks good to me.

PR #1539, opened 2016-03-09
Test the deletion of buffers that are attached to the currently bound VAO

bkotzz:
We want to bind the VAO before deleting the buffers. This means the deleteBuffer calls are attempting to delete the buffers currently attached to the VAO. In this case, the spec mandates that the underlying object should not be deleted, as they are still in use. The previous code was always deleting buffers that were not attached to the current VAO. This change exposes bugs in both Chrome and Firefox.

### DIFF
--- a/sdk/tests/conformance2/vertex_arrays/vertex-array-object.html
+++ b/sdk/tests/conformance2/vertex_arrays/vertex-array-object.html
@@ -476,17 +476,22 @@ function runDeleteTests() {
     // delete the color buffers AND the position buffer.
     gl.bindVertexArray(null);
     for (var ii = 0; ii < colorBuffers.length; ++ii) {
-      gl.deleteBuffer(colorBuffers[ii]);
-      gl.deleteBuffer(elementBuffers[ii]);
       gl.bindVertexArray(vaos[ii]);
-      var boundBuffer = gl.getVertexAttrib(1, gl.VERTEX_ATTRIB_ARRAY_BUFFER_BINDING);
+
+      gl.deleteBuffer(colorBuffers[ii]);
+      gl.deleteBuffer(positionBuffer);
+
       // The buffers should still be valid at this point, since it was attached to the VAO
-      if(boundBuffer != colorBuffers[ii]) {
+      var boundPositionBuffer = gl.getVertexAttrib(0, gl.VERTEX_ATTRIB_ARRAY_BUFFER_BINDING);
+      if(boundPositionBuffer != positionBuffer) { 
+        testFailed("buffer removed too early");
+      }
+      var boundColorBuffer = gl.getVertexAttrib(1, gl.VERTEX_ATTRIB_ARRAY_BUFFER_BINDING);
+      if(boundColorBuffer != colorBuffers[ii]) { 
         testFailed("buffer removed too early");
       }
     }
-    gl.deleteBuffer(positionBuffer);
-
+	
     // Render with the deleted buffers. As they are referenced by VAOs they
     // must still be around.
     for (var ii = 0; ii < colors.length; ++ii) {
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/vertex_arrays/vertex-array-object.html
```diff
@@ -476,17 +476,22 @@
     // delete the color buffers AND the position buffer.
     gl.bindVertexArray(null);
     for (var ii = 0; ii < colorBuffers.length; ++ii) {
-      gl.deleteBuffer(colorBuffers[ii]);
-      gl.deleteBuffer(elementBuffers[ii]);
```
zhenyao: Can we not remove this line?

sdk/tests/conformance2/vertex_arrays/vertex-array-object.html
```diff
@@ -476,17 +476,22 @@
     // delete the color buffers AND the position buffer.
     gl.bindVertexArray(null);
```
zhenyao: This line is useless now.

bkotzz: That's true

sdk/tests/conformance2/vertex_arrays/vertex-array-object.html
bkotzz: If we want to delete element buffers too, shouldn't we check if they're still valid? And then make sure they're invalid after the draw call?

Also, this loop only runs three times, but we have 4 separate VAOs with their own element buffers, should we delete for the fourth one as well?

sdk/tests/conformance2/vertex_arrays/vertex-array-object.html
```diff
@@ -476,17 +476,22 @@
...
     for (var ii = 0; ii < colorBuffers.length; ++ii) {
-      gl.deleteBuffer(colorBuffers[ii]);
-      gl.deleteBuffer(elementBuffers[ii]);
       gl.bindVertexArray(vaos[ii]);
-      var boundBuffer = gl.getVertexAttrib(1, gl.VERTEX_ATTRIB_ARRAY_BUFFER_BINDING);
+
+      gl.deleteBuffer(colorBuffers[ii]);
+      gl.deleteBuffer(positionBuffer);
```
qkmiao: When a buffer, texture, or renderbuffer object is deleted, it is unbound from any
bind points it is bound to in the current context. Since you move deleteBuffer after bindVertexArray, these buffers are  unbound from VAO. You cannot use these buffers below. See https://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf, page 315 "D.1.2 Automatic Unbinding of Deleted Objects".

bkotzz: Maybe I'm misunderstanding the spec here, but in D.1.3, it says the buffer won't be deleted while it is in use, and it defines "in use" as a buffer object being attached to a VAO

---

When a buffer, texture, sampler, renderbuffer, query, or sync object is deleted, its
name immediately becomes invalid (e.g. is marked unused), but the underlying
object will not be deleted until it is no longer in use. A buffer, texture, sampler, or
renderbuffer object is in use while it is attached to any container object (such as a
buffer object attached to a vertex array object, or a renderbuffer or texture attached
to a framebuffer object) or bound to a context bind point in any context.

bkotzz: My understanding here is that there is an automatic unbinding when an object is **deleted**, but an object can't be **deleted** while it is still in use

qkmiao: If the buffer is automatic unbound, it should be not "in use" status. 
You can first deleteBuffer then bindVertexArray. You can see the deleted buffer can still be used.

kenrussell: If the VAO is bound to the current context (via bindVertexArray), *then* the buffer is supposed to be automatically unbound from the VAO during DeleteBuffers. Otherwise, the DeleteBuffers call is supposed to act as an "unref" operation, but the buffer is supposed to remain alive until the VAO is deleted.

Firefox and Chrome are correctly reporting that the test fails with this pull request applied -- the pull request is incorrect, sorry.

PR #1540, opened 2016-03-10
Fix fragment shader to work for GPUs with no highp support.

willeastcott:


kenrussell:
Thanks Will. I meant to get to this today but ran out of time. Thanks for contributing the fix.

PR #1541, opened 2016-03-10
Only delete buffers that are not bound in the current context

bkotzz:
When the loop ends, vao[2] is bound to the current context, so when we delete the position buffer, it should be unbound, and therefore unusable in the subsequent draw call for vao[2]. Since all VAOs have a position buffer attached, we bind null before deleting the position buffer, so that all VAOs still have a bound position buffer for the draw call.


bkotzz:
As a follow up, Chrome was successfully able to draw even though the position buffer should have been unbound. Are we interested in adding tests where we delete buffers that are bound to the current context, and check that they automatically get unbound?

kenrussell:
Yes, definitely. Please add more tests like this. Thanks.

PR #1544, opened 2016-03-11
Make it easier to test around unpack_alignment driver bug.

zhenyao:


zhenyao:
Really simple code to make testing easier. Merging myself.

kenrussell:
Very good. LGTM after the fact.

PR #1545, opened 2016-03-14
Set last edit date automatically

msc-:
Add smudge and clean filters to automatically add the last edit date ("author date" in Git speak) on checkout and remove it on commit.

For this to work, the first time you pull these changes you need to do the following with the current
directory set to your clone of the repo

```bash
./.gitconfig.sh
rm specs/latest/*/index.html
git checkout !$
```

The first command adds a line to the local git config file (.git/config) in your repo clone that includes the config file for the "dater" filter (.gitconfig). The remaining commands force a new checkout of the index.html files to smudge them with the date.

I have only targeted `latest` for automatic dating as inserting the necessary keywords into the 1.0.* versions would cause their last edit dates to change to today.

The date now includes time and timezone. There is not much flexibility about date presentation in git log. The choices are:

- Mon Mar 14 15:12:30 2016 +0900
- Mon, 14 Mar 2016 15:12:30 +0900
- 2016-03-14 15:12:30 +0900
- 2016-03-14

The first format is currently used.

It is likely possible to pipe the timestamp into another program to reformat it if none of the above is acceptable.


msc-:
I have tested this on Linux (Ubuntu Trusty) and Windows (10). It works fine. On Windows I tried the Git for Windows shell and TortoiseGit. It works in both. Note that TortoiseGitMerge for a long time had a bug that it did not run the clean filter so would show a difference. That has finally been fixed. I tested with TortoiseGit 2.0.0.0. I originally developed this on OS X (10.11) so that is fine too.

kenrussell:
Mark, thank you and sorry for taking so long to review this. This is excellent. I've tested it on Mac OS and Windows. Once this is merged, I'll add a "gitconfig.bat" which works with Git for Windows, but outside of the Cygwin shell. (I found that setting the include.path to '../.gitconfig' didn't work in this configuration, but '..\.gitconfig' did.)

I'm going to sync up with the Khronos webmaster before merging this so that he can make the needed changes on the server side right afterward. Thank you again!

msc-:
Good idea to sync up. James is already aware this is coming because I asked him what OS the server is running so I could make sure the filter script will run there.

kenrussell:
Thanks again for this contribution Mark. Synced up with James. Merging.

PR #1546, opened 2016-03-14
fix bug for primitive restart test

Richard-Yunchao:
I suppose that this is a typo. See the original code in native deqp: https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/gles3/functional/es3fPrimitiveRestartTests.cpp#545. 

This small change can fix bugs in primitiverestart.html in webgl deqp. 

PTAL. Thanks!


kenrussell:
Very nice Yunchao. Thank you for tracking this down. It looks like a hack that was put in place because of a misdiagnosis of the cause of test failures. Merging.

PR #1549, opened 2016-03-16
Draw correct fullscreen square in rasterizerdiscard.html

qkmiao:
This aligns with C++ code:
https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fRasterizerDiscardTests.cpp#347


kenrussell:
Thanks for fixing this.

PR #1550, opened 2016-03-16
Adding new test to vertex array object. 

bkotzz:
New test where we delete buffers that are attached to the currently bound VAO, expecting that they get automatically unbound, and shouldn't be accessible or usable in a draw call. We expect these buffers to exist until they are unbound from their last binding point.


kenrussell:
Excellent. Thank you for adding this test.

PR #1551, opened 2016-03-16
Ternary operators not permitted on arrays

bkotzz:
From section 5.7 in the ESSL spec, ternary operators are not permitted to operate on arrays or structures.

https://www.khronos.org/registry/gles/specs/3.2/GLSL_ES_Specification_3.20.withchanges.pdf

### DIFF
--- a/sdk/tests/conformance2/glsl3/array-complex-indexing.html
+++ b/sdk/tests/conformance2/glsl3/array-complex-indexing.html
@@ -77,17 +77,6 @@ void main() {
   color = (a == 2.0) ? vec4(0, 1.0, 0, 1.0) : vec4(1.0, 0, 0, 1.0);
 }
 </script>
-<script id="fshader-ternary" type="x-shader/x-fragment">#version 300 es
-precision mediump float;
-out vec4 color;
-
-void main() {
-  float a[2] = float[2](1.0, 1.0);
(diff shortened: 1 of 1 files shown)


kenrussell:
I thought this was one gray area in the spec where clarification was being made. @Oletus is that correct?

Oletus:
Yes, the latest proposal was to allow ternary operator on arrays. This should work with the majority of existing GLES drivers too, though old ones from some specific vendors might have trouble with it.

bkotzz:
It's failing on Chrome and Firefox right now. Is the new proposal coming out soon enough that it's worth leaving?

kenrussell:
Olli: assigning this to you to decide how to handle.

Oletus:
The new spec has actually been out for a little while now, just hadn't noticed that when I made my last comment: https://www.khronos.org/registry/gles/specs/3.0/GLSL_ES_Specification_3.00.pdf

The final decision was to make ternary operator on arrays optional in GLSL ES 3.00, and WebGL will keep disallowing them. There's also already a negative test that conflicts with the one being removed here, so this test can simply be removed. I'm merging this patch.

bkotzz:
I think the newest one is actually 3.04 but I could be wrong. Thanks.

Oletus:
The shading language got a 3.00.6 update recently, spec is still 3.0.4.

PR #1553, opened 2016-03-21
Fix sdk/tests/extra/canvas-compositing-test.html.

kenrussell:


kenrussell:
Apparently broken during a refactor. Trivial change; merging myself.

PR #1554, opened 2016-03-22
Fix bugs in multisample.html

qkmiao:
1. Add support to skip unavailable tests.
2. Reset GL states in deinit to avoid affecting subsequent tests.

With this patch, all tests passed in multisample.html.
C++ code:
https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/tes3TestPackage.cpp#69
https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fMultisampleTests.cpp#281

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
@@ -212,6 +212,51 @@ goog.scope(function() {
     /** Copy the constructor */
     es3fMultisampleTests.MultisampleCase.prototype.constructor = es3fMultisampleTests.MultisampleCase;
 
+    /* Rest states */
+    es3fMultisampleTests.MultisampleCase.prototype.deinit = function() {
+        gl.colorMask(true, true, true, true);
+        gl.depthMask(true);
+
+        gl.clearColor(0.0, 0.0, 0.0, 0.0);
+        gl.clearDepth(1.0);
+        gl.clearStencil(0);
+
+        gl.disable(gl.STENCIL_TEST);
+        gl.disable(gl.DEPTH_TEST);
+        gl.disable(gl.BLEND)
+        gl.disable(gl.SAMPLE_COVERAGE);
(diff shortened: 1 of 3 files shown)


sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
```diff
@@ -212,6 +212,51 @@ goog.scope(function() {
...
+
+        gl.clearColor(0.0, 0.0, 0.0, 0.0);
+        gl.clearDepth(1.0);
+        gl.clearStencil(0);
+
+        gl.disable(gl.STENCIL_TEST);
+        gl.disable(gl.DEPTH_TEST);
+        gl.disable(gl.BLEND)
+        gl.disable(gl.SAMPLE_COVERAGE);
+        gl.disable(gl.SAMPLE_ALPHA_TO_COVERAGE);
```
kenrussell: Is it worth a comment about why these calls are needed here but not in the native test?

kenrussell:
Thank you. Looks good overall. One question above, which can be addressed in a future pull request. Merging now.

sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
qkmiao: In TestCaseWrapper::deinit, glu::resetState() will do this cleanup at the end of each test.
https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/tes3TestPackage.cpp#69
https://android.googlesource.com/platform/external/deqp/+/master/framework/opengl/gluStateReset.cpp#35

kenrussell: Thanks for pointing that out. Would it be possible to refactor the JavaScript port in this area to be closer to the C++ code's organization? The intent was that they behave nearly identically.

qkmiao: Yes, I do agree with you we should keep JS port same as C++ port as much as possible. I will add this task to my TODO list.

PR #1555, opened 2016-03-23
Skip unvailable test case in shaderloop.html

qkmiao:
C++ code: https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fShaderLoopTests.cpp#1198


kenrussell:
LGTM. Thank you.

PR #1556, opened 2016-03-23
Create WebGL 2 context by default in conformance2 tests

Oletus:
This makes running the tests more convenient: ?webglVersion=2
doesn't need to be added to the URL to run the tests standalone.


kenrussell:
Thank you, this will make running these tests standalone much more convenient.

PR #1558, opened 2016-03-23
Convert expected colors from linear -> SRGB when reading from an SRGB texture.

bkotzz:
When we have a texture with internal format SRGB8_ALPHA8, the texture uses SRGB color encoding, and doing a glClearColor/glClear operation converts from linear to SRGB. When we do a glReadPixels, the read format is RGBA, as glReadPixels doesn't support SRGB8_ALPHA8 reads, so no SRGB -> linear conversion occurs. When we compare against the expected color, we must convert linear -> SRGB. We must convert depending on the internal format (as the internal type tells us if the pixels are in SRGB format), and regardless of the read type (since the read type can't be SRGB8_ALPHA8 anyways).

### DIFF
--- a/sdk/tests/conformance2/reading/read-pixels-from-fbo-test.html
+++ b/sdk/tests/conformance2/reading/read-pixels-from-fbo-test.html
@@ -129,7 +129,7 @@ function convertToSRGB(val) {
   }
 }
 
-function denormalizeColor(srcInternalFormat, destFormat, destType, color) {
+function denormalizeColor(srcInternalFormat, destType, color) {
   var result = color.slice();
   var tol = 0;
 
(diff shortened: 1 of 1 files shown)


zhenyao:
I agree the current code isn't correct because the readback format/type can't be sRGB.

However, I couldn't find spec for the behavior when the fbo internal format is in sRGB space, and whether ReadPxiels(RGBA, UNSIGNED_TYPE) already performed the converts or not.  I expect it to already be converted, but from your patch, you expect it not.  Do you have spec to support that?

zhenyao:
Never mind.  I found the resolution in https://www.opengl.org/registry/specs/EXT/texture_sRGB.txt that the GL's pixel path is bind of color space.  So LGTM

zhenyao:
Unfortunately the two GL drivers I tested (Linux NVidia, Mac AMD) both don't do color version to sRGB with glClearColor()

@Oletus @grorg

zhenyao:
Never mind.  Apparently we need to Enable(FRAMEBUFFER_SRGB)  on desktop GL.

PR #1559, opened 2016-03-23
Make timestamps optional in timestamp WebGL test.

iewell:
Allow the counter bits for timestamps to be 0 for the WebGL
EXT_disjoint_timer_query conformance test and disable the timestamp
portion of the test if that is the case. Certain backing drivers for
WebGL such as the OSX drivers and ANGLE on D3D11 do not support
timestamps and it is within the spec to disable timestamp functionality
in the extension by setting the number of counter bits for the
timestamps to be 0. This updates the test to check if the timestamp
counter bits is set to 0 and to skip the timestamp portion of the test
if that is the case.

### DIFF
--- a/sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
+++ b/sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
@@ -51,6 +51,7 @@ var elapsed_query = null;
 var timestamp_query1 = null;
 var timestamp_query2 = null;
 var availability_retry = 500;
+var timestamp_counter_bits = 0; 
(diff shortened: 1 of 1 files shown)


iewell:
@kenrussell

kenrussell:
Thank you for updating the test. Your changes look good. The test is getting pretty complicated at this point, with a lot of dependencies between variables declared and initialized early in the test, and code later in the test. Please think it over and see if you can figure out a way to make the individual timestamp and elapsed time tests more modular. Thanks. Merging this now.

PR #1561, opened 2016-03-23
Split texturefilter* tests into multiple ones.

zhenyao:
This is because each one of them takes too long to run.

This patch only splits a few of them.  Will need more work to split them
all to reasonable running time per test.


zhenyao:
Merging myself since this is trivia split up.  So I can continue working on more splitting. (worried this tedious manual work may get lost)

kenrussell:
Thanks for doing this Mo. LGTM after the fact.

PR #1563, opened 2016-03-24
Update GLSL ES 3.00 references in WebGL 2.0 to 3.00.6

Oletus:
This GLSL ES revision was made after taking feedback from the WebGL
WG. The spec references that are updated in this patch were reviewed
to make sure they are still correct.


kenrussell:
Thanks, looks good.

PR #1564, opened 2016-03-24
Fix transform feedback boolean output varying

qkmiao:
Boolean output is converted to an int type varying whose name is
constructed by adding 'o_' before the original name. Correct name should
be specified for transform feedback output.

C++ code: 
https://android.googlesource.com/platform/external/deqp/+/master/modules/glshared/glsShaderExecUtil.cpp#446

Also see code about creating vertex shader in line 156 of glsShaderExecUtil.js:
https://www.khronos.org/registry/webgl/sdk/tests/deqp/modules/shared/glsShaderExecUtil.js


zhenyao:
lgtm

PR #1565, opened 2016-03-24
Set correct inputs for special cases

qkmiao:
Scalarsize should be considered when set inputs.

C++ code: 
https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fShaderCommonFunctionTests.cpp#532
The C++ code has bug too. It doesn't move ptr for different special values.

Other place in the same file does the same thing:
see code in es3fShaderCommonFunctionTests.TruncCase.prototype.getInputValues().

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fShaderCommonFunctionTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fShaderCommonFunctionTests.js
@@ -514,9 +514,21 @@ goog.scope(function() {
         if (gluShaderUtil.isDataTypeFloatOrVec(type)) {
             // Special cases.
             // [dag] The special cases are 1, -1, and 0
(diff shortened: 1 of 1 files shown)


kenrussell:
Thanks Qiankun for this patch. Which specific tests does this fix? Are tests broken in the native dEQP harness because of the bug in the C++ code?

qkmiao:
It fixes a lot of tests in deqp/functional/gles3/shadercommonfunction.html. 
The native dEQP doesn't fail, because the unset input values was initialized to 0. Real output and ref output are correctly calculated based on 0.

kenrussell:
Thanks Qiankun for tracking this down. The changes look correct and I'll alert @phaulos that it looks like there's a bug in the native tests.

PR #1566, opened 2016-03-24
Use correct trunc function to align with C++ code

qkmiao:
C++ code:
https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fShaderCommonFunctionTests.cpp#1251


zhenyao:
lgtm

PR #1567, opened 2016-03-24
Add a python script to generate all texturefilter* tests

zhenyao:
It's not fully done yet.  The current status of this script is
to make minimum change to the existing files.


toji:
LGTM

PR #1571, opened 2016-03-24
Performing some optimization on tcuFloat.

toji:
tcuFloat has been identified as a major bottleneck in multiple DEQP tests. These optimizations (most specifically the change to `tcuFloat.numberToHalfFloat`) speed up some tests like `texturefiltering_3d_formats` around 100x. I think other tests may benefit from similar conversions of `tcuFloat.numberToFloat10` and `tcuFloat.numberToFloat11`.

### DIFF
--- a/sdk/tests/deqp/framework/common/tcuFloat.js
+++ b/sdk/tests/deqp/framework/common/tcuFloat.js
@@ -60,7 +60,7 @@ tcuFloat.FloatDescription = function(exponentBits, mantissaBits, exponentBias, f
 tcuFloat.FloatDescription.prototype.zero = function(sign) {
     return tcuFloat.newDeFloatFromParameters(
         deMath.shiftLeft((sign > 0 ? 0 : 1), (this.ExponentBits + this.MantissaBits)),
-        new tcuFloat.FloatDescription(this.ExponentBits, this.MantissaBits, this.ExponentBias, this.Flags)
+        this
     );
 };
 
@@ -72,7 +72,7 @@ tcuFloat.FloatDescription.prototype.zero = function(sign) {
 tcuFloat.FloatDescription.prototype.inf = function(sign) {
     return tcuFloat.newDeFloatFromParameters(((sign > 0 ? 0 : 1) << (this.ExponentBits + this.MantissaBits)) |
         deMath.shiftLeft(((1 << this.ExponentBits) - 1), this.MantissaBits), //Unless using very large exponent types, native shift is safe here, i guess.
-        new tcuFloat.FloatDescription(this.ExponentBits, this.MantissaBits, this.ExponentBias, this.Flags)
+        this
     );
 };
 
@@ -82,7 +82,7 @@ tcuFloat.FloatDescription.prototype.inf = function(sign) {
  */
 tcuFloat.FloatDescription.prototype.nan = function() {
     return tcuFloat.newDeFloatFromParameters(deMath.shiftLeft(1, (this.ExponentBits + this.MantissaBits)) - 1,
-        new tcuFloat.FloatDescription(this.ExponentBits, this.MantissaBits, this.ExponentBias, this.Flags)
+        this
     );
 };
 
@@ -123,7 +123,7 @@ tcuFloat.FloatDescription.prototype.construct = function(sign, exponent, mantiss
             ),
             deMath.BinaryOp.OR
         ),
-        new tcuFloat.FloatDescription(this.ExponentBits, this.MantissaBits, this.ExponentBias, this.Flags)
+        this
     );
 };
 
@@ -156,7 +156,7 @@ tcuFloat.FloatDescription.prototype.constructBits = function(sign, exponent, man
             mantissaBits,
             deMath.BinaryOp.OR
         ),
-        new tcuFloat.FloatDescription(this.ExponentBits, this.MantissaBits, this.ExponentBias, this.Flags)
+        this
     );
 };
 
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/framework/common/tcuFloat.js
```diff
@@ -60,7 +60,7 @@ tcuFloat.FloatDescription = function(exponentBits, mantissaBits, exponentBias, f
 tcuFloat.FloatDescription.prototype.zero = function(sign) {
     return tcuFloat.newDeFloatFromParameters(
         deMath.shiftLeft((sign > 0 ? 0 : 1), (this.ExponentBits + this.MantissaBits)),
-        new tcuFloat.FloatDescription(this.ExponentBits, this.MantissaBits, this.ExponentBias, this.Flags)
+        this
```
kenrussell: Were all of these methods just allocating unnecessary FloatDescription objects all along? Or is there a subtle semantic change here?

sdk/tests/deqp/framework/common/tcuFloat.js
```diff
@@ -596,37 +633,81 @@ tcuFloat.numberToFloat11 = function(value) {
...
-tcuFloat.float11ToNumber = function(float11) {
-    var x = tcuFloat.newDeFloatFromParameters(float11, tcuFloat.description11);
-    return x.getValue();
-};
+tcuFloat.float11ToNumber = (function() {
+    var x = tcuFloat.newDeFloatFromParameters(0, tcuFloat.description11);
+    return function(float11) {
+        x.deFloatParametersNumber(float11);
+        return x.getValue();
+    };
```
kenrussell: To clarify: is one of the main optimizations to reuse the same deFloat instance over and over again for these conversions?

sdk/tests/deqp/framework/common/tcuFloat.js
```diff
@@ -596,37 +633,81 @@ tcuFloat.numberToFloat11 = function(value) {
...
+        if( val < 0x33000000 )  { // too small for subnormal
+            return sign; // becomes +/-0
+        }
+        val = ( fbits & 0x7fffffff ) >> 23; // tmp exp for subnormal calc
+        return sign | ( ( fbits & 0x7fffff | 0x800000 ) // add subnormal bit
+             + ( 0x800000 >>> val - 102 ) // round depending on cut off
+             >> 126 - val ); // div by 2^(1-(exp-127+15)) and >> 13 | exp=0
+    };
+})();
+// Previously was: return tcuFloat.newFloat16(value).bits();
```
kenrussell: Is there any way to reuse a deFloat instance, as it looks like was done above, so that the specialized half-float code doesn't need to be added? The possibility of slight bugs or behavioral differences compared to the original code is concerning.

zhenyao:
That deFloat only accounts for 24%.  The major cutback (I think) is that we go through this direct float-to-half-float path instead of the generic float encoding path.

zhenyao:
This is probably a deviation from deqp tests, but we test 10 formats in texturefiltering tests, and we could probably implement all the 10 short cuts like this one here.  As @toji pointed out, it is a 100x optimization, and will save almost an hour of running time in WebGL deqp tests.

We just need to make sure the new encoding coder output the same thing as the deqp c++ tests.

My two cents.

kenrussell:
Awesome work! This will make these tests finally practical to run.

A few questions. Also, could you please spell out very clearly in this pull request's description what all of the optimizations were? For example, "cache and re-use deFloat instances during all toNumber conversions", etc. Thanks. That will help when looking back at this code.

Please request review from @zhenyao as well and please feel free to merge by yourselves.

kenrussell:
If the numberToHalfFloat optimization is the key one, then certainly let's incorporate it -- the time savings are staggering.

sdk/tests/deqp/framework/common/tcuFloat.js
toji: As far as I can tell there's no purpose to creating new copies of the descriptor. The descriptor properties are never altered, so these seem to be unnecessary copies. I'm guessing it's that was directly ported from the C++ code where it likely made more sense.

sdk/tests/deqp/framework/common/tcuFloat.js
toji: Yes, exactly. Specifically, these functions were creating new deFloats, which in turn did very expensive TypedArray allocations (sometimes multiple times) which were then immediately discarded after producing the value. By only allocating the deFloat once we avoid TypedArray allocation churn.

sdk/tests/deqp/framework/common/tcuFloat.js
zhenyao: Can you format the above copied code a bit?

Let's merge this in for now and get the tests running on bots.

The original code path is still available.  Let's not delete it in case this new path turns out to be buggy in corner cases.

zhenyao:
Feel free to merge yourself after clean up the format a little.

sdk/tests/deqp/framework/common/tcuFloat.js
toji: I've tried that and it leads to some side effects that I don't yet fully understand. It seems that calling `floatInstance.deFloatNumber(value); return tcuFloat.description16.convert(floatInstance);` Changes the attributes of the floatInstance in such a way that subsequent calls to those same methods don't produce the expected output. I plan to look into that further.

Regardless of that, however, `tcuFloatDescriptor.convert` allocates new `tcuFloat`s, which in turn allocates new `TypedArray`s, which is where the primary bottleneck is coming from.

PR #1572, opened 2016-03-25
Generate -0 for roundeven reference value in shadercommonfunction.html

qkmiao:
-0 is treated as different with 0 for high precision case. WebGL do
return -0 but the reference output is still 0. This patch updates the
reference output calculation. 

C++ dEQP also has such bug, but due to the
input isn't set correctly -0 case will not appear in C++ dEQP.
For input initialization bug, please see discussion in https://github.com/KhronosGroup/WebGL/pull/1565.


zhenyao:
great

PR #1575, opened 2016-03-25
Handle multiple input and output case

qkmiao:
This aligns with C++ code.

https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fShaderCommonFunctionTests.cpp#324

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fShaderCommonFunctionTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fShaderCommonFunctionTests.js
@@ -276,27 +276,34 @@ goog.scope(function() {
     es3fShaderCommonFunctionTests.CommonFunctionCase.prototype.iterate = function() {
         /** @type {number} */ var numInputScalars = es3fShaderCommonFunctionTests.computeTotalScalarSize(this.m_spec.inputs);
         /** @type {number} */ var numOutputScalars = es3fShaderCommonFunctionTests.computeTotalScalarSize(this.m_spec.outputs);
-        /** @type {goog.TypedArray} */ var inputData;
-        /** @type {goog.TypedArray} */ var outputData;
+        /** @type {Array<goog.TypedArray>} */ var inputData = [];
+        /** @type {Array<goog.TypedArray>} */ var outputData = [];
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fShaderCommonFunctionTests.js
```diff
@@ -276,27 +276,34 @@ goog.scope(function() {
...
         /** @type {number} */ var numInputScalars = es3fShaderCommonFunctionTests.computeTotalScalarSize(this.m_spec.inputs);
         /** @type {number} */ var numOutputScalars = es3fShaderCommonFunctionTests.computeTotalScalarSize(this.m_spec.outputs);
-        /** @type {goog.TypedArray} */ var inputData;
-        /** @type {goog.TypedArray} */ var outputData;
+        /** @type {Array<goog.TypedArray>} */ var inputData = [];
+        /** @type {Array<goog.TypedArray>} */ var outputData = [];
         /** @type {gluShaderUtil.DataType} */ var inputType = this.m_spec.inputs[0].varType.getBasicType();
         /** @type {gluShaderUtil.DataType} */ var outputType = this.m_spec.outputs[0].varType.getBasicType();
-        /** @type {Array<number>} */ var inputValues;
+        /** @type {Array<Array<number>>} */ var inputValues;
```
zhenyao: Is this really necessary?  It looks to me getInputValues() always only return one array, never multiple arrays. So maybe Array<number> is OK? This could simply the code quite a bit.

qkmiao: You are right. No multiple input tests currently. Only have multiple output tests. I just align with C++ code for input.

qkmiao: Only unary functions are implemented currently. In future, binary and ternary function should be added. I see a comment near line 1824: // min, max, clamp, mix, step, smoothstep.

zhenyao: OK

PR #1576, opened 2016-03-26
Batch file for installation of .gitconfig on Windows without Cygwin.

kenrussell:
Follow-on to Mark Callow's https://github.com/KhronosGroup/WebGL/pull/1545 , for Windows users that do not use Cygwin.


msc-:
Thanks for the .bat file but is it really necessary? The existing script works fine in Git Bash, which I used for my testing and which is a standard part of Git for Windows. I do not have Cygwin. Is there anyone using Windows and Git who does not use Git for Windows?

kenrussell:
In-house we run Git for Windows from the Windows command prompt. Its bundled "bash.exe" and similar commands are not installed on the %PATH%. When invoking your .gitconfig.sh on Windows, did you manually run git's bundled bash.exe?

msc-:
The filter commands use `bash -c "..."` so that they will work provided the directory of `bash.exe` is in your %PATH%. I think I manually added that to my %PATH%. If the filters are working for you, you must already have it in your %PATH%. However the script for adding the filters to .git/config is a regular bash script. From the windows command prompt it would have to be called as `bash -c .gitconfig.sh`.

I have no objection to having a .bat file. README.md should be updated to give both commands.

kenrussell:
Thanks Mark. Will update README.md in this pull request after your #1545 lands. (Still waiting for confirmation from the webmaster.) My plan is to rename .gitconfig.sh to install-gitconfig.sh, in case you have any comment about that.

I think Chromium's batch file wrapper for Git for Windows (from depot_tools: https://www.chromium.org/developers/how-tos/install-depot-tools ) sets up the PATH each time a Git command runs so that the subordinate commands like bash are available. That might be why I have git but not bash on my path. Still, the batch file for installation seems to be convenient. Will document both.

kenrussell:
Updated README.md. Merging now.

msc-:
I just fixed a small typo in README.md that was hiding the call to install-gitconfig.bat. Since it was such a simple change and clearly an error I fixed it directly on GitHub.

kenrussell:
OK, thanks Mark. In general I'd appreciate it even for these small fixes if you'd push a pull request and then merge it yourself. People who watch the repo get email notifications about pull requests, but not direct commits to the repo.

msc-:
> On Mar 30, 2016, at 9:55 AM, Ken Russell <notifications@github.com> wrote:
> 
> OK, thanks Mark. In general I'd appreciate it even for these small fixes if you'd push a pull request and then merge it yourself. People who watch the repo get email notifications about pull requests, but not direct commits to the repo.
> 
> 


I didn’t realize notifications relied on pull requests. I’ll make sure to do so in the future.

Regards

    -Mark

PR #1577, opened 2016-03-28
Fix wrong types in shaderpackingfunction.html

qkmiao:
This patches fix several bugs:
1. Clear status of tcuFloat when setting a new value.
2. Wrong type for output and ref output.
3. Support NaN value.

With this patch, all tests passed in shaderpackingfunction.html.

C++:
https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fShaderPackingFunctionTests.cpp

https://android.googlesource.com/platform/external/deqp/+/master/framework/common/tcuFloat.hpp#242

### DIFF
--- a/sdk/tests/deqp/framework/common/tcuFloat.js
+++ b/sdk/tests/deqp/framework/common/tcuFloat.js
@@ -279,6 +279,10 @@ tcuFloat.deFloat = function() {
 
     this.m_buffer = null;
     this.m_array = null;
+    this.bitValue = undefined;
+    this.signValue = undefined;
+    this.expValue = undefined;
+    this.mantissaValue = undefined;
 
     this.m_value = 0;
 };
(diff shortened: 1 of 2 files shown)


sdk/tests/deqp/framework/common/tcuFloat.js
```diff
@@ -415,8 +415,7 @@ tcuFloat.deFloat.prototype.getBitRange = function(begin, end) {
  * @return {number}
  */
 tcuFloat.deFloat.prototype.bits = function() {
-    if (typeof this.bitValue === 'undefined')
-        this.bitValue = deMath.arrayToNumber(this.array());
+    this.bitValue = deMath.arrayToNumber(this.array());
```
zhenyao: Does this affect perf of running texture* tests?  We found out that this deFloat class is the main bottleneck (took like 95%+ of running time) in these tests and probably other tests.

If it does, can we implement a dirty bit? So we don't have to recompute each time.

zhenyao:
Mostly look good, but I have a concern about perf.  Can you test it out?

qkmiao:
It's better to cache the value all the time. I updated the pr to do clear each time we set a new jsnumber.

zhenyao:
lgtm
thanks

PR #1578, opened 2016-03-29
Set default context version for WebGL 1.0 and 2.0 texture tests.

kenrussell:
This allows these tests to be run as the test harness would do so without the ?webglVersion query string, while still allowing that query string to override the default version. In this way, the WebGL 1.0 tests can still be run against WebGL 2.0 contexts.

This is a follow-on to #1556, which made a similar change to the rest of the WebGL 2.0 conformance tests.

### DIFF
--- a/sdk/tests/conformance/textures/canvas/tex-image-and-sub-image-2d-with-canvas-rgb-rgb-unsigned_byte.html
+++ b/sdk/tests/conformance/textures/canvas/tex-image-and-sub-image-2d-with-canvas-rgb-rgb-unsigned_byte.html
@@ -51,7 +51,7 @@ function testPrologue(gl) {
     return true;
 }
 
-generateTest("RGB", "RGB", "UNSIGNED_BYTE", testPrologue, "../../../resources/")();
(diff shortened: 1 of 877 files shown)


kenrussell:
NOTE: this change is not complete; please do not merge. The generator script needs to be run, which will change hundreds of files and which can not be reviewed effectively. I will do this and merge myself after your reviews.

@zhenyao @Oletus, in particular, please review.

@Richard-Yunchao @toji @jdashg @grorg FYI.

The motivation for this change comes from debugging https://codereview.chromium.org/1822643002/ ; I lost over an hour today because I forgot that ?webglVersion=2 was required in order for these tests to run correctly. (I still haven't figured out the root cause for the test failures in that patch.)

kenrussell:
Also, for the record: I spot checked every type of test (canvas, image_bitmap_from_blob, etc.) in conformance/textures/ and conformance2/textures/ with one format/internalformat/type combination, though I didn't run the entire conformance suite after this change (including the generated changes).

zhenyao:
Thanks for catching and fixing this.  LGTM.  Feel free to merge yourself when the auto-generated changes are included.

kenrussell:
Thanks for your review Mo. Ran the generator script. Merging.

PR #1579, opened 2016-03-29
Fix wrong parameter VAO

qkmiao:
gluDrawUtil.VertexArrayBinding requires stride and offset for VAO binding.

C++ code:
https://android.googlesource.com/platform/external/deqp/+/master/framework/opengl/gluDrawUtil.cpp#220

With this patch, part of tests in shadermatrix.html passed. But there are still some other failures.

### DIFF
--- a/sdk/tests/deqp/framework/opengl/gluDrawUtil.js
+++ b/sdk/tests/deqp/framework/opengl/gluDrawUtil.js
@@ -87,10 +87,10 @@ gluDrawUtil.namedBindingsToProgramLocations = function(gl, program, inputArray, 
             //assert(binding.location >= 0);
             var location = gl.getAttribLocation(program, cur.name);
             if (location >= 0) {
-                if (cur.offset)
-                    location += cur.offset;
+                if (cur.location >= 0)
+                    location += cur.location;
                 // Add binding.location as an offset to accomodate matrices.
-                outputArray.push(new gluDrawUtil.VertexArrayBinding(cur.type, location, cur.components, cur.elements, cur.data));
+                outputArray.push(new gluDrawUtil.VertexArrayBinding(cur.type, location, cur.components, cur.elements, cur.data, cur.stride, cur.offset));
             }
         } else {
             outputArray.push(cur);
@@ -493,7 +493,7 @@ gluDrawUtil.newFloatVertexArrayBinding = function(name, numComponents, numElemen
  */
 gluDrawUtil.newFloatColumnVertexArrayBinding = function(name, column, rows, numElements, stride, data) {
     var bindingPoint = gluDrawUtil.bindingPointFromName(name);
-    bindingPoint.offset = column;
+    bindingPoint.location = column;
     var arrayPointer = new gluDrawUtil.VertexArrayPointer(gluDrawUtil.VertexComponentType.VTX_COMP_FLOAT,
         gluDrawUtil.VertexComponentConversion.VTX_COMP_CONVERT_NONE, rows, numElements, stride, data);
     return gluDrawUtil.vabFromBindingPointAndArrayPointer(bindingPoint, arrayPointer);


sdk/tests/deqp/framework/opengl/gluDrawUtil.js
```diff
@@ -40,7 +40,7 @@ var gluShaderProgram = framework.opengl.gluShaderProgram;
  */
 gluDrawUtil.VertexArrayBinding = function(type, location, components, elements, data, stride, offset) {
     this.type = type;
-    this.location = location === undefined ? -1 : location;
+    this.location = location === undefined ? 0 : location;
```
zhenyao: Why is this needed?  Who called with location undefined? From the caller side, gl.getAttribLocation() always returns a number.

kenrussell: It looks to me like this maps to the "location" field in the native dEQP's BindingPoint class:
https://android.googlesource.com/platform/external/deqp/+/master/framework/opengl/gluDrawUtil.hpp#105

and that (given the change below to the addition of "location"), setting it to -1 was causing the wrong value to be added / computed for attributes' locations in these tests.

Qiankun, is that correct?

kenrussell:
This LGTM assuming it's been tested, but Mo's concerns should be addressed before landing.

sdk/tests/deqp/framework/opengl/gluDrawUtil.js
qkmiao: You are correct, ken.  In gluDrawUtil.bindingPointFromName = function(name, location), location may be undefined, we should initialize it to 0. 

For Mo's concern, I didn't find the undefined case for location argument in gluDrawUtil.VertexArrayBinding. Should we keep it as before or change to "this.location = location;"?

kenrussell: It would be safest to leave the code as:
this.location = location === undefined ? 0 : location;
since the previous code explicitly tested against "undefined". I don't know how all the callers work.

Looking through gluDrawUtil.js, there may be other suspect pieces of code. gluDrawUtil.bindingPointFromName sets the location to -1 if the location is undefined (I see you modified it below as well), and functions like gluDrawUtil.vabFromBindingPointAndArrayPointer take the BindingPoint and set the location of the new gluDrawUtil.VertexArrayBinding to it.

-1 is the canonical "undefined attribute location" value. For this reason, I wonder whether a more reliable fix would be to change gluDrawUtil.namedBindingsToProgramLocations to check:
  if (cur.location >= 0)
rather than
  if (cur.location)

and leave this and gluDrawUtil.bindingPointFromName alone. What do you think?

qkmiao: Thanks for this suggestion.  I updated this PR.

kenrussell: Can you now undo this change? Do the tests still pass?

sdk/tests/deqp/framework/opengl/gluDrawUtil.js
```diff
@@ -433,7 +433,7 @@ gluDrawUtil.bindingPointFromLocation = function(location) {
  * return {gluDrawUtil.BindingPoint}
  */
 gluDrawUtil.bindingPointFromName = function(name, location) {
-    location = location === undefined ? -1 : location;
+    location = location === undefined ? 0 : location;
```
kenrussell: Can you now undo this change as well?

sdk/tests/deqp/framework/opengl/gluDrawUtil.js
qkmiao: The tests still passed after undo these two changes. When undo "if(cur.location >= 0)", the tests failed again.

kenrussell: That's what I would expect. Please undo this change and the change to bindingPointFromName. Thanks.

qkmiao: I updated this pr. Thank you, ken.

kenrussell:
Very good. Thank you Qiankun. LGTM

I leave this to @zhenyao to re-review and merge.

zhenyao:
Thanks @kenrussell @qkmiao for getting to the bottom of this and coming up this clean fix.
lgtm

PR #1580, opened 2016-03-29
Fix a bug with sRGB and RED texture uploading through ImageBitmap.

zhenyao:
When we premultiply a pixel value 255 in sRGB space with alpha = 0.5,
the result in sRGB space should be 55 instead of 128.


zhenyao:
@kenrussell @xidachen please review. with this, the SRGB8 and SRGBA_ALPHA8 tests pass.

zhenyao:
Note that the latest patch fix all existing image-bitmap test failures (both the sRGB and RED formats)

kenrussell:
Bravo. LGTM.

PR #1581, opened 2016-03-30
Fix display of primitive type name in primitive restart tests.

kenrussell:


kenrussell:
Trivial fix. Merging myself.

PR #1583, opened 2016-03-30
Make the code aligned with native deqp, in order to fix shader compilation failure

Richard-Yunchao:
See the corresponding code in native deqp: https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/glshared/glsBuiltinPrecisionTests.cpp#4936. 

PTAL, Thanks!


zhenyao:
lgtm
thanks

kenrussell:
Very good catch @Richard-Yunchao . LGTM after the fact.

PR #1584, opened 2016-03-31
fix error to make builtinprecision31.html pass

Richard-Yunchao:
A small change. The builtin function should be 'roundEven', instead of 'roundeven'. After the change applied, builtinprecision31.html can pass in Chromium.


kenrussell:
Nice catch. Thanks for fixing this. LGTM

PR #1585, opened 2016-03-31
Fix notes about installing gitconfig.

msc-:
This is a simple correction but Ken, please check the information is valid for your Windows environment before merging.


kenrussell:
Thanks Mark, this looks fine.

PR #1586, opened 2016-03-31
Add a test to cover an ANGLE rendering bug.

null77:
This bug would occur after the app would render several times with the
same vertex attributes and buffers, but using a different start offset.
One of the buffers would likely have to be DYNAMIC.

See http://anglebug.com/1327 and http://crbug.com/594509

### DIFF
--- a/sdk/tests/conformance/rendering/draw-with-changing-start-vertex-bug.html
+++ b/sdk/tests/conformance/rendering/draw-with-changing-start-vertex-bug.html
@@ -1,0 +1,133 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/rendering/draw-with-changing-start-vertex-bug.html
```diff
@@ -0,0 +1,129 @@
...
+varying mediump vec4 color;
+void main(void)
+{
+    gl_FragColor = color;
+}
+</script>
+
+<script>
+"use strict";
+description("Test calling drawArrays with repeatedly with a different start vertex");
```
zhenyao: Could you add a comment linking to the angle/chromium bug IDs?  It's in the commit message, but it's easier if it's in the test.

sdk/tests/conformance/rendering/draw-with-changing-start-vertex-bug.html
```diff
@@ -0,0 +1,129 @@
...
+
+var program = wtu.setupProgram(gl, ["vshader", "fshader"], ["position", "test", "expected"]);
+
+var vertexCount = 24;
+var testData = new Float32Array(vertexCount);
+var expectedData = new Float32Array(vertexCount);
+
+for (var index = 0; index < vertexCount; ++index) {
+    testData[index] = index;
+    expectedData[index] = index;
```
zhenyao: It's no big deal, but if they are just the same, you only need one.

zhenyao:
Looks good in general with some minor suggestions.  Thanks for getting this into WebGL conformance suite.

null77:
Addressed comments and added a merged patch.

zhenyao:
Thanks. lgtm

kenrussell:
Very nice. Thanks for putting together a reduced test case for this. LGTM after the fact.

PR #1588, opened 2016-04-01
Optimized tcuFloat.numberToFloat10 and numberToFloat11

toji:
Allows tests like deqp/functional/gles3/texturefiltering_3d_formats_07.html and friends to go significantly faster. (From around 8 seconds runtime to < 2 seconds on my system.) `tcuFloat.convertFloat32Inline` is a specialization of `tcuFloat.FloatDescription.convert` that takes a floating point number in and returns a number formatted with the description provided. Because we know the input format and don't expect a tcuFloat back out it can run much much faster.


zhenyao:
Looks good to me.

This is really awesome optimization!

PR #1589, opened 2016-04-01
Optimize some code in tcuTexLookupVerifier.js

zhenyao:
This cut off 2/3 of running time for tests like deqp/functional/gles3/texturefiltering_3d_formats_29.html

(the test is still too slow and needs further optimization though)


toji:
LGTM, thanks!

PR #1590, opened 2016-04-01
Fix a bug in tcuFloat.js

zhenyao:
It's introduced in https://github.com/KhronosGroup/WebGL/pull/1588

(and I failed to catch it during review)


zhenyao:
This is a really simple fix.  Merging myself.

toji:
Ah, yeah. Sorry about that. Thanks for the fix!

PR #1592, opened 2016-04-06
avoid init irrelevant cases

Richard-Yunchao:
PTAL. Thanks!

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fBuiltinPrecisionTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fBuiltinPrecisionTests.js
@@ -35,14 +35,15 @@ goog.scope(function() {
 
     /**
      * @param {*} context
+     * @param {number} caseId test case Id
      * @return {tcuTestCase.DeqpTest}
      */
-    es3fBuiltinPrecisionTests.createBuiltinPrecisionTests = function(context) {
+    es3fBuiltinPrecisionTests.createBuiltinPrecisionTests = function(context, caseId) {
         /** @type {tcuTestCase.DeqpTest} */
         var group = tcuTestCase.newTest('precision', 'Builtin precision tests');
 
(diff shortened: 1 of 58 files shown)


kenrussell:
Is this correct? How is this argument interpreted?

kenrussell:
+ @zhenyao

sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
```diff
@@ -5253,69 +5253,67 @@ var setParentClass = function(child, parent) {
     /**
      * @return {glsBuiltinPrecisionTests.CaseFactories}
```
zhenyao: Please update the documentation for caseId here.

sdk/tests/deqp/functional/gles3/es3fBuiltinPrecisionTests.js
```diff
@@ -37,12 +37,12 @@ goog.scope(function() {
      * @param {*} context
      * @return {tcuTestCase.DeqpTest}
```
zhenyao: Please update the documentation for caseId here.

zhenyao:
Mostly looks good.  Could you please update the two documentation places?

Richard-Yunchao:
Thanks for your review, @kenrussell and @zhenyao . The code has updated accordingly, please have another look. 
Each builtin-precision file (builtinprecision00.html to builtinprecision55.html) test one builtin function, say min, max, mod, ceil, clamp and so forth. The original code will init all builtin functions before the test (but it only test one function). Then uses 'range' to filter irrelverant cases. It is complex and time-consuming. Now each builtin-precision test file only initiate the case it tests. It is simple, direct and clear now: builtinprecision*.html (caseId) -> es3fBuiltinPrecision.js (caseId) -> glsBuiltinPrecision.js (caseId).

zhenyao:
lgtm

PR #1593, opened 2016-04-06
Shorten texture test filenames.

kenrussell:
Shorten texture test filenames.

Some of these file names are over 255 characters which is (still, in the year 2016) causing problems on Windows machines.

Remove the element type from the test's file name; it's already present in the directory containing the test. Also remove the image-and-sub-image string. These will comfortably get all of the test names under this limit.


kenrussell:
@zhenyao please review.

@xidachen FYI.

This pull request currently only contains the simple change to the generator. I'll run it and delete the old files and add the new ones before merging.

sdk/tests/py/tex_image_test_generator.py
```diff
@@ -127,7 +127,7 @@
     
 def GenerateFilename(dimension, element_type, internal_format, format, type):
   """Generate test filename."""
-  filename = ("tex-image-and-sub-image-" + dimension + "d-with-" + element_type + "-" +
+  filename = ("tex-image-and-sub-image-" + dimension + "d-" +
```
zhenyao: Maybe also remove "image-and-sub-image-"?

zhenyao:
I have one suggestion, but lgtm one way or the other.  Feel free to generate the whole CL and land yourself.

sdk/tests/py/tex_image_test_generator.py
kenrussell: Good idea. Will do.

kenrussell:
Updated the pull request with the massive renaming. Spot-checked the tests in the conformance test runner (both 1.0 and 2.0 tests). Merging now.

PR #1594, opened 2016-04-07
Round float color to uint8 and clamp it in [0, 255]

qkmiao:
This aligns with C++ code:
https://android.googlesource.com/platform/external/deqp/+/master/framework/common/tcuFuzzyImageCompare.cpp#58
https://android.googlesource.com/platform/external/deqp/+/master/framework/common/tcuFuzzyImageCompare.cpp#186


zhenyao:
lgtm
one question: does this affect some tests passing/failing?

qkmiao:
Thanks for merging. This doesn't affect tests failing. It just reduces image difference between ref image and result image for failing tests. I found the issue when I was debugging shadertexturefunction00.html.

PR #1595, opened 2016-04-07
Fix the gl_VertexID test reference not being rendered.

Kangz:
PTAL, this fixes the gl_VertexID test reference in two parts:
 - Fix it not rendering anything because the software FS output to the wrong variable
 - Add triangle rasterization because quad rasterization didn't match with what the GPU produced.

The drawQuads -> drawTriangles is very significant as many tests inderectly used drawQuads to generate reference images. The dEQP functional draw test have equivalent results with Chrome stable between both version, and the simple_reference_renderer test results are more correct now.

While this is in review, I'm running the WebGL2 CTS on Chrome stable both with and without this change to see if any new failure is introduced. New failures would either point to bugs in the drawTriangles, or bugs in the tests when they used drawQuads to render triangles.

I still need to run the closure compiler to typecheck the change but I don't expect any errors.

cc @kenrussell

### DIFF
--- a/sdk/tests/deqp/framework/opengl/simplereference/sglrReferenceContext.js
+++ b/sdk/tests/deqp/framework/opengl/simplereference/sglrReferenceContext.js
@@ -3785,7 +3785,7 @@ goog.scope(function() {
 
         var primitiveType = sglrReferenceUtils.mapGLPrimitiveType(primitive);
 
-        var renderFunction = rrRenderer.drawQuads;
+        var renderFunction = rrRenderer.drawTriangles;
         if (primitiveType == rrRenderer.PrimitiveType.LINES ||
             primitiveType == rrRenderer.PrimitiveType.LINE_STRIP ||
             primitiveType == rrRenderer.PrimitiveType.LINE_LOOP)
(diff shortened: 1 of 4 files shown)


kenrussell:
Thank you @Kangz for this excellent piece of debugging work. Apologies that you were the one that ran into the bugs in this area. The JS code is quite different in structure from the original C++ but all of your changes look good.

Please indicate when you're done with your testing and we'll merge it then.

Please also note that I am pretty sure the closure compiler scripts are broken at the moment -- I have a to-do item about pulling in the top-of-tree Closure dependencies into this workspace, and then filing a bug about their pulling the webgl2.js extern file into their workspace. Don't worry too much about that. The more important thing is that the tests currently using the referencerasterizer improve in correctness with your change.

Kangz:
I was trying to run the whole test suite with and without this patch, but it was causing WebGL to "hit a snack". Now trying with Chromium's run_gpu_tests on Windows. Something in the builtin variable test is breaking the compositor in Chromium Linux (opened a bug for it).

Kangz:
Found a bug in the PR with testing (rrUtil is not using rrRenderer directly). Now Chromium's run_gpu_tests <args for webgl2 deqp tests> passes all the tests on NVIDIA Windows. Merging.

PR #1597, opened 2016-04-08
fix errors about infinity

Richard-Yunchao:
This logic error lead to many failures in builtinprecision*.html. After the change applied, most of the failed files under the builtinprecision category can pass now. 

PTAL. Thanks!


Richard-Yunchao:
Sorry, but this is not correct. The original implementation is aligned with the native one. I will submit another pr soon.

PR #1599, opened 2016-04-08
Misc optimizations for DEQP tests

toji:
Found that a fair amount of time was being spent in some tests creating intermediate color objects, so I've put in place some optimizations that allow colors to be returned as an unsigned int. This adds some bit shifting math but the reduction in temporary object creation appears to be worth it. For example,  shaved of about 5 seconds a piece from shaders/swizzles.html and shaders/conversions.html on my machine, and a full **20 seconds** (out of 75) off of gles3/draw.html. Other tests may benefit, but I haven't checked rigorously.

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsDrawTests.js
+++ b/sdk/tests/deqp/modules/shared/glsDrawTests.js
@@ -3170,12 +3170,15 @@ goog.scope(function() {
     };
 
     /**
-     * @param {tcuRGBA.RGBA} c
+     * @param {number} c1
+     * @param {number} c2
+     * @param {Array<number>} threshold
      * @return {boolean}
      */
-    glsDrawTests.isBlack = function(c) {
-        // ignore alpha channel
-        return c.getRed() == 0 && c.getGreen() == 0 && c.getBlue() == 0;
+    glsDrawTests.compareUintRGB8 = function(c1, c2, threshold) {
+        return (Math.abs(((c1 >> 16) & 0xff) - ((c2 >> 16) & 0xff)) <= threshold[0] &&  // Red
+                Math.abs(((c1 >> 8) & 0xff) - ((c2 >> 8) & 0xff)) <= threshold[1] &&    // Green
+                Math.abs((c1 & 0xff) - (c2 & 0xff)) <= threshold[2]);                   // Blue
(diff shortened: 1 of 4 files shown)


sdk/tests/deqp/modules/shared/glsDrawTests.js
```diff
@@ -3246,19 +3249,19 @@ goog.scope(function() {
         /** @type {tcuRGBA.RGBA} */ var c3;
```
zhenyao: You will need to update type documentation of c1, c2, c3

sdk/tests/deqp/modules/shared/glsDrawTests.js
```diff
@@ -3388,29 +3386,29 @@ goog.scope(function() {
 
         for (var y = 1; y < height - 1; ++y)
         for (var x = 1; x < width - 1; ++x) {
-            /** @type {tcuRGBA.RGBA} */ var refPixel = tcuRGBA.newRGBAFromArray(reference.getPixel(x, y));
-            /** @type {tcuRGBA.RGBA} */ var screenPixel = tcuRGBA.newRGBAFromArray(result.getPixel(x, y));
+            /** @type {tcuRGBA.RGBA} */ var refPixel = reference.getPixelUintRGB8(x, y);
+            /** @type {tcuRGBA.RGBA} */ var screenPixel = result.getPixelUintRGB8(x, y);
```
zhenyao: Same here, type documentation needs to be updated

sdk/tests/deqp/modules/shared/glsDrawTests.js
```diff
@@ -3467,16 +3465,16 @@ goog.scope(function() {
...
             /** @type {boolean} */ var isOkScreenPixel = glsDrawTests.pixelNeighborhoodContainsColor(reference, x, y, screenPixel, compareThreshold); // reference image has a matching pixel nearby (~= If something is drawn on screen, it must be drawn to reference too.)
```
zhenyao: Same here, update the type documentation

zhenyao:
Great work.  Code looks good to me, but documentations need to be updated.  Feel free to merge yourself once they are addressed.

PR #1604, opened 2016-04-12
Fix texture initialization for shadertexturefunction05.html

qkmiao:
C++ code:
https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fShaderTextureFunctionTests.cpp#728


zhenyao:
Thanks for catching this!

PR #1606, opened 2016-04-13
Fix the logic error in the js implementation of modf() function

Richard-Yunchao:
Please have a look. Thanks.

I have written a native code to modify the modf() functionality. The result turned out that the int_part of the modf(x) equals to ceil(x) if x is negative.

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
+++ b/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
@@ -1117,7 +1117,9 @@ var setParentClass = function(child, parent) {
     glsBuiltinPrecisionTests.ApplyVar.prototype.doEvaluate = function(ctx) {
         return this.m_func.applyFunction(ctx,
                     ctx.env.lookup(this.m_args.a), ctx.env.lookup(this.m_args.b),
-                    ctx.env.lookup(this.m_args.c), ctx.env.lookup(this.m_args.d));
+                    ctx.env.lookup(this.m_args.c), ctx.env.lookup(this.m_args.d),
+                    [this.m_args.a.getName(), this.m_args.b.getName(),
+                    this.m_args.c.getName(), this.m_args.d.getName()]);
     };
 
     /**
@@ -1225,8 +1227,8 @@ var setParentClass = function(child, parent) {
      * @param {glsBuiltinPrecisionTests.Intervals=} Iarg3
      * @return {glsBuiltinPrecisionTests.Intervals}
      */
-    glsBuiltinPrecisionTests.Func.prototype.applyFunction = function(ctx, Iarg0, Iarg1, Iarg2, Iarg3) {
-        return this.applyArgs(ctx, new glsBuiltinPrecisionTests.Tuple4(Iarg0, Iarg1, Iarg2, Iarg3));
+    glsBuiltinPrecisionTests.Func.prototype.applyFunction = function(ctx, Iarg0, Iarg1, Iarg2, Iarg3, variablenames) {
+        return this.applyArgs(ctx, new glsBuiltinPrecisionTests.Tuple4(Iarg0, Iarg1, Iarg2, Iarg3), variablenames);
     };
 
     /**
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
In addition, the native implementation use reference/pointer to change the data stored in ctx.env.m_map, while the js implementation lost the reference when use ctx.env.lookup() to construct a tuple4 variable in doEvaluate() and applyFunction(), although js can transfer objects directly during arguments passing when call into a function.

With this change applied, builtinprecision35 can pass now.

sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
```diff
@@ -1117,7 +1117,9 @@ var setParentClass = function(child, parent) {
     glsBuiltinPrecisionTests.ApplyVar.prototype.doEvaluate = function(ctx) {
         return this.m_func.applyFunction(ctx,
                     ctx.env.lookup(this.m_args.a), ctx.env.lookup(this.m_args.b),
```
Richard-Yunchao: The lookup() function will return a reference/address of an object in native implementation. However, in the js implementation, it returns the value of the object. So the object itself is lost during argument passing in js implementation. 
See the implementation of deEvaluate() in native code https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/glshared/glsBuiltinPrecisionTests.cpp#1404. And it can assign value to this referenced object returned by lookup(), see https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/glshared/glsBuiltinPrecisionTests.cpp#4556. But in JS implementation, it is invalid to do so. 

As a result, the assignment on line 4571 in this js file can not change the value stored in the original object (ctx.env.m_map). It is useless.

sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
```diff
@@ -4555,20 +4557,20 @@ var setParentClass = function(child, parent) {
         return 'modf';
     };
 
-    glsBuiltinPrecisionTests.Modf.prototype.doApply = function(ctx, iargs) {
+    glsBuiltinPrecisionTests.Modf.prototype.doApply = function(ctx, iargs, variablenames) {
         var intPart;
         var func1 = function(x) {
-            intPart = Math.floor(x);
+            intPart = x >= 0 ? Math.floor(x) : Math.ceil(x);
```
qkmiao: Does Math.trunc() work here?

Richard-Yunchao: Yes, it works. Thanks for your suggestion, @qkmiao .

sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
```diff
@@ -4555,20 +4557,20 @@ var setParentClass = function(child, parent) {
...
         var func2 = function(x) {
-            return Math.floor(x);
+            return Math.trunc(x);
         };
 
         var fracIV = tcuInterval.applyMonotone1p(func1, iargs.a);
         var wholeIV = tcuInterval.applyMonotone1p(func2, iargs.a);
 
-        iargs.b = wholeIV;
+        ctx.env.m_map[variablenames[1]] = wholeIV;
```
zhenyao: A question: if you always only use one variable name, why do you need to pass four in?

Richard-Yunchao: Yes. I have the same consideration. In fact, for this specific bug, I can use 
        ctx.env.m_map["out1"] = wholeIV. 
The variablenames[1] is definitely to be "out1".  
But there are failures in some other builtinprecision related test files. I suppose some other failures may have the same issue, then we can use whatever variable names we want in those cases. Besides, passing all 4 arguments is not a heavy operation. WDYT? @zhenyao .

zhenyao: Sounds good.

PR #1607, opened 2016-04-13
Relax pixel comparison tolerance.

zhenyao:
0.5 maps to 127 or 128, but in RGBA4, if it's 8, 8 / 15 * 255 is 136.

This makes some test cases pass on Mac OSX.


kenrussell:
Excellent. LGTM

PR #1608, opened 2016-04-14
Unbind vao before delete position buffer.

mephisto41:
In previous loop, we will bind vao for checking vertex arrtib. If we
don't unbind the vao before deleting position buffer, the link between
this vao and position buffer will be broken.

### DIFF
--- a/conformance-suites/1.0.2/conformance/extensions/oes-vertex-array-object.html
+++ b/conformance-suites/1.0.2/conformance/extensions/oes-vertex-array-object.html
@@ -539,6 +539,8 @@ function runDeleteTests() {
         testFailed("buffer removed too early");
       }
     }
+
+    ext.bindVertexArrayOES(null);
     gl.deleteBuffer(positionBuffer);
 
     // Render with the deleted buffers. As they are referenced by VAOs they
--- a/conformance-suites/1.0.3/conformance/extensions/oes-vertex-array-object.html
+++ b/conformance-suites/1.0.3/conformance/extensions/oes-vertex-array-object.html
@@ -539,6 +539,8 @@ function runDeleteTests() {
         testFailed("buffer removed even though it is still attached to a VAO");
       }
     }
+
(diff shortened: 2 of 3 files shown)


kenrussell:
Thanks for this patch. It looks like #1520 accidentally introduced this bug. CC'ing @zhenyao @qkmiao.

I can see that this fixes the test failure in Firefox. I'm pretty sure that Chrome is doing the wrong thing here which is why it was passing before. It would be really ideal if you would go back and write a negative test which exposes the bug in Chrome. I think that sdk/tests/conformance2/vertex_arrays/vertex-array-object.html was just updated with a similar test already.

Also, I note that Firefox Nightly on OS X fails runArrayBufferBindTests without this fix, which is strange because that test should be using completely different buffer objects and VAOs than runDeleteTests. I suspect there's another bug in Firefox and would encourage you to investigate it and write a test for it.

mephisto41:
The later problem is runDeleteTests() test generate a GL error without this fix and didn't consume it. So later test will call setupProgram which found a GL error and failed to setup program.

conformance-suites/1.0.2/conformance/extensions/oes-vertex-array-object.html
```diff
@@ -539,6 +539,8 @@
         testFailed("buffer removed too early");
       }
```
qkmiao: Should we do unbind here?

kenrussell: Sorry for the delay replying. I don't think it's necessary to do the unbind at this point. colorBuffers[ii] and elementBuffers[ii] are only referenced by vaos[ii] and not any other entry in the vaos array. Therefore leaving vaos[ii] bound will not affect the deletion or unbinding of colorBuffers[ii+1] and elementBuffers[ii+1]. Unbinding the last bound vao at the end of the loop should be sufficient.

PR #1611, opened 2016-04-14
Correct result and reference pixels for cube texture filtering

qkmiao:
Align with C++ code. This fixes texturefiltering_cube_formats_04.html.

C++ code:
https://android.googlesource.com/platform/external/deqp/+/master/modules/glshared/glsTextureTestUtil.cpp#2043

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsTextureTestUtil.js
+++ b/sdk/tests/deqp/modules/shared/glsTextureTestUtil.js
@@ -1006,12 +1006,24 @@ glsTextureTestUtil.execSample = function(src, params, texCoord, lod) {
  * @param {Array<number>} pixel
  * @param {Array<number>} scale
  * @param {Array<number>} bias
- * @return {Array<number>}
  */
 glsTextureTestUtil.applyScaleAndBias = function(pixel, scale, bias) {
-    var pixel1 = deMath.multiply(pixel, scale);
-    var pixel2 = deMath.add(pixel1, bias);
-    return pixel2;
+    pixel[0] = pixel[0] * scale[0] + bias[0];
+    pixel[1] = pixel[1] * scale[1] + bias[1];
+    pixel[2] = pixel[2] * scale[2] + bias[2];
+    pixel[3] = pixel[3] * scale[3] + bias[3];
+};
+
+/**
+ * @param {Array<number>} pixel
+ * @param {Array<number>} scale
+ * @param {Array<number>} bias
+ */
+glsTextureTestUtil.deapplyScaleAndBias = function(pixel, scale, bias) {
+    pixel[0] = (pixel[0] - bias[0]) / scale[0];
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/modules/shared/glsTextureTestUtil.js
```diff
@@ -1922,8 +1922,10 @@ glsTextureTestUtil.computeTextureLookupDiffCube = function(
             qpWatchDog_touch(watchDog);*/
 
         for (var px = 0; px < result.getWidth(); px++) {
-            var resPix = result.getPixel(px, py);
-            var refPix = reference.getPixel(px, py);
+            /** @type {Array<number>} */
+            var resPix = deMath.divide(deMath.subtract(result.getPixel(px, py), sampleParams.colorBias), sampleParams.colorScale);
+            /** @type {Array<number>} */
+            var refPix = deMath.divide(deMath.subtract(reference.getPixel(px, py), sampleParams.colorBias), sampleParams.colorScale);
```
zhenyao: Thanks for tracking this down.

Can we do inline computation instead of using deMath?  deMath operations are the bottleneck of perf because they create new vectors at each operation.

qkmiao:
Thanks @zhenyao for pointing the performance issue. I updated this pr. Please review again.

sdk/tests/deqp/modules/shared/glsTextureTestUtil.js
```diff
@@ -2527,9 +2549,17 @@ glsTextureTestUtil.computeTextureLookupDiff2DArray = function(result, reference,
 
         for (var px = 0; px < result.getWidth(); px++) {
             /** @type {Array<number>} */
-            var resPix = deMath.divide(deMath.subtract(result.getPixel(px, py), sampleParams.colorBias), sampleParams.colorScale);
+            resPix[0] = (resPix[0] - sampleParams.colorBias[0]) / sampleParams.colorScale[0];
```
zhenyao: You forget to declare resPix and initialize it with result.getPixel(px, py) here.

sdk/tests/deqp/modules/shared/glsTextureTestUtil.js
```diff
@@ -1922,8 +1925,18 @@ glsTextureTestUtil.computeTextureLookupDiffCube = function(
             qpWatchDog_touch(watchDog);*/
 
         for (var px = 0; px < result.getWidth(); px++) {
+            /** @type {Array<number>} */
             var resPix = result.getPixel(px, py);
+            resPix[0] = (resPix[0] - sampleParams.colorBias[0]) / sampleParams.colorScale[0];
```
zhenyao: This similar processing repeated six times in this CL.  Could you define an inline helper function so the code looks cleaner without harming perf?

qkmiao:
Thanks for review. Updated the pr.

zhenyao:
Great!

PR #1612, opened 2016-04-14
Fix the error against builtinprecision30.html in webgl deqp.

Richard-Yunchao:
In JS implementation of round(), (-inf) - (-inf) equals to NaN.
While it is -0 in native deqp case.

So when the input of round() is inf or -inf, the fract part is neither >=0.5 nor <= 0.5. Then it returns a wrong result to the caller. 

PTAL, thanks!

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
+++ b/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
@@ -1655,7 +1655,10 @@ var setParentClass = function(child, parent) {
         var fract = x - truncated;
         var ret = new tcuInterval.Interval();
 
-        if (Math.abs(fract) <= 0.5)
+        // When x is inf or -inf, truncated would be inf or -inf too. Then fract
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
```diff
@@ -1655,7 +1655,7 @@ var setParentClass = function(child, parent) {
         var fract = x - truncated;
```
Richard-Yunchao: The JS implementation is exactly the same with the native code, see the function applyPoint() in native deqp: https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/glshared/glsBuiltinPrecisionTests.cpp#3168. The the result of this line differs from that in native code. When the input x equals to inf / -inf, the result is NaN. Then neither of these two ifs below (between line 1658 and line 1661) will be executed.

kenrussell:
Excellent diagnosis Yunchao. LGTM after the fact.

PR #1613, opened 2016-04-14
Improve global variable initialization test

Oletus:
Add a test where a ternary operation condition references a uniform,
and another test where a uniform struct is assigned to a global.

An option to specify uniform values for GLSL tests is added for the
second test. This could be also used to refactor other existing tests
that currently are not using the GLSL test helpers because they need
to set uniforms.

### DIFF
--- a/sdk/tests/js/glsl-conformance-test.js
+++ b/sdk/tests/js/glsl-conformance-test.js
@@ -70,15 +70,25 @@ var fShaderDB = {};
  * the parameters for one shader out, in which case the default shader will be
  * used.
  * vShaderSource: the source code for vertex shader
+ * vShaderId: id of an element containing vertex shader source code. Used if
+ *   vShaderSource is not specified.
  * vShaderSuccess: true if vertex shader compilation should
  *   succeed.
  * fShaderSource: the source code for fragment shader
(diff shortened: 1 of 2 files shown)


Oletus:
This should cover a few more cases that trigger particularly tricky code paths in ANGLE.

sdk/tests/js/glsl-conformance-test.js
```diff
@@ -256,6 +256,15 @@ function runOneTest(gl, info) {
...
   gl.useProgram(program);
+
+  if (info.uniforms !== undefined) {
+    for (var i = 0; i < info.uniforms.length; ++i) {
+      var uniformLocation = gl.getUniformLocation(program, info.uniforms[i].name);
+      gl[info.uniforms[i].functionName](uniformLocation, info.uniforms[i].value);
+      debug(info.uniforms[i].name + ' set to ' + info.uniforms[i].value);
+    }
+  }
+
```
zhenyao: Please update the documentation of info above runOneTest() function.

zhenyao:
The code looks good.  Please update the documentation, then feel free to merge yourself.

Oletus:
Amended the commit with updated documentation.

PR #1614, opened 2016-04-14
Fix a bug in defaultvertexattribute.html

zhenyao:
The bug exists in the original c++ tests.


toji:
LGTM! Good catch.

zhenyao:
@kenrussell do you have instructions how to submit fix to the original c++ deqp tests?

kenrussell:
Great catch Mo. Unfortunately it's not trivial to submit fixes because the open-source dEQP:
https://source.android.com/devices/graphics/testing.html
https://android.googlesource.com/platform/external/deqp
(deqp-dev branch)

 is quite far behind the internal repository.

Let me put you in touch with the dEQP team. The best way to get this fixed may be to just file a bug with them.

PR #1615, opened 2016-04-15
Fix sRGB format cube texture lookup

qkmiao:
This fixes deqp/functional/gles3/texturefiltering_cube_formats_08.html.

C++ code:
isNearestSampleResultValid() call lookup<ScalarType>(level, sampler, x, y, coordZ). Due looup is a C++ template funciton the function really call is "inline Vector<float, 4> lookup (const ConstPixelBufferAccess& access, const Sampler& sampler, int i, int j, int k)". 

https://android.googlesource.com/platform/external/deqp/+/master/framework/common/tcuTexLookupVerifier.cpp#541
https://android.googlesource.com/platform/external/deqp/+/master/framework/common/tcuTexLookupVerifier.cpp#65

This patch isn't aligned with C++ code because JS doesn't support template function. Any suggestions about the fix?

### DIFF
--- a/sdk/tests/deqp/framework/common/tcuTexLookupVerifier.js
+++ b/sdk/tests/deqp/framework/common/tcuTexLookupVerifier.js
@@ -702,7 +702,12 @@ goog.scope(function() {
 
         for (var i = minI; i <= maxI; i++) {
             /** @type {number} */ var x = tcuTexVerifierUtil.wrap(sampler.wrapS, i, level.getWidth());
-            /** @type {Array<number>} */ var color = tcuTexLookupVerifier.lookupScalar(level, sampler, x, coordY, 0);
+            /** @type {Array<number>} */ var color;
+            if (tcuTexLookupVerifier.isSRGB(level.getFormat())) {
+                color = tcuTexLookupVerifier.lookupFloat(level, sampler, x, coordY, 0);
+            } else {
+                color = tcuTexLookupVerifier.lookupScalar(level, sampler, x, coordY, 0);
+            }
 
             if (tcuTexLookupVerifier.isColorValid(prec, color, result))
                 return true;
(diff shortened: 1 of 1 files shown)


zhenyao:
@toji might have a better idea if we can emulate c++ template in js
but this fix looks acceptable to me

sdk/tests/deqp/framework/common/tcuTexLookupVerifier.js
```diff
@@ -741,7 +741,12 @@ goog.scope(function() {
         for (var i = minI; i <= maxI; i++) {
             /** @type {number} */ var x = tcuTexVerifierUtil.wrap(sampler.wrapS, i, level.getWidth());
             /** @type {number} */ var y = tcuTexVerifierUtil.wrap(sampler.wrapT, j, level.getHeight());
-            /** @type {Array<number>} */ var color = tcuTexLookupVerifier.lookupScalar(level, sampler, x, y, coordZ);
+            /** @type {Array<number>} */ var color;
+            if (tcuTexLookupVerifier.isSRGB(level.getFormat())) {
+                color = tcuTexLookupVerifier.lookupFloat(level, sampler, x, y, coordZ);
+            } else {
+                color = tcuTexLookupVerifier.lookupScalar(level, sampler, x, y, coordZ);
+            }
```
kenrussell: If I'm understanding this fix correctly, don't all three of these functions need the same fix?

isNearestSampleResultValid_CoordXYAsNumber
isNearestSampleResultValid_CoordAsVec2AndInt (this change)
isNearestSampleResultValid_CoordAsVec3

?

I think the template code should have been ported differently, probably by making the template arguments real run-time arguments to the functions. However, for the moment, let's make the minimal fix as you've done, but let's do it everywhere needed. Thanks.

qkmiao: Thanks @zhenyao @kenrussell . I updated the pr following your comments. PTAL.

kenrussell:
Thanks for tracking this down. Looks good.

PR #1616, opened 2016-04-15
Correct elements order for lookup quad

qkmiao:
This makes all texturefiltering_2d_array_formats_*.html pass except texturefiltering_2d_array_formats_48.html.

C++ code:
https://android.googlesource.com/platform/external/deqp/+/master/framework/common/tcuTexLookupVerifier.cpp#101


zhenyao:
Great!

kenrussell:
Excellent catch! LGTM after the fact.

PR #1617, opened 2016-04-15
Spec update for WebGL 2.

zhenyao:
Demand after invalidate{Sub}Framebuffer, the affected region's
contents either stay unchanged or become all 0.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -3117,6 +3117,18 @@ match the <em>type</em> according to the <a href="#TEXTURE_PIXELS_TYPE_TABLE">ab
         implementations.
     </div>
 
+    <h3>Framebuffer contents after invalidation</h3>
+    <p>
+        In OpenGL ES 3.0, after calling <code>invalidateFramebuffer</code> or <code>invalidateSubFramebuffer</code>,
+        the affected region's contents are undefined. In WebGL 2.0, it is required for the contents to either stay
+        unchanged or become all 0.
+    </p>
+
+    <div class="note rationale">
+        It is acceptable for WebGL 2.0 implementations to make <code>invalidateFramebuffer</code> or
+        <code>invalidateSubFramebuffer</code> a no-op.
+    </div>
+
 <!-- ======================================================================================================= -->
 
     <h2>References</h2>


zhenyao:
Please take a look.

1) is discussed and agreed during last working group call.

2) is a driver issue that's not easy to work around. As discussed before in the working group, we are putting limits on the spec for corner cases if there are driver issues that are not easy to work around.

jdashg:
Please split this PR. I do not think (2) is necessary, but let's not discuss that here.

zhenyao:
OK, removed the texture level part and updated the subject and description already.

kenrussell:
LGTM

msc-:
invalidate{Sub}Framebuffer is a performance optimization to save tilers having to copy the framebuffer into the tile buffer. It has no other purpose as far as I am aware. Therefore isn't it better to just make them no-ops in WebGL instead of adding this requirement?

Is this proposal testable?

msc-:
I see that the new text says it is acceptable for these functions to be no-ops. This statement is listed as Rationale but it is not.

However my question about testability remains and is rather important. Specs should not include requirements that can't be tested.

While this addresses behavior stated as undefined in the OpenGL specs, it does not address the real security issue which is, is it possible to leak data from the application that previously used the tile buffer? It certainly was in the days before GPUs had MMUs. With modern GPUs I don't know. Even if it is, I think the requirement that framebuffers be zeroed on allocation and when preserveDrawingBuffer is false probably covers it as it means the WebGL implementation will have used the tile buffer to clear the frame buffer before the application has a chance to call invalidate.

zhenyao:
The current propose is testable.  We just need to verify the read back content is either the same as before the invalidation (no op) or it's cleared to all 0 (we did pass the call down to the driver and WebGL implementation is responsible to make sure no undefined content leak).

msc-:
Glad to hear it is testable. But your recent comment is indicating a new requirement: that the buffer be set to 0 if the call is passed to the driver. I didn't read that in the MR language. It is undesireable. In probably all cases, invalidation will not change the contents of the framebuffer. It just stops the copy to the tile buffer that would normally occur during multipass rendering. The framebuffer attachments remain allocated to the FBO and context until such time as the attachment is deleted and the attachment or FBO unbound.

Note that the ES 3.1 spec says _effectively_ undefined and that the GL 4.5 spec does not include this language.

zhenyao:
I have no object to just define both functions as no-op in WebGL 2 if other browser vendors agree to that.  However, if we end up passing the calls to the driver, no matter what the intention of the ES3 spec writers is, we don't trust the drivers not to leak vrams if it is not clearly defined in a verifiable way in the GL and ES spec, and we will have to clean it before using it, like we do with all textures, renderbuffers, etc.

Most apps clear before render, so if implementation is smart enough, there is really no wasted work even if we require undefined buffers to be actually all cleared to zero.

msc-:
The normal use for these functions is to call them immediately before unbinding the FBO. to indicate to the tiler that their contents are transient and need not be saved from the tile buffer to the actual framebuffer. The unbind acts as a signal to the tiler to perform the render. Since the buffer has been invalidated the tiler will probably not write _anything_ to it even if the application has called gl.clear() for that buffer.

Writing 0 to the buffer when it is allocated will prevent data leakage allowing `invalidate{Sub}Framebuffer` to be safely used as intended to optimize rendering.

Writing 0 to the buffer every time `invalidate{Sub}Framebuffer` is called would be a performance disaster and is unnecessary.

Zeroing newly allocated buffers is already mandated by WebGL. Any optimizations relying on the application calling gl.clear() will have to be changed to turn off that optimization if invalidate is being used.

At first glance testing seems straightforward: 
```
gl.bindFramebuffer(x);
// Draw some stuff
gl.readPixels(...); // Get pixel values before the call
gl.invalidateFramebuffer(...)
gl.bindFramebuffer(0);  // Flush the rendering to framebuffer x
gl.bindFramebuffer(x);  // Bind the buffer again so we can test its contents
gl.readPixels(...);
// Compare result pixels to previous values or 0.
```

There are 2 assumptions here:

1. A tiler never writes anything to an invalidated buffer.
2. gl.readPixels reads from the frame buffer not the tile buffer.

2 seems a reasonable assumption. I don't know if 1 is true.

This update does not define _after_. It needs to. Clearly calling `invalidate{Sub}Framebuffer` does nothing to the framebuffer contents so a requirement/test for the state immediately after calling the function is pointless.

Testing seems difficult. If the test calls gl.finish() before reading the _before_ pixels, a non-tiler will likely fail the test because it won't have completed rendering so some values in the _after_ pixels will be different. If it does call gl.finish() it will force a tiler to flush its render queue and render everything before the call to invalidate turning it into a no-op and making the test pointless.

zhenyao:
I assume you mean "If the test DOES NOT call gl.finish() before reading the before pixels"?

I am confused. My understanding is, tiler or non tiler, when we call gl.readPixels, there should be no undefined results: all rendering should be written to the fbo before reading back the pixels. You are saying a gl.finish is required to guarantee that? We did see some drivers on Android behave like this and it is considered as a driver bug.

I also don't think the test is pointless even with gl.readPixels flushing the rendering pipeline. What you said is what you expect the driver to behave, but as far as it's not in clear language in the spec, we can't assume that will always be the case for all drivers.  The spec says after glInvalidate{Sub}Framebuffer, the fbo content is undefined.  WebGL spec needs to clear this undefined behavior. Either make glInvalidate{Sub}Framebuffer no op, or define it in a verifiable way. gl.readPixels is only one path.  We could also render to the fbo right after glInvalidate{Sub}Framebuffer. It may not be the wise thing to do in an app, but it is WebGL spec and implementation's responsibility to cover all paths with no vram leak possibility.

I don't understand why you say the "after" is not defined in this pull request. WebGL does not define anything according to flush from tiler buffer to fbo, because that's ES internal detail and is not exposed. WebGL defines "after" as after gl.invalidateFramebuffer call.

msc-:
Yes, I meant DOES NOT in the first mention of `gl.finish()`.

By the time `readPixels` reaches the head of the pipeline all preceding rendering commands are supposed to have been executed so a glFinish() is not necessary. I  was temporarily confused because I was looking for mention of an implied finish in the GL ES spec under `ReadPixels` and there isn't any.

If the code does something like
```
/* create an FBO and attachments */
gl.bindFramebuffer(x);
gl.readPixels(...);
gl.invalidateFramebuffer(...)
gl.readPixels(...);
```
The values in the framebuffer will be the same before and after since `invalidateFramebuffer` will merely set a flag in the state to tell it not to copy out the tile buffer or be a no op. So while this tests the letter of the specification it doesn't really test anything different than if you deleted the last 2 lines.

If you add some rendering between `gl.invalidateFramebuffer(...)` and the 2nd  `gl.readPixels(...)` you'd be testing a different part of the implementation. With the added rendering the test result would no longer meet the spec requirement on a non-tiler. The values before invalidate and after the rendering could legitimately be different. The difference between these is 2 test scenarios is what I think we need to capture by defining _after_.

jdashg:
"The values in the framebuffer will be the same before and after since invalidateFramebuffer will "
This is not guaranteed by the spec.

This PR is the new language we want. Feel free to continue discussion here or in a new issue.

zhenyao:
re: If you add some rendering between gl.invalidateFramebuffer(...) and the 2nd gl.readPixels(...) you'd be testing a different part of the implementation.

For this, what I have in mind is to adjust the viewport, and render into some area, but at least leave some invalidated area untouched by the rendering, and verify its either stay the same as before the invalidation (since rendering didn't touch them) or being cleared to zero (lazily by WebGL implementations).

msc-:
@jdashg, yes I know the ES spec does not guarantee this. @zhenyao I like your test plan but the requirement in the proposed spec. language can be met without rendering after calling invalidate. My concern is that calling readPixels immediately after invalidate means invalidate will be a nop at that point on virtually all implementations of OpenGL {,ES}. 

I have to go out now. Later I'll see if I can come up with some language that tightens the requirement.

specs/latest/2.0/index.html
```diff
@@ -3117,6 +3117,18 @@ <h2 class="no-toc">Table of contents</h2>
...
     </div>
 
+    <h3>Framebuffer contents after invalidation</h3>
+    <p>
+        In OpenGL ES 3.0, after calling <code>invalidateFramebuffer</code> or <code>invalidateSubFramebuffer</code>,
+        the affected region's contents are undefined. In WebGL 2.0, it is required for the contents to either stay
+        unchanged or become all 0.
+    </p>
+
+    <div class="note rationale">
```
msc-: Should just be class="note". It not describing a rationale.

specs/latest/2.0/index.html
```diff
@@ -3117,6 +3117,18 @@ <h2 class="no-toc">Table of contents</h2>
         implementations.
     </div>
 
+    <h3>Framebuffer contents after invalidation</h3>
+    <p>
+        In OpenGL ES 3.0, after calling <code>invalidateFramebuffer</code> or <code>invalidateSubFramebuffer</code>,
+        the affected region's contents are undefined. In WebGL 2.0, it is required for the contents to either stay
```
msc-: "affected region's contents ..." to "affected region's contents become effectively undefined."

specs/latest/2.0/index.html
```diff
@@ -3117,6 +3117,18 @@ <h2 class="no-toc">Table of contents</h2>
...
+        unchanged or become all 0.
```
msc-: Change to "... unchanged, reflect the results of subsequent rendering operations or become all 0.

Add a note:

As the immediate effect of calls to  <code>invalidateFramebuffer</code> or <code>invalidateSubFramebuffer</code> in OpenGL ES is likely to at most result in setting a state bit that is checked during later rendering, tests of this requirement are encouraged generate rendering before checking region's contents.

zhenyao: I understand where you come from, but "reflect the results of subsequent rendering operations" isn't really a state right after the invalidation call. When the spec defines "become all 0", it is really not asking implementations to insert a clear() call, but rather, only to set up a bit indicating the content is undefined.  Only when this undefined content is about to be exposed to a user, then we need to do some clearing.

All other suggested changes are good and I'll get a pull request to incorporate them.

PR #1618, opened 2016-04-15
Add a new test for mac driver bug on fbo completeness

zhenyao:
Remove that situation from an unrelated test.


kenrussell:
Looks great. Excellent work diagnosing this.

CC @grorg as FYI. We will try to file a Radar about this.

PR #1622, opened 2016-04-19
Add framebuffer tests to oes-texture-half-float

RafaelCintron:
These tests ensure:
1) For LUMINANCE, LUMINANCE_ALPHA and ALPHA half float textures:
- checkFramebufferStatus should return FRAMEBUFFER_INCOMPLETE_ATTACHMENT.
- getParameter with IMPLEMENTATION_COLOR_READ_FORMAT and IMPLEMENTATION_COLOR_READ_TYPE should return null and flag INVALID_OPERATION
- readPixels should flag a FRAMEBUFFER_INCOMPLETE_ATTACHMENT error.

2) For RGBA and RGB half float buffers
- RGBA and FLOAT should always be an accepted format and type combination for readPixels.
- IMPLEMENTATION_COLOR_READ_FORMAT should be RGB or RGBA.
- IMPLEMENTATION_COLOR_READ_TYPE must be UNSIGNED_BYTE, HALF_FLOAT_OES, or one of the UNSIGNED_SHORT combinations
- If the type of the arrayBuffer does not match the glType passed into readPixels, fail the test.

### DIFF
--- a/sdk/tests/conformance/extensions/oes-texture-half-float.html
+++ b/sdk/tests/conformance/extensions/oes-texture-half-float.html
@@ -177,6 +177,8 @@ if (!gl) {
         runRenderTest(gl.RGBA, [10000, 10000, 10000, 10000], null);
         runRenderTest(gl.RGB, [10000, 10000, 10000, 1], null);
 
+        runFramebufferTest();
+
         // Check of getExtension() returns same object
         runUniqueObjectTest();
     }
@@ -206,6 +208,19 @@ function getFormatName(format)
         return "ALPHA";
     else if (format == gl.LUMINANCE_ALPHA)
         return "LUMINANCE_ALPHA";
+}
+
+function getTypeName(type) {
+    if (type === gl.UNSIGNED_BYTE)
+        return "UNSIGNED_BYTE";
+    else if (type === ext.HALF_FLOAT_OES)
+        return "HALF_FLOAT_OES";
+    else if (type === gl.UNSIGNED_SHORT_4_4_4_4)
+        return "UNSIGNED_SHORT_4_4_4_4";
+    else if (type === gl.UNSIGNED_SHORT_5_5_5_1)
+        return "UNSIGNED_SHORT_5_6_5";   
+    else if (type === gl.FLOAT)
+        return "FLOAT";
 }
 
 function allocateTexture()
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/extensions/oes-texture-half-float.html
```diff
@@ -335,6 +350,137 @@
...
+    debug("Framebuffer Tests");
+    
+    var texture = allocateTexture();
+    gl.bindTexture(gl.TEXTURE_2D, texture);
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, 1, 1, 0, gl.RGBA, ext.HALF_FLOAT_OES, null);
+
+    var fbo = gl.createFramebuffer();
+    gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture, 0);
+    if (gl.checkFramebufferStatus(gl.FRAMEBUFFER) != gl.FRAMEBUFFER_COMPLETE) {
```
kenrussell: I don't think that renderability of RGBA floating-point textures necessarily implies renderability of RGB format textures; see below.

sdk/tests/conformance/extensions/oes-texture-half-float.html
```diff
@@ -335,6 +350,137 @@
...
+    arrayBufferHalfFloatInput[1] = 0x3400; // 0.25 in half float
+    arrayBufferHalfFloatInput[2] = 0x3800; // 0.50 in half float
+    arrayBufferHalfFloatInput[3] = 0x3A00; // 0.75 in half float
+    
+    [gl.RGBA, gl.RGB].forEach(function(goodFormat) {
+        debug(getFormatName(goodFormat) + " framebuffer tests");
+        debug("");
+        
+        gl.texImage2D(gl.TEXTURE_2D, 0, goodFormat, 1, 1, 0, goodFormat, ext.HALF_FLOAT_OES, arrayBufferHalfFloatInput);
+        shouldBe("gl.checkFramebufferStatus(gl.FRAMEBUFFER)", "gl.FRAMEBUFFER_COMPLETE");
```
kenrussell: I'm not sure that it's correct to assume that RGB format floating-point textures are renderable. Would you agree to short-circuit this loop if it turns out the framebuffer is incomplete in that case? (Basically, the shouldBe should only apply to the RBGA format.)

kenrussell:
Thank you Rafael for assembling this test case and I'm very sorry for taking so long to review it.

Please see inline comments. Would you agree with updating this pull request to take them into account?

Perhaps in a follow-on pull request, should we fix oes-texture-float.html to parallel the format of this test?

sdk/tests/conformance/extensions/oes-texture-half-float.html
RafaelCintron: The reason I made RGB part of the list is because the EXT_color_buffer_half_float says that both RGB16F and RGBA16F become available as color renderable formats. Unless I am missing something, that wording doesn't imply you can support RGBA but not RGB.

Earlier in this conformance test, it calls renderTest with both RGB and RGBA formats. Granted, it gives the implementation a passing grade if the framebuffer is not complete but I think that's more about not supporting float framebuffers in general, not about only supporting RGBA and not RGB.

FWIW, both Chrome and Firefox on Windows allow both RGB and RGBA half float framebuffers.

kenrussell: I see. Thanks for your careful study. Sounds good. Let's merge this.

Would you do the favor of taking another look at oes-texture-float.html and bringing it up to parity with this test?

RafaelCintron: Sure. I'll see what I can do.

zhenyao:
ES2 spec does not say anything about GetInteger(IMPLEMENTATION_COLOR_READ_FORMAT/TYPE) generates INVALID_OPERATION when fbo is incomplete.  ES3 does.

This pull request verifies the ES3 behavior on WebGL 1.

I agree especially with the float/half_float texture extensions, ES3 behavior is more desirable, but should we update the WebGL 1 spec to clarify this? @RafaelCintron

RafaelCintron:
I agree the ES 3 spec is more clear about the desired behavior compared to ES 2. Updating the WebGL 1 spec is fine with me.

PR #1623, opened 2016-04-20
Add a skip list to the dEQP test runner.

kenrussell:
This allows tests to be skipped if they are failing because of
insurmountable graphics driver bugs. Deliberately, entries apply to all
operating systems, browsers and GPU types. The mechanism is deliberately
orthogonal to the existing support for the "filter" query variable.

The implementation is unfortunately complicated because of confusing
mutual recursion between some of the inner methods of tcuTestCase, and
the desire to report reasonable warnings when tests are skipped.
Untangling these methods is left for future work; patches welcome.

Some initial failure suppressions are added in this pull request.
Happily, NVIDIA's recently-released 364.15 driver fixes at least some,
if not all, of these failures. The tests that are now passing will be
removed in the next pull request. (It was desired to have an example in
the version history of how to use the skip list.)

Regenerated deqp-deps.js via "build.py depfile". Did not run the Closure
typechecker on these changes; believe the script for running it needs to
be updated. (This is on my to-do list.)


kenrussell:
@zhenyao @toji @jdashg @grorg @Oletus @RafaelCintron @Richard-Yunchao @qkmiao FYI, I am merging this myself in order to next update the skip list and remove now-passing tests. Comments welcome; will address them in following pull requests. This has been tested fairly extensively locally.

sdk/tests/deqp/framework/common/tcuTestCase.js
```diff
@@ -217,8 +219,47 @@ goog.scope(function() {
...
+        var tryAgain = false;
+        var test = null;
+        do {
+            tryAgain = false;
+            test = this._nextIgnoringSkipList(pattern);
+            if (test != null) {
+                // See whether the skip list vetoes the execution of
+                // this test.
+                var fullTestName = test.fullName();
+                var skipDisposition = tcuSkipList.getSkipStatus(fullTestName);
```
zhenyao: I am concerned about the perf impact, because it's for each test case.  Do you see a slow down because of this?  If not, then it is good.

zhenyao:
lgtm after the fact
one question about perf impact

sdk/tests/deqp/framework/common/tcuTestCase.js
kenrussell: Thanks for raising this concern. I just tried both putting back in the many transform feedback failure suppressions from #1623 , and commenting out this block of code which tests the skip list at all. There was no visible change in the per-test run time. The skip list is deliberately implemented using JavaScript's built-in property lookup, which should be hash table-based and should be fast.

PR #1625, opened 2016-04-20
Correct name of skipped tests

qkmiao:
This fixes the wrong name of skipped tests and extra skip info at the
end of result page of each html file.

@kenrussell  I found some bugs about the skipped list. PTAL.

### DIFF
--- a/sdk/tests/deqp/framework/common/tcuTestCase.js
+++ b/sdk/tests/deqp/framework/common/tcuTestCase.js
@@ -242,6 +242,7 @@ goog.scope(function() {
                 var skipDisposition = tcuSkipList.getSkipStatus(fullTestName);
                 if (skipDisposition.skip) {
                     tryAgain = true;
+                    setCurrentTestName(fullTestName);
                     checkMessage(false, 'Skipping test due to tcuSkipList: ' + fullTestName);
                 }
             }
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/framework/common/tcuTestCase.js
```diff
@@ -271,7 +272,6 @@ goog.scope(function() {
 
         // If no more children, get the next brother
         if (test == null && this.parentTest != null) {
-            this.currentTestNdx = 0;
```
kenrussell: Out of curiosity what bug did this fix? I noticed this in the original code and it seemed both confusing and suspicious, but I decided not to touch it to avoid potentially breaking something.

kenrussell:
Thanks Qiankun. I tested your patch locally and it seems to be working well.

sdk/tests/deqp/framework/common/tcuTestCase.js
qkmiao: It will print wrong skip test information at the end of a test html. Have you seen them before? You can try texturefunction* tests to see them. These information is generated when go through brother tests of currentTest's parent (though these tests will not be executed because out of range).

kenrussell: Thanks for the information. I didn't notice them, but was mainly testing with the transform feedback tests, and using a filter to only run a subset of the tests.

PR #1626, opened 2016-04-20
simple implementation of alternative for mix() built-in function

Richard-Yunchao:
PTAL. Thanks a lot!

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
+++ b/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
@@ -3968,6 +3968,10 @@ var setParentClass = function(child, parent) {
             this.m_body[ndx].execute(funCtx);
 
         var ret = this.m_ret.evaluate(funCtx);
+        if (this.m_ret_alternative !== undefined) {
+            var ret_alternative = this.m_ret_alternative.evaluate(funCtx);
+            ret = tcuInterval.withIntervals(ret, ret_alternative);
+        }
 
         // \todo [lauri] Store references instead of values in environment
         args.a = funEnv.lookup(this.m_var0);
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
```diff
@@ -4609,6 +4616,17 @@ var setParentClass = function(child, parent) {
...
+        // x + (y - x) * a
+        var x = args.a;
+        var y = args.b;
+        var a = args.c;
+        var v0 = app(new glsBuiltinPrecisionTests.Sub(), y, x);
+        var v1 = app(new glsBuiltinPrecisionTests.Mul(), a, v0);
+        var v2 = app(new glsBuiltinPrecisionTests.Add(), x, v1);
+        return v2;
+    };
+
```
Richard-Yunchao: Native deqp has an alternative implementation for mix() builtin precision. They are:
1). ret0 = x * (1 - a) + y * a
2). ret1 = x + (y - x) * a. 
If the result of mix(x, y, a) from shader is in this range [min(ret0, ret1), max(ret0, ret1)], It will think that the result is correct. 

See the native deqp code https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/glshared/glsBuiltinPrecisionTests.cpp#3268 and the class of Alternative::doApply and the definition of "|“ (union operation) in tcuInterval in native deqp: https://android.googlesource.com/platform/external/deqp/+/deqp-dev/framework/common/tcuInterval.hpp#88 

However, in JS implementation, there is no "alternative" class at all. Considering that there are only a few users of "Alternative" class, and only mix() failed. For simplicity, I use another function to calculate mix() with method 2) above, then union the result with the original one (just like what it does in native deqp) in DerivedFunc.doApply().

zhenyao:
Thanks for explaining this fix clearly. Looks good to me.

kenrussell:
Thanks for tracking this down and for the clear explanation Yunchao. LGTM after the fact.

PR #1629, opened 2016-04-22
Move texturefiltering* tests to its own folder

zhenyao:
Also, combined a bunch of tests together now that we optimize them and
they finish in much shorter time.

Note that the auto-generated changes are not included in this patch, for
easier review.


toji:
LGTM

PR #1630, opened 2016-04-22
Autogenerated changes related to texturefiltering* tests.

zhenyao:


zhenyao:
Merging myself since it's all auto-generated and I tested them locally already

PR #1633, opened 2016-04-26
Fix wrong sign bit calculation in tcuFloat.js

qkmiao:
The bug causes shaderderivate.html regression. It's introduced in
https://github.com/KhronosGroup/WebGL/pull/1588. It also exposed
an uninitialization error in es3fShaderDerivateTests.js.


kenrussell:
Thanks for tracking this down @qkmiao . Looks good.

PR #1634, opened 2016-04-26
Fix several bugs in fborender.html

Jiawei-Shao:
(1) Add maps from internalFormat to format before function texImage2D in sglrGLContext.js and sglrReferenceContext.js
(2) Fix an index error in es3fFboRenderTest.SharedColorbufferClearsTest
(3) Fix a misuse of function bindFramebuffer in es3fFboRenderTest.js:1044
(4) Fix bugs in function getWidth and getHeight in sglrGLContext.js since the return values of these functions should be the width and the height of current viewport instead of the ones of current canvas
(5) In function iterate in es3fFboRenderTest.js, remove the call of gles3Frame.setSize since after setSize is called the frame will be cleared
(6) In function iterate in es3fFboRenderTest.js, set default value of failReason to "" instead of null since a null pointer exception occurs when a message for success should be shown
This patch will let 14 (of 125) cases pass in fborender.html

### DIFF
--- a/sdk/tests/deqp/framework/opengl/simplereference/sglrGLContext.js
+++ b/sdk/tests/deqp/framework/opengl/simplereference/sglrGLContext.js
@@ -127,7 +127,10 @@ goog.scope(function() {
          * @return {number}
          */
         var getWidth = function() {
-            return context.drawingBufferWidth;
+            if(viewport)
+                return viewport[2];
+            else
+                return context.drawingBufferWidth;
         };
         wrap['getWidth'] = getWidth;
 
@@ -135,7 +138,10 @@ goog.scope(function() {
          * @return {number}
          */
         var getHeight = function() {
-            return context.drawingBufferHeight;
+            if(viewport)
+                return viewport[3];
+            else
+                return context.drawingBufferHeight;
         };
         wrap['getHeight'] = getHeight;
 
@@ -162,7 +168,40 @@ goog.scope(function() {
         };
         wrap['readPixels'] = readPixels;
 
-        return wrap;
+        /**
+         * @param {number} target
+         * @param {number} level
+         * @param {number} internalFormat
+         * @param {number} width
+         * @param {number} height
+         */
+        var texImage2DDelegate = function(target, level, internalFormat, width, height) {
+            var format;
+            var dataType;
+
+            switch(internalFormat)
+            {
+                case gl.ALPHA:
+                case gl.LUMINANCE:
+                case gl.LUMINANCE_ALPHA:
+                case gl.RGB:
+                case gl.RGBA:
+                    format = internalFormat;
+                    dataType = gl.UNSIGNED_BYTE;
+                    break;
+                default:
+                {
+                    var transferFmt = gluTextureUtil.getTransferFormat(gluTextureUtil.mapGLInternalFormat(internalFormat));
+                    format = transferFmt.format;
+                    dataType = transferFmt.dataType;
+                    break;
+                }
+             }
+             context.texImage2D(target, level, internalFormat, width, height, 0, format, dataType, null);
+         };
+         wrap['texImage2DDelegate'] = texImage2DDelegate;
+
+         return wrap;
     };
 
     /**
--- a/sdk/tests/deqp/framework/opengl/simplereference/sglrReferenceContext.js
+++ b/sdk/tests/deqp/framework/opengl/simplereference/sglrReferenceContext.js
@@ -4045,6 +4045,38 @@ goog.scope(function() {
     * @param {number} internalFormat
     * @param {number} width
     * @param {number} height
+    */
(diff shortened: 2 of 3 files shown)


sdk/tests/deqp/functional/gles3/es3fFboRenderTest.js
```diff
@@ -413,9 +413,8 @@ goog.scope(function() {
             name = this.m_context.createTexture();
 
         this.m_context.bindTexture(gl.TEXTURE_2D, name);
-        this.m_context.texImage2D(
-            gl.TEXTURE_2D, 0, format, width, height,
-            0, gl.RGBA, gl.UNSIGNED_BYTE, null
+        this.m_context.texImage2DDelegate(
+            gl.TEXTURE_2D, 0, format, width, height
```
zhenyao: You deleted an extra line here?  Did you test the CL?

sdk/tests/deqp/framework/opengl/simplereference/sglrGLContext.js
```diff
@@ -127,15 +127,15 @@ goog.scope(function() {
          * @return {number}
          */
         var getWidth = function() {
-            return context.drawingBufferWidth;
+            return viewport[2];
```
zhenyao: note from line 89 that viewport is optional

sdk/tests/deqp/framework/opengl/simplereference/sglrGLContext.js
```diff
@@ -162,7 +162,40 @@ goog.scope(function() {
...
 
-        return wrap;
+        /**
+         * @param {number} target
+         * @param {number} level
+         * @param {number} internalFormat
+         * @param {number} width
+         * @param {number} height
+         */
+        var texImage2DDelegate = function(target, level, internalFormat, width, height) {
```
zhenyao: Can the function name just be texImage2D? To align with c++ code.

sdk/tests/deqp/framework/opengl/simplereference/sglrReferenceContext.js
```diff
@@ -4045,6 +4045,38 @@ goog.scope(function() {
     * @param {number} internalFormat
     * @param {number} width
     * @param {number} height
+    */
+    sglrReferenceContext.ReferenceContext.prototype.texImage2DDelegate = function (target, level, internalFormat, width, height) {
```
zhenyao: Sam here: can the function name just be texImage2D? To align with c++ code.

zhenyao: Also, it's a pity we have to implement this twice. But the c++ class inheritance isn't implemented on the js side, with no sgrlContext implemented.

zhenyao:
Good work.  Mostly look good with some minor comments.  Specially 'viewport' seems optional, so will the unconditional use cause some tests to throw an exception?

sdk/tests/deqp/framework/opengl/simplereference/sglrReferenceContext.js
qkmiao: There is already a function named texImage2D in sglrReferenceContext.ReferenceContext.prototype.texImage2D. Due JS doesn't support overload, so this name is used here.

sdk/tests/deqp/framework/opengl/simplereference/sglrGLContext.js
qkmiao: Thanks for pointing this. What about check viewport before using it? If viewport exists, return viewport size otherwise report drawingBuffer size.

sdk/tests/deqp/functional/gles3/es3fFboRenderTest.js
```diff
@@ -1476,7 +1471,7 @@ goog.scope(function() {
         /** @type {WebGLProgram} */
         var texFromFboShaderID = context.createProgram(texFromFboShader);
         /** @type {WebGLProgram} */
-        var flatShaderID = context.createProgram(texFromFboShader);
+        var flatShaderID = context.createProgram(flatShader);
```
qkmiao: C++ code:
https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fFboRenderTest.cpp#1123

@zhenyao  the new code fixed your comments and added this new fix.

sdk/tests/deqp/framework/opengl/simplereference/sglrReferenceContext.js
zhenyao: Ah OK, my js really sucks.

zhenyao:
lgtm

sdk/tests/deqp/framework/opengl/simplereference/sglrReferenceContext.js
Jiawei-Shao: Thanks for all your comments! @zhenyao @qkmiao

Issue #1635, opened 2016-04-26
deqp/data/gles2/shaders/scoping test, case local_variable_hides_function_parameter is invalid


shannonwoods:
This test requires the following shader to compile:
int func (int inp, int x) { int x = 5; return inp + x - 5; }

This is invalid, because the declaration of "int x" in the function body falls within the same scope as the function parameter of the same name, making it an illegal redeclaration. Note that this test directly conflicts with conformance/glsl/misc/shader-with-functional-scoping, which requires that the following shader fail:
// fragment shader with invalid functional scoping should fail
precision mediump float;
int f(int k) {
  int k = k + 3;
  return k;
}

The dEQP-sourced test case was ported from a case which is not in the mustpass list in AOSP. Function parameter shadowing was specifically discussed by the OpenGL ES WG (at https://cvs.khronos.org/bugzilla/show_bug.cgi?id=8656), which determined that this must provoke a compilation error. This test should be removed from the WebGL set.

zhenyao:
Removed.

PR #1636, opened 2016-04-26
Remove local_variable_hides_function_parameter test case

zhenyao:
from deqp/data/gles2/shaders/scoping

See https://github.com/KhronosGroup/WebGL/issues/1635


toji:
LGTM

PR #1637, opened 2016-04-27
Fix texSubImage from pixel unpack buffer parameter list in spec

Oletus:
The parameters should include the dimensions of the texture when
uploading from a pixel unpack buffer.


kenrussell:
Thanks for catching this. These prototypes were copied from the ones taking HTML elements, which is why those parameters were lost.

PR #1631, opened 2016-04-25
WebGL 2.0 supports transpose for uniformatirx functions

qkmiao:
ES3 spec allows transpose argument to be true now.

PR #1638, opened 2016-04-27
Fix a bug introduced in https://github.com/KhronosGroup/WebGL/pull/1631

zhenyao:


kenrussell:
Thanks for catching this Mo and sorry I was sloppy in my earlier review. LGTM after the fact.

PR #1641, opened 2016-04-28
Expand tests for S3TC compressed textures to cover TexStorage2D.

kenrussell:
This is the first time that uninitialized compressed textures can be
exposed to the JavaScript level, and they must be tested. The new
tests assert that the texture contains either transparent or opaque
black, depending on the specific DXT internal format.

See http://crbug.com/605129 for fixes for this test in Chromium.

### DIFF
--- a/sdk/tests/conformance/extensions/webgl-compressed-texture-s3tc.html
+++ b/sdk/tests/conformance/extensions/webgl-compressed-texture-s3tc.html
@@ -213,13 +213,15 @@ function testDXT1_RGB() {
             height: 4,
             channels: 3,
             data: img_4x4_rgb_dxt1,
-            format: ext.COMPRESSED_RGB_S3TC_DXT1_EXT
+            format: ext.COMPRESSED_RGB_S3TC_DXT1_EXT,
+            hasAlpha: false,
         },
         {   width: 8,
             height: 8,
             channels: 3,
             data: img_8x8_rgb_dxt1,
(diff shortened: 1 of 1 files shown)


kenrussell:
@zhenyao in particular please review.

Please see https://codereview.chromium.org/1925093002 for the Chromium patch which makes the new tests pass.

sdk/tests/conformance/extensions/webgl-compressed-texture-s3tc.html
```diff
@@ -435,7 +451,10 @@
   }
 }
 
-function testDXTTexture(test) {
+function testDXTTexture(test, useTexStorage) {
+    if (!useTexStorage)
+        return;
```
zhenyao: This needs to be removed.  Assume you add it for easier testing?

zhenyao:
Mostly looks good to me.  Feel free to merge yourself after removing the debugging hack.

sdk/tests/conformance/extensions/webgl-compressed-texture-s3tc.html
kenrussell: Thank you for catching that! I'm sorry I missed it when cleaning up the CL. Yes, I had to take out most of the test's code when trying to figure out where the errors were coming from initially.

kenrussell:
Re-tested with the upcoming revised fix for https://codereview.chromium.org/1925093002/ . Merging now.

PR #1642, opened 2016-04-28
Fix texture initialization for texturespecification tests

qkmiao:
This fixes texturespecification01.html on chromeOS. But
texturespecification01.html still fails on my linux desktop with nVidia
graphics card. The failure is about texture filtering. I suppose it's a
driver bug: wrong level is selected as input texture for nVidia.

C++ code:
https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fTextureSpecificationTests.cpp#928

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fTextureSpecificationTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fTextureSpecificationTests.js
@@ -1366,7 +1366,7 @@ goog.scope(function() {
 
         images = rnd.shuffle(images);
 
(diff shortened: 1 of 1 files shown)


qkmiao:
I started fixing texturespecification*. JS code structure is quite different from C++ code for software rasterization. It's a bit difficult to fix failures in these test. There is excellent work about this part before: https://github.com/KhronosGroup/WebGL/pull/1595. I will continue to make the test move forward.

kenrussell:
@Kangz do you think you could help review this? Also @zhenyao . Thanks.

zhenyao:
lgtm

PR #1643, opened 2016-04-29
Correct some issues about fbo depth in deqp/functional/gles3/es3fFboD…

xhcao:
…epthbufferTests.js

This html file could pass on intel gpu, but still fail on nvidia gpu. The failed issue is related to floating-point depth buffer.

### DIFF
--- a/sdk/tests/deqp/framework/referencerenderer/rrFragmentOperations.js
+++ b/sdk/tests/deqp/framework/referencerenderer/rrFragmentOperations.js
@@ -193,11 +193,27 @@ rrFragmentOperations.executeDepthCompare = function(inputFragments, depthFunc, d
       for (var i = 0; i < inputFragments.length; i++) {
             var frag = inputFragments[i];
             if (frag.isAlive) {
-                var fragSampleNdx = 0;
-                var depthBufferValue = depthBuffer.getPixDepth(fragSampleNdx, frag.pixelCoord[0], frag.pixelCoord[1]);
-                var sampleDepthFloat = frag.sampleDepths[fragSampleNdx];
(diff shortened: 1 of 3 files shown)


qkmiao:
Thanks, LGTM.  One nit: the title of the pr is truncated.

zhenyao:
lgtm

sdk/tests/deqp/framework/referencerenderer/rrFragmentOperations.js
```diff
@@ -193,11 +193,27 @@ rrFragmentOperations.executeDepthCompare = function(inputFragments, depthFunc, d
...
+                    var depthBufferValue = depthBuffer.getPixDepth(fragSampleNdx, frag.pixelCoord[0], frag.pixelCoord[1]);
+                    var sampleDepthFloat = frag.sampleDepths[fragSampleNdx];
+                    var buffer = new Uint32Array(2);
+                    var access = new tcuTexture.PixelBufferAccess({format:depthBuffer.m_format,
+                                                                   width: 1,
+                                                                   height: 1,
+                                                                   depth: 1,
+                                                                   data: buffer});
+                    access.setPixDepth(sampleDepthFloat, 0, 0, 0);
+                    var sampleDepth = access.getPixDepth(0, 0, 0);
```
zhenyao: This change caused a test case in deqp/functional/gles3/fbostencilbuffer.html to fail.  Now comparing with the c++ code again, the c++ code here use getPixelUint instead of getPixDepth here. I am not sure if this is root cause, but I will revert this part of this patch to avoid regression on a passing test. @xhcao Please take a look at this case when you have time. Thanks.

PR #1644, opened 2016-04-29
Clear gl state at the end of each test 

Jiawei-Shao:


sdk/tests/deqp/functional/gles3/es3fFboRenderTest.js
```diff
@@ -1007,6 +1009,7 @@ goog.scope(function() {
         rrUtil.drawQuad(
             context, texShaderID, [-1.0, -1.0, 0.0], [1.0, 1.0, 0.0]
         );
+        context.disable(gl.BLEND);
```
Jiawei-Shao: Blending should not always be enabled, or it will let the following tests go wrong

qkmiao:
Could you move the gl state reset to the deinit() function of the test did the state set, other than doing clear at the beginning of next test? Since it may also effect other test. See state.currentTest.deinit() in deqp/framework/common/tcuTestCase.js. Please refer es3fMultisampleTests.MultisampleCase.prototype.deinit().

Jiawei-Shao:
Thanks for comments from @qkmiao !

Jiawei-Shao:
@zhenyao Please help to review

sdk/tests/deqp/functional/gles3/es3fFboRenderTest.js
```diff
@@ -704,7 +704,48 @@ goog.scope(function() {
         );
     };
 
+    /**
+     * deinit
+     */
     es3fFboRenderTest.FboRenderCase.prototype.deinit = function() {
```
Jiawei-Shao: There are many errors caused by uncleared states set by former executed test cases. Here I added codes to clear gl states.
Referenced from: 
https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js#L216

sdk/tests/deqp/functional/gles3/es3fFboRenderTest.js
Jiawei-Shao: @qkmiao  Please help to review

zhenyao: Is this still necessary now that you disable BLEND in deinit?

Jiawei-Shao: Thanks to point out this. It will not affect the following drawings and other test cases and I think I should keep consistent with C++ code.
https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fFboRenderTest.cpp#688

zhenyao:
lgtm

PR #1646, opened 2016-05-04
Add a context size change test.

pavolas:
Add a context size change test.

The test makes an antialiased, no alpha context. Then it resizes it, and checks
that clearing/reading the resized context still works.

### DIFF
--- a/sdk/tests/conformance/context/00_test_list.txt
+++ b/sdk/tests/conformance/context/00_test_list.txt
@@ -1,7 +1,7 @@
 --max-version 1.9.9 constants-and-properties.html
 --min-version 1.0.2 context-attribute-preserve-drawing-buffer.html
 context-attributes-alpha-depth-stencil-antialias.html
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/context/00_test_list.txt
```diff
@@ -1,6 +1,7 @@
 --max-version 1.9.9 constants-and-properties.html
 --min-version 1.0.2 context-attribute-preserve-drawing-buffer.html
 context-attributes-alpha-depth-stencil-antialias.html
+context-size-change.html
```
kenrussell: Please add --min-version 1.0.4 as a prefix (it's just the way new tests are added at this point).

kenrussell:
Thanks, looks good. Could you please make the one tiny fix above?

pavolas:
Done.

kenrussell:
Excellent. Thanks for adding this test.

PR #1647, opened 2016-05-06
Add conformance test for extensions no longer available in WebGL 2.0.

kenrussell:
These include both extensions promoted to core, and extensions subsumed by other extensions.

Explicitly point out that EXT_color_buffer_half_float and WEBGL_color_buffer_float are no longer available in WebGL 2.0, subsumed by EXT_color_buffer_float.


kenrussell:
Please review. @zhenyao @jdashg in particular. Thanks.

zhenyao:
Looks good.

PR #1648, opened 2016-05-06
Improve indexed buffer binding testing in conformance2/buffers/

zhenyao:


sdk/tests/conformance2/buffers/bound-buffer-size-change-test.html
```diff
@@ -0,0 +1,135 @@
...
+debug("");
+debug("bindBufferRange with UNIFORM_BUFFER target");
+var buffer4 = gl.createBuffer();
+var offset = gl.getParameter(gl.UNIFORM_BUFFER_OFFSET_ALIGNMENT);
+gl.bindBufferRange(gl.UNIFORM_BUFFER, 1, buffer4, offset, 12);
+wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+    "Calling bindBufferRange on a buffer where no storage is allocated should succeed.");
+shouldBe("gl.getParameter(gl.UNIFORM_BUFFER_BINDING)", "buffer4");
+shouldBe("gl.getIndexedParameter(gl.UNIFORM_BUFFER_BINDING, 1)", "buffer4");
+shouldBe("gl.getIndexedParameter(gl.UNIFORM_BUFFER_SIZE, 1)", "12");
```
kenrussell: Is it worth adding a comment that this tests the specified ES behavior -- that the driver is responsible for making sure no out-of-range writes occur?

kenrussell:
Very nice test. Please merge yourself once you're satisfied with it. LGTM

sdk/tests/conformance2/buffers/bound-buffer-size-change-test.html
zhenyao: Done

PR #1649, opened 2016-05-06
Fix a bug that UNIFORM_BUFFER_SIZE or TRANSFORM_BUFFER_SIZE should

zhenyao:
return 0 if it's not specified at binding time (for example, through
BinBufferBase)


toji:
LGTM!

PR #1651, opened 2016-05-07
Further formatting.

zhenyao:


zhenyao:
Merging myself since no logic change.

PR #1653, opened 2016-05-09
Fix transform feedback buffer binding issue

qkmiao:
In WebGL, a buffer bound to TRANSFORM_FEEDBACK_BUFFER cannot be bound to
any other targets. See section 5.1 Buffer Object Binding in WebGL 2.0
spec. This is a difference between GLES and WebGL. This patch detaches
the buffer from TRANSFORM_FEEDBACK_BUFFER before other binding, then
reattaches it after other binding.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/lifetime.html
+++ b/sdk/tests/deqp/functional/gles3/lifetime.html
@@ -20,7 +20,7 @@ setCurrentTestName(testName);
 description("Functional test: " + testName + ".");
 
 var wtu = WebGLTestUtils;
-/** @type {WebGL2RenderingContext} */ var gl = wtu.create3DContextWithWrapperThatThrowsOnGLError('canvas', null, 2);
+/** @type {WebGL2RenderingContext} */ var gl = wtu.create3DContext('canvas', null, 2);
 
     try{
         functional.gles3.es3fLifetimeTests.run(gl);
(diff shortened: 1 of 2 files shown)


sdk/tests/deqp/functional/gles3/es3fLifetimeTests.js
```diff
@@ -435,10 +435,13 @@ es3fLifetimeTests.TfDeleteActiveTest.prototype.iterate = function() {
...
     var errCode = gl.NONE;
-    gl.deleteTransformFeedback(tf);
-    errCode = gl.getError();
-    assertMsgOptions(errCode == gl.INVALID_OPERATION,
-        'Deleting active transform feedback must produce INVALID_OPERATION', false, true);
+    try {
+        gl.deleteTransformFeedback(tf);
+    }
+    catch (err) {
+        assertMsgOptions(err.error == gl.INVALID_OPERATION,
```
kenrussell: The only reason an exception is being thrown here is that in sdk/tests/deqp/functional/gles3/lifetime.html, create3DContextWithWrapperThatThrowsOnGLError is being used. I don't think it is a good idea to rely on this fact; that was mainly done for debugging. Please change that to wtu.create3DContext(...) instead and leave this line unchanged. Thanks.

kenrussell:
Thanks @qkmiao for this patch. Please make the one change above. I tested your patch locally with that modification and all of the tests in deqp/functional/gles3/lifetime.html still pass. Thanks.

qkmiao:
Thanks @kenrussell for review. I updated this pr as you suggested. PTAL.

kenrussell:
Excellent. Thank you for persisting with this patch. LGTM

PR #1655, opened 2016-05-09
Fixed whitespace errors in oes-texture-half-float.html.

kenrussell:


kenrussell:
Whitespace changes only. Merging myself.

PR #1656, opened 2016-05-09
Allow FLOAT as one of the implementation color read types.

kenrussell:
There is nothing in the specifications which forbids this, and we've
found that Intel's Linux OpenGL driver, which supports the
GL_OES_read_format extension, reports RGBA/FLOAT as the implementation
color read format and type. Relax the newly-expanded test.

### DIFF
--- a/sdk/tests/conformance/extensions/oes-texture-half-float.html
+++ b/sdk/tests/conformance/extensions/oes-texture-half-float.html
@@ -457,12 +457,18 @@ function runFramebufferTest() {
             "IMPLEMENTATION_COLOR_READ_FORMAT should be color renderable: RGBA or RGB. Received: " + getFormatName(implementationColorReadFormat));
 
         var implementationColorReadType = gl.getParameter(gl.IMPLEMENTATION_COLOR_READ_TYPE);
+
+        // There is nothing in the specifications that keeps the
+        // implementation color read format and type from being the
+        // same as the implicitly supported one. For this reason, keep
+        // gl.FLOAT as one of the valid options.
         assertMsg(implementationColorReadType === gl.UNSIGNED_BYTE ||
+                  implementationColorReadType === gl.FLOAT ||
                   implementationColorReadType === ext.HALF_FLOAT_OES ||
                   implementationColorReadType === gl.UNSIGNED_SHORT_4_4_4_4 ||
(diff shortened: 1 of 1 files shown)


kenrussell:
@zhenyao please review; as discussed last week.

@RafaelCintron based on our reading of the specifications we are pretty sure this change is both correct, and needed on some drivers. Please raise any concerns you may have. Thanks.

zhenyao:
LGTM

In theory a driver can just return RGB/FLOAT, which is definitely valid.

RafaelCintron:
@kenrussell @zhenyao When I first wrote the code that checks IMPLEMENTATION_COLOR_READ_TYPE, I debated whether to make it generous or limit it to just HALF_FLOAT_OES.

The rationale for not allowing integer types is because no implementation should be doing half float textures with 8bits (or fewer) per channel. The loss of precision would be too great.  Separately from this, we should consider restricting the test to only allow float types.

The rationale for not allowing FLOAT to be an implementation color read type is that it somewhat defeats the purpose of having a half_float extension. If the developer wants full floats, then they should be using the float extension as the primary or as a fallback. It's strange to me that Intel would be returning FLOAT as the implementation color read type. But, I don't feel as strongly about allowing FLOAT as I do about preventing integer types, especially since Intel already returns FLOAT for the color read type.

WDYT?

zhenyao:
@RafaelCintron I understand and agree with your rationale, but it's not supported by the spec, so we can't prevent underlying drivers from returning FLOAT.  When I reviewed this patch, I thought about this but was hoping no driver would do that.  Apparently some drivers do return FLOAT, so we will have to allow it as it is spec compliant.

kenrussell:
The fact that the implementation color read format and type may return integer types doesn't reflect the precision of the half-float texture used to back the framebuffer. It only reflects that the driver will perform the conversion to the integer type upon readback. Since it's defined that RGBA/FLOAT are supported as the main ReadPixels format/type combination, it's always possible to get back the data from the GPU with full precision for either FP32 or FP16 framebuffers.

jdashg:
My take has been that the driver can truly return whatever it wants for IMPL format/type.
Forbidding ints to try to prevent drivers faking float support isn't the goal here, since we test (or should test) actual float behavior separately.

PR #1666, opened 2016-05-10
Create WebGL 2 context by default in conformance2 tests using js/tests

Oletus:
Some WebGL 2 tests would default to a WebGL 1 context if
?webglVersion=2 was not added to the URL. Fix this to make the tests
easier to run standalone.

### DIFF
--- a/sdk/tests/js/tests/gl-enum-tests.js
+++ b/sdk/tests/js/tests/gl-enum-tests.js
@@ -32,7 +32,7 @@ debug("");
 debug("Canvas.getContext");
 
 var wtu = WebGLTestUtils;
-var gl = wtu.create3DContext("canvas");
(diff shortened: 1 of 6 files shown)


zhenyao:
lgtm

kenrussell:
LGTM after the fact. Just checking however: the versions of these tests in the WebGL 1.0 directory (under conformance/, not conformance2/) will still upgrade to a WebGL 2.0 context with ?webglVersion=2 , correct?

zhenyao:
Looking into the code, the answer is not.  However, we really should not run the WebGL 1 version of the test with webglVersion=2, because we already have a webgl2 version running.  So these WebGL1 versions should be marked as max-version=1.9.9.  In this sense, this CL is still good.

kenrussell:
OK.

PR #1669, opened 2016-05-10
Add default value of "" to WebGLContextEventInit::statusMessage.

jdashg:
Members of EventInit dicts must either have default values or be 'required' values.


kenrussell:
LGTM

PR #1671, opened 2016-05-11
Fix a bug in es3fTextureSpecificationTests.js

zhenyao:
The offset should be 0, but instead, it is set to the buffer size.
So the ArrayBufferView passed to the teximage call become size 0.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fTextureSpecificationTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fTextureSpecificationTests.js
@@ -2639,8 +2639,7 @@ goog.scope(function() {
                 format: this.m_texFormat,
                 width: this.m_width,
                 height: this.m_height,
-                data: data,
-                offset: this.m_texFormat.getPixelSize() * this.m_width * this.m_height
+                data: data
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fTextureSpecificationTests.js
```diff
@@ -2639,8 +2639,7 @@ goog.scope(function() {
                 format: this.m_texFormat,
                 width: this.m_width,
                 height: this.m_height,
-                data: data,
-                offset: this.m_texFormat.getPixelSize() * this.m_width * this.m_height
+                data: data
             });
```
zhenyao: c++ code for comparison: https://code.google.com/p/chromium/codesearch#chromium/src/third_party/deqp/src/modules/gles3/functional/es3fTextureSpecificationTests.cpp&l=1656

zhenyao:
@kenrussell @toji Please take a look.

kenrussell:
Wow. Hard to understand how any of the tests could have been passing. Great work tracking this down. LGTM

PR #1673, opened 2016-05-11
Realize function deinit() in tcuTestCase.js

Jiawei-Shao:
Many cases fail because some related gl states fail to be reset after the completion of last case, such as cases in FboRender and FboMultisamples, etc. 
Original deinit() functions in es3fFboRenderTest.js and es3fFboTestCase.js are moved to function deinit() in tcuTestCase.js.
16 (of 46) cases in FboMultisample will pass after this patch

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fFboRenderTest.js
+++ b/sdk/tests/deqp/functional/gles3/es3fFboRenderTest.js
@@ -708,44 +708,14 @@ goog.scope(function() {
      * deinit
      */
     es3fFboRenderTest.FboRenderCase.prototype.deinit = function() {
-        gl.colorMask(true, true, true, true);
-        gl.depthMask(true);
-
         gl.clearColor(0.0, 0.0, 0.0, 0.0);
         gl.clearDepth(1.0);
         gl.clearStencil(0);
 
         gl.disable(gl.STENCIL_TEST);
         gl.disable(gl.DEPTH_TEST);
-        gl.disable(gl.BLEND)
-        gl.disable(gl.SAMPLE_COVERAGE);
-        gl.disable(gl.SAMPLE_ALPHA_TO_COVERAGE);
+        gl.disable(gl.BLEND);
 
-        if (this.m_program) {
-            gl.deleteProgram(this.m_program.getProgram());
-            this.m_program = null;
-        }
-        if (this.m_msColorRbo) {
-          gl.deleteRenderbuffer(this.m_msColorRbo);
-          this.m_msColorRbo = null;
-        }
-        if (this.m_msDepthStencilRbo) {
-          gl.deleteRenderbuffer(this.m_msDepthStencilRbo);
-          this.m_msDepthStencilRbo = null;
-        }
-        if (this.m_resolveColorRbo) {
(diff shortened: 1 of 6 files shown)


sdk/tests/deqp/framework/common/tcuTestCase.js
```diff
@@ -161,9 +161,50 @@ goog.scope(function() {
...
+     * Deinit
      */
-    tcuTestCase.DeqpTest.prototype.deinit = function() {};
+    tcuTestCase.DeqpTest.prototype.deinit = function() {
+        // Multisampling state
+        gl.disable(gl.SAMPLE_COVERAGE);
+        gl.disable(gl.SAMPLE_ALPHA_TO_COVERAGE);
+
+        // Texture state
+        gl.bindTexture(gl.TEXTURE_2D, null);
```
kenrussell: This is unlikely to be sufficient, given that at least some of the tests call activeTexture(). I'm not sure whether the best thing to do would be to iterate through all of the texture units, or remove this from here and do it in the deinit() methods of test subclasses that need it.

kenrussell: Note that we need to be careful not to slow down test execution too much with these state resets.

sdk/tests/deqp/functional/gles3/es3fFboRenderTest.js
```diff
@@ -704,52 +704,6 @@ goog.scope(function() {
...
-        }
-
-        if (this.m_msFbo) {
-          gl.deleteFramebuffer(this.m_msFbo);
-          this.m_msFbo = null;
-        }
-        if (this.m_resolveFbo) {
-          gl.deleteFramebuffer(this.m_resolveFbo);
-          this.m_resolveFbo = null;
-        }
```
kenrussell: Would it be better to leave the deletion of these objects in this subclass's deinit() method?

Is there a straightforward way to do a superclass call in this framework?

kenrussell:
@zhenyao sorry, I didn't see your review for some reason -- looks like you caught all of the issues I found and more.

sdk/tests/deqp/framework/common/tcuTestCase.js
Jiawei-Shao: Thanks for your comments. I will do more investigation then.

sdk/tests/deqp/functional/gles3/es3fFboRenderTest.js
Jiawei-Shao: These codes are referenced directly from es3fMultisample by me:
https://github.com/Jiawei-Shao/WebGL/commit/98cb5837fe8d9dc2268517f53d8a6f8b5f921b92
https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js#L216
Now I find in fact these objects don't appear in the execution of all cases in FboRender so it is unnecessary to keep these codes any more.

zhenyao:
Mostly looks good with one question and one minor suggestion (adding a comment to explain why the code differ from c++ version)

zhenyao:
lgtm

kenrussell:
LGTM too. @zhenyao would you like to merge this?

PR #1674, opened 2016-05-11
Disallow FLOAT_32_UNSIGNED_INT_24_8_REV in TexSubImage{2|3}D

zhenyao:
@kenrussell @jdashg Please review.


zhenyao:
Note that the texImage{2|3}D part is already there.  texSubImage{2|3}D part is simply an negligence to be fixed.

kenrussell:
Looks good to me. Let's let @jdashg review too.

jdashg:
I'm not sure I agree with the rationale, but I can accept that this normalizes the spec's treatment of the matter.
I think we should look into relaxing this later.

Actually, isn't this a result of D3D-ANGLE's limitations as seen in ANGLE_depth_texture. IIRC, we couldn't upload depth data in D3D, hence ANGLE_depth_texture's limitations.
If this is still the case, this should be the rationale for forbidding depth (and stencil?) uploads.

Please check with ANGLE before merging the wrong rationale.

zhenyao:
I think the limitation you mentioned is D3D9 only and not in D3D11, but @null77 can you confirm?

zhenyao:
OK, I'll remove the rationale and land this first.

zhenyao:
The rationale removed. Merging this.

null77:
It's certainly quite easy to upload depth and stencil data in D3D11. I believe this is not possible in D3D9, so as you said Mo, it makes sense that the restriction was created for D3D9.

PR #1675, opened 2016-05-11
Fix bugs for texsubimage tests in texturespecification/

zhenyao:


sdk/tests/deqp/framework/opengl/simplereference/sglrReferenceContext.js
```diff
@@ -4442,12 +4442,6 @@ goog.scope(function() {
                  target == gl.TEXTURE_CUBE_MAP_POSITIVE_Y ||
                  target == gl.TEXTURE_CUBE_MAP_NEGATIVE_Z ||
                  target == gl.TEXTURE_CUBE_MAP_POSITIVE_Z) {
-            // Validate size and level.
-            if (this.conditionalSetError(width != height || width > this.m_limits.maxTextureCubeSize || depth != 1, gl.INVALID_VALUE))
-                return;
-            if (this.conditionalSetError(level > Math.floor(Math.log2(this.m_limits.maxTextureCubeSize)), gl.INVALID_VALUE))
-                return;
-
```
zhenyao: This is just wrong.  Checking width == height for texsubimage calls causes all cubemap reference image generation to fail.  The C++ code doesn't have these checks - likely copied over from teximage part and forgot to delete.

https://code.google.com/p/chromium/codesearch#chromium/src/third_party/deqp/src/framework/opengl/simplereference/sglrReferenceContext.cpp&l=1344

sdk/tests/deqp/functional/gles3/es3fTextureSpecificationTests.js
```diff
@@ -2781,22 +2781,22 @@ goog.scope(function() {
...
+        tcuTextureUtil.fillWithGrid(access, 4, [1, 0, 0, 1], [0, 1, 0, 1]
         );
         this.m_context.pixelStorei(gl.UNPACK_ALIGNMENT, this.m_alignment);
         for (var f in tcuTexture.CubeFace) {
             var face = tcuTexture.CubeFace[f];
             this.m_context.texSubImage2D(
                 es3fTextureSpecificationTests.s_cubeMapFaces[face],
                 0, this.m_subX, this.m_subY, this.m_subW, this.m_subH,
-                this.m_format, this.m_dataType, new Uint8Array(data)
+                this.m_format, this.m_dataType, access.getDataPtr()
```
zhenyao: Of course we will need this.  The ArrayBufferType has to match that of the format.

zhenyao:
This patch makes texturespecification/texsubimage2d_align.html all passing.  It could help other test cases  too.

sdk/tests/deqp/framework/opengl/simplereference/sglrReferenceContext.js
kenrussell: Great catch.

sdk/tests/deqp/functional/gles3/es3fTextureSpecificationTests.js
kenrussell: Very good.

kenrussell:
Excellent work tracking these down and fixing them. LGTM.

PR #1676, opened 2016-05-12
Fix texturespecification/teximage2d_align.html

zhenyao:
The c++ tests specifies number of levels, whereas the js tests compute number
of levels based on width/height. So js tests include the 1 pixel level, which
caused rendering to 3x1 or 2x1 destinations failing.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fTextureSpecificationTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fTextureSpecificationTests.js
@@ -405,7 +405,7 @@ goog.scope(function() {
             for (var ndx = 0; ndx < 2; ndx++) {
                 /** @type {tcuSurface.Surface} */
                 var dst = ndx ? reference : result;
-                ctx = ndx ? refContext : webgl2Context;
+                var ctx = ndx ? refContext : webgl2Context;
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fTextureSpecificationTests.js
```diff
@@ -1415,10 +1415,7 @@ goog.scope(function() {
         es3fTextureSpecificationTests.Texture2DSpecCase.call(
             this, name, desc, gluTextureUtil.mapGLTransferFormat(
                 format, dataType
-            ), width, height,
-            es3fTextureSpecificationTests.maxLevelCount(
-                width, height
-            )
+            ), width, height, numLevels
         );
```
zhenyao: c++ code: https://code.google.com/p/chromium/codesearch#chromium/src/third_party/deqp/src/modules/gles3/functional/es3fTextureSpecificationTests.cpp&l=953

zhenyao:
@kenrussell Please review. This took me three hours to pin down the issue. Now I found it, it seems like there is clear evidence leading to the root cause. It's fun, like doing detective work. :)

kenrussell:
Excellent catch @zhenyao ! LGTM.

PR #1677, opened 2016-05-12
Fix negative one index for WebGL 2.0

qkmiao:
Primitive restart is always enabled in WebGL 2.0. -1 index is treated as
a primitive restart and nothing is rendered. This is different with
WebGL 1.0.


zhenyao:
Looks good.

PR #1678, opened 2016-05-12
Allow non-null FLOAT_32_UNSIGNED_INT_24_8_REV uploads.

jdashg:
It sounds like the only reason we disabled this is because there's no good match for the type of the ArrayBufferView to be uploaded. We do seem to allow other depth/stencil upload types.

We should just let people upload it if they manage to marshal it, perhaps just allowing ArrayBuffer here.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -1230,7 +1230,7 @@ WebGL2RenderingContext implements WebGL2RenderingContextBase;
           </span>
         </p>
       </dt>
-      <dd>        
+      <dd>
         <p>Accepts internal formats from OpenGL ES 3.0 as detailed in the <a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-4.4.2">specification</a> and <a href="http://www.khronos.org/opengles/sdk/docs/man3/html/glRenderbufferStorage.xhtml" class="nonnormative">man page</a>.</p>
         <p>To be backward compatible with WebGL 1, also accepts internal format <code>DEPTH_STENCIL</code>, which should be mapped to <code>DEPTH24_STENCIL8</code> by implementations.</p>
       </dd>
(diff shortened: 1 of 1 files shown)


specs/latest/2.0/index.html
```diff
@@ -1376,6 +1375,7 @@ <h2 class="no-toc">Table of contents</h2>
           <tr><td>Uint32Array</td><td>UNSIGNED_INT_24_8</td></tr>
           <tr><td>Uint16Array</td><td>HALF_FLOAT</td></tr>
           <tr><td>Float32Array</td><td>FLOAT</td></tr>
+          <tr><td>ArrayBuffer</td><td>FLOAT_32_UNSIGNED_INT_24_8_REV</td></tr>
```
kenrussell: ArrayBuffer is not compatible with the ArrayBufferView argument in the IDL. This change does not look good as is.

I am dubious that the added value of supporting uploads for this texture type is worth the perturbation to the IDL. Another overload would have to be added in order to support just this texture type -- assuming we don't want to use one of the other existing ArrayBufferView types for it (which would probably be confusing). The best fit would be either Float32Array or Uint32Array.

lexaknyazev:
From https://github.com/KhronosGroup/WebGL/issues/2667#issuecomment-409387227:
> In the last WebGL working group meeting it was resolved to change the spec to allow ArrayBuffer for these entry points that currently only allow ArrayBufferView.

Is it still expected? If so, this PR could be extended to cover that issue.

kenrussell:
We do still aim to make that change, but right now all browser vendors are making a push to pass the top-of-tree conformance suite before taking the next snapshot. It's not a good time right now to make significant API changes - and this would be one.

lexaknyazev:
@kenrussell @jdashg 
Do you still expect to implement this change, along with #2667?

jdashg:
Yes, some day. :) Thanks for the bump!

PR #1680, opened 2016-05-12
Similar fix to make texturespecification/teximage2d_pbo.html pass on …

zhenyao:
…Mac.

It still fails on Linux with GL errors. Will look into that separately.


zhenyao:

kenrussell:
Great work. LGTM

PR #1681, opened 2016-05-13
Fix array index bug in WebGL 2.0

qkmiao:
In WebGL 2.0, primitive restart is always enabled. Avoid using primitive
restart fixed index for element index.


zhenyao:
lgtm

PR #1685, opened 2016-05-18
Fix the ES3 IntegerStateQuery blendfunc tests reset

Kangz:
Previously the initial value for that blendfunc parameter was used to
reset all of the blendfunc parameters, which was wrong because the
default for SRC parameters is ONE while the DST parameters's is ZERO

Here we take care to only touch the one parameter being modified. If we wanted to stay 100% close to dEQP could instead not use an initial value and reset the blendfunc with the known default values. I didn't choose this approach because it would make the testing of the initial value of blendfunc useless as it would be reset just before.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fIntegerStateQueryTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fIntegerStateQueryTests.js
@@ -883,12 +883,12 @@ goog.scope(function() {
 		switch (this.m_testTargetName) {
 			case gl.BLEND_SRC_RGB:
 			case gl.BLEND_SRC_ALPHA:
-				gl.blendFunc(func, gl.ZERO);
+				gl.blendFunc(func, gl.getParameter(gl.BLEND_DST_RGB));
 				break;
 
 			case gl.BLEND_DST_RGB:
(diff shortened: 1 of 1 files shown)


zhenyao:
Looks good!

kenrussell:
Thanks for this fix @Kangz and for reviewing it @zhenyao . I trust that this has been tested locally. The question in my mind is that deinit() is intended to reset all affected states to their default values, correct? It seems to me that querying the parameters defeats the purpose of this method, and will introduce implicit state dependencies between the tests in this file. Is it possible that that will be a side-effect of this change?

Kangz:
Thanks @zhenyao for the review. The problem with BlendFunc or BlendFuncSeparate is that multiple states are changed in a single call. The ```deinit``` logic could reset all the states at once, however it means we wouldn't be able to test for the initial value of the DST state as it would be reset to the correct init value at the end of the SRC test. Querying the values and passing them again to BlendFunc allows to keep the current state of DST for later tests.
Let me know if a complete reset is better, the patch would be trivial too.

zhenyao:
We can add a separate test in conformance2/state to verify the initial states, so here we can remain close to the c++ deqp.  Does that sound like a solution?

PR #1686, opened 2016-05-19
add point-size-varyings.html

emnh:
This check attempts to provoke a [bug in ANGLE](https://bugs.chromium.org/p/angleproject/issues/detail?id=1380) on Windows. 

The test passes in Edge on Windows and in Firefox and Chrome on Linux, and fails in Chrome (version 50.0.2661.102) on Windows and in Firefox (version 46.0.1) on Windows. I heard it also fails in Chrome on OS X. 

When gl_PointSize is mentioned but not set in vertex shader and gl_PointCoord is mentioned but not read in fragment shader, rendering a triangle coloured with varyings fails. The check is based on triangle.html in same folder.

### DIFF
--- a/sdk/tests/conformance/rendering/00_test_list.txt
+++ b/sdk/tests/conformance/rendering/00_test_list.txt
@@ -18,6 +18,7 @@ multisample-corruption.html
 --min-version 1.0.3 negative-one-index.html
 --min-version 1.0.3 point-no-attributes.html
 point-size.html
+--min-version 1.0.4 point-size-varyings.html
 --min-version 1.0.3 point-with-gl-pointcoord-in-fragment-shader.html
 --min-version 1.0.3 polygon-offset.html
 --min-version 1.0.2 simple.html
--- a/sdk/tests/conformance/rendering/point-size-varyings.html
+++ b/sdk/tests/conformance/rendering/point-size-varyings.html
@@ -1,0 +1,111 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Rendering Test</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"> </script>
+</head>
+<body>
+<canvas id="example" width="50" height="50">
+There is supposed to be an example drawing here, but it's not important.
+</canvas>
+<div id="description"></div>
+<div id="console"></div>
+<script id="vshader" type="x-shader/x-vertex">
+precision highp float;
+uniform float isCloud; // never set
+attribute vec4 vPosition;
+varying vec2 vUV;
+void main()
+{
+    gl_Position = vPosition;
+    vUV = vPosition.xy * 2.0;
+    if (isCloud == 1.0) {
+      gl_PointSize = 1.0;
+    }
+}
+</script>
+
+<script id="fshader" type="x-shader/x-fragment">
+precision highp float;
+varying vec2 vUV;
+uniform float isCloud; // never set
+void main()
+{
+    gl_FragColor = vec4(vUV.xy,0.0,1.0);
+    if (isCloud == 1.0) {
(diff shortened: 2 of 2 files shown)


kenrussell:
I'm reluctant to add this test. It touches a corner case of the specification that nobody has run into in the past several years, and working around it could require a lot of code in the WebGL implementation. In the worst case, it would be necessary to recompile programs based on the kind of primitive that is being rendered, and eliminate references to gl_PointSize and gl_PointCoord if not rendering points. The working group's time is better spent elsewhere, not working around this bug.

I don't see failures on OS X, but am running on an NVIDIA GPU and there could well be problems with Intel or AMD GPUs.

Looking for opinions from @zhenyao @jdashg @RafaelCintron @grorg .

jdashg:
I'll double-check the spec to see what's actually (technically) mandated.

(Aside: Ugh, we should have dropped support for POINTS. These things keep cropping up)

zhenyao:
Since if draw primitive is not points , then gl_PointCoord is undefined, then we should generate an INVALID_OPERATION in draw calls with a non point primitive and gl_PointCoord is statically read in fragment shader.

jdashg:
Here's a more minimal testcase:
https://github.com/jdashg/WebGL/tree/point-size-varyings

@zhenyao 's proposal would resolve this at the expense of breaking content. (such as the origin of this testcase)
Unfortunately, I'm not coming up with any better ideas without requiring a decent amount of voodoo and/or overhead.
I think we should merge this test (preferably the minimal version), as it does test what should happen, spec-as-written.

If we want to change the spec (I think so), INVALID_OP seems like the best solution, but we'll need to keep an eye on webcompat as we roll it out.

Finally, regardless of adding INVALID_OP, ANGLE has a bug they should probably fix. :)

jdashg:
I will add that gl_PointSize may alternatively be actually assigned (as opposed to statically assigned), and this bug will still repro in ANGLE.

jdashg:
I issued a PR to simplify the test:
https://github.com/emnh/WebGL/pull/1

zhenyao:
I do not worry too much about breaking contents, because this is undefined values, and any webgl content relying on undefined values should be fixed.

That said, this undefined behavior puts no threat, so I am OK whether we allow it or not.

jdashg:
The problem is that the content is *not* relying on undefined values. In the example this was pulled from, gl_PointCoord is only actually read when drawing POINTS, and is not actually (but is statically) read when drawing non-POINTS.

This is well-formed, and should work according to the current spec.
Whether we should disallow static reads of gl_PointCoord for non-POINTS should really be a separate issue.

jdashg:
@kenrussell Do you agree with the reasoning for including this test, even if we do change the behavior later?

kenrussell:
I still don't think it is a good idea to include this test.

The original test was more targeted, since it conditionally assigned gl_PointSize and read gl_PointCoord. The simplified version always reads gl_PointSize, and writes gl_FragColor twice. A compiler can easily optimize out the first write and therefore the read of gl_PointCoord. This is not possible in the original test case. In the original test, if the conditional write of gl_PointSize is removed, the test starts working, and similar for the conditional read of gl_PointCoord. It's much harder to make the original test case actually work.

It's difficult to reason about why ANGLE would be broken for this test case. It's likely that mobile GPUs, which historically have not included dynamic branching and execute all code paths with conditional kill flags of certain values, would have problems with static (but not dynamic) references to these built-in variables.

The strict interpretation Mo suggests sounds acceptable if it is changed in the following way:

 - If a program's vertex and fragment shaders either statically write gl_PointSize or read gl_PointCoord, and this program is used to draw any primitive type except POINTS, INVALID_OPERATION is generated.

I am concerned about the potential impact on web compatibility. Testing should be done to make sure this new rule doesn't break any of the major game engines. I maintain that this corner case is not worth the working group's effort, at least not at this time.

Again, if anything is going to be merged, I think it should be the original test case. It should be modified to require INVALID_OPERATION and renamed to something more obvious like referencing-point-only-glsl-variables-while-rendering-triangles.html.

jdashg:
Let's do this one step at a time.
If we want to make a *spec change*, I think we should do this in another issue/PR.

I'll add back in the unused branch, though it's not fool-proof either. (it could be jit-recompiled-out anyways)

I disagree with @kenrussell's proposal. There's nothing unsafe about writing to gl_PointSize in non-POINTS draw calls, thus there's no reason to restrict this. We should only be restricting reads from gl_PointCoord, since this is unsafe outside of POINTS draws.

kenrussell:
The original test case did not write to gl_PointSize dynamically, only statically -- the write was in an un-taken branch. If the un-taken branch's body was commented out, the combined vertex and fragment shaders rendered correctly, even though the fragment shader statically read from gl_PointCoord (again, in an un-taken branch).

Clearly, the presence of the unused write to gl_PointSize had an effect, so it seems to me that it is necessary to also prohibit writes to that variable for non-POINTS draws. What do you think?

Opinions @grorg @RafaelCintron ?

jdashg:
The presence of gl_PointSize only changes behavior in ANGLE. It's just a bug in ANGLE.
The spec doesn't provide any reason gl_PointSize needs to be treated specially.

kenrussell:
The WebGL spec contains restrictions that reflect the behavior of drivers on multiple platforms. Since there's a problem on ANGLE, it's easy to believe that there may be similar problems on non-Windows platforms.

jdashg:
I disagree. This is very likely to be limited to ANGLE.
And if we find it elsewhere, we will work around it.

kenrussell:
@null77 should probably comment on how ANGLE-specific this bug appears to be.

RafaelCintron:
I agree we should have a test for this. Edge currently ensures gl_PointCoord.xy is forced to 0,0 when drawing any topology besides POINTS. Writes to gl_PointSize are ignored. Thus far, we haven't received interop bugs by doing this.

Leaving gl_PointCoord undefined in these instances could lead to information disclosure bugs, which is undesirable.

RafaelCintron:
If we haven't already done so, we should have tests for both point and triangle drawing against all of the combinations of gl_PointSize and gl_PointCoord:
- Both gl_PointSize and gl_PointCoord present
- No gl_PointSize present, gl_PointCoord present.
- Neither gl_PointSize nor gl_PointCoord present.
- Negative gl_PointSize.
- gl_PointSize present, gl_PointCoord not present.

null77:
Ken et al., this bug is pretty ANGLE-specific. The problem is our code currently assumes in D3D11 (non 9_3) that when the vertex shader writes to gl_PointSize and reads gl_PointCoord in the fragment shader, the app will render with GL_POINTS. If it satisfies the first two conditions but breaks the third, D3D11 will complain that a fragment (pixel) shader input is missing for gl_PointCoord, which we would have written in the geometry shader for GL_POINTS. That being said if gl_PointCoord would be undefined, I think it makes sense to initialize it to (0,0) or (0.5,0.5).

null77:
Also just a note that we can fix this using some more dynamic generation of shaders. We can output a different vertex shader when rendering with GL_POINTS and using gl_PointCoord, and when not using GL_POINTS and using gl_PointCoord. It's a bit award but a sensible fix.

kenrussell:
Per comments above and offline discussion in the working group, merging this pull request now to add the test. Will follow up with another pull request making the test's name more descriptive.

PR #1687, opened 2016-05-19
Make more texturespecification tests pass

zhenyao:
including:
  basic_teximage3d.html
  texstorage2d.html
  texstorage3d.html


kenrussell:
LGTM. Thanks for tracking down these problems.

PR #1688, opened 2016-05-19
Better fix for the IntegerStateQuery blendfunc test

Kangz:
Make a better version of the fix for the IntegerStateQuery as suggested in https://github.com/KhronosGroup/WebGL/pull/1685

### DIFF
--- a/sdk/tests/conformance/state/00_test_list.txt
+++ b/sdk/tests/conformance/state/00_test_list.txt
@@ -3,6 +3,6 @@ gl-enable-enum-test.html
 gl-get-calls.html
 gl-geterror.html
 --max-version 1.9.9 gl-getstring.html
+--min-version 1.0.4 gl-initial-state.html
 --max-version 1.9.9 gl-object-get-calls.html
 --min-version 1.0.3 state-uneffected-after-compositing.html
-
--- a/sdk/tests/conformance/state/gl-initial-state.html
+++ b/sdk/tests/conformance/state/gl-initial-state.html
@@ -1,0 +1,79 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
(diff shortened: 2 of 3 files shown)


sdk/tests/conformance2/state/gl-initial-state.html
```diff
@@ -0,0 +1,79 @@
...
+<canvas id="canvas" width="2" height="2"> </canvas>
+<script>
+"use strict";
+description("This test checks the initial value of the OpenGL state");
+
+debug("");
+debug("Canvas.getContext");
+
+var wtu = WebGLTestUtils;
+var gl = wtu.create3DContext("canvas", null, 2);
```
zhenyao: Sorry to mislead you, but can you move this test to conformance/state/ and leave this create3DContext without a version parameter, so we could run it in both WebGL1 and WebGL 2? For the initial settings in this test, they are also available in WebGL1.

Kangz: Done

sdk/tests/conformance/state/00_test_list.txt
```diff
@@ -5,4 +5,4 @@ gl-geterror.html
 --max-version 1.9.9 gl-getstring.html
 --max-version 1.9.9 gl-object-get-calls.html
 --min-version 1.0.3 state-uneffected-after-compositing.html
-
+gl-initial-state.html
```
kenrussell: Please alphabetize and add --min-version 1.0.4. Thanks.

Kangz: Done

kenrussell:
LGTM. Thanks. Let's let @zhenyao re-review and merge.

zhenyao:
lgtm.

Kangz:
Thank you for the merge, please tell me when this is rolled so I can remove the expectation.

PR #1689, opened 2016-05-19
Implement copyTex{Sub}Image2D in sglrReferenceContext

zhenyao:
This makes the following two tests passing (at least on Linux):
  texturespecification/basic_copyteximage2d.htm
  texturespecification/basic_copytexsubimage2d.htm


zhenyao:
@kenrussell These two functions are strictly following their c++ version. Please take a look.

kenrussell:
This is superb Mo! LGTM!

PR #1703, opened 2016-05-23
Optimize texturespecification tests.

zhenyao:
THis cuts running time from 645s to 568s on my local Linux,
about 12% gain.


zhenyao:

sdk/tests/deqp/framework/common/tcuTexture.js
```diff
@@ -2033,6 +2033,12 @@ tcuTexture.PixelBufferAccess.prototype.clear = function(color, x, y, z) {
     var width = range_x[1] - range_x[0];
     var height = range_y[1] - range_y[0];
     var depth = range_z[1] - range_z[0];
+    if (x === undefined && y === undefined && z == undefined &&
```
kenrussell: Presumably the "z" test should use the triple-equals operator like the other two.

kenrussell:
Excellent work. LGTM

PR #1706, opened 2016-05-23
Remove debug wrapper from passing tests.

zhenyao:
Also, remove debug logging to console from primitiverestart test,
because it's polluting the bot output.


kenrussell:
LGTM

PR #1707, opened 2016-05-24
check gl.getVertexAttrb and gl.getVertexAttribOffset

greggman:
It looks like this was never tested. It fails in Firefox

Should probably test more than attribute 0? Also should probably test rendering with each combination?


zhenyao:
lgtm

kenrussell:
Thanks for catching this Gregg. LGTM after the fact.

kenrussell:
Also, this is a good start. It can be expanded out later.

dmikis:
Why `vertexAttribPointer` call should take effect on attribute state while vertex array is disables on that attribute? Shouldn't it this test require `enableVertexAttribArray` call?

zhenyao:
No, whether a vertex attribute array is enabled or not is irrelevant to its
state setup

dmikis:
I've figured that out and removed the comment.

PR #1708, opened 2016-05-24
Split textureformat.html into multiple tests of reasonable run time.

zhenyao:
Also, remove depth32f_stencil8 from test list.

(Note this CL does not include all the generated files for easier review)


zhenyao:
Merging myself.

PR #1710, opened 2016-05-25
Fix an error in FboInvalidateTests

Jiawei-Shao:
In InvalidateFboUnbindReadCase the result picture should be rendered from the buffer that is not invalidated.


zhenyao:
lgtm

PR #1713, opened 2016-05-26
Add expando loss test for second query object

qkmiao:
@kenrussell  please help to review this.

### DIFF
--- a/sdk/tests/conformance2/misc/expando-loss-2.html
+++ b/sdk/tests/conformance2/misc/expando-loss-2.html
@@ -115,21 +115,37 @@ function testBasicBindings() {
 function testQueries() {
     debug('Query');
 
-    var query = gl.createQuery();
-    setTestExpandos(query);
-    gl.beginQuery(gl.ANY_SAMPLES_PASSED, query);
+    expandoValue = "First query";
+    var query1 = gl.createQuery();
+    setTestExpandos(query1);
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/misc/expando-loss-2.html
```diff
@@ -115,21 +115,37 @@
...
+    expandoValue = "First query";
+    verifyTestExpandos(retrievedQuery1, "Query");
+    shouldBeType(retrievedQuery1, 'WebGLQuery');
+
+    expandoValue = "Second query";
+    verifyTestExpandos(retrievedQuery2, "Query");
+    shouldBeType(retrievedQuery2, 'WebGLQuery');
 
     gl.endQuery(gl.ANY_SAMPLES_PASSED);
+    gl.endQuery(gl.TRANSFORM_FEEDBACK_PRIMITIVES_WRITTEN);
```
kenrussell: Do these queries need to be ended in the reverse order they were begun?

kenrussell:
Looks great. Thank you for adding this test. LGTM

sdk/tests/conformance2/misc/expando-loss-2.html
qkmiao: As I known, it doesn't matter what's the order.

PR #1714, opened 2016-05-26
indexing to shader output is not correct in outerproduct

Richard-Yunchao:
I have come back to office this week from honeymoon, :). Long time not to touch the code. Please take a look at this small patch, @zhenyao and @kenrussell . It can fix all failures for outerproduct.html against vertex shader tests. 

Thanks a lot!

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
+++ b/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
@@ -2986,7 +2986,7 @@ var setParentClass = function(child, parent) {
             var size = reference.rows * reference.cols;
             for (var i = 0; i < reference.rows; i++)
                 for (var j = 0; j < reference.cols; j++)
-                    ret.set(i, j, output[size * index + j * reference.cols + i]);
+                    ret.set(i, j, output[size * index + j * reference.rows + i]);
             return ret;
         }
 


sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
```diff
@@ -2986,7 +2986,7 @@ var setParentClass = function(child, parent) {
             var size = reference.rows * reference.cols;
             for (var i = 0; i < reference.rows; i++)
                 for (var j = 0; j < reference.cols; j++)
-                    ret.set(i, j, output[size * index + j * reference.cols + i]);
+                    ret.set(i, j, output[size * index + j * reference.rows + i]);
```
Richard-Yunchao: This small function set() sets the value from output array of builtin function in shader through transform feedback. It works in this way: ret[i, j] = output[size * index + j * reference.cols + i]; 
Considering that reference.rows = 2, reference.cols = 3, and index = 0 (get the first result for outerproduct(vec2, vec3). For the original implementation, Then the first result of mat3x2 would be: 
ret[0, 0] = output[0], 
ret[0, 1] = output[3], 
ret[0, 2] = output[6]
ret[1, 0] = output[1], 
ret[1. 1] = output[4], 
ret[1, 2] = output[7]. 
They are not correct, output[6] and output[7] are out of output[0] ~ output[5]. 

After the patch applied, the result would be:
ret[0, 0] = output[0], 
ret[0, 1] = output[2], 
ret[0, 2] = output[4]
ret[1, 0] = output[1], 
ret[1. 1] = output[3], 
ret[1, 2] = output[5]. 

If we use 
ret.set[i, j, output[size * index + i * reference.cols + j]); 
All indices for output array are among (0, 5), but it is not logically correct.

zhenyao: Welcome back!

I don't think it's apparent why ret.set(i, j, output[size * index + j * reference.rows + i]) is logically correct whereas ret.set[i, j, output[size * index + i * reference.cols + j]) is not.  It really depends on how the data is laid out in |output|.  Can you explain a bit more and maybe add a comment in the code here?

Richard-Yunchao: Yeah, It depends on the layout. I mean, in the shader, it is in column-major order, but in JS (like c/c++/java), array is always in row-major order. 
Here comes the example. Let's take outerproduct(vec2, vec3) for an example too. The result would be
c0 * r0,     c0 * r1,      c0 * r2
c1 * r0,     c1 * r1,      c1 * r2 

In shader, the layout of the index would be (colum-majored, treat it located in 1 dimension array): 
0, 2, 4
1, 3, 5.   
But in JS, it is in row-major order. So when we get the result from shader, the index should be 0, 2, 4, 1, 3, 5. Exactly the same with the index after my patch applied. 

Please correct me if I am wrong. :)

zhenyao:
lgtm
Thanks for the explanation.

PR #1715, opened 2016-05-27
Clarified recently-added test of point-specific shader variables.

kenrussell:
- Renamed the test.
 - Added back in variants that only conditionally write gl_PointSize.
 - Changed the writes of gl_PointSize to write 1.0, since that's the only guaranteed supported maximum point size.

Future work to test point rendering in these scenarios is TBD. At least some of this functionality is already covered by other tests.


kenrussell:
Please review @zhenyao @jdashg @RafaelCintron in particular. Thanks.

zhenyao:
lgtm
Thanks for the thorough tests

jdashg:
lgtm

PR #1716, opened 2016-05-27
Remove redundant array allocation to fix the bugs for outerproduct

Richard-Yunchao:
This change can fix all failures in outerproduct,html, with some coding style change like indentation. 

PTAL. Thanks.

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsShaderExecUtil.js
+++ b/sdk/tests/deqp/modules/shared/glsShaderExecUtil.js
@@ -613,32 +613,32 @@ goog.scope(function() {
         vertexArrays.push(gluDrawUtil.newFloatVertexArrayBinding('a_position', 2, numValues, 0, positions));
 
         for (var inputNdx = 0; inputNdx < this.m_inputs.length; inputNdx++) {
-        symbol = this.m_inputs[inputNdx];
-        var attribName = 'a_' + symbol.name;
-        var ptr = inputs[inputNdx];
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/modules/shared/glsShaderExecUtil.js
```diff
@@ -685,10 +685,11 @@ goog.scope(function() {
...
                 gl.readPixels(0, 0, framebufferW, framebufferH, transferFormat.format, transferFormat.dataType, tmpBuf.getAccess().getDataPtr());
 
-                if (outSize == 4 && outNumLocs == 1)
+                if (outSize == 4 && outNumLocs == 1) {
                     outputs[outNdx] = new Uint8Array(tmpBuf.getAccess().getBuffer());
-                else {
-                    outputs[outNdx] = new Uint32Array(numValues * outVecSize);
+                } else {
+                    if (locNdx == 0)
+                        outputs[outNdx] = new Uint32Array(numValues * outVecSize);
```
Richard-Yunchao: The inner 'for' loop from line 684 to line 700 will insert data to the same array by setting offset during each iteration. We just need to allocate one array for each output (It is the result of one 'out' variable in fragment shader) at the beginning. If we allocate multiple arrays in the inner 'for' loop, data stored in arrays (except the last one) are lost.

zhenyao:
lgtm

Thanks for tracking down this bug.

PR #1717, opened 2016-05-27
Deleted objects should not exist

qkmiao:
Name of deleted buffer immediately becomes invalid. Though the
continue using the object ("D.1.3 Deleted Object and Object Name
Lifetimes", page 315 ES spec 3.0.4).

Also in section 3 "WebGL Resources" in WebGL 1.0 spec,
"If authors wish to control when the underlying
resource is released then the delete call can be made explicity".
So the deleted object cannot be used any more.

### DIFF
--- a/sdk/tests/conformance2/vertex_arrays/vertex-array-object.html
+++ b/sdk/tests/conformance2/vertex_arrays/vertex-array-object.html
@@ -586,13 +586,8 @@ function runBoundDeleteTests() {
       var isPositionBuffer = gl.isBuffer(positionBuffer);
       var isColorBuffer    = gl.isBuffer(colorBuffer);
 
-      if(ii < 3) {
-        if(!isPositionBuffer) testFailed("Position buffer should still exist until last ref removed");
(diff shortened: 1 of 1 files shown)


qkmiao:
@kenrussell @zhenyao @bkotzz , please help to review. Thanks.

zhenyao:
lgtm

bkotzz:
I might be misunderstanding this, but in this case the same position buffer is attached to four different VAOs. Calling delete on the position buffer means it is unattached from the currently bound VAO, but still attached to the other VAOs. Is the name of the buffer really invalid at this point, if it is still being referred to by other VAOs? If so, does that mean the other VAOs are storing an invalid name for their position buffer attachment?

bkotzz:
Okay, the third paragraph of D.1.3 clears things up. Deleting the buffer object means the name is invalid in all contexts, even though it still may be bound in the non-current context, and (oddly enough) still may be returned by genBuffer calls. So isBuffer should always be false after deletion.

PR #1718, opened 2016-05-27
Fix several bugs and add missed Texture2DArrayShadow cases into es3fTextureShadowTests

Jiawei-Shao:


kenrussell:
Excellent work Jiawei. I didn't compare this against the C++ sources but the fixes look good and I trust you've verified them locally. LGTM

PR #1719, opened 2016-05-27
deqp/f/gles3/pbo: Check framebuffer completness before rendering

Kangz:
This fixes a failure on Chrome on OSX where the rgb5_a1 framebuffer is
not complete.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
+++ b/sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
@@ -287,6 +287,11 @@ var tcuImageCompare = framework.common.tcuImageCompare;
                 break;
         }
 
+        if (gl.checkFramebufferStatus(gl.FRAMEBUFFER) != gl.FRAMEBUFFER_COMPLETE) {
+            testSkippedOptions("Cannot test pixel buffer readPixels on an incomplete framebuffer", true);
+            return tcuTestCase.IterateResult.STOP;
+        }
+
         this.clearColor(this.m_colorScale * 0.4, this.m_colorScale * 1.0, this.m_colorScale * 0.5, this.m_colorScale * 1.0);
 
         if (this.m_useColorClears) {


sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
```diff
@@ -287,6 +287,11 @@ var tcuImageCompare = framework.common.tcuImageCompare;
                 break;
         }
 
+        if (gl.checkFramebufferStatus(gl.FRAMEBUFFER) != gl.FRAMEBUFFER_COMPLETE) {
```
kenrussell: There are definitely framebuffer attachment formats that are required to result in a complete framebuffer. It looks to me like this could allow the implementation to behave badly and report them all as incomplete. Do you agree? If so, could you modify the tests so they have a notion of whether they're optional or not?

kenrussell:
Question above. Also, it looks like there's a typo "m_framebuffeType" throughout this test. Should we fix it?

Kangz:
I'll fix the typo. With respect to required framebuffer formats, I think another test should be checking them so that the logic can be kept simple in this one. Do you know if such a test already exists? Otherwise I'll add one.

zhenyao:
No, this is the first time I heard of this RGB555_A1 fbo incomplete issue on Mac, so maybe not a test  yet to catch this.

To me, we should not change a deqp test port so it can pass on a buggy implementation.

Kangz:
From what I understand this is not a bug, a driver has the right to declare an FBO incomplete for pretty much any reason.

zhenyao:
Not for ES3.

zhenyao:
We can discuss to change WebGL2 spec not to require it, as it will be hard to emulate on top of desktop GL drivers where the spec allows reject a combination.

Kangz:
Good point, I only had checked the desktop specification. We could work around this problem in the WebGL2 implementation by silently changing GL_RGB5_A1 to some other format but it would get complex quickly.

Kangz:
See #1755

Kangz:
Void as per the discussion on #1755

PR #1725, opened 2016-05-31
Broadcast writes to gl_FragColor to enabled color buffers

qkmiao:
Without specifying draw buffers, gl_FragColor only writes to color
number zero (GLES3 spec section 3.9.2 Shader Outputs). But with
specifying draw buffers with drawBuffer, gl_FragColor should write to
all enabled color buffers (GLES3 spec section 4.2.1 Selecting Buffers
for Writing). Also refer Issues 3) in Ext_draw_buffers extension.

### DIFF
--- a/sdk/tests/conformance2/rendering/draw-buffers.html
+++ b/sdk/tests/conformance2/rendering/draw-buffers.html
@@ -223,7 +223,7 @@ function runDrawTests() {
   var nones = makeArray(maxUsable, gl.NONE);
 
   [fb, fb2, halfFB1, halfFB2, endsFB, middleFB].forEach(function(fbo) {
-    gl.bindFramebuffer(gl.FRAMEBUFFER, fb);
+    gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
     gl.drawBuffers(bufs);
   });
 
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/rendering/draw-buffers.html
```diff
@@ -375,31 +375,37 @@
...
+  // By default, gl_FragColor only writes to color number zero.
+  // drawBuffers can specify the color buffer that gl_FragColor should write into.
+  // See GLES3 spec section 4.2.1 Selecting Buffers for Writing.
+  debug("test that gl_FragColor writes to color buffer specified by drawBuffers");
   gl.bindFramebuffer(gl.FRAMEBUFFER, fb);
-  gl.drawBuffers(bufs);
+  var bufNum = 4;
+  var blueBufs = makeColorAttachmentArray(bufNum);
+  gl.drawBuffers(blueBufs);
   gl.useProgram(blueProgramESSL1);
```
zhenyao: It is said in section 4.2.1 that for ESSL1 fragment shaders, only at most one draw buffer is written into. Only for ESSL3 fragment shaders, multiple draw buffers may be written into.  Can you test both cases?

sdk/tests/conformance2/rendering/draw-buffers.html
```diff
@@ -375,31 +375,37 @@
 
   checkAttachmentsForColor(attachments, [0, 255, 0, 255]);
 
-  // GLES3 spec section 3.9.2 Shader Outputs
-  debug("test that gl_FragColor only writes to color number zero");
+  // By default, gl_FragColor only writes to color number zero.
```
zhenyao: Why remove the original test of the default behavior? I think it's still valuable.

sdk/tests/conformance2/rendering/draw-buffers.html
kenrussell: Excellent point Mo. This is a significant difference between ES 2.0 + EXT_draw_buffers, and ES 3.0 -- ESSL 1.00 shaders can't access more than one draw buffer in ES 3.0.

What is your reading of the specification in this area? Looking at page 186 of es_spec_3.0.4.pdf, it seems to me that if an ESSL 1.00 shader attempts to write to another index of gl_FragData other than zero, the behavior is not actually defined. The shader doesn't fail compilation, etc. This is briefly discussed in section 1.5 "Compatibility" of GLSL_ES_Specification_3.00.4.pdf. I don't know what the rules are about indexing gl_FragData by a variable rather than a constant. One way or the other, it seems that a clarification to the WebGL 2.0 spec is needed. What do you think that clarification should look like? Should it only allow indexing into gl_FragData by the constant 0 in ESSL 1.00 shaders in WebGL 2.0?

qkmiao:
Thanks @kenrussell @zhenyao . I updated the pull request according your comments. So I think there are bugs in driver to deal with gl_FragColor in ES 3.0 + ESSL 1.00. Can browser work around it?

kenrussell:
What are the bugs you see? Is gl_FragColor not being broadcast as specified to all of the color attachments?

Is there a desktop OpenGL state which controls whether broadcasting occurs? I recall hearing this topic discussed recently, so I think it is probably still a current topic.

qkmiao:
ES SL 3.0 should broadcast while ES 3.0 + ESSL 1.00 should not broadcast. I tested Linux with nVidia and ChromeOS with Mesa. Both of them do broadcast for ES 3.0 + ESSL 1.00.

I will try to find the OpenGL state you mentioned.

sdk/tests/conformance2/rendering/draw-buffers.html
zhenyao: Only allowing gl_FragData[0] for ESSL 1.0 fragment shaders sounds good to me. Any attempt to access gl_FragData with a index that's not a constant 0 should fail the compile.

zhenyao: This is actually already specified: https://www.khronos.org/registry/webgl/specs/latest/2.0/#5.13

kenrussell: Aha. Thanks Mo for finding that. Good that this was caught earlier.

kenrussell:
Thanks @qkmiao for this fix and @zhenyao for the clarification. This LGTM and I'm merging it now.

Upon further thought, I think the browser could work around this simply by calling glDrawBuffers just before the draw call, if necessary (i.e., the shader is using ESSL 1.00, it writes gl_FragColor, there are multiple color attachments to the FBO, and the bug exists in the driver).

PR #1726, opened 2016-05-31
Issue a warning instead of fail in fbo completeness checking.

zhenyao:
The c++ code does the same.

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsFboCompletenessTests.js
+++ b/sdk/tests/deqp/modules/shared/glsFboCompletenessTests.js
@@ -737,7 +737,7 @@ goog.scope(function() {
                     testFailedOptions('Framebuffer checked as incomplete, but with wrong status', true);
                 }
             } else if (fboStatus != gl.FRAMEBUFFER_COMPLETE && statuses.isFBOStatusValid(gl.FRAMEBUFFER_COMPLETE)) {
-                testFailedOptions('Framebuffer object could have checked as complete but did not.', true);
+                testPassedOptions('Warning: framebuffer object could have checked as complete but did not.', true);
             } else {
                 // pass
                 testPassed();


sdk/tests/deqp/modules/shared/glsFboCompletenessTests.js
```diff
@@ -737,7 +737,7 @@ goog.scope(function() {
                     testFailedOptions('Framebuffer checked as incomplete, but with wrong status', true);
                 }
             } else if (fboStatus != gl.FRAMEBUFFER_COMPLETE && statuses.isFBOStatusValid(gl.FRAMEBUFFER_COMPLETE)) {
-                testFailedOptions('Framebuffer object could have checked as complete but did not.', true);
+                testPassedOptions('Warning: framebuffer object could have checked as complete but did not.', true);
```
zhenyao: See c++ code: https://code.google.com/p/chromium/codesearch#chromium/src/third_party/deqp/src/modules/glshared/glsFboCompletenessTests.cpp&l=663

zhenyao:
This patch together with https://codereview.chromium.org/2020473002/ and https://codereview.chromium.org/2024133003/, fbocompleteness.com passes in Chrome.

@kenrussell please take a look.

sdk/tests/deqp/modules/shared/glsFboCompletenessTests.js
kenrussell: Looks good, but would you consider using checkMessage (sdk/tests/js/js-test-pre.js) so that a warning is more visibly issued? At your discretion of course. Not sure about which primitives (checkMessage, testPassed, etc.) are best for this purpose.

kenrussell:
Thanks for tracking down this behavioral difference. One comment. Please merge yourself after you've given it more thought.

sdk/tests/deqp/modules/shared/glsFboCompletenessTests.js
zhenyao: I believe the deqp tests actually overdo by issuing a warning.  Here is the situation.  The test is for framebuffer with multiple attachments, each with different "samples". When the framebuffer is tested to be incomplete, the tests issue a warning, saying that the fbo could be complete if the underlying driver actually upgrade all the "samples" to the same "RENDERBUFFER_SAMPLES" (which is allowed by the spec). This is very unlikely and I think WebGL2 spec should just return INCOMPLETE to avoid different behaviors among driver implementations and the tests should just verify INCOMPLETE is the status, but that would need to be agreed by other WebGL2 implementors.

kenrussell: Thanks for this clarification. Given this and how many other issues need to be resolved, let's leave this as a passing test as you suggest.

PR #1727, opened 2016-06-01
Attach buffer to VAO before draw call

qkmiao:
If no buffer is bound to enabled attribution, draw call will genetate
invalid operation error. Refer section 6.5 Enabled Vertex Attributes and
Range Checking in WebGL 1.0 spec.

See some discussion in https://codereview.chromium.org/2019513004/

### DIFF
--- a/sdk/tests/conformance/misc/webgl-specific.html
+++ b/sdk/tests/conformance/misc/webgl-specific.html
@@ -49,6 +49,7 @@ gl.useProgram(program);
 var vertexObject = gl.createBuffer();
 gl.bindBuffer(gl.ARRAY_BUFFER, vertexObject);
 gl.enableVertexAttribArray(0);
+gl.vertexAttribPointer(0, 4, gl.FLOAT, false, 0, 0);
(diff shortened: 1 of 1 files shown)


kenrussell:
Thanks for this patch, Qiankun, but per comments on https://codereview.chromium.org/2019513004/ could you at least please also expand conformance2/vertex_arrays/vertex-array-object.html to explicitly test this behavior? Otherwise there is zero test coverage for this part of the spec. Thanks.

( @zhenyao any opinions on where the new test should go?)

qkmiao:
I think the test your suggested is already in conformance2/vertex_arrays/vertex-array-object.html. See test for "Draw call should fail with unbound position and color buffers".  https://codereview.chromium.org/2019513004/  is just for fixing it.

kenrussell:
Ah, I see. Thanks. Sorry for my confusion.

@zhenyao , if this looks acceptable to you, I'll leave it to you to merge.

zhenyao:
lgtm
We can add another test case for the removed behavior after working group makes a decision on it.

PR #1728, opened 2016-06-01
fix bugs against builtin precision for matrix operation

Richard-Yunchao:
This bug is a little deep, which takes me a lot of time these days. Now, all matrix related failures in builtin precision can pass. They are matirxcompmult.html, transpose.html, determinant.html and inverse.html. 

PTAL. Thanks!

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsShaderExecUtil.js
+++ b/sdk/tests/deqp/modules/shared/glsShaderExecUtil.js
@@ -446,7 +446,11 @@ goog.scope(function() {
             else if (gluShaderUtil.isDataTypeMatrix(basicType)) {
                 /** @type {number} */ var numRows = gluShaderUtil.getDataTypeMatrixNumRows(basicType);
                 /** @type {number} */ var numCols = gluShaderUtil.getDataTypeMatrixNumColumns(basicType);
-                /** @type {number} */ var stride = numRows * numCols * 4;//sizeof(float);
+                // A matrix consists of several (column-major) vectors. A buffer is created for
+                // every vector in gluDrawUtil.draw() below. Data in every buffer will be tightly
+                // packed. So the stride should be 0. This is different from the code in native
+                // deqp, which use only one buffer for a matrix, the data is interleaved.
+                /** @type {number} */ var stride = 0;
 
                 for (var colNdx = 0; colNdx < numCols; ++colNdx)
                     vertexArrays.push(gluDrawUtil.newFloatColumnVertexArrayBinding(symbol.name,
@@ -628,7 +632,11 @@ goog.scope(function() {
             else if (gluShaderUtil.isDataTypeMatrix(basicType)) {
                 var numRows = gluShaderUtil.getDataTypeMatrixNumRows(basicType);
                 var numCols = gluShaderUtil.getDataTypeMatrixNumColumns(basicType);
-                var stride = numRows * numCols * 4;
+                // A matrix consists of several (column-major) vectors. A buffer is created for
+                // every vector in gluDrawUtil.draw() below. Data in every buffer will be tightly
+                // packed. So the stride should be 0. This is different from the code in native
+                // deqp, which use only one buffer for a matrix, the data is interleaved.
+                var stride = 0;
 
                 for (var colNdx = 0; colNdx < numCols; ++colNdx)
                     vertexArrays.push(gluDrawUtil.newFloatColumnVertexArrayBinding(attribName,


sdk/tests/deqp/modules/shared/glsShaderExecUtil.js
```diff
@@ -446,7 +446,11 @@ goog.scope(function() {
             else if (gluShaderUtil.isDataTypeMatrix(basicType)) {
                 /** @type {number} */ var numRows = gluShaderUtil.getDataTypeMatrixNumRows(basicType);
                 /** @type {number} */ var numCols = gluShaderUtil.getDataTypeMatrixNumColumns(basicType);
-                /** @type {number} */ var stride = numRows * numCols * 4;//sizeof(float);
+                // A matrix consists of several (column-major) vectors. A buffer is created for
+                // every vector in gluDrawUtil.draw() below. Data in every buffer will be tightly
+                // packed. So the stride should be 0. This is different from the code in native
+                // deqp, which use only one buffer for a matrix, the data is interleaved.
+                /** @type {number} */ var stride = 0;
```
Richard-Yunchao: The WebGL deqp use multiple buffers to upload data to a matrix, see the line 118 ~ line 121 in gluDrawUtil.draw() in https://www.khronos.org/registry/webgl/sdk/tests/deqp/framework/opengl/gluDrawUtil.js. It uses a loop to create multiple buffers for one matrix. So data in every buffer is tightly packed. The stride should be 0. 

The code snippet in native deqp use only one buffer for a matrix, see the code https://android.googlesource.com/platform/external/deqp/+/deqp-dev/framework/opengl/gluDrawUtil.cpp#542. So the data in this buffer is interleaved, we should set appropriate stride and offset.
The call stack is:
gluDrawUtil::draw -> gluDrawUtil::drawFromVAOBuffers -> gluDrawUtil::drawFromBuffers

Richard-Yunchao: Seems that the code has been updated a lot in glsShaderExecUtil in the native deqp. Geometry shader and compute shader are added into this file...

zhenyao:
great work tracking this done.
lgtm

sdk/tests/deqp/modules/shared/glsShaderExecUtil.js
kenrussell: Excellent work @Richard-Yunchao tracking down this discrepancy between the native code and the JavaScript port. I don't know how difficult it would be, but do you think it would be worth re-porting the code from scratch from the C++ code? If so, do you have an estimate for how much work it would be? We could potentially ask Mobica to revisit it.

kenrussell:
Excellent work @Richard-Yunchao . Thank you for patiently tracking down this unfortunate discrepancy between the C++ code and the JavaScript port.

sdk/tests/deqp/modules/shared/glsShaderExecUtil.js
Richard-Yunchao: @kenrussell , I have no idea about the exact workload of porting the latest C++ deqp code to JS. But when we debug the errors introduced by porting, I think it is really a big effort. You know, there are tremendous differences between C++ and JS, and some subtle differences between GLES and webgl, and so on. All of these will lead to lots of redesign and/or workarounds. I can not say they are very difficult, but it is definitely time-consuming. 

@gyagp want to try to use some tool (emscripten or something like emscripten) to auto-generate the JS code from C++, then manually rewrite the auto-porting code and refine the patterns of the tool. But he has no time currently. You know, in this way, we can reuse the tool and make the tool better and better every time we port the code from C++ to JS when WebGL is upgraded. However, I am not sure whether this method can save some time. WDYT?

kenrussell: Thanks for your feedback @Richard-Yunchao regarding how much work it would be to re-port this code. I think we will stick with its current structure for now. It was a great effort to port it over in the first place and I think it is most expedient to diagnose and fix the bugs in the current code.

I don't think it is feasible to write a tool to autogenerate the JavaScript from C++. The working group rejected the idea of using Emscripten to get the native dEQP tests running; the code is unreadable, and doesn't integrate well with surrounding JavaScript code. A custom tool to port code would be a lot of work and it would be unclear whether it would actually function as expected. This is why we contracted Mobica to do the dEQP port by hand in the first place.

Richard-Yunchao: Got it, @kenrussell. I think you guys have had an assessment about the workload of porting manually vs using a tool and refine the tool as well as the generated code. I also doubt the feasibility of using such a tool when we discussed these two solutions among Intel guys. :)

PR #1729, opened 2016-06-01
At least one enabled attribute with 0 divisor for instanced drawing

qkmiao:
This is a WebGL 2.0 limitation different from C++ code. See section 5.6
Enabled Attribute in WebGL 2.0 spec.


zhenyao:
lgtm

kenrussell:
Good work tracking this down @qkmiao .

PR #1730, opened 2016-06-01
Some anti-alias algorithms may not effect on diagonal line pixels

xhcao:

### DIFF
--- a/sdk/tests/conformance/context/context-attributes-alpha-depth-stencil-antialias.html
+++ b/sdk/tests/conformance/context/context-attributes-alpha-depth-stencil-antialias.html
@@ -62,7 +62,8 @@ void main()
 var wtu = WebGLTestUtils;
 var gl;
 var contextAttribs = null;
-var pixel = [0, 0, 0, 1];
+var pixel_1 = [0, 0, 0, 1];
+var pixel_2 = [0, 0, 0, 1];
 var correctColor = null;
 var framebuffer;
 var fbHasColor;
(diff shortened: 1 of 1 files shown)


zhenyao:
LGTM

sdk/tests/conformance/context/context-attributes-alpha-depth-stencil-antialias.html
```diff
@@ -316,10 +317,17 @@
...
-    var buf = new Uint8Array(1 * 1 * 4);
-    gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, buf);
-    pixel[0] = buf[0];
-    shouldBe("pixel[0] != 255 && pixel[0] != 0", "contextAttribs.antialias");
+    var buf_1 = new Uint8Array(1 * 1 * 4);
+    var buf_2 = new Uint8Array(1 * 1 * 4);
+    gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, buf_1);
+    gl.readPixels(0, 1, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, buf_2);
+    pixel_1[0] = buf_1[0];
+    pixel_2[0] = buf_2[0];
```
kenrussell: Looking more closely, the copies into these "pixel_1" and "pixel_2" arrays are unnecessary. If you'd like to clean these up in a subsequent commit that would be great.

kenrussell:
Thanks. Looks good after the fact.

sdk/tests/conformance/context/context-attributes-alpha-depth-stencil-antialias.html
zhenyao: Right.  Thanks for catching that.  It is an issue in the original code, but now we are touching it, we should clean it up.

xhcao: @kenrussell @zhenyao "shouldBe" function need to a global variable for its first parameter. If directly transfer "buf_1" to it, will report "buf_1 is not defined".

zhenyao: Thanks.  Then it is fine as is.

kenrussell: Ah darn. Thanks for pointing that out. Yes, it's fine as is. We should really change these primitives to avoid using eval, but that's future work.

PR #1733, opened 2016-06-01
Fix references to instancing and drawing commands.

kenrussell:
These are in section 2.9.3, not 2.8.3. Perhaps they moved between dot releases of the ES 3.0 spec.


kenrussell:
Noticed this bug while reviewing #1729 . Trivial fix, merging myself. FYI @jdashg @grorg @RafaelCintron @zhenyao .

PR #1734, opened 2016-06-01
Fix references to m_allowComplete in FBO completeness tests.

kenrussell:


kenrussell:
PTAL @zhenyao @Kangz .

The affected tests still fail on Mac NVIDIA, but with the correct error message now.

zhenyao:
lgtm
Thanks for fixing this!

Kangz:
LGTM too

PR #1736, opened 2016-06-03
Polish test result information

qkmiao:


qkmiao:
@gyagp I changed the result information 
from: (16 of 44 skipped in 10000 ms)
to: (28 of 44 passed, 16 skipped in 10000 ms).
Is it what you want?

@kenrussell @zhenyao  PTAL.

zhenyao:
lgtm
but let @kenrussell take a look too

kenrussell:
Thanks, LGTM.

In the long run, all of the non-skipped tests will have to pass, but this summary information does seem useful.

PR #1737, opened 2016-06-03
fix the bug for builtin_precision.clamp

Richard-Yunchao:
Built-in function clamp(x, min, max) will clamp x to the range [min, max]. However, if min > max, the result is undefined, we should not check the result's precision in this situation. This small change fix a bug to detect this edge case, which finally make clamp.html pass. 
It also follows the design of native deqp.
PTAL. Thanks!


zhenyao:
looks good.

PR #1738, opened 2016-06-03
Fix conformance2/textures/misc/copy-texture-image.html

zhenyao:
for sized color internalformats, component sizes have to match the src format.


zhenyao:
@kenrussell PTAL

sdk/tests/conformance2/textures/misc/copy-texture-image.html
```diff
@@ -168,9 +174,14 @@
     }
 
     goodByteFormats.forEach(function(internalformat) {
-        var srcTexFormatsTypes = { internalformat: gl.RGBA, format: gl.RGBA, type: gl.UNSIGNED_BYTE };
+        var srcTexFormatsTypes = { internalformat: gl.RGBA8, format: gl.RGBA, type: gl.UNSIGNED_BYTE };
```
kenrussell: Is it worth continuing to test copying from a texture with unsized internal format to a texture with a sized internal format? It's a new situation in the WebGL 2.0 tests.

kenrussell:
LGTM with comment. If you think the changes are ready to go please merge yourself. Thanks.

sdk/tests/conformance2/textures/misc/copy-texture-image.html
zhenyao: Sounds good.  I'll add back the original test cases.

zhenyao:
Added the unsigned src format cases back. Merging myself.

PR #1739, opened 2016-06-03
FboMultisample Fix

anzurio:
- Fixing the masks which extracts the different channels on bilinearCompare.
- Using a different random seed because the bilinearCompare (even in C++'s deqp) does not handle a few boundary cases correctly.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fFboMultisampleTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fFboMultisampleTests.js
@@ -165,6 +165,10 @@ var DE_ASSERT = function(x) {
 
         // Render random-colored quads.
         /** @const {number} */ var numQuads = 8;
+
+        // The choice of random seed affects the correctness of the tests,
+        // because there are some boundary conditions which aren't handled
+        // correctly even in the C++ dEQP tests.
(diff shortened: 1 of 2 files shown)


sdk/tests/deqp/functional/gles3/es3fFboMultisampleTests.js
```diff
@@ -165,7 +165,7 @@ var DE_ASSERT = function(x) {
 
         // Render random-colored quads.
         /** @const {number} */ var numQuads = 8;
-        /** @type {deRandom.Random} */ var rnd = new deRandom.Random(9);
+        /** @type {deRandom.Random} */ var rnd = new deRandom.Random(7);
```
kenrussell: Could you please add a comment here stating what you've mentioned in the commit message? Something like:

// The choice of random seed affects the correctness of the tests,
// because there are some boundary conditions which aren't handled
// correctly even in the C++ dEQP tests.

Thanks.

kenrussell:
Looks good to me. Thank you for tracking this down and fixing it. One comment. Let's let @zhenyao review and merge this.

zhenyao:
Please address Ken's comment. Then we'll merge the patch.  Thanks for getting this fixed.

zhenyao:
lgtm

PR #1741, opened 2016-06-06
Some anti-alias algorithms may not effect on diagonal line pixels,

xhcao:
add this change to suite1.0.3

Hi, zhenyao  & ken,
Because we test anti-alias feature based on suite1.0.3, I am not sure that it is suitable to add the anti-alias change to suite1.0.3. If not, please ignore this patch. Thank you.

### DIFF
--- a/conformance-suites/1.0.0/conformance/context-attributes-alpha-depth-stencil-antialias.html
+++ b/conformance-suites/1.0.0/conformance/context-attributes-alpha-depth-stencil-antialias.html
@@ -39,6 +39,8 @@ var successfullyParsed = false;
 var webGL = null;
 var contextAttribs = null;
 var pixel = [0, 0, 0, 1];
+var pixel_1 = [0, 0, 0, 1];
+var pixel_2 = [0, 0, 0, 1];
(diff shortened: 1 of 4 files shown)


zhenyao:
I think it's OK. It relaxes the passing condition a little, so no way to regress on any browsers.  I think it should also be ported back to 1.0.2 and 1.0.1 if it applies.

Let's hear what @kenrussell thinks.

kenrussell:
Thanks for checking. It's fine to backport the change to this test to 1.0.3 and the earlier conformance tests, since this is relaxing a too-strict test. Would you also backport it to 1.0.2 and 1.0.1? Also, would you reference the original pull request, #1730 , in the commit message? Thanks. You can do this with "git commit --amend" and "git push origin [branchname] --force" per https://www.khronos.org/webgl/wiki/Using_Github_To_Contribute .

xhcao:
I had ported to 1.0.3, 1.0.2, 1.0.1 and 1.0.0.  Please review it again, thank you.

zhenyao:
lgtm
Thank you.

PR #1742, opened 2016-06-06
update the precision for trigomometric functions

Richard-Yunchao:
The precision calculated by JS against trigomometric functions is not accurate, making some cases failed. This change updates the precision for trigomometric functions according to the latest native deqp: https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/glshared/glsBuiltinPrecisionTests.cpp#2290. It can fix a number of failures in webgl deqp files, say, sin.html, cos.html, tan.html, and so on. 

PTAL. Thanks!


zhenyao:
lgtm

PR #1746, opened 2016-06-06
Fix fboDepthBuffer

anzurio:
- All clamp tests were failing because clamping was
not actually being performed.


zhenyao:
This looks good and align with c++ version.

PR #1748, opened 2016-06-07
Add out-of-bounds-index-buffers.html conformance test

RafaelCintron:
Out-of-bounds-index-buffers.html tests whether implementations successfully handle
the "Out-of-Range Array Accesses" section of the WebGL 1 specification.

The test accomplishes this by having a vertex shader with two attributes: a position
and value. The position attribute will come from large position vertex buffer
filled with tri-strips that form quads which cover the entire canvas. This vertex
buffer will not be indexed off the end. The value attribute will come from a tiny
vertex buffer that is filled with a known-ahead-of-time "random" value. This buffer
will be indexed off the end. The shader tests whether the random value is either
found in the vertex buffer or is the zero index. Upon success, the shader outputs
green. On failure, the shader outputs red.

The test passes if implementations either:
1) Flag INVALID_OPERATION and do not output anything to the canvas OR
2) Flag NO_ERROR and output green for all canvas pixels.

Future improvements can be made that test more data types than just FLOAT as well
datatypes with different numbers of components.

### DIFF
--- a/sdk/tests/conformance/misc/out-of-bounds-index-buffers.html
+++ b/sdk/tests/conformance/misc/out-of-bounds-index-buffers.html
@@ -1,0 +1,156 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+-->
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/misc/out-of-bounds-index-buffers.html
```diff
@@ -0,0 +1,154 @@
...
+var glProgram = wtu.setupProgram(gl, ["vsCheckOutOfBounds", wtu.simpleVertexColorFragmentShader], ["position", "vecRandom"]);
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "Shader and buffer setup should generate errors");
+
+debug("Test -- Index off the end of the vertex buffer near the beginning of the out of bounds area.");
+drawAndVerifyOutOfBoundsIndex(/*StartIndex*/4);
+
+debug("");
+
+debug("Test -- Index off the end of the vertex buffer near the end of the out of bounds area.")
+drawAndVerifyOutOfBoundsIndex(/*StartIndex*/numberOfQuads - 4);
```
kenrussell: I think it'd be better to have the body of the test after the definitions of the helper functions. Would you mind moving the helpers higher up in the file?

kenrussell:
Excellent Rafael. Thank you for putting this test together. A couple of minor comments.

What shall we do with conformance/rendering/draw-arrays-out-of-bounds.html and conformance/rendering/draw-elements-out-of-bounds.html ?

sdk/tests/conformance/misc/out-of-bounds-index-buffers.html
```diff
@@ -0,0 +1,154 @@
...
+<canvas id="canvas" width="8" height="8" style="width: 100px; height: 100px;"></canvas>
+<script id="vsCheckOutOfBounds" type="x-shader/x-vertex">
+    attribute vec2 position;
+    attribute vec4 vecRandom;
+    varying vec4 v_color;
+
+    // Per the spec, each component can either contain existing contents
+    // of the buffer or 0.
+    bool testFloatComponent(float component) {
+        return (component == 0.2 || component == 0.0);
```
RafaelCintron: I think so.  We're not doing any math operations in the shader. We're just reading attributes and comparing against the exact value we set in the vertex buffer.

sdk/tests/conformance/misc/out-of-bounds-index-buffers.html
RafaelCintron: Done!

RafaelCintron:
@kenrussell , My thinking on draw-arrays-out-of-bounds.html and draw-elements-out-of-bounds.html was to change them to also accept NO_ERROR in addition to INVALID_OPERATION.

kenrussell:
@RafaelCintron sounds okay regarding draw-arrays-out-of-bounds.html and draw-elements-out-of-bounds.html ; any pull requests you can offer for those would be appreciated. I guess this test will be the one that explicitly verifies the values for out-of-range vertex fetches.

Sorry for the delay re-reviewing; merging now.

RafaelCintron:
No worries about the delay, @kenrussell .  Thank you very much for the review.

PR #1754, opened 2016-06-07
Split fbocolorbuffer.html into multiple tests.

zhenyao:


zhenyao:
Tests are still running around 30 sec, but I feel we should optimize the tests to bring down the running time instead of further splitting them.

kenrussell:
LGTM after the fact.

PR #1755, opened 2016-06-08
WebGL 2 spec: make RGB5_A1 not color-renderable

Kangz:
The Mac OpenGL driver makes framebuffers with GL_RGB5_A1 attachments
incomplete. While the WebGL implementations could replace all instances
of GL_RGB5_A1 with GL_RGBA8, this would break the strict precision
guarantees of WebGL.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -1087,6 +1087,7 @@ WebGL2RenderingContext implements WebGL2RenderingContextBase;
         <p>Only differences from <a href="../1.0/index.html#CHECK_FRAMEBUFFER_STATUS">checkFramebufferStatus</a> in WebGL 1 are described here.</p>
         <p><em>target</em> must be <code>DRAW_FRAMEBUFFER</code>, <code>READ_FRAMEBUFFER</code> or <code>FRAMEBUFFER</code>. <code>FRAMEBUFFER</code> is equivalent to <code>DRAW_FRAMEBUFFER</code>.</p>
         <p>Returns <code>FRAMEBUFFER_UNSUPPORTED</code> if depth and stencil attachments, if present, are not the same image. See <a href="#FBO_ATTACHMENTS">Framebuffer Object Attachments</a> for detailed discussion.</p>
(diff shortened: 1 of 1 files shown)


kenrussell:
@grorg @jdashg in particular, please review. Also @RafaelCintron @zhenyao @Oletus , please provide opinions.

This particular internal format is problematic from a portability standpoint. We would like to carve it off to retain WebGL's precise semantics for the other formats.

zhenyao:
To be precise, it's Mac NVidia only.  @grorg Is it possible to find out why? If it's a bug Apple is willing to fix and include in the next update, then we don't have to exclude it.

Kangz:
From the F2F discussion, it seems that it would be acceptable to silently promote this format to RGBA8888 (ANGLE does that on Win7 already). This would remove the need for this change to the spec. What do you think?

dmikis:
Won't it produce unexpected results with alpha? If I understand correctly, RGB5_A1 can be fully opaque for 1.0 output and fully transparent otherwise, whereas RGBA8 can be semitransparent for alpha values between 0 and 1.

Kangz:
It depends, WebGL strictly enforces precision on data upload (TexImage2D) so alpha 0.5 would become 1. However the format is just a guideline on the minimum number of bits, so rendering alpha 0.5 to a RGB5_A1 texture backed with an RGBA8 texture would sample as 0.5, which is correct.

kenrussell:
@Kangz sorry for the delay, was OOO. It sounds acceptable to do the upgrade from RGB5_A1 to RGBA8 behind the scenes; I'm not sure if many people will find that they expected banding in their rendering but instead received nice smooth gradients.

daoshengmu:
@Kangz , it happens to me in Firefox as well. Could you tell me does Chrome decide to convert GL_RGB5_A1 framebuffer to GL_RGBA8 internally?

Kangz:
I'm not sure what we decided to do on OSX. @zhenyao do you know if we upgrade RGB5A1 to RGBA8?

zhenyao:
We didn't.  NVidia drivers on OSX is not in a good shape, and this is only one of the conformance issues.  We simply just focus on AMD and Intel.

PR #1756, opened 2016-06-08
update the JS code for Div and Modf

Richard-Yunchao:
Align the JS code with native C++ code for Div() and Modf(). There is no improvement and no regression for test cases from the test result. 

PTAL. Thanks a lot!

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
+++ b/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
@@ -2322,7 +2322,17 @@ var setParentClass = function(child, parent) {
     };
 
     glsBuiltinPrecisionTests.Div.prototype.applyPoint = function(ctx, x, y) {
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
```diff
@@ -4580,6 +4590,12 @@ var setParentClass = function(child, parent) {
         var fracIV = tcuInterval.applyMonotone1p(func1, iargs.a);
         var wholeIV = tcuInterval.applyMonotone1p(func2, iargs.a);
 
+        if (!iargs.a.isFinite()) {
```
Richard-Yunchao: See the native code at https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/glshared/glsBuiltinPrecisionTests.cpp#3300

zhenyao:
lgtm

PR #1759, opened 2016-06-09
Added example querying number of timestamp bits on suggestion from Fl…

kenrussell:
…orian Boesch.

### DIFF
--- a/extensions/EXT_disjoint_timer_query/extension.xml
+++ b/extensions/EXT_disjoint_timer_query/extension.xml
@@ -212,7 +212,7 @@
 
         ext.endQueryEXT(ext.TIME_ELAPSED_EXT);
 
-        // At some point in the future, after returning control to the browser and being called again:
+        // ...at some point in the future, after returning control to the browser and being called again:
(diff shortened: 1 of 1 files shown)


kenrussell:

extensions/EXT_disjoint_timer_query/extension.xml
```diff
@@ -254,6 +253,69 @@
...
+        var timeElapsedQuery;
+        var startQuery;
+        var endQuery;
+
+        var useTimestamps = false;
+
+        if (ext.getQueryEXT(ext.TIMESTAMP_EXT, ext.QUERY_COUNTER_BITS_EXT) > 0) {
+          useTimestamps = true;
+        }
+
```
jdashg: Clear gl.getParameter(ext.GPU_DISJOINT_EXT) right before starting queries.
It's technically cleared in the previous example, but it's good for this example to be self-contained.

kenrussell: Thanks. Making this change now.

kenrussell:
Merging now.

PR #1764, opened 2016-06-10
Added WEBGL_compressed_texture_s3tc_srgb proposal

Groovounet:
Hi,

I am currently working on [linear rendering](http://docs.unity3d.com/Manual/LinearLighting.html) for mobile which led me to look at WebGL to figure out what's missing to enable linear rendering on this platform as well and the much higher rendering quality that correct mathematics provides.

WebGL 2.0 supports sRGB framebuffers and sRGB textures which is required for linear rendering. 

However, the supported texture formats are either uncompressed sRGB which consumes a lot of bandwidth and space or ETC2 sRGB textures which are not actually supported by the hardware on desktop and expected to be uncompressed.

With our WebGL support with rely on DXT to benefit hardware compressed texture hence for linear rendering we need sRGB DXT.

This proposal is an extension to expose sRGB DXT compressed formats in WebGL.

Additionally, on desktop with WebGL 1.0, only Chrome supports [EXT_sRGB](https://www.khronos.org/registry/webgl/extensions/EXT_sRGB/) while in practice all desktop machines support EXT_sRGB and WEBGL_compressed_texture_s3tc_srgb in hardware which seems like an opportunity for visibly much higher rendering quality for the WebGL ecosystem.

Thanks,
Christophe

### DIFF
--- a/extensions/proposals/WEBGL_compressed_texture_s3tc_srgb/extension.xml
+++ b/extensions/proposals/WEBGL_compressed_texture_s3tc_srgb/extension.xml
@@ -1,0 +1,107 @@
+<?xml version="1.0"?>
+
+<proposal href="proposals/WEBGL_compressed_texture_s3tc_srgb/">
+  <name>WEBGL_compressed_texture_s3tc_srgb</name>
+  <contact>
+    <a href="https://www.khronos.org/webgl/public-mailing-list/">WebGL working group</a> (public_webgl 'at' khronos.org)
+  </contact>
+  <contributors>
+    <contributor>Members of the WebGL working group</contributor>
+  </contributors>
+  <number>1</number>
+  <depends>
+    <api version="1.0"/>
+  </depends>
+  <overview>
+    <p>
+      This extension exposes the sRGB compressed texture formats defined in the 
+      <a href="https://www.opengl.org/registry/specs/EXT/texture_sRGB.txt">
+      EXT_texture_sRGB</a> OpenGL extension to WebGL.
+    </p>
+    <features>
+      <feature>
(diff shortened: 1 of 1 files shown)


extensions/proposals/WEBGL_compressed_texture_s3tc_srgb/extension.xml
```diff
@@ -0,0 +1,107 @@
...
+  <name>WEBGL_compressed_texture_s3tc_srgb</name>
+  <contact>
+    <a href="https://www.khronos.org/webgl/public-mailing-list/">WebGL working group</a> (public_webgl 'at' khronos.org)
+  </contact>
+  <contributors>
+    <contributor>Members of the WebGL working group</contributor>
+  </contributors>
+  <number>1</number>
+  <depends>
+    <api version="1.0"/>
```
kenrussell: For WebGL 1.0, should this extension require the WebGL EXT_sRGB extension to be enabled? Or should it be orthogonal, and be available if the underlying EXT_texture_compression_s3tc and EXT_texture_sRGB extensions are present? It will probably be simpler if it can be enabled independently.

kenrussell:
Thanks for putting this together Christophe. It looks good; let's merge for further discussion.

extensions/proposals/WEBGL_compressed_texture_s3tc_srgb/extension.xml
Groovounet: I think it's fine if it's enabled independently because there isn't direct interactions.

Of course, for linear rendering we would enable EXT_sRGB, EXT_texture_compression_s3tc and WEBGL_compressed_texture_s3tc_srgb

PR #1765, opened 2016-06-12
Fix sRGB formatted output comparison

qkmiao:
When we do gl.readPixels from a renderbuffer with SRGB8_ALPHA8 internal
format, due readPixels doesn't support SRGB8_ALPHA8 format reading, RGBA
read format is used. So, the result of readPixels is in RGBA format.
Reference pixels and result pixels should be in same format before doing
comparison. This patch converts reference pixels to RGBA format and
keeps threshold unchanged. Otherwise, if we change RGBA pixels to sRGB,
we should also change the threshold.

This is different behavior with C++ dEQP. In C++ dEQP, glReadPixels can
return sRGB results depends on if FRAMEBUFFER_SRGB was enabled or not.
I refered https://github.com/KhronosGroup/WebGL/pull/1558 to implement
patch.

### DIFF
--- a/sdk/tests/deqp/framework/common/tcuImageCompare.js
+++ b/sdk/tests/deqp/framework/common/tcuImageCompare.js
@@ -340,6 +340,7 @@ tcuImageCompare.floatUlpThresholdCompare = function(imageSetName, imageSetDesc, 
  * @return {boolean}
  */
 tcuImageCompare.floatThresholdCompare = function(imageSetName, imageSetDesc, reference, result, threshold) {
+    /** @type {boolean} */ var isSRGB = reference.getFormat().order == tcuTexture.ChannelOrder.sRGB || reference.getFormat().order == tcuTexture.ChannelOrder.sRGBA;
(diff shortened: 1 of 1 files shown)


zhenyao:
We should always enable FRAMEBUFFER_SRGB in Chrome's ES3.  I think the test is actually correct, and no need to fix it.  I landed a CL to enable FRAMEBUFFER_SRGB a while back but got reverted due to driver bugs.  We should reland it soon.

qkmiao:
I agree with you. I think the after enabling FRAMEBUFFER_SRGB several tests related to sRGB will pass.

qkmiao:
This bug should be fixed in Chrome. Close this pr now.

msc-:
Yes this change was a mistake. glReadPixels does not do sRGB to linear conversions when reading from a framebuffer whose FRAMEBUFFER_ATTACHMENT_COLOR_ENCODING is SRGB. You'll get the sRGB pixels values even though the format is RGBA.

PR #1766, opened 2016-06-12
Backport: Attach buffer to VAO before draw call

qkmiao:
Original pr is at: https://github.com/KhronosGroup/WebGL/pull/1727.


qkmiao:
Should we also short-circuit the validation check for Draws with 0 elements in chrome as what Firefox and Edge do? And Shoud we add such validation tests to WebGL 1.0 conformance tests?

zhenyao:
Yes we should fix that in Chrome, and we should have a test for WebGL 1. (This is resolved in the latest working group f2f meeting that we should honor the current spec.)

PR #1767, opened 2016-06-12
Fix ext-SRGB test

qkmiao:
The tests intend to testing color conversion when writing to frame
buffer backed with sRGB texture attachment. But gl.readPixels doesn't
support sRGB reading. sRGB color is converted to RGAB after readPixels.
So we can only test after tow rounds conversion of RGBA->sRGB->RGBA, the
color keeps unchanged. See
https://github.com/KhronosGroup/WebGL/pull/1558 and
https://github.com/KhronosGroup/WebGL/pull/1765 which have similar issue.

### DIFF
--- a/sdk/tests/conformance/extensions/ext-sRGB.html
+++ b/sdk/tests/conformance/extensions/ext-sRGB.html
@@ -336,9 +336,9 @@ function runFramebufferTextureConversionTest(format) {
     // Draw
     var conversions = [
       [   0,   0 ],
-      [  13,  63 ],
-      [  54, 127 ],
-      [ 133, 191 ],
+      [  13,  13 ],
+      [  54,  54 ],
+      [ 133, 133 ],
       [ 255, 255 ]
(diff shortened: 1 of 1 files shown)


qkmiao:
FYI @xhcao @gyagp

qkmiao:
I tested this test with --use-gl=angle and --use-gl=egl which had different behavior with --use-gl=desktop. For angle and egl context, readPixels can return sRGB format color even passing RGBA format as argument. And we don't need this patch for angle and egl context. @kenrussell @zhenyao , can you clarify which is correct behavior? Thanks!

zhenyao:
Same here, we should enable FRAMEBUFFER_SRGB in chrome to make the test passing, not changing the test.  Otherwise, when we enable FRAMEBUFFER_SRGB, the test actually will fail with this fix.

qkmiao:
The bug should be fixed in chrome. Close this pr.

PR #1768, opened 2016-06-12
Fix deqp builtin precision bugs in radians.html

Richard-Yunchao:
Div operation will lead to precision loss, we should not use
a constant value to store the result of constant division for
derived functions, for example: PI / 180. We should use Div
function in derived functions, which will bring in the precision
calculation.

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
+++ b/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
@@ -4297,7 +4297,9 @@ var setParentClass = function(child, parent) {
     };
 
     glsBuiltinPrecisionTests.Radians.prototype.doExpand = function(ctx, args) {
-        var val = new glsBuiltinPrecisionTests.Constant(Math.PI / 180);
+        var val = app(new glsBuiltinPrecisionTests.Div(),
+                                                      new glsBuiltinPrecisionTests.Constant(Math.PI),
+                                                      new glsBuiltinPrecisionTests.Constant(180));
         return new glsBuiltinPrecisionTests.Apply('float',
                                                   new glsBuiltinPrecisionTests.Mul(),
                                                   val,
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
```diff
@@ -4297,7 +4297,9 @@ var setParentClass = function(child, parent) {
     };
 
     glsBuiltinPrecisionTests.Radians.prototype.doExpand = function(ctx, args) {
-        var val = new glsBuiltinPrecisionTests.Constant(Math.PI / 180);
+        var val = app(new glsBuiltinPrecisionTests.Div(),
+                                                      new glsBuiltinPrecisionTests.Constant(Math.PI),
+                                                      new glsBuiltinPrecisionTests.Constant(180));
         return new glsBuiltinPrecisionTests.Apply('float',
                                                   new glsBuiltinPrecisionTests.Mul(),
                                                   val,
```
Richard-Yunchao: See the native code: https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/glshared/glsBuiltinPrecisionTests.cpp#2235. It will use Div() operation, not a constant value. The Div() operation will bring in precision calculation, see https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/glshared/glsBuiltinPrecisionTests.cpp#2047. 

If we use a constant value like the original JS implementation, the precision loss caused by the division of PI / 180.0 will not be considered. So the final precision range is very small, making the tests failed. However, the output from radians() in shader in WebGL is exactly the same with that in native. And the result is thought to be acceptable in native deqp. The only reason is the precision ranges are different.

sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
```diff
@@ -4321,7 +4323,9 @@ var setParentClass = function(child, parent) {
     };
 
     glsBuiltinPrecisionTests.Degrees.prototype.doExpand = function(ctx, args) {
-        var val = new glsBuiltinPrecisionTests.Constant(180 / Math.PI);
+        var val = app(new glsBuiltinPrecisionTests.Div(),
+                                                      new glsBuiltinPrecisionTests.Constant(180),
+                                                      new glsBuiltinPrecisionTests.Constant(Math.PI));
         return new glsBuiltinPrecisionTests.Apply('float',
                                                   new glsBuiltinPrecisionTests.Mul(),
                                                   val,
```
Richard-Yunchao: The same reason, although the test cases against degrees() can pass on many platforms. See the native code: https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/glshared/glsBuiltinPrecisionTests.cpp#2236

zhenyao:
lgtm
Thanks for tracking this down.

PR #1769, opened 2016-06-12
Fix sync.html

zhenyao:
WebGL behavior is different from ES3.


zhenyao:

qkmiao:
Sorry @zhenyao . As discussed in https://github.com/KhronosGroup/WebGL/pull/1506. This pr seems not a correct fix. I send another pr in https://github.com/KhronosGroup/WebGL/pull/1771.  We can verify sync status in next few frames until it succeeds or exceeds max retry times.

zhenyao:
Ah I forgot about that discussion.  Closing this.

PR #1770, opened 2016-06-13
Fix occlusion query test

qkmiao:
Query's result must not be made available in the current frame. So
occlusion query test does query result check in next frame, so does
pixles check. By default, color buffer should be cleared before a new
frame. So set preserveDrawingBuffer as true when creating WebGL context
to preserve color buffer in next frame. Please refer section 2.2 "The
Drawing Buffer" in WebGL 1.o spec and section 5.23 "Queries' results
must not be make available in the current frame" in WebGL 2.0 spec.


zhenyao:
Awesome!

kenrussell:
Great work! Thanks for tracking this down.

PR #1771, opened 2016-06-13
Verify sync status in next several frames

qkmiao:
It's legal for a sync object to become signaled within a given frame.
Reverify sync status within several frames after gl.finish. See some
discussions in https://github.com/KhronosGroup/WebGL/pull/1506.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fSyncTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fSyncTests.js
@@ -33,6 +33,7 @@ goog.scope(function() {
     var deString = framework.delibs.debase.deString;
 
     /** @const {number} */ es3fSyncTests.NUM_CASE_ITERATIONS = 5;
+    /** @const {number} */ es3fSyncTests.MAX_VERIRY_WAIT = 5;
 
     /**
      * @enum
@@ -47,6 +48,14 @@ goog.scope(function() {
     es3fSyncTests.CaseOptions = {
         FLUSH_BEFORE_WAIT: 1,
         FINISH_BEFORE_WAIT: 2
+    };
+
+    /** @enum
+     */
+    es3fSyncTests.State = {
+        DRAW: 0,
+        VERIFY: 1,
+        FINISH: 2
     };
 
     /**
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fSyncTests.js
```diff
@@ -33,6 +33,7 @@ goog.scope(function() {
     var deString = framework.delibs.debase.deString;
 
     /** @const {number} */ es3fSyncTests.NUM_CASE_ITERATIONS = 5;
+    /** @const {number} */ es3fSyncTests.MAX_NUM_VERIRY = 10;
```
zhenyao: Will this introduce flakiness into the test?  because there is really no guarantee that it returns in 10 frames.

kenrussell: In es3fTransformFeedbackTests.js, the strategy used was to spin for a maximum of a certain number of seconds (5). Could that be done here too? I think it would make the test less potentially flaky; there's no reason it should take 5 seconds for a sync object to be signaled.

kenrussell:
Thanks for putting this together. I share @zhenyao 's concern about flakiness; could we try a slightly different approach?

sdk/tests/deqp/functional/gles3/es3fSyncTests.js
```diff
@@ -33,6 +33,7 @@ goog.scope(function() {
     var deString = framework.delibs.debase.deString;
 
     /** @const {number} */ es3fSyncTests.NUM_CASE_ITERATIONS = 5;
+    /** @const {number} */ es3fSyncTests.MAX_VERIRY_WAIT = 5;
```
qkmiao: I tested at my side the waiting time is less than 1 second.

qkmiao:
Thanks @zhenyao @kenrussell for your suggestion. I updated this pr. PTAL.

sdk/tests/deqp/functional/gles3/es3fSyncTests.js
```diff
@@ -170,11 +179,15 @@ goog.scope(function() {
             gl.flush();
         if (this.m_caseOptions & es3fSyncTests.CaseOptions.FINISH_BEFORE_WAIT)
             gl.finish();
+        this.m_state = es3fSyncTests.State.VERIFY;
+    };
 
-        // Wait for sync object
 
+    es3fSyncTests.FenceSyncCase.prototype.verify = function() {
         /** @type {number} */ var waitValue = 0;
```
zhenyao: Can we clean up the bufferedLogToConsole in this function a little bit?  There could be overwhelmingly large number of them in the console and in the webgl log.

zhenyao:
Mostly lgtm with one cleanup request.

sdk/tests/deqp/functional/gles3/es3fSyncTests.js
qkmiao: It seems not print to console unless the test fails. I didn't see these logs in console.

zhenyao:
OK, then let's merge.

PR #1772, opened 2016-06-13
More bug fixes in TextureShadowTests

Jiawei-Shao:
1. deMath.max and deMath.min should be used to compare arrays instead of numbers
2. Add missed tcuTexture.CubeFace.CUBEFACE_NEGATIVE_X in function es3fTextureShadowTests.clampFloatingPointTextureCube()

All cases in TextureShadowTests can pass with this patch.


zhenyao:
Great work!

zhenyao:
By the way, I tested textureshadow/ tests after merging your CL.  Many cube cases are still failing on my Mac (retina with AMD).  It might just be driver issues, but it would be great if you could take a look at that and figure out the root of these failures.

kenrussell:
Very nice @Jiawei-Shao . Thanks for tracking down these bugs.

Jiawei-Shao:
@zhenyao OK I get it

PR #1773, opened 2016-06-13
Draw with no buffer bound to VAO generates INVALID_OPERATION

qkmiao:


qkmiao:
This tests a chrome fix:  Validate bound buffer for draw calls https://codereview.chromium.org/2019513004/.

sdk/tests/conformance/extensions/oes-vertex-array-object.html
```diff
@@ -546,22 +547,102 @@
...
-        testFailed("buffer not properly cleaned up after VAO deletion");
-      }
+        // The buffers should no longer be valid now that the VAOs are deleted
+        if(gl.isBuffer(colorBuffers[ii])) {
+            testFailed("buffer not properly cleaned up after VAO deletion");
+        }
+    }
+}
+
+function runBoundDeleteTests() {
```
qkmiao: Only this runBoundDeleteTests is a new test which is backported from https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/conformance2/vertex_arrays/vertex-array-object.html#L514. Other part in this file is just formatting.

zhenyao:
lgtm

sdk/tests/conformance/rendering/gl-drawarrays.html
```diff
@@ -74,6 +74,10 @@
     gl.bindBuffer(gl.ARRAY_BUFFER, vertexObject);
     gl.bufferData(gl.ARRAY_BUFFER, new Float32Array([ 0,0.5,0, -0.5,-0.5,0, 0.5,-0.5,0 ]), gl.STATIC_DRAW);
     gl.enableVertexAttribArray(0);
+
+    checkDrawArrays(gl.TRIANGLES, 3,
+                    gl.INVALID_OPERATION, "gl.DrawArrays with no buffer attached to VAO shoud return INVALID_OPERATION");
```
kenrussell: typo: shoud -> should

kenrussell:
Very nice. Thanks for strengthening these tests. LGTM after the fact.

PR #1777, opened 2016-06-14
Fixed typo in name of constant in sync tests.

kenrussell:
This bug would cause the comparison "[number] > undefined" to be
repeatedly reevaluated, which would always return false. Thus the
timeout would never be applied.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fSyncTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fSyncTests.js
@@ -33,7 +33,7 @@ goog.scope(function() {
     var deString = framework.delibs.debase.deString;
 
     /** @const {number} */ es3fSyncTests.NUM_CASE_ITERATIONS = 5;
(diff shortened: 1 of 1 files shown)


kenrussell:
@zhenyao please review.

@qkmiao FYI. It'd be hard to catch a bug like this without running the Closure compiler against all of the sources. (I'm not sure that that works right now...will look.)

kenrussell:
Related to #1771 .

zhenyao:
Ah thanks for catching this. LGTM.

I am curious, since this constant isn't defined, how does the test pass (assuming now this.MAX_VERIFY_WAIT is 0?) ? I run it against all try bots and they all pass.

kenrussell:
The test passed because it was evaluating:

  if ([number] > undefined) {
    // test failed
  }

It turns out that this if-test always returns false, so the sync object would be tested over and over again until its result became available. In other words, the timeout wasn't actually being applied.

PR #1778, opened 2016-06-14
RGB16F isn't color renderable in WebGL2.

zhenyao:
We don't expose EXT_color_buffer_half_float and it is not part of the
EXT_color_buffer_float extension.

### DIFF
--- a/sdk/tests/conformance2/extensions/ext-color-buffer-float.html
+++ b/sdk/tests/conformance2/extensions/ext-color-buffer-float.html
@@ -133,7 +133,7 @@ function runFloatTextureRenderTargetTest(enabled, internalFormat, format, testPr
     var width = 2;
     var height = 2;
     gl.texImage2D(gl.TEXTURE_2D, 0, internalFormat, width, height, 0, format, gl.FLOAT, null);
-    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "floating-point texture allocation should succeed if EXT_color_buffer_float is enabled");
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "floating-point texture allocation should succeed");
 
     // Try to use this texture as a render target.
     var fbo = gl.createFramebuffer();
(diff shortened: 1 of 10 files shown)


zhenyao:
With this fix, deqp/functional/gles3/fbocolorbuffer/clear.html  still fails the case color.clear.r11f_g11f_b10f

It looks like the reference image isn't correct, but I will look into fixing that bug in a separate CL.

sdk/tests/deqp/functional/gles3/es3fFboTestCase.js
```diff
@@ -374,6 +373,9 @@ var DE_ASSERT = function(x) {
             case gl.R32F:
                 out.push('EXT_color_buffer_float');
                 break;
+            case gl.RGB16F:
+                // EXT_color_buffer_half_float is not exposed in ES3.
```
kenrussell: Perhaps better to say "WebGL 2.0" rather than "ES3" here.

sdk/tests/conformance2/extensions/ext-color-buffer-float.html
```diff
@@ -226,6 +226,41 @@
...
+    if (completeStatus == gl.FRAMEBUFFER_COMPLETE)
+        testFailed("RGB16F render target should not be supported with or without enabling EXT_color_buffer_float");
+    else
+        testPassed("RGB16F render target should not be supported with or without enabling EXT_color_buffer_float");
+    gl.deleteTexture(texture);
+
+    var colorbuffer = gl.createRenderbuffer();
+    gl.bindRenderbuffer(gl.RENDERBUFFER, colorbuffer);
+    gl.renderbufferStorage(gl.RENDERBUFFER, gl.RGB16F, width, height);
+    wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "RGB16F renderbuffer allocation should fail with or without enableing EXT_color_buffer_float");
```
kenrussell: Minor typo: enableing -> enabling

kenrussell:
Thanks for putting this together Mo. LGTM overall. One minor issue with one of the comments. Please merge yourself.

I'll update extensions/EXT_color_buffer_float/extension.xml with a note after this lands.

sdk/tests/deqp/functional/gles3/es3fFboTestCase.js
zhenyao: done

sdk/tests/conformance2/extensions/ext-color-buffer-float.html
zhenyao: done

PR #1781, opened 2016-06-15
Loops with side effects

Kangz:
These are variations on loops that would either cause a GPU reset on
macOS 10.9, or give bad results. Seems fixed in macOS 10.10


zhenyao:
lgtm

PR #1782, opened 2016-06-16
Fix gles3 draw

Kangz:
PTAL, this fixes gles3/draw.html

### DIFF
--- a/sdk/tests/deqp/framework/referencerenderer/rrVertexAttrib.js
+++ b/sdk/tests/deqp/framework/referencerenderer/rrVertexAttrib.js
@@ -245,7 +245,7 @@ var rrGenericVector = framework.referencerenderer.rrGenericVector;
                 dst = new arrayType(dst);
             }
 
-            rrVertexAttrib.read(dst, vertexAttrib.type, vertexAttrib.size, new Uint8Array(vertexAttrib.pointer).subarray(byteOffset));
+            rrVertexAttrib.read(dst, vertexAttrib.type, vertexAttrib.size, new Uint8Array(vertexAttrib.pointer, byteOffset));
         } else {
             dst = new arrayType(/** @type {Array<number>} */ vertexAttrib.generic.data);
         }
--- a/sdk/tests/deqp/framework/referencerenderer/rrVertexAttrib.js
+++ b/sdk/tests/deqp/framework/referencerenderer/rrVertexAttrib.js
@@ -220,6 +220,19 @@ var rrGenericVector = framework.referencerenderer.rrGenericVector;
         DE_ASSERT(rrVertexAttrib.isValidVertexAttrib(vertexAttrib));
         /** @type {goog.NumberArray} */ var dst;
 
+        var arrayType = null;
+        switch (genericType) {
+            case rrGenericVector.GenericVecType.INT32:
+                arrayType = Int32Array;
+                break;
+            case rrGenericVector.GenericVecType.UINT32:
+                arrayType = Uint32Array;
+                break;
+            case rrGenericVector.GenericVecType.FLOAT:
+                arrayType = Float32Array;
+                break;
+        }
+
         if (vertexAttrib.pointer) {
             /** @type {number} */ var elementNdx = (vertexAttrib.instanceDivisor != 0) ? (instanceNdx / vertexAttrib.instanceDivisor) : vertexNdx;
             /** @type {number} */ var compSize = rrVertexAttrib.getComponentSize(vertexAttrib.type);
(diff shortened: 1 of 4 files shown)


Kangz:
cc @kenrussell @zhenyao

Kangz:
Ah, it also makes the random draw tests harder to pass (it draws more geometry). That's a side effect of removing the workarounds for the reference rasterizer only supporting quads

sdk/tests/deqp/framework/referencerenderer/rrVertexAttrib.js
```diff
@@ -220,6 +220,19 @@ var rrGenericVector = framework.referencerenderer.rrGenericVector;
         DE_ASSERT(rrVertexAttrib.isValidVertexAttrib(vertexAttrib));
         /** @type {goog.NumberArray} */ var dst;
 
+        var arrayType = null;
```
Kangz: The C++ code is different from the original, and modified version and use templating.

sdk/tests/deqp/functional/gles3/es3fDrawTests.js
```diff
@@ -88,15 +88,14 @@ goog.scope(function() {
     es3fDrawTests.addTestIterations = function(test, baseSpec, type) {
         var spec = /** @type {glsDrawTests.DrawTestSpec} */ (deUtil.clone(baseSpec));
```
Kangz: Changes to this file make stuff closer to the C++ version

sdk/tests/deqp/modules/shared/glsDrawTests.js
```diff
@@ -1443,59 +1443,7 @@ goog.scope(function() {
...
-                    components[componentNdx] = glsDrawTests.GLValue.getRandom(vertexParameters.rnd, vertexParameters.min, vertexParameters.max);
-                }
-            }
-        }
-        return components;
-    };
-
-    /**
-     * createBasicArray - Modified JS version to ensure quad generation for compatibility with refrast.
+     * createBasicArray
```
Kangz: Change in this file make the code massively closer to the C++ version, some small differences are noted below

sdk/tests/deqp/modules/shared/glsDrawTests.js
```diff
@@ -1524,228 +1475,56 @@ goog.scope(function() {
...
-                        vertex4[oppositeaxis] = vertex2[oppositeaxis];
-                        vertex4[parallelaxis] = vertex3[parallelaxis];
-
-                        quadVertices = [vertex1, vertex2, vertex3, vertex4];
+            var components = [];
+
+            for (var componentNdx = 0; componentNdx < componentCount; componentNdx++) {
+                var getRandomComponent = function() {
+                    // For packed formats we can't use GLValue
+                    if (packed) {
```
Kangz: The C++ version use templating on the vertex attribute type, while here we use GLValue that doesn't handle the (U)INT_2_10_10_10_REV formats so we special case these formats with the packed variable like the previous code was doing in various places.

sdk/tests/deqp/modules/shared/glsDrawTests.js
```diff
@@ -1524,228 +1475,56 @@ goog.scope(function() {
...
-                }
-                case glsDrawTests.DrawTestSpec.Primitive.TRIANGLE_FAN: {
-                    if (vertexNdx < first + 4) {
-                        vertex1 = generateVertex(components, vertexParameters);
+                };
 
-                        parallelaxis = rnd.getBool() ? 0 : 1;
-                        oppositeaxis = parallelaxis ? 0 : 1;
+                var component = getRandomComponent();
+                var componentFloat = (component instanceof glsDrawTests.GLValue) ? component.toFloat() : component;
```
Kangz: The dEQP code doesn't use componentFloat and uses multiple loops on componentNdx. The logic here is he same but expressed slightly differently.

kenrussell:
Awesome. LGTM from my standpoint. @zhenyao has been reviewing all of the dEQP fixes so let's let him take a look and merge.

sdk/tests/deqp/functional/gles3/es3fDrawTests.js
```diff
@@ -109,10 +108,10 @@ goog.scope(function() {
             test.addIteration(spec, 'instance count = ' + spec.instanceCount);
         } else if (type == es3fDrawTests.TestIterationType.INDEX_RANGE) {
             spec.indexMin = 0;
-            spec.indexMax = 24;
+            spec.indexMax = 23;
             test.addIteration(spec, 'index range = [' + spec.indexMin + ', ' + spec.indexMax + ']');
 
-            spec.indexMin = 24;
+            spec.indexMin = 23;
             spec.indexMax = 41;
```
zhenyao: c++ code this is 40. Why align all other numbers but not this one?

zhenyao:
This is great work.  Mostly looks good with one question.

sdk/tests/deqp/functional/gles3/es3fDrawTests.js
Kangz: I didn't realized this one had changed, will align with the C++ code.

Kangz: Done.

zhenyao:
LGTM
Thank you!

PR #1783, opened 2016-06-17
Split transform feedback tests into semantically useful groups.

kenrussell:


kenrussell:
@zhenyao please review. Tested locally and manually. Verified the right tests showed up in the right HTML files.

zhenyao:
You will also want to delete the original transformfeedback.html

kenrussell:
Thanks for the feedback. transformfeedback.html is deleted by this patch. It showed up as a rename of the original .html file into one of the autogenerated ones, plus edits.

zhenyao:
OK, thanks for pointing this out - I didn't realize.

PR #1785, opened 2016-06-17
Remove shader tests on line continuation

qkmiao:
There is no line continuation character in GLSL 3.0. See Section 3.1
"Character Set" in GLSL spec 3.30.6.


zhenyao:
I checked GLSL spec 3.00.4, '\' is legal.

qkmiao:
Yes, GLSL ES spec 3.00.4 supports '\'. Close this pr.

PR #1788, opened 2016-06-20
Fix backslash in shader source

qkmiao:


sdk/tests/deqp/modules/shared/glsShaderLibrary.js
```diff
@@ -145,9 +145,7 @@ var gluShaderUtil = framework.opengl.gluShaderUtil;
         if (trimFront)
             str = str.replace(/^\s*\n/, '');
         return str
-            .replace('\\n', '\n')
-            .replace('\\t', '\t')
-            .replace(/\\/g, '');
+            .replace(/\\\\/g, '\\');
```
qkmiao: See C++ code:
https://android.googlesource.com/platform/external/deqp/+/deqp-dev/framework/opengl/gluShaderLibrary.cpp#537

tow \ is converted to one \: "\\\\" -> "\\", e.g. "\\note" should be converted to "\note" other than
 "
ote".

zhenyao: What about replacing "\n" to '\n' and "\t" to '\t"? That seems lost with this change?

qkmiao: It's difficult to do replacement for '\n', '\t', '\\' with repalce() at the same time. Align the code with C++.

zhenyao:
lgtm

PR #1790, opened 2016-06-21
Clarify that an error is generated when drawing without program

Oletus:
In GLES 2.0 and 3.0 this is undefined behavior. The WebGL 1.0 spec
already specified this for drawArrays and drawElements, but this patch
documents it also in the spec differences section and in the WebGL 2.0
spec.

There are already tests for this, including tests for WebGL 2.0 in
dEQP.


zhenyao:
lgtm

PR #1792, opened 2016-06-22
Bulk remove EOL whitespace.

jdashg:
It's a bitter pill, but let's take it now.


jdashg:
@kenrussell @grorg

kenrussell:
I'm dubious of the value of touching the older conformance suite snapshots, but if you think this is desirable, OK.

PR #1793, opened 2016-06-22
fix the bugs in fboMultisample.html against sRGB color format

Richard-Yunchao:
The failure is a little deep, the variables and the shaders feed to real renderer and renference renderer are exactly the same. However, the generated reference image looks much sharper than the real result, I guess that either the msaa is not applied to the reference image, or the reference image doesn't change the colos space from srgb8_alhph8 to linear (rgba). After tracing the native code, it turns out the deduction is true. 

Now fboMultisample.html can pass. PTAL. Thanks!

### DIFF
--- a/sdk/tests/deqp/framework/referencerenderer/rrRenderer.js
+++ b/sdk/tests/deqp/framework/referencerenderer/rrRenderer.js
@@ -386,7 +386,7 @@ void FragmentProcessor::render (const rr::MultisamplePixelBufferAccess& msColorB
     var stencilState = state.stencilStates[fragmentFacing];
     var colorMaskFactor = [state.colorMask[0] ? 1 : 0, state.colorMask[1] ? 1 : 0, state.colorMask[2] ? 1 : 0, state.colorMask[3] ? 1 : 0];
     var colorMaskNegationFactor = [state.colorMask[0] ? false : true, state.colorMask[1] ? false : true, state.colorMask[2] ? false : true, state.colorMask[3] ? false : true];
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/framework/referencerenderer/rrRenderer.js
```diff
@@ -386,7 +386,7 @@ void FragmentProcessor::render (const rr::MultisamplePixelBufferAccess& msColorB
     var stencilState = state.stencilStates[fragmentFacing];
     var colorMaskFactor = [state.colorMask[0] ? 1 : 0, state.colorMask[1] ? 1 : 0, state.colorMask[2] ? 1 : 0, state.colorMask[3] ? 1 : 0];
     var colorMaskNegationFactor = [state.colorMask[0] ? false : true, state.colorMask[1] ? false : true, state.colorMask[2] ? false : true, state.colorMask[3] ? false : true];
-    var sRGBTarget = false;
+    var sRGBTarget = state.sRGBEnabled && colorBuffer.getFormat().isSRGB();
```
Richard-Yunchao: See the native c++ code: https://android.googlesource.com/platform/external/deqp/+/deqp-dev/framework/referencerenderer/rrFragmentOperations.cpp#777. 

Some files between JS and native C++ are not exactly the same, but this function in JS (rrRenderer.js) is ported from the one in rrFragmentOperation.cc.

zhenyao:
Great job!

kenrussell:
Superb work @Richard-Yunchao !

PR #1795, opened 2016-06-22
Fix lint errors in sdk

renchenglei:
Clean up the lint errors under sdk folder. @Richard-Yunchao @Honry @zqzhang ptal, thanks a lot.


zhenyao:
LGTM

PR #1796, opened 2016-06-22
Create an RGBA from an array of floats

anzurio:
- Added a function to convert an array of floats into an RGBA object.
- Used the aforemention function to solve some framebufferblit tests'
  issues.


zhenyao:
lgtm

PR #1811, opened 2016-06-24
Add more logging to ImageBitmap test

xidachen:
A WebGL conformance test for ImageBitmap seems to time out some times:
https://bugs.chromium.org/p/chromium/issues/detail?id=596622

Here we add more logs to that test so that next time when it timed out, we will have enough info to know which point it timed out.

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-with-image-bitmap-utils.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-with-image-bitmap-utils.js
@@ -53,10 +53,10 @@ function runOneIterationImageBitmapTest(useTexSubImage2D, bindingTarget, program
     }
 
     if (optionsVal.is3D)
-        debug('Testing texSubImage3D' + ' with flipY=' + flipY + ' and premultiplyAlpha=' + premultiplyAlpha +
+        console.log('Testing texSubImage3D' + ' with flipY=' + flipY + ' and premultiplyAlpha=' + premultiplyAlpha +
                 ', bindingTarget=' + (bindingTarget == gl.TEXTURE_3D ? 'TEXTURE_3D' : 'TEXTURE_2D_ARRAY'));
     else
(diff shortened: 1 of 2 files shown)


zhenyao:
lgtm

sdk/tests/js/tests/tex-image-and-sub-image-with-image-bitmap-utils.js
```diff
@@ -53,10 +53,10 @@ function runOneIterationImageBitmapTest(useTexSubImage2D, bindingTarget, program
     }
 
     if (optionsVal.is3D)
-        debug('Testing texSubImage3D' + ' with flipY=' + flipY + ' and premultiplyAlpha=' + premultiplyAlpha +
+        console.log('Testing texSubImage3D' + ' with flipY=' + flipY + ' and premultiplyAlpha=' + premultiplyAlpha +
                 ', bindingTarget=' + (bindingTarget == gl.TEXTURE_3D ? 'TEXTURE_3D' : 'TEXTURE_2D_ARRAY'));
     else
-        debug('Testing ' + (useTexSubImage2D ? 'texSubImage2D' : 'texImage2D') +
+        console.log('Testing ' + (useTexSubImage2D ? 'texSubImage2D' : 'texImage2D') +
```
kenrussell: This will change the output of the test, moving it from the web page to the console and making it more difficult to interactively tell which part of the test is running. It would be better to compute the string:

var str = 'Testing ' + ...

then:
debug(str);
console.log(str);

Could you make this change in a follow-on pull request?

kenrussell:
Basically LGTM after the fact with one nit. Would appreciate a follow-up.

sdk/tests/js/tests/tex-image-and-sub-image-with-image-bitmap-utils.js
zhenyao: I sent a pull request here: https://github.com/KhronosGroup/WebGL/pull/1813

PR #1812, opened 2016-06-24
Fix a bug in draw-buffers.html

zhenyao:
In ES3, when there is only a single output, it defaults to location zero. It
should not broadcast to all buffers.


zhenyao:
@kenrussell @qkmiao Please take a look

kenrussell:
This looks correct to me based on reading of the ES 3.0.4 and ESSL 3.00.4 specifications.

kenrussell:
Merging now.

PR #1815, opened 2016-06-27
Fix lint errors in conformance-suites/1.0.3 which can't be checked by tools

renchenglei:
@zhenyao PTAL, thanks~


kenrussell:
LGTM based on skimming through, but @zhenyao should look and merge.

zhenyao:
Looks good.  Thanks for the cleaning up.

PR #1816, opened 2016-06-28
deqp/gles3/fragmentoutput: for ints, the threshold is UINT_MAX

Kangz:
Image comparison was done with a threshold of 1 for unwritten integer
channels, which forced them to be in the [-1, 1] range, This was not
correct because the integer formats are not normalized, use UINT_MAX
instead.

This makes the code match the C++ version of dEQP.


zhenyao:
LGTM

PR #1817, opened 2016-06-29
Improve webgl-draw-buffers.html

zhenyao:


zhenyao:

sdk/tests/conformance/extensions/webgl-draw-buffers.html
```diff
@@ -553,17 +560,30 @@
   if (maxUsable > 1) {
     var bufs1 = makeColorAttachmentArray(maxUsable);
     var bufs2 = makeColorAttachmentArray(maxUsable);
+    var bufs3 = makeColorAttachmentArray(maxUsable);
+    var bufs4 = makeColorAttachmentArray(maxUsable);
```
kenrussell: How about a one-line comment before each "bufs" variable saying what each intends to test? (For example, "First half of color buffers disabled", "Second half", etc.)

kenrussell:
LGTM. One small comment. Please merge yourself. Thanks.

sdk/tests/conformance/extensions/webgl-draw-buffers.html
zhenyao: Done.

PR #1818, opened 2016-06-29
Revise WEBGL_draw_buffers behavior.

kenrussell:
While upgrading to WebGL 2.0 it has been discovered that some of the behaviors defined by WebGL 1.0 with the WEBGL_draw_buffers extension will impose severe performance penalties as implementations are upgraded. Revise the extension and the conformance test to avoid these penalties.

### DIFF
--- a/extensions/WEBGL_draw_buffers/extension.xml
+++ b/extensions/WEBGL_draw_buffers/extension.xml
@@ -68,7 +68,10 @@
         Likewise the shading language preprocessor <code>#define GL_EXT_draw_buffers</code>, will be defined to 1 if the extension is supported.
       </addendum>
       <addendum>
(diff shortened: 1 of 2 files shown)


kenrussell:
@zhenyao @jdashg @grorg @RafaelCintron , in particular, please review.

The test hasn't been fully updated yet to test all of the new behaviors. If there is agreement that this is an acceptable direction, it will be.

This change will enable WebGL 1.0 implementations to perform very simple modifications in the shader translator to match the specified behavior. The previous behavior, requiring broadcasting of gl_FragColor in ESSL 1.00 shaders in WebGL 1.0 + WEBGL_draw_buffers when EXT_draw_buffers was not enabled in fragment shaders, would impose a severe performance penalty in all WebGL 1.0 content running on later OpenGL implementations.

Please review and comment. Thanks.

zhenyao:
LGTM

jdashg:
This seems fine.

zhenyao:
Let's merge now.  If someone feels differently, we can always revise.

Issue #1819, opened 2016-06-29
Mac Intel drivers do not sample cubemap integer textures correctly with seamless cubemap enabled


zhenyao:
The bug only applies to texture size of 64x64 or smaller.

zhenyao:
Correction: the bug actually exists in any texture sizes.  It's just when the size is big enough and the rendering target isn't, then the incorrect corner pixel (top right, tex coords (1, 1)) won't show.

kenrussell:
Recent drivers seem to have fixed this problem and @kainino0x has removed the workarounds in #2796.

PR #1820, opened 2016-06-30
Move the failing case (sampling integer cubemap textures) to a separa…

zhenyao:
…te tests.

The purpose is to isolate the failing case so we can skip in WebGL 2.0.0 and leave it
for future versions where the driver bug can be fixed or worked around.

https://github.com/KhronosGroup/WebGL/issues/1819

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
@@ -90,7 +90,7 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
       ctx.canvas.height = 2;
       setCanvasToRedGreen(ctx);
     }
-  
+
     function runOneIteration(canvas, useTexSubImage2D, flipY, program, bindingTarget, opt_texture)
     {
         debug('Testing ' + (useTexSubImage2D ? 'texSubImage2D' : 'texImage2D') + ' with flipY=' +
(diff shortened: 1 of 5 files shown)


zhenyao:
@kenrussell Please review.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
```diff
@@ -154,14 +165,14 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
...
+            wtu.checkCanvasRect(gl, 0, bottom, (skipCorner && !flipY) ? halfWidth : width, halfHeight, redColor,
                     "shouldBe " + redColor);
             debug("Checking " + (flipY ? "bottom" : "top"));
-            wtu.checkCanvasRect(gl, 0, top, width, halfHeight, greenColor,
+            wtu.checkCanvasRect(gl, 0, top, (skipCorner && flipY) ? halfWidth : width, halfHeight, greenColor,
                     "shouldBe " + greenColor);
         }
 
-        if (false) {
+        if (true) {
```
kenrussell: Is the debugging image still desired? If so, that's fine.

kenrussell:
Excellent work Mo diagnosing this bug and splitting off the failing test cases. LGTM from my standpoint.

One question, but it's minor, and could be addressed later. Merging now.

Richard-Yunchao:
I think we should not skip these tests in WebGL 2.0.1. WDYT? @zhenyao , @kenrussell . We can file radar to Apple if necessary.

FYI. @gyagp

kenrussell:
@Richard-Yunchao @gyagp These tests are enabled in the WebGL 2.0.1 test suite.

The associated Chromium bug is http://crbug.com/658930 and the Radar was 27231650. It looks like it's been fixed on Intel GEN9 GPUs on macOS but not on GEN7. If fixing these issues on GEN7 is a priority then please help us file a new Radar. Thanks.

PR #1823, opened 2016-06-30
Add getWildcardSkipStatus function

JiangYizhou:
using getSkipsStatus function can skip a list of test cases by testname*

### DIFF
--- a/sdk/tests/deqp/framework/common/tcuSkipList.js
+++ b/sdk/tests/deqp/framework/common/tcuSkipList.js
@@ -43,6 +43,7 @@ goog.scope(function() {
     var tcuSkipList = framework.common.tcuSkipList;
 
     var _skipEntries = {};
+    var _wildcardSkipEntries = {};
     var _reason = "";
 
     function _setReason(reason) {
@@ -50,7 +51,12 @@ goog.scope(function() {
     }
 
     function _skip(testName) {
-        _skipEntries[testName] = _reason;
+        if(testName.indexOf("*") >= 0){
+            testName = testName.split("*")[0];
+            _wildcardSkipEntries[testName] = _reason;
+        }else{
+            _skipEntries[testName] = _reason;
+        }
     }
 
     var runSkippedTests = false;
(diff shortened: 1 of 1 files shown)


JiangYizhou:
@Richard-Yunchao @zhenyao PTAL

JiangYizhou:
@Richard-Yunchao PTAL

Richard-Yunchao:
LTGM

sdk/tests/deqp/framework/common/tcuSkipList.js
```diff
@@ -96,10 +97,28 @@ goog.scope(function() {
...
 
+    /*
+     * Gets the skip status of the given tests like testpath*
+     * object with the properties "skip", a boolean, and "reason", a
+     * string.
+    */
+    tcuSkipList.getSkipsStatus = function(testName) {
+        var skipEntry;
+        for (var key in _skipEntries) {
+            if (testName.indexOf(key) >=0 ) {
```
zhenyao: I am a bit concerned about the perf impact of this change because it's for every test case within a single test.

kenrussell: It has been the case in the past that lookups of test expectations with wildcards has been a major source of slowdowns of the WebGL conformance suite. Please see http://crbug.com/522234 and https://codereview.chromium.org/1310123002 .

Sorry, but I don't think we should make this change. There shouldn't be that many entries in the skip list, and I think those tests can be added explicitly. How many do you predict will be needed?

Richard-Yunchao: The native hw-issues.txt and driver-issues.txt in deqp in aosp also use * to skip a bunch of test cases. I will let Yizhou have more investigations about this. Thanks for your feedbacks, @zhenyao and @kenrussell .

kenrussell:
What is the plan for this pull request? Would it be OK to state that in the JavaScript port of the dEQP tests, we will not use wildcards for test suppressions, and close this pull request?

Richard-Yunchao:
@kenrussell , Yizhou try to collect some performance data for this pr. I look forward to her performance data of the original test time vs the test time with this pr and wildcards for test suppressions.

Richard-Yunchao:
The code LGTM again.
 
@kenrussell , @zhenyao , Yizhou has test the original code vs this pr to investigate the perf impact yersterday. The data show that the run time is actually the same. For instance, she run all/deqp/data/gles3/shaders w/ and w/o this pr several times to get the average number, the total test time are actually the same: 1min 43s. There are lots of cases in the category of all/deqp/data/gles3/shaders, but the real run time for each test is short, so the time to prepare the test (time to run into the skip list is included here) is relative long. You know, this kind of cases will expose the potential perf impact for this pr. But the finally data show that there is no obvious perf impact.  

I guess that the string comparison in this pr is lightweight in JS. Moreover, the original code in JS use object comparison (although it looks like an array), which would be not faster than string comparison. These features may differ from the C++ implementation of wildcards you point out in Chromium. 

This is just FYI. It is up to you to merge it or not. If not, we would glad to close this pr.

kenrussell:
@Richard-Yunchao @JiangYizhou thanks for doing this performance measurement. However, the real performance impact will happen when many wildcard suppressions are added to the skip list. Could you please try adding the ones you had in mind from the native dEQP, and re-measure the time to run that suite? Thanks.

Richard-Yunchao:
Thanks for your suggestions, @kenrussell . In fact, I have already suggested @JiangYizhou to add a couple of extra tests into the skip list (the original one vs wildcards) in the test data above. We will add more cases to stress the tests  and re-measure the time.

Richard-Yunchao:
@kenrussell , Yizhou has added 10+ tests with wildcard into skip list to test the time. Then she added corresponding tests one by one into skip list to thest the time. The average time for they are basically the same: 1min 43s when she run all/deqp/data/gles3/shaders (3K test cases in this category). Besides, the total test time basically keep the same as the test results without these new added cases.  we have listed the latter in comments above. The overhead is not as big as we expected. 
In addition, there are not many test cases (those failed in both native deqp and in webgl deqp) to add into the skip list for webgl 2.0.0 shippment, considering that we began to add some test cases into 2.0.1.

sdk/tests/deqp/framework/common/tcuSkipList.js
```diff
@@ -43,14 +43,20 @@ goog.scope(function() {
     var tcuSkipList = framework.common.tcuSkipList;
 
     var _skipEntries = {};
+    var _skipList = {};
```
kenrussell: Please rename this to something more descriptive like "_wildcardSkipEntries".

sdk/tests/deqp/framework/common/tcuSkipList.js
```diff
@@ -127,10 +120,28 @@ goog.scope(function() {
...
             return { 'skip': true, 'reason': skipEntry };
         }
     }
 
+    /*
+     * Gets the skip status of the given tests like testpath*
+     * object with the properties "skip", a boolean, and "reason", a
+     * string.
+    */
+    tcuSkipList.getSkipsStatus = function(testName) {
```
kenrussell: Please rename to an internal function (using "_" at the beginning) and to something more descriptive like "_getWildcardSkipStatus".

kenrussell:
OK. Thanks @JiangYizhou and @Richard-Yunchao for testing. Could you please make a couple of renaming changes? We'll merge this afterward. Thanks.

JiangYizhou:
@kenrussell I have updated accordingly. PTAL.

kenrussell:
Thanks for updating the code. Looks good. Merging.

PR #1824, opened 2016-06-30
Skip top right corner pixel comparison in textureformat cubemap test …

zhenyao:
…cases.

This is to make the test pass with Mac Intel driver bug.

A separate test case has been added to catch the driver bug.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fTextureFormatTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fTextureFormatTests.js
@@ -307,7 +307,16 @@ es3fTextureFormatTests.TextureCubeFormatCase.prototype.testFace = function(face)
         this.m_texture.getRefTexture(), texCoord, renderParams);
 
     // Compare and log.
-    var isOk = glsTextureTestUtil.compareImages(referenceFrame, renderedFrame, threshold);
+    var skipPixels = null;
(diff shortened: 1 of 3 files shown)


zhenyao:
@kenrussell Please review.  This makes textureformat/sized_color_cube_* cases all pass.

sdk/tests/deqp/functional/gles3/es3fTextureFormatTests.js
```diff
@@ -307,7 +307,12 @@ es3fTextureFormatTests.TextureCubeFormatCase.prototype.testFace = function(face)
         this.m_texture.getRefTexture(), texCoord, renderParams);
 
     // Compare and log.
-    var isOk = glsTextureTestUtil.compareImages(referenceFrame, renderedFrame, threshold);
+    // Skip top right pixel due to Mac Intel driver bug.
+    // https://github.com/KhronosGroup/WebGL/issues/1819
+    var skipPixels = [
+        [this.m_width - 1, this.m_height - 1]
+    ];
```
kenrussell: Can skipPixels be applied only when an integer format is used? It would be best to minimize the number of situations where testing is reduced, if most formats are working on this GPU.

zhenyao: Done.  Please take another look.

kenrussell:
Great. Thanks. LGTM

PR #1825, opened 2016-06-30
Relax tolerance a bit to allow RGB565 RGB5_A1 RGB10_A2 format readPix…

zhenyao:
…els to pass

On Mac Intel.


zhenyao:
This makes conformance2/reading/read-pixels-from-fbo-test.html pass on Mac Intel.
@kenrussell  Please take a look.

kenrussell:
LGTM!

PR #1826, opened 2016-06-30
deqp/gles3/fborender: Bind to create doesn't work in WebGL

Kangz:
C++ dEQP uses bind to create in the fborender recreate tests. This
doesn't work in WebGL so instead we explicitly use glCreateTexture and
glCreateRenderbuffer and (grossly) set the fbo state to these new
objects.


zhenyao:
Great work!

kenrussell:
Great work tracking this down Corentin. LGTM after the fact.

PR #1827, opened 2016-07-01
Test line-continuation character in shader for WebGL 2.0

qkmiao:
Line-continuation character (backslash) is supported in WebGL 2.0. See
ESSL spec 3.00.4 in section 3.1 Character Set. This patch adds test
which is not covered by deqp/data/gles3/shaders/preprocessor.html.


zhenyao:
This looks fine to me, but @kenrussell should also have a look.

kenrussell:
LGTM. Thanks for adding this test.

PR #1828, opened 2016-07-01
Improve compound assignment type combination testing

Oletus:
1. Refactor the test logic into a JavaScript file
2. Test that vecX *= matX is a valid operation
3. Test non-square matrices in ESSL 3.00 shaders

The WebGL 2.0 version of the test reveals a bug in ANGLE that got
fixed recenty.


zhenyao:
LGTM

kenrussell:
Great work improving these tests. LGTM after the fact too.

PR #1829, opened 2016-07-01
Uniform block does not be initialized

xhcao:
The case cannot pass on mesa+i965, but pass on nvidia platform. It seems that shader uniform block members' default value is 0 on mesa+i965, but is 1 on nvidia driver.

### DIFF
--- a/sdk/tests/conformance2/buffers/uniform-buffers.html
+++ b/sdk/tests/conformance2/buffers/uniform-buffers.html
@@ -193,48 +193,68 @@ function runDrawTest() {
         return;
     }
 
-    var blockIndex = gl.getUniformBlockIndex(program, "UBOData");
-    var blockSize = gl.getActiveUniformBlockParameter(program, blockIndex, gl.UNIFORM_BLOCK_DATA_SIZE);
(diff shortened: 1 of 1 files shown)


zhenyao:
LGTM. Thanks for fixing this.

@xhcao Can you dig a bit further into the spec and see if the uniform default values are defined or undefined?

kenrussell:
I found the appropriate text in section 2.12.6 "Uniform variables" of the ES 3.0.4 spec:

"If any active uniform block is not backed by a sufficiently large buffer object, the results of shader execution are undefined, and may result in GL interruption or termination."

This is unfortunate, and will have to be addressed in the WebGL 2.0 spec before shipment. My first instinct is to generate an INVALID_OPERATION error for draw calls with programs that have active uniform blocks which don't have sufficient storage bound to them. KHR_robust_buffer_access_behavior probably defines this situation well enough to write a conformance test for it (i.e., fetches of the uniforms that go beyond the end of the storage return the zero value), but this can't be the only defined behavior in WebGL 2.0; it would be impossible to emulate this behavior when KHR_robust_buffer_access_behavior isn't present.

PR #1831, opened 2016-07-02
Add test asserting that global invariant pragma does not leak state.

kenrussell:
This test fails in Chrome Canary on Mac OS, at least. Does not fail in Firefox or Safari. Not sure about Chrome on other platforms yet. Filed http://crbug.com/625363 to track it.


zhenyao:
Thanks for tracking this down. LGTM

kenrussell:
This bug is insane. Haven't debugged it fully yet.

kenrussell:
Thanks @zhenyao for reviewing this just before a holiday weekend.

PR #1835, opened 2016-07-04
Add invalid case for line-continuation character

qkmiao:
This fixes comments#11 in https://codereview.chromium.org/2082483002/.

### DIFF
--- a/sdk/tests/conformance2/glsl3/shader-with-invalid-characters.html
+++ b/sdk/tests/conformance2/glsl3/shader-with-invalid-characters.html
@@ -49,9 +49,46 @@ void main()
     gl_Position = vec4(1.0, 0.0, 0.0, 1.0);
 }
 </script>
+<script id="vertexShaderBackslashValid" type="text/something-not-javascript">
+void main()
+{
+    gl_Position = \
+        vec4(1.0, 0.0, 0.0, 1.0);
+}
+</script>
+<script id="vertexShaderBackslashInvalid" type="text/something-not-javascript">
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/glsl3/shader-with-invalid-characters.html
```diff
@@ -49,9 +49,36 @@
...
+void main()
+{
+    gl_Position = \
+        vec4(1.0, 0.0, 0.0, 1.0);
+}
+</script>
+<script id="vertexShaderBackslashInvalid" type="text/something-not-javascript">
+void main()
+{
+    gl_Position = \ vec4(1.0, 0.0, 0.0, 1.0);
```
zhenyao: I think a more meaningful test is after \ there are a few white space
that could be common mistake to make

In this situation, it's better you put source directly as a string in the source instead of a html element - to make sure lint tools won't get rid of the whitespace in the end of a line

qkmiao: Another test was added. PTAL.

zhenyao:
lgtm

kenrussell:
Thanks for adding these tests @qkmiao . Per discussion on #1836 , I hope we are going in the right direction when adding these tests. There aren't many dEQP tests which verify behavior of the line continuation character in the middle of the line. It seems the behavior is mainly defined by the dEQP tests and not very well defined by the ESSL 3.00.4 spec.

PR #1836, opened 2016-07-04
Remove unsupported test for WebGL 2.0

qkmiao:
Backslash in middle of a line isn't supported by WebGL 2.0. WebGL will
report invalid value error before compiling shader.


zhenyao:
This makes me think if we are doing the right thing in Chrome.  DEQP can be considered as part of the spec. So if it requires this, maybe WebGL2 should not reject it. The spec interpretation is ambiguous.

@kenrussell What's your opinion?

zhenyao:
@qkmiao Per offline discussion with kbr, we agree that this case should be left in and we should make Chrome pass this test.

PR #1837, opened 2016-07-04
fix an error from pull #1795

renchenglei:
@zhenyao When I cleaned the lint errors in sdk folders, I modified shader-lib-test.html by mistake. I'm so sorry, PTAL. 
BTW: There are still 14576 lint errors in sdk folders. Most of them are in *.frag and *.vert files. Wheather it need help?

### DIFF
--- a/sdk/tests/deqp/data/gles2/shaders/shader-lib-test.html
+++ b/sdk/tests/deqp/data/gles2/shaders/shader-lib-test.html
@@ -158,7 +158,7 @@
                         test_result(
                                 "Checking string parser. Passing \"\"hello world\"extra\". Expecting \"hello world\"",
                                 result == "hello world", result
-ts/deqp/data/gles2/shaders/qualification_order.html                       );
+                        );
                         return true;
                 },
                 function() {


Kangz:
Happens :) LGTM

zhenyao:
No need to apologize.  I am just surprised that I already rolled this change into chromium but no test caught this bug.

renchenglei:
@zhenyao I have scan all the folders. The shader-lib-test.html is not in the test lists or used by other files. So maybe it's the reason.

zhenyao:
Thanks.  Good to know.

kenrussell:
How are the lint errors being discovered? Does your team run a script, is there an IDE or tool which highlights them, etc.?

If they can be cleaned up non-destructively, it seems fine to me to do one pass that cleans them up.

renchenglei:
@kenrussell w3c/web-platform-tests has a [tool](https://github.com/w3c/wpt-tools/tree/09258ba099c2cd0f7c107a5d4e8b5f458c209a47/lint) to check lint error. We integrated the tool into our [test suite](https://github.com/crosswalk-project/crosswalk-test-suite) for catching common mistakes.  By running the tool, we can know where lint error occurs and which type error it is. The most lint errors in WebGL tests are TRAILING WHITESPACE, INDENT TABS or CR AT EOL. You can use ‘sed' command to fix them. For example:
TRAILING WHITESPACE   sed -i 's/[ \t]*$//' /path/to/file
INDENT TABS                     sed -i 's/\t		/    /g' /path/to/file
CR AT EOL                          sed -i 's/\r//g' /path/to/file
The 'sed' command can fix TRAILING WHITESPACE and CR AT EOL lint error directly. But it's better to check the INDENT TABS lint errors before pull request. For example:
        var x=0; //use 'table'
        var y=0; //use 'space'
After sed -i 's/\t		/    /g' /path/to/file:
    var x=0; //change 'table' to 'space'
        var y=0; //no change. We need delete four 'sapce' by hand to keep alignment.
Now the lint tool can check html, htm, xhtml, xhtm, json and xml files. But there are many INDENT TABS lint errors in *.frag and *.vert files under conformance/ogles/GL. You can modify [lint.py](https://github.com/crosswalk-project/crosswalk-test-suite/blob/master/tools/lint/lint.py) by:
 if re.compile('\.html$|\.htm$|\.xhtml$|\.xhtm$').search(abs_path):  //line 257, before
 if re.compile('\.html$|\.htm$|\.xhtml$|\.xhtm$|\.vert$|\.frag$').search(abs_path):  //line 257, after
Now, you can also catch lint errors in *.vert and *.frag files. 
Our lint tool comes from https://github.com/w3c/wpt-tools/ with some modification. There is a detail doc to introduce the lint tool(https://github.com/crosswalk-project/crosswalk-test-suite/tree/master/tools/lint)
FYI. :)

kenrussell:
@renchenglei if you would like to integrate this script into the workspace perhaps under sdk/tests/py/ that would be welcome.

renchenglei:
@kenrussell OK, NP. :)

PR #1838, opened 2016-07-04
Fix a flakiness in gl-object-get-calls.js

zhenyao:
After fencSync, the commands can be completed even if the getSyncParameter is
called right away.


zhenyao:
I've seen this test being flaky from time to time, and today I took a look at the output.  This seems to be the cause.

@kenrussell @qkmiao Can one of you take a look?

zhenyao:
@Kangz Can you take a look? Others are either sleeping or on celebrating holidays.

Kangz:
LGTM, this seems especially true since the code above doesn't do any GPU work.

zhenyao:
Thank you.

qkmiao:
LGTM after the fact. It just a different case than what I met before. In my case, the sync status was changed to "SIGNALED" after several frames.

kenrussell:
Sorry for the delay reviewing.

This seems OK for now but exposes some undefined behavior in the WebGL 2.0 spec. Queries have been specified to not become available in the current frame: see https://www.khronos.org/registry/webgl/specs/latest/2.0/#5.23 . The same should probably be done for fences. Perhaps add a TODO about this in the test, at least?

PR #1839, opened 2016-07-05
Fix ambiguous test cases in tex-sub-image-2d-bad-args.html

Richard-Yunchao:
When we test against the argument 'format', we should keep the argument 'type' correct. 

PTAL. Thanks a lot!

### DIFF
--- a/sdk/tests/conformance/textures/misc/tex-sub-image-2d-bad-args.html
+++ b/sdk/tests/conformance/textures/misc/tex-sub-image-2d-bad-args.html
@@ -100,15 +100,15 @@ if (contextVersion < 2) {
     gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGB, gl.UNSIGNED_SHORT_5_6_5, c);
     wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "type not same as original RGB");
 }
-gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_SHORT_4_4_4_4, c);
-wtu.glErrorShouldBe(gl, gl.NO_ERROR, "make texture RGBA 4_4_4_4");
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/textures/misc/tex-sub-image-2d-bad-args.html
```diff
@@ -104,11 +104,11 @@
 wtu.glErrorShouldBe(gl, gl.NO_ERROR, "make texture RGBA 4_4_4_4");
 gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGBA, gl.UNSIGNED_SHORT_4_4_4_4, c);
 wtu.glErrorShouldBe(gl, gl.NO_ERROR, "format same as original RGBA 4_4_4_4");
-gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGB, gl.UNSIGNED_BYTE, c);
-wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "format not same as original RGBA 4_4_4_4");
+gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGB, gl.UNSIGNED_SHORT_4_4_4_4, c);
```
zhenyao: I am not sure this is a meaningful change, because RGB/UNSIGNED_SHORT_4_4_4_4 are not valid combination, so you didn't really isolate a failing cause.

sdk/tests/conformance/textures/misc/tex-sub-image-2d-bad-args.html
```diff
@@ -121,19 +121,19 @@
...
```
zhenyao: Same here.  I agree the current case is not perfect, but I am not sure the change makes it any better.

Richard-Yunchao:
Thanks for your review, @zhenyao . Code has been updated accordingly. Now format/type combinations are OK too. 
PTAL. Thanks a lot!

zhenyao:
lgtm

kenrussell:
LGTM too. Thanks for the fix.

PR #1843, opened 2016-07-06
Allow 0 and 1 in gles3/f/fbodepthbuffer.depth_write_clamp

Kangz:
The test checked that gl_FragDepth is clamped to (0, 1) before it is
written to the depth buffer while the specification says the depth
should be clamped to [0, 1].

This was found while running the corresponding WebGL 2 CTS test on the
NVIDIA OpenGL driver.

Please do not land before the respective dEQP CL is landed in https://android-review.googlesource.com/#/c/245910/


kenrussell:
Great work tracking this down and thanks for submitting an upstream fix to the dEQP. LGTM too.

Kangz:
Not the actual fix, see #1854 for the fix.

PR #1844, opened 2016-07-06
Allow 0 gles3/f/framebufferblit.depth_stencil

Kangz:
In the _scaled cases, the areas of dstFbo that have not been blitted
onto still have depth = 0 from the texShader draw.

PTAL. I'd like to land this only when the upstream fix has been landed at https://android-review.googlesource.com/#/c/245889/


kenrussell:
Very nice. LGTM. Please merge yourself after the upstream fix lands, and let's let @zhenyao review too.

zhenyao:
lgtm

Kangz:
Not the actual fix, see #1854 for the fix.

PR #1846, opened 2016-07-07
Remove incorrect preprocessor.pragmas.pragma_fragment test.

kenrussell:
This test is buggy and is in the skip list for the native dEQP (in
android/cts/master/src/gles2-failures.txt). The test is never going to
be updated (the invariace matching rules are different between ESSL 1.00
and 3.00) so to avoid confusion the test is being removed rather than
added to tcuSkipList.js.

See http://crbug.com/625363 .


kenrussell:
@zhenyao please review.

zhenyao:
LGTM

Issue #1832, opened 2016-07-02
Mac OSX drivers handle R11F_G11F_B10F incorrectly


zhenyao:
A bunch of deqp tests fail because of this.

Let's leave these test cases out in WebGL 2.0.0 and wait for a driver fix to roll out.

zhenyao:
This is fixed in 10.12.2 (16C67)

PR #1847, opened 2016-07-07
Add a test for format R11F_G11F_B10F.

zhenyao:
See https://github.com/KhronosGroup/WebGL/issues/1832.

### DIFF
--- a/sdk/tests/conformance2/reading/format-r11f-g11f-b10f.html
+++ b/sdk/tests/conformance2/reading/format-r11f-g11f-b10f.html
@@ -1,0 +1,165 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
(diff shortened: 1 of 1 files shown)


zhenyao:
@kenrussell Please review
@Richard-Yunchao @qkmiao FYI.

sdk/tests/conformance2/reading/format-r11f-g11f-b10f.html
```diff
@@ -0,0 +1,165 @@
...
+  data[1] = testG ? value : 0;
+  data[2] = testB ? value : 0;
+  data[3] = 1; // Doesn't really matter for RGB formats.
+  return data;
+}
+
+function clearAndVerifyColor(width, height, testR, testG, testB, value) {
+  var data = setupColor(testR, testG, testB, value);
+  // TODO(zmo): are we too tolerant here?
+  var tol = [36, 36, 36];
```
kenrussell: Not sure. These formats have 5- and 6-bit mantissas, respectively. Not sure how closely the given values can be represented. If these tests are passing on at least one driver then I think it is reasonable to assume they should pass on Mac OS as well.

kenrussell:
LGTM.

Richard-Yunchao:
I suppose that we should add this new test into 00_test_list.txt: https://github.com/KhronosGroup/WebGL/pull/1848

sdk/tests/conformance2/reading/format-r11f-g11f-b10f.html
Richard-Yunchao: The test case fails both on Mac Intel and Linux Nvidia when r/g/b equals to 8191. 

According to the spec of GLES 3.0, take unsigned 11bit floating point number as an example (see spec at https://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.1.3). The accuracy of different data range will vary. If the data is among 4096 ~ 8191, the exponent would be 27, the accuracy for every one step of mantissa would be 4096/64 = 64. Moreover, the driver will round to  the floor. So all data among 8128 to 8191 would round to 8128 (exponent E is 27, mantissa M is 63). So the deviation for 8191 would be 8191 - 8128 = 63, which is bigger than the tolerant number you set. So the case fails. 

I would like to list some possible biggest deviation for input data for unsigned 11bit floating point number: 
unsigned 11bit floating point number       possbile biggest deviation
     ....                                                             ...
1024 - 2047                                                   15
2048 - 4095                                                   31
4096 - 8191                                                   63
8192 - 16383                                                 127
   ...                                                                  ...

I tried the data 2047, 4095, 8191, 16383 respectively. The result conforms to my deduction.

Richard-Yunchao: I try to fix the bug at https://github.com/KhronosGroup/WebGL/pull/1850. Could you have a look? @zhenyao , @kenrussell .

zhenyao: Thanks for the analysis and CL.

As for adding to test list, I purposely left it out because we are not going to include this for WebGL 2.0.0.

PR #1848, opened 2016-07-07
add format-r11f-g11f-b10f.html into test list

Richard-Yunchao:
PTAL.


zhenyao:
I purposely exclude this as we won't be able to pass this in WebGL 2.0.0.

kenrussell:
Let's close this. Perhaps we should add a comment in 00_test_list.txt indicating that it's deliberately excluded though.

PR #1849, opened 2016-07-07
Allow line-continuation character in middle of line

qkmiao:
Remove invalid test cases according discussion in
https://github.com/KhronosGroup/WebGL/pull/1836.


zhenyao:
lgtm

PR #1850, opened 2016-07-07
Set tolerance for format r11f-g11f-b10f to fix bug in format-r11f-g11f-b10f.html.

Richard-Yunchao:
PTAL. Thanks a lot!

### DIFF
--- a/sdk/tests/conformance2/reading/format-r11f-g11f-b10f.html
+++ b/sdk/tests/conformance2/reading/format-r11f-g11f-b10f.html
@@ -69,10 +69,29 @@ function setupColor(testR, testG, testB, value) {
   return data;
 }
 
+// The definition of <Unsinged 11-Bit Floating-Point Number> in GLES 3.0.4:
+// https://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.1.3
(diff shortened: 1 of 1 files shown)


zhenyao:
Great work! Thanks for the improvement!

sdk/tests/conformance2/reading/format-r11f-g11f-b10f.html
```diff
@@ -95,25 +114,34 @@
   clearAndVerifyColor(width, height, true, false, false, 1000);
   clearAndVerifyColor(width, height, true, false, false, 2500);
   clearAndVerifyColor(width, height, true, false, false, 5000);
-  clearAndVerifyColor(width, height, true, false, false, 8191);
-  clearAndVerifyColor(width, height, true, false, false, 8192);
```
zhenyao: Just curious: why do you remove 8192?  As that value leads to Infinity on Mac rather than NaN, so slightly different.

Richard-Yunchao: Sorry, I am not aware of the slight difference between Inf and NaN on Mac when I run this pr. Glad to see that you have added 8192 into the test values. @zhenyao

PR #1852, opened 2016-07-07
Update WebGL 2 spec.

zhenyao:
To include tex upload from pbo functions and readPixels from unpack buffer.

@kenrussell @jdashg PTAL


kenrussell:
Excellent. Thanks for remembering to re-run the IDL generation script.

PR #1853, opened 2016-07-07
Add two Mac driver bug tests to 2.0.1+

zhenyao:
Let's mark it as 2.0.1 for now (in case we forget about them)

@kenrussell PTAL
@Richard-Yunchao FYI


kenrussell:
Clever. :) LGTM

PR #1854, opened 2016-07-07
Finally

Kangz:
PTAL, this is the actual fix for fbodepthbuffer and framebufferblit.depth_stencil.

Basically when doing a depth compare of 0.5 and 0.5 in the tests, what was actually happening was that we were comparing a depth of 0.5 for the fragment and 0.5 + epsilon for the depth buffer because it was quantized.

To fix this we convert the incoming fragment buffer depth to the depthbuffer format and back, making it incur the same loss of precision.

The rrUtil change is a noop cleanup I did thinking it might have been the root cause of this bug.

### DIFF
--- a/sdk/tests/deqp/framework/referencerenderer/rrFragmentOperations.js
+++ b/sdk/tests/deqp/framework/referencerenderer/rrFragmentOperations.js
@@ -196,7 +196,18 @@ rrFragmentOperations.executeDepthCompare = function(inputFragments, depthFunc, d
                 var fragSampleNdx = 0;
                 var depthBufferValue = depthBuffer.getPixDepth(fragSampleNdx, frag.pixelCoord[0], frag.pixelCoord[1]);
                 var sampleDepthFloat = frag.sampleDepths[fragSampleNdx];
-                var sampleDepth = deMath.clamp(sampleDepthFloat, 0.0, 1.0);
+
+                /* convert input float to target buffer format for comparison */
+                var access = new tcuTexture.PixelBufferAccess({
+                    format: depthBuffer.getFormat(),
+                    width: 1,
+                    height: 1,
+                    depth: 1,
+                    data: new ArrayBuffer(8)
+                });
+                access.setPixDepth(sampleDepthFloat, 0, 0, 0);
+                var sampleDepth = access.getPixDepth(0, 0, 0);
+
                 frag.depthPassed = expression(sampleDepth, depthBufferValue);
             }
         }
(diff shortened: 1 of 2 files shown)


sdk/tests/deqp/framework/referencerenderer/rrFragmentOperations.js
```diff
@@ -196,7 +196,18 @@ rrFragmentOperations.executeDepthCompare = function(inputFragments, depthFunc, d
                 var fragSampleNdx = 0;
                 var depthBufferValue = depthBuffer.getPixDepth(fragSampleNdx, frag.pixelCoord[0], frag.pixelCoord[1]);
                 var sampleDepthFloat = frag.sampleDepths[fragSampleNdx];
-                var sampleDepth = deMath.clamp(sampleDepthFloat, 0.0, 1.0);
+
+                /* convert input float to target buffer format for comparison */
```
Kangz: Matches https://android.googlesource.com/platform/external/deqp/+/master/framework/referencerenderer/rrFragmentOperations.cpp#194

sdk/tests/deqp/framework/referencerenderer/rrUtil.js
```diff
@@ -91,7 +88,17 @@ goog.scope(function() {
             ctx.bindBuffer(gl.ARRAY_BUFFER, null);
         }
 
-        ctx.drawQuads(gl.TRIANGLES, 0, 6);
+        {
```
Kangz: Matches https://android.googlesource.com/platform/external/deqp/+/master/framework/opengl/simplereference/sglrContextUtil.cpp#97

zhenyao:
This is great work!  LGTM
Let's merge and roll to see if this would cause any regressions in other tests (I reverted a CL on this a while ago due to causing a regression in another test).

sdk/tests/deqp/framework/referencerenderer/rrFragmentOperations.js
```diff
@@ -196,7 +196,18 @@ rrFragmentOperations.executeDepthCompare = function(inputFragments, depthFunc, d
...
+                var access = new tcuTexture.PixelBufferAccess({
+                    format: depthBuffer.getFormat(),
+                    width: 1,
+                    height: 1,
+                    depth: 1,
+                    data: new ArrayBuffer(8)
+                });
```
kenrussell: Could the allocation of the PixelBufferAccess object be hoisted outside this loop so allocations aren't performed per-pixel?

kenrussell:
Awesome work Corentin! The optimization can wait until it's known whether this patch causes any regressions. Thanks for tracking this down!

zhenyao:
@Kangz This seems to break fbodepthbuffer.html, but I think this is the right direction to move forward, so let's not revert this.  Can you take a look at the failure cases?

Kangz:
Which platforms does it break? It passed up to driver bugs on NVIDIA Linux.

zhenyao:
On Win AMD / NVidia.  Mac bots aren't running, but at least it's failing on my Mac AMD.

Kangz:
See #1856 for a follow up fix.

PR #1856, opened 2016-07-08
rrFragmentOutput: fix depth compare for floats

Kangz:
A previous commit forced the fragment depth to be converted to the depth
buffer format. The behavior didn't match the C++ dEQP code for float
formats where the incoming fragment depth is just clamped.

This also reduces the amount of allocation done when we convert to the depthbuffer format.

### DIFF
--- a/sdk/tests/deqp/framework/referencerenderer/rrFragmentOperations.js
+++ b/sdk/tests/deqp/framework/referencerenderer/rrFragmentOperations.js
@@ -189,6 +189,21 @@ rrFragmentOperations.executeDepthCompare = function(inputFragments, depthFunc, d
     /**
      * @param {function(number=,number=):boolean} expression
      */
+    var convertToDepthBuffer = false;
+
+    var access;
+    if (depthBuffer.getFormat().type != tcuTexture.ChannelType.FLOAT &&
+        depthBuffer.getFormat().type != tcuTexture.ChannelType.FLOAT_UNSIGNED_INT_24_8_REV) {
+        access = new tcuTexture.PixelBufferAccess({
+            format: depthBuffer.getFormat(),
+            width: 1,
+            height: 1,
+            depth: 1,
+            data: new ArrayBuffer(8)
+        });
+        convertToDepthBuffer = true;
+    }
+
     var sample_register_depth_compare = function(expression) {
       for (var i = 0; i < inputFragments.length; i++) {
             var frag = inputFragments[i];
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/framework/referencerenderer/rrFragmentOperations.js
```diff
@@ -197,16 +212,14 @@ rrFragmentOperations.executeDepthCompare = function(inputFragments, depthFunc, d
...
-                access.setPixDepth(sampleDepthFloat, 0, 0, 0);
-                var sampleDepth = access.getPixDepth(0, 0, 0);
+                var sampleDepth;
+                if (convertToDepthBuffer) {
+                    /* convert input float to target buffer format for comparison */
+                    access.setPixDepth(sampleDepthFloat, 0, 0, 0);
+                    sampleDepth = access.getPixDepth(0, 0, 0);
+                } else {
+                    sampleDepth = deMath.clamp(sampleDepthFloat, 0.0, 1.0);
+                }
```
kenrussell: Looking a little more closely at the C++ code:
https://android.googlesource.com/platform/external/deqp/+/deqp-dev/framework/referencerenderer/rrFragmentOperations.cpp#195

it looks like the floating-point sample depth is always stored using the PixelBufferAccess, and then it's fetched as an unsigned int for the depth comparison. With the current code structure in the JavaScript port, does it make sense to try to mimic this structure? It is technically feasible to reinterpret the bits of a floating-point number as an integer by storing into a Float32Array and reading back from the same ArrayBuffer via Uint32Array.

kenrussell:
LGTM; excellent work. One question; if the answer is that the code is correct, please merge yourself. Thanks.

sdk/tests/deqp/framework/referencerenderer/rrFragmentOperations.js
Kangz: There are two macros in this function and this [line](https://android.googlesource.com/platform/external/deqp/+/deqp-dev/framework/referencerenderer/rrFragmentOperations.cpp#212) chooses between the two. Float depth buffer formats use a different macro than the one implemented in #1854, which is what this PR implements.

Kangz:
Thank you for the review!

sdk/tests/deqp/framework/referencerenderer/rrFragmentOperations.js
kenrussell: Ah I see. Thanks for pointing that out.

zhenyao:
Great!

PR #1857, opened 2016-07-08
Define a few undefined behaviors related to DrawBuffers.

zhenyao:
Also add a few test cases for that.

### DIFF
--- a/sdk/tests/conformance/extensions/webgl-draw-buffers.html
+++ b/sdk/tests/conformance/extensions/webgl-draw-buffers.html
@@ -54,6 +54,11 @@ void main() {
     }
 }
 </script>
+<script id="fshaderNoWrite" type="x-shader/x-fragment">
+#extension GL_EXT_draw_buffers : require
+void main() {
+}
+</script>
 <script id="fshaderRed" type="x-shader/x-fragment">
 precision mediump float;
 void main() {
@@ -109,6 +114,7 @@ var canvas = document.getElementById("canvas");
 var output = document.getElementById("console");
 var gl = wtu.create3DContext(canvas);
 var ext = null;
+var programWithMaxDrawBuffersEqualOne = null;
 
 var extensionConstants = [
   { name: "MAX_COLOR_ATTACHMENTS_WEBGL", enum: 0x8CDF, expectedFn: function(v) { return v >= 4; }, passMsg: " should be >= 4"},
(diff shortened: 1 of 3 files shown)


zhenyao:
@Kangz @kenrussell @jdashg @grorg @RafaelCintron PTAL

sdk/tests/conformance/extensions/webgl-draw-buffers.html
```diff
@@ -300,6 +305,14 @@
   wtu.checkCanvas(gl, [0, 255, 0, 255], "should be green");
   gl.deleteProgram(program);
   wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no errors");
+
+  debug("");
+  debug("test that gl_MaxDrawBuffers is frozen at link time and enabling extenion won't change it.");
```
kenrussell: typo: extenion -> the extension

specs/latest/2.0/index.html
```diff
@@ -2809,6 +2809,20 @@ <h2 class="no-toc">Table of contents</h2>
...
+    <p>
+        If a fragment shader writes to none of <code>gl_FragColor</code>, <code>gl_FragData</code>, nor any
+        user-defined output variables, the values of the fragment colors following shader execution are untouched.
+        If some, but not all user-defined output variables are written, the values of fragment colors corresponding
+        to unwritten variables are similarly untouched.
+    </p>
+
+    <p>
+        If the values written by the fragment shader do not match the format(s) of the corresponding
+        color buffer(s), and the color buffer states are not set to <code>NONE</code> by <code>DrawBuffers</code>,
```
kenrussell: Is it worth adding some non-normative text with an example of this mismatch? For example, clearly stating that the shader attempts to write an integer value to a floating-point color buffer, or vice versa?

kenrussell:
LGTM from my standpoint. A couple of comments.

specs/latest/2.0/index.html
zhenyao: Good suggestion. Done.

sdk/tests/conformance/extensions/webgl-draw-buffers.html
zhenyao: Done

zhenyao:
Addressed kbr's comments.  Merging now.

Others, specifically @jdashg , if you feel differently, we can revise later.

PR #1858, opened 2016-07-08
Move builtinprecision/* tests to WebGL 2.0.1.

zhenyao:
@jdashg @kenrussell @Kangz PTAL

### DIFF
--- a/sdk/tests/deqp/functional/gles3/00_test_list.txt
+++ b/sdk/tests/deqp/functional/gles3/00_test_list.txt
@@ -1,4 +1,4 @@
-builtinprecision/00_test_list.txt
+--min-version 2.0.1 builtinprecision/00_test_list.txt
 draw/00_test_list.txt
 fbocolorbuffer/00_test_list.txt
 fboinvalidate/00_test_list.txt


kenrussell:
For more background on this decision:

 - These tests take over an hour to run in their current form, which is too long.
 - There are many failures on various GPUs and operating systems; requiring these tests to pass before shipping WebGL 2.0 in its initial form will delay shipment by weeks.
 - It's unclear how much incremental value for the average developer these tests will add. Nonetheless, we will aim to enable them in the next iteration. In WebGL 1.0, the shader tests from the OpenGL ES suite weren't incorporated until much later (version 1.0.2).

LGTM. Merging now.

kenrussell:
CC: @Richard-Yunchao @qkmiao

Richard-Yunchao:
Make sense. 

BTW, the builtin precision related tests take a long time in native too. This means that It is hard to shorten the run time they take. So we still need some mechanism like sharding (What kbr is doing in Chromium) in version 2.0.1.

jdashg:
I guess it's a game of semantics, but I don't think pushing these to 2.0.1 is the right idea.

That they take ~1hr to run shouldn't really matter from a conformance standpoint, since we only *need* to run the full conformance suite for conformance submissions. 

I have similar complaints about many of the deqp tests, but I don't think moving those is a good idea either.

FWIW we run neither of these on our CI right now, since they generally aren't going to catch regressions that aren't caught elsewhere.

zhenyao:
The reason I moved it out is they catch many shader function driver bugs that we won't be able to workaround easily and we should not block WebGL 2 on it.

The reason I put it as 2.0.1 is I hope we can work with driver vendors to get them fixed when we are ready to snapshot 2.0.1, but we could push them further down the road map if we need to.

Also, these tests need some optimization, so any help is greatly appreciated.

kenrussell:
Moving these tests out to 2.0.1 was already discussed, and agreed upon, within the working group. We will discuss this again in Thursday's conference call.

PR #1859, opened 2016-07-08
Remove pragma_vertex from deqp/data/gles2/shaders/preprocessor.test.

kenrussell:
This test is in the skip list in the native dEQP, and fails on Mac OS
on all GPU types. Remove it and leave a comment behind.

BUG=chromium:625363


kenrussell:
@zhenyao please review.

This will allow http://crbug.com/625363 to be finally closed, and this test enabled on all platforms.

zhenyao:
lgtm

PR #1860, opened 2016-07-08
Add missing compressedTex{Sub}Image{2|3}D functions.

zhenyao:
Just realized I missed these functions.

@kenrussell PTAL
@jdashg FYI


kenrussell:
LGTM. Thanks for noticing and cleaning these up.

PR #1861, opened 2016-07-08
Clarify a few undefined behaviors in WebGL 2 spec.

zhenyao:
@kenrussell @jdashg @grorg @RafaelCintron PTAL

Once this is approved, I'll add test cases.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -1846,6 +1846,32 @@ match the <em>type</em> according to the <a href="#TEXTURE_PIXELS_TYPE_TABLE">ab
         Set the rate at which the vertex attribute identified by <code>index</code> advances when drawing.
       </dd>
       <dt>
+        <p class="idl-code">void drawArrays(GLenum mode, GLint first, GLsizei count)
+          <span class="gl-spec">
+            (<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.9.3">OpenGL ES 3.0.4 &sect;2.9.3</a>,
+            <a href="http://www.khronos.org/opengles/sdk/docs/man3/html/glDrawArrays.xhtml" class="nonnormative">man page</a>)
+          </span>
+        </p>
+      </dt>
+      <dd>
+        If an active vertex attribute's base type does not match the <code>vertexAttrib</code> function that specifies its value, an
+        <code>INVALID_OPERATION</code> error will be generated. See <a href="#ATTRIBUTE_TYPE_MATCH">VertexAttrib function must
+        match shader attribute type</a>.
+      </dd>
+      <dt>
+        <p class="idl-code">void drawElements(GLenum mode, GLsizei count, GLenum type, GLintptr offset)
+          <span class="gl-spec">
+            (<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.9.3">OpenGL ES 3.0.4 &sect;2.9.3</a>,
+            <a href="http://www.khronos.org/opengles/sdk/docs/man3/html/glDrawElements.xhtml" class="nonnormative">man page</a>)
+          </span>
+        </p>
+      </dt>
+      <dd>
+        If an active vertex attribute's base type does not match the <code>vertexAttrib</code> function that specifies its value, an
+        <code>INVALID_OPERATION</code> error will be generated. See <a href="#ATTRIBUTE_TYPE_MATCH">VertexAttrib function must
+        match shader attribute type</a>.
+      </dd>
+      <dt>
         <p class="idl-code">void drawArraysInstanced(GLenum mode, GLint first, GLsizei count, GLsizei instanceCount)
           <span class="gl-spec">
             (<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.9.3">OpenGL ES 3.0.4 &sect;2.9.3</a>,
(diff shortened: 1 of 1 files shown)


specs/latest/2.0/index.html
```diff
@@ -3192,6 +3192,37 @@ <h2 class="no-toc">Table of contents</h2>
...
+    <p>
+        If <code>vertexAttribI4ui</code> or <code>vertexAttribI4uiv</code> is used and the base type of
+        the shader attribute at slot <code>index</code> is floating-point or signed integer,
+        generates an <code>INVALID_OPERATION</code> error.
+    </p>
+    <p>
+        If <code>vertexAttribI4i</code> or <code>vertexAttribI4iv</code> is used and the base type of
+        the shader attribute at slot <code>index</code> is floating-point or unsigned integer,
+        generates an <code>INVALID_OPERATION</code> error.
+    </p>
```
kenrussell: I don't see how it will be possible to enforce these; the vertex attribute state is part of that of the bound vertex array object, not the program object. Therefore, calls to UseProgram or BindVertexArray will change this state, so an error can't be generated upon call to vertexAttrib*. Am I missing something?

specs/latest/2.0/index.html
```diff
@@ -3192,6 +3192,37 @@ <h2 class="no-toc">Table of contents</h2>
...
+
+    <h3>Transform Feedback Primitive Capture</h3>
+
+    <p>
+        If any output variable is specified to be streamed to a buffer object but not actually
+        written by a vertex shader, the buffer remains untouched.
+    </p>
```
kenrussell: Could you point to the section of the ES 3.0 spec that leaves this undefined? Is it feasible to guarantee this behavior? How would it be done?

specs/latest/2.0/index.html
```diff
@@ -3192,6 +3192,37 @@ <h2 class="no-toc">Table of contents</h2>
...
+
+    <h3>gl_FragDepth</h3>
+
+    <p>
+        If a fragment shader statically assigns a value to <code>gl_FragDepth</code>, for any fragment
+        where statements assigning a value to gl_FragDepth are not executed, the value 0 is used.
+    </p>
```
kenrussell: Could you point to the ES 3.0 spec text that leaves this behavior undefined? How will this be handled? By assigning to gl_FragDepth at the beginning of the shader, and then leaving the user's shader untouched (so that if gl_FragDepth is conditionally assigned, it gets the 0 value assigned at the beginning)?

kenrussell:
Thanks for putting this together. A few questions.

specs/latest/2.0/index.html
zhenyao: You are right.  I didn't think this through.  It seems the check needs to happen at draw time (!)

specs/latest/2.0/index.html
zhenyao: Spec reference added.  You are right.  It seems impossible to leave the corresponding buffer ranges untouched if drivers behave differently.  Let's set it to 0 instead.

specs/latest/2.0/index.html
zhenyao: Spec reference added.  Yes, assigning it to 0 at the beginning should work.

specs/latest/2.0/index.html
```diff
@@ -3202,6 +3202,60 @@ <h2 class="no-toc">Table of contents</h2>
         <code>FLOAT_32_UNSIGNED_INT_24_8_REV</code>, generates an <code>INVALID_ENUM</code>.
     </p>
 
+    <h3>VertexAttrib function must match shader attribute type</h3>
+
+    <p>
+        If any of the following situations are true, a draw call generates an <code>INVALID_OPERATION</code> error:
```
kenrussell: Perhaps clarify to "calls to drawArrays and drawElements". This should also be linked from higher up in the spec, at the point of definition of those APIs.

specs/latest/2.0/index.html
```diff
@@ -3202,6 +3202,60 @@ <h2 class="no-toc">Table of contents</h2>
...
+        (<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.8">
+        OpenGL ES 3.0.4 &sect;2.8</a>)</span>.
+    </div>
+
+    <h3>Transform Feedback Primitive Capture</h3>
+
+    <p>
+        If any output variable is specified to be streamed to a buffer object but not actually
+        written by a vertex shader, the value is set to 0.
+    </p>
```
kenrussell: Could you link to this "h3" from section 3.7.15 "Transform feedback"?

specs/latest/2.0/index.html
kenrussell: Thanks.

specs/latest/2.0/index.html
zhenyao: Done

specs/latest/2.0/index.html
zhenyao: Done.

zhenyao:
Revised.  @kenrussell Please take another look.

zhenyao:
Note that I added two tests to verify the two changes here:

https://github.com/KhronosGroup/WebGL/pull/1863
https://github.com/KhronosGroup/WebGL/pull/1864

specs/latest/2.0/index.html
```diff
@@ -3202,6 +3246,62 @@ <h2 class="no-toc">Table of contents</h2>
...
+        integer);
+      </li>
+      <li>
+        <code>vertexAttribI4ui</code> or <code>vertexAttribI4uiv</code> is used and the base type of
+        the shader attribute at slot <code>index</code> is floating-point or signed integer;
+      </li>
+      <li>
+        <code>vertexAttribI4i</code> or <code>vertexAttribI4iv</code> is used and the base type of
+        the shader attribute at slot <code>index</code> is floating-point or unsigned integer.
+      </li>
```
kenrussell: The case also needs to be handled of VertexAttribIPointer with floating-point vertex attributes, and VertexAttribPointer and integer vertex attributes.

kenrussell: Also, could you reference the spec section where this is left undefined? Section 2.8, "Vertex Specification", ES 3.0.4 page 24; and Section 2.9, "Vertex Arrays", page 26.

kenrussell: ...sorry, I missed where you did this for section 2.8 already below. Could you add the VertexAttrib*Pointer cases and reference Section 2.9 too?

kenrussell:
Thanks for pushing this forward; sorry, I missed something the first time through.

specs/latest/2.0/index.html
zhenyao: That is actually unclear from the spec.  Because if we want to distinguish int / uint, we will have to look into the function parameter |type|, not just the function type.  This feels the right thing, only it's not backed up by the ES3 spec. What do you feel?

kenrussell: It's the same basic problem, and is backed up by the ES 3.0 spec. The values loaded into the shader attribute are undefined if the array for it was *not* specified by:
 - VertexAttribPointer for floating-point base type attributes;
 - VertexAttribIPointer with type BYTE, SHORT, or INT for signed integer vertex attributes;
 - VertexAttribIPointer with type UNSIGNED_BYTE, UNSIGNED_SHORT, and UNSIGNED_INT for unsigned integer base type attributes.

zhenyao: Done.

kenrussell:
Excellent. LGTM

PR #1863, opened 2016-07-11
Add a test to verify that an active attrib's base type

zhenyao:
needs to match the function type that sets its value.

This is a WebGL2 specific behavior.
See https://github.com/KhronosGroup/WebGL/pull/1861

### DIFF
--- a/sdk/tests/conformance2/rendering/attrib-type-match.html
+++ b/sdk/tests/conformance2/rendering/attrib-type-match.html
@@ -1,0 +1,484 @@
+﻿<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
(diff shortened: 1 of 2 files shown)


zhenyao:
@kenrussell Please review.

kenrussell:
LGTM - but per comment on #1861 , VertexAttribPointer and VertexAttribIPointer should be tested too. (These are more commonly used than the single-value initialization functions.)

zhenyao:
I updated the test to include the vertexAttrib{I}Pointer cases.  Please take another look.

sdk/tests/conformance2/rendering/attrib-type-match.html
```diff
@@ -0,0 +1,484 @@
...
+                                  -1, -2,
+                                  -1, -2,
+                                  -1, -2,
+                                  -1, -2,
+                                  -1, -2]);
+    gl.bufferData(gl.ARRAY_BUFFER, offsetI, gl.STATIC_DRAW);
+
+    var offsetUBuffer = gl.createBuffer();
+    gl.bindBuffer(gl.ARRAY_BUFFER, offsetUBuffer);
+    var offsetU = new Int32Array([1, 2,
```
kenrussell: This could be Uint32Array - not a big deal.

zhenyao: Right.  It's a typo bug. I'll fix it.

kenrussell:
The new tests LGTM. Thanks for including them. One minor comment. Please feel free to merge yourself.

PR #1864, opened 2016-07-11
Add a test to verify that transformfeedback unwritten output

zhenyao:
defaults to zero.

Note that this test fails in Linux NVidia with random output values.

This is WebGL2 specific behavior.
See https://github.com/KhronosGroup/WebGL/pull/1861

### DIFF
--- a/sdk/tests/conformance2/transform_feedback/00_test_list.txt
+++ b/sdk/tests/conformance2/transform_feedback/00_test_list.txt
@@ -1,1 +1,2 @@
 transform_feedback.html
+unwritten-output-defaults-to-zero.html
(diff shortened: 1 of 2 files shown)


zhenyao:
@kenrussell Please review.

kenrussell:
Excellent. LGTM. Please merge yourself once the other pull request is merged.

Richard-Yunchao:
@zhenyao , should we do sth in Chromium (and in other browsers) to get the correct result? By default, the value for any output variable specified to be streamed to a buffer object but not actually written by a vertex shader is undefined.

zhenyao:
That needs to be handled in ANGLE's shader translator, where we initialize all output variables to 0.

PR #1865, opened 2016-07-11
Fix missing entry in WebGL 2.0 test list

Oletus:
compound-assignment-type-combination test for WebGL 2 was missing from
the test runner test list.


zhenyao:
lgtm

PR #1866, opened 2016-07-12
Update DrawBuffers related behaviors (regarding unwritten output variables)

zhenyao:
@kenrussell @grorg @jdashg @RafaelCintron Please take a look.

### DIFF
--- a/extensions/WEBGL_draw_buffers/extension.xml
+++ b/extensions/WEBGL_draw_buffers/extension.xml
@@ -74,10 +74,12 @@
         If the WEBGL_draw_buffers extension is enabled, but the fragment shader does not contain the <code>#extension GL_EXT_draw_buffers</code> directive to enable it, then writes to <code>gl_FragColor</code> are only written to <code>COLOR_ATTACHMENT0_WEBGL</code>, and not broadcast to all color attachments. In this scenario, other color attachments are guaranteed to remain untouched.
       </addendum>
       <addendum>
-        If a fragment shader writes to none of <code>gl_FragColor</code>, <code>gl_FragData</code>, the values of
-        the fragment colors following shader execution are untouched. If some, but not all <code>gl_FragData</code>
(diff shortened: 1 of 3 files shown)


extensions/WEBGL_draw_buffers/extension.xml
```diff
@@ -74,10 +74,12 @@
...
-        If a fragment shader writes to none of <code>gl_FragColor</code>, <code>gl_FragData</code>, the values of
-        the fragment colors following shader execution are untouched. If some, but not all <code>gl_FragData</code>
-        output variables are written, the values of fragment colors corresponding to unwritten variables are
-        similarly untouched.
+        If a fragment shader writes to neither <code>gl_FragColor</code> nor <code>gl_FragData</code>, the values of
+        the fragment colors following shader execution are untouched.
+
+        If a fragment shader contains the <code>#extension GL_EXT_draw_buffers</code> directive, all
+        <code>gl_FragData</code> variables (from <code>gl_FragData[0]</code> to <code>gl_FragData[MAX_DRAW_BUFFERS_WEBGL - 1]</code>)
+        defaults to zero if no values are written to them during a shader execution.
```
kenrussell: typo: defaults -> default

kenrussell:
Thanks for revising this Mo. LGTM

Others: Mo and I discovered that the other way of resolving this undefined behavior was very difficult to understand and explain, and involved extensive static analysis. This alternative, of always initializing user-defined output variables to 0 and leaving other attachments untouched, should be easily implementable in the shader translator and WebGL implementations, and is easy to explain. It should also have no, or minimal, compatibility impact. (There's a chance that applications using WEBGL_draw_buffers and relying on particular undefined behavior may be impacted.)

extensions/WEBGL_draw_buffers/extension.xml
zhenyao: Done.

PR #1867, opened 2016-07-12
WebGL 2.0 allows line-continuation character

qkmiao:


sdk/tests/conformance/misc/invalid-passed-params.html
```diff
@@ -143,6 +144,9 @@
 debug("");
 debug("Test shaderSource() with invalid characters");
 for (var i = 0; i < invalidSet.length; ++i) {
+  // Backslash as line-continuation is allowed in WebGL 2.0.
+  if (contextVersion > 1 && invalidSet[i] == '\\')
+    continue;
```
zhenyao: Instead of adding logic here, you should only add '\\' to the invalidSet if contextVersion > 1.  That's a cleaner way.

qkmiao: I want '\\' in variable names can be tested by other part in the file. Here only tests invalid shader source.

zhenyao: I see.

kenrussell:
LGTM after the fact.

PR #1868, opened 2016-07-12
Add a test for tricky loop conditions

Oletus:
Some types of transformations in shader translation may need to add
new statements to implement certain types of expressions. In case an
expression inside a loop condition or loop expression requires this
kind of a transformation, it needs special handling. Add a test for
bugs in this area.

There are known failures in this test on ANGLE's DirectX backend and
NVIDIA's OpenGL drivers.

### DIFF
--- a/sdk/tests/conformance2/glsl3/00_test_list.txt
+++ b/sdk/tests/conformance2/glsl3/00_test_list.txt
@@ -24,5 +24,6 @@ shader-with-1025-character-identifier.frag.html
 shader-with-invalid-characters.html
 short-circuiting-in-loop-condition.html
 texture-offset-out-of-range.html
+tricky-loop-conditions.html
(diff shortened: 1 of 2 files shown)


zhenyao:
Thanks for the test cases. LGTM

Kangz:
Stop adding tests that fail on NVIDIA Linux or we'll never get to 100 % passing! :P

zhenyao:
We should welcome tests that exposes driver bugs or implementation bugs.  Don't worry, all such driver bugs that are non-critical will be put under --min-version=2.0.1.

kenrussell:
@Kangz was joking. :) These are great test cases Olli and they'll help ensure consistency of WebGL 2.0 / ES 3.0. Please keep them coming. LGTM after the fact.

PR #1869, opened 2016-07-12
Add a test for incrementing array elements

Oletus:
This hits a bug in NVIDIA 364.19 OpenGL drivers. The bug is specific
to arrays that are initialized when they are declared, so it does not
affect ESSL 1.00.


Oletus:
The bug should be fixed in newer drivers, though I didn't verify this yet.

zhenyao:
lgtm

kenrussell:
Very nice. LGTM too after the fact.

PR #1872, opened 2016-07-12
Update shader tests

Kangz:
@zhenyao PTAL, this fixes the preprocessor.html Chromium failures as well as part of the linkage tests. (the rest are addressed in the interface blocks CL).

### DIFF
--- a/sdk/tests/deqp/data/gles3/shaders/linkage.test
+++ b/sdk/tests/deqp/data/gles3/shaders/linkage.test
@@ -464,20 +464,16 @@ group varying "Varying linkage"
 			""
 		end
 
-		# different interpolation
-		case differing_interpolation_1
+		# interpolation_mismatch
+		case interpolation_mismatch_1
 			version 300 es
 			desc "varying interpolation different (flat vs. smooth)"
 			expect link_fail
-			values
-			{
-				input float in0		= [ -1.25 | -25.0 | 1.0 | 2.25 | 3.4 | 16.0 ];
-				output float out0	= [ -1.25 | -25.0 | 1.0 | 2.25 | 3.4 | 16.0 ];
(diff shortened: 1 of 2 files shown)


zhenyao:
Can you explain a bit why the change?  Every deviation from the native deqp should be documented (at least a comment in the test itself).

Kangz:
Added notes to the preprocessor tests that were not updated. The other changes to preprocessor.test are  upstream changes.

kenrussell:
LGTM assuming these match the upstream tests now but @zhenyao should re-review. Thanks for updating these tests.

zhenyao:
Thank you. LGTM.

sdk/tests/deqp/data/gles3/shaders/preprocessor.test
```diff
@@ -1647,7 +1647,6 @@ group conditional_inclusion "Conditional Inclusion Tests"
 
 	case basic_2
 		version 300 es
-		expect compile_fail
 		values { output float out0 = 1.0; }
```
qkmiao: These changes may violate current WebGL 2.0 spec, see https://github.com/KhronosGroup/WebGL/pull/1523.
FYI: @Oletus

sdk/tests/deqp/data/gles3/shaders/linkage.test
```diff
@@ -3494,43 +3607,6 @@ group uniform "Uniform linkage"
 	end
 
 	group block "Uniform blocks"
-		case precision_mismatch
```
qkmiao: Chrome still fails on this test. So there should be precision mismatch bugs for uniform blocks. Should we move it to conformance2 if it doesn't exist in native dEQP?

Issue #1870, opened 2016-07-12
D3D11 missing shadow sampler functions


vonture:
After investigation failing WebGL2 tests in ANGLE, I determined that a few sampler functions related to shadow samplers are missing from HLSL.  The missing functions are:

Vertex Shader: Any sampling from sampler2DShadow and samplerCubeShadow sampler. The corresponding HLSL function, [SampleCmp](https://msdn.microsoft.com/en-us/library/windows/desktop/bb509696.aspx), is not available in the vertex shader.  ANGLE currently fails shader compilation if these functions are used.

Fragment Shader: Sampling from a sampler2DShadow or samplerCubeShadow sampler using the textureLod, textureLodOffset, textureProjLod, textureGrad, textureProjGrad, textureProjGradOffset functions (pretty much any sampling function that uses Lod or Grad).  The HLSL SampleCmp function is not overloaded as much as the other HLSL sampling functions and does not include these parameters.  ANGLE currently ignores these parameters and does the equivalent of the regular texture function.

As far as I can tell, the only solution is to emulate this missing functionality by passing parameters of the sampler (compare mode and compare func) into the shader as uniforms and using a switch statement to perform the operations.  This work will be done but for the sake of pushing WebGL2 out faster, I propose that these tests are moved to the 2.0.1 suite.

In the mean-time, is there a preference for making the behaviour when using these functions consistent? Should ANGLE fail shader compilation in all of these cases or should the vertex shader case be updated to return some dummy value?

For reference, the failing webgl tests are:
deqp/functional/gles3/shadertexturefunction/texturelod.html
deqp/functional/gles3/shadertexturefunction/texturelodoffset.html
deqp/functional/gles3/shadertexturefunction/textureprojlod.html
deqp/functional/gles3/shadertexturefunction/textureprojlodoffset.html
deqp/functional/gles3/shadertexturefunction/texturegrad.html
deqp/functional/gles3/shadertexturefunction/texturegradoffset.html
deqp/functional/gles3/shadertexturefunction/textureprojgrad.html
deqp/functional/gles3/shadertexturefunction/textureprojgradoffset.html

And the specific failing deqp tests are:
dEQP-GLES3.functional.shaders.texture_functions.texturelod.sampler2dshadow_vertex
dEQP-GLES3.functional.shaders.texture_functions.texturelod.sampler2dshadow_fragment
dEQP-GLES3.functional.shaders.texture_functions.texturelodoffset.sampler2dshadow_vertex
[...]

PR #1873, opened 2016-07-12
Skip a bunch of shadow sampler function tests in deqp

zhenyao:
It's not supported in D3D11. Let's move them out of WebGL 2.0.0.

See https://github.com/KhronosGroup/WebGL/issues/1870

They should be added back for later versions of WebGL 2.


zhenyao:
@vonture Please review.

zhenyao:
@kenrussell @null77 @Kangz @jdashg FYI

vonture:
LGTM

kenrussell:
LGTM after the fact.

PR #1874, opened 2016-07-13
Interpolation qualifier in all linked shaders must match

qkmiao:
GLSL ES 3.00.4 specifies that "The type and presence of the
interpolation qualifiers and storage qualifiers and invariant qualifiers
of variables with the same name declared in all linked shaders must
match, otherwise the link command will fail." in Section 4.3.9 Interpolation.

The restriction for interpolation qualifier is relaxed to "It is a
link-time error if, within the same stage, the interpolation qualifiers
of variables of the same name do not match." in Section 4.5
Interpolation Qualifiers, GLSLangSpec 4.40.

### DIFF
--- a/sdk/tests/deqp/data/gles3/shaders/linkage.test
+++ b/sdk/tests/deqp/data/gles3/shaders/linkage.test
@@ -495,8 +495,12 @@ group varying "Varying linkage"
 
 		# different interpolation
 		case differing_interpolation_2
+			# This deviates from C++ dEQP here. GLSL ES 3.00.4 specifies interpolation
+			# qualifier of variables with the same name declared in all linked shaders
+			# must match, in Section 4.3.9 Interpolation.
 			version 300 es
 			desc "varying interpolation different (smooth vs. centroid)"
+			expect link_fail
 			values
 			{
 				input float in0		= [ -1.25 | -25.0 | 1.0 | 2.25 | 3.4 | 16.0 ];


sdk/tests/deqp/data/gles3/shaders/linkage.test
```diff
@@ -495,8 +495,12 @@ group varying "Varying linkage"
 
 		# different interpolation
 		case differing_interpolation_2
+			# This deviates from C++ dEQP here. GLSL ES 3.00.4 specifies interpolation
+			# qualifier of variables with the same name declared in all linked shaders
+			# must match, in Section 4.3.9 Interpolation.
 			version 300 es
 			desc "varying interpolation different (smooth vs. centroid)"
+			expect link_fail
```
qkmiao: C++ dEQP passed without link_fail:
https://android.googlesource.com/platform/external/deqp/+/deqp-dev/data/gles3/shaders/linkage.test#497

But according GLSL spec "Section 4.3.9 Interpolation", "The type and presence of the interpolation qualifiers and storage qualifiers and invariant qualifiers of variables with the same name declared in all linked shaders must match, otherwise the link command will fail."

So marked as link_fail here. The link_fail error is reported in command buffer before it was sent to driver. So it behaviors differently than C++.

qkmiao:
@kenrussell @zhenyao @Kangz please review.

sdk/tests/deqp/data/gles3/shaders/linkage.test
Kangz: I think this shouldn't fail because centroid is an auxiliary interpolation qualifier. See https://cs.chromium.org/chromium/src/third_party/angle/src/compiler/translator/ShaderVars.cpp?q=ShaderVars&sq=package:chromium&l=26

qkmiao: Thanks for pointing that. 
I just see your patch in angle should fix this bug: https://chromium-review.googlesource.com/#/c/359607/.  Close this pr.

PR #1877, opened 2016-07-13
Fix test using sequence operator on arrays

Kangz:


zhenyao:
@Oletus I thought this is legal?

Kangz:
It isn't in WebGL, see https://www.khronos.org/registry/webgl/specs/latest/2.0/#5.26

zhenyao:
Then instead of removing the test case, we should test the compile / link fails.

Kangz:
It is tested in https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/conformance2/glsl3/forbidden-operators.html#L78 and the translator fix for forbidden-operators had a failure in complex-indexing, leading to this PR.

zhenyao:
OK, thanks.

PR #1878, opened 2016-07-13
Revert changes to preprocessor.test

Kangz:
This is functionally equivalent to a revert of a82c23a05 but adds notes
on the expected to fail tests.

PTAL @Oletus @qkmiao @zhenyao


zhenyao:
LGTM

PR #1880, opened 2016-07-14
Print description for failed test case

gyagp:
opt_msg is ignored for failed case, so the description of it will have
no way to display. This PR fixes this so that we can easily find the
failed case with description in output.

### DIFF
--- a/sdk/tests/js/webgl-test-utils.js
+++ b/sdk/tests/js/webgl-test-utils.js
@@ -1160,7 +1160,7 @@ var checkCanvasRectColor = function(gl, x, y, width, height, color, opt_errorRan
  * @param {number} height width of region to check.
  * @param {!Array.<number>} color The color expected. A 4 element array where
  *        each element is in the range 0 to 255.
- * @param {string} opt_msg Message to associate with success. Eg
+ * @param {string} opt_msg Message to associate with success or failure. Eg
  *        ("should be red").
  * @param {number} opt_errorRange Optional. Acceptable error in
  *        color checking. 0 by default.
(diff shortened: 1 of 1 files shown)


gyagp:
For example, conformance/state/state-uneffected-after-compositing.html contains several test cases, and it fails at the 4th one. 
Before the PR, the output is as below:
PASS drawing with texture should be green
PASS drawing with texture after composite without rebinding should be green
PASS reading from fbo with attached texture should be green
FAIL at (0, 0) expected: 0,255,0,255 was 0,0,0,0

We only know it fails but the description of failed case doesn't show, which makes not straightforward to check the related place in source code. But actually, the case in source code is as below, and it already passes the description of case.
wtu.checkCanvasRect(gl, 0, 0, 1, 1, [0, 255, 0, 255], "reading from fbo after composite without rebinding should be green");

After the fix, the output in webpage is as  below:
PASS drawing with texture should be green
PASS drawing with texture after composite without rebinding should be green
PASS reading from fbo with attached texture should be green
FAIL reading from fbo after composite without rebinding should be green
at (0, 0) expected: 0,255,0,255 was 0,0,0,0

The description of failed case ("reading from fbo after composite without rebinding should be green") will display after "FAIL", so that we can easily find the place failed in source code.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -1174,7 +1174,12 @@ var checkCanvasRect = function(gl, x, y, width, height, color, opt_msg, opt_erro
           msg = "should be " + color.toString();
         testPassed(msg);
       },
-      testFailed,
+      function(differentMsg) {
+        var msg = opt_msg;
+        if (msg === undefined)
+          msg = "should be " + color.toString();
+          testFailed(msg + "\n" + differentMsg);
```
zhenyao: wrong indent here.

zhenyao:
LGTM with a format nit fixed. Thank you.

kenrussell:
Thanks @gyagp for this fix. LGTM too with Mo's comment addressed.

gyagp:
Sorry for the indention. It's fixed now.

zhenyao:
lgtm

PR #1881, opened 2016-07-14
Add program link info for debug purpose

qkmiao:
The log is only printed to console when test fails. This is useful for
debugging failing cases.


sdk/tests/deqp/framework/opengl/gluShaderProgram.js
```diff
@@ -229,6 +229,7 @@ gluShaderProgram.Program.prototype.link = function() {
     assertMsgOptions(this.gl.getError() == this.gl.NO_ERROR, 'gl.getProgramParameter()', false, true);
     this.info.linkOk = linkStatus;
     this.info.infoLog = this.gl.getProgramInfoLog(this.program);
+    bufferedLogToConsole("program linking: " + this.info.infoLog);
```
zhenyao: My concern is we will pollute the already too huge stdout in our bots.  Can we limit this to only if link fails?

qkmiao: Done. Thanks.

zhenyao:
lgtm

PR #1882, opened 2016-07-15
update attrib type in shader to match with attrib type set by vertexAttribI4{u}i{v}

Richard-Yunchao:
update the test cases, in order to make them conform to the spec: vertex attrib function must match shader attrib type: https://www.khronos.org/registry/webgl/specs/latest/2.0/#5.31

I'd like to submit this pr to fix a couple of test cases at first. Then merge more cases in one pr later.


Richard-Yunchao:
The bug was found by this Chromium patch: https://codereview.chromium.org/2148723004/. In addtion, there is a crbug to track this issue: https://bugs.chromium.org/p/chromium/issues/detail?id=628459. 

PTAL, thanks a lot!

zhenyao:
lgtm

PR #1883, opened 2016-07-15
Add some tolerance to PBO RGB10_A2 case

gyagp:
When using PBO, the result rounded by Intel GPU would have slight
difference with the one from CPU, but it still follows the spec well.
See detailed discussion at
https://bugs.freedesktop.org/show_bug.cgi?id=89314.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
+++ b/sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
@@ -425,8 +425,16 @@ var tcuImageCompare = framework.common.tcuImageCompare;
 
         var isOk = false;
 
-        if (floatCompare)
-            isOk = tcuImageCompare.floatThresholdCompare('Result comparison', 'Result of read pixels to memory compared with result of read pixels to buffer', readReference.getLevel(0), readResult, [0.0, 0.0, 0.0, 0.0]);
+        if (floatCompare) {
+            // The result of rgb10_a2 from Intel GPU would have slight difference with the one from CPU, so some tolerance is added here.
+            // Detailed discussion in Mesa upstream can be found at https://bugs.freedesktop.org/show_bug.cgi?id=89314.
+            var threshold;
+            if (tcuTestCase.runner.currentTest.name == 'rgb10_a2_triangles')
+                threshold = [0.004, 0.004, 0.004, 0.0];
+            else
+                threshold = [0.0, 0.0, 0.0, 0.0];
+            isOk = tcuImageCompare.floatThresholdCompare('Result comparison', 'Result of read pixels to memory compared with result of read pixels to buffer', readReference.getLevel(0), readResult, threshold);
+        }
         else
             isOk = tcuImageCompare.intThresholdCompare('Result comparison', 'Result of read pixels to memory compared with result of read pixels to buffer', readReference.getLevel(0), readResult, [0, 0, 0, 0]);
 


gyagp:
Only one case related to RGB10_A2 has problem within all the cases, and it also fails with native dEQP test. The error is reported as below:
Start testcase: pixel_buffer_object.renderbuffer.rgb10_a2_triangles
Image comparison failed: max difference = 0.0039215686274509665,0.0039215686274509665,0.0039215686274509665,0, threshold = 0,0,0,0

According to discussion from Mesa upstream at https://bugs.freedesktop.org/show_bug.cgi?id=89314, this is the expected behavior of Intel GPU when doing round with PBO, and also follows the spec. The actual result has just slight difference with the expected value, so I suggest to add some tolerance here.

zhenyao:
looks good.  can you add a comment in the code to include the mesa bug link?

zhenyao:
I confirm Mac drivers have the same precision issue. This CL fixes Mac failures too.

gyagp:
The Mesa issue link is added into code. Is the Mac based on Intel GPU?

zhenyao:
No, it's Mac Intel/NVidia/AMD.  LGTM

sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
```diff
@@ -425,8 +425,16 @@ var tcuImageCompare = framework.common.tcuImageCompare;
 
         var isOk = false;
 
-        if (floatCompare)
-            isOk = tcuImageCompare.floatThresholdCompare('Result comparison', 'Result of read pixels to memory compared with result of read pixels to buffer', readReference.getLevel(0), readResult, [0.0, 0.0, 0.0, 0.0]);
+        if (floatCompare) {
+            // The result of rgb10_a2 from Intel GPU would have slight difference with the one from CPU, so some tolerance is added here.
+            // Detailed discussion in Mesa upstream can be found at https://bugs.freedesktop.org/show_bug.cgi?id=89314.
+            var threshold;
+            if (tcuTestCase.runner.currentTest.name == 'rgb10_a2_triangles')
```
kenrussell: Is there any more robust way to check for this aside from looking at the test's name? Or is it only this exact test (and not the clear tests, for example, in this file) that need this workaround?

kenrussell:
Thanks for tracking this down and LGTM overall after the fact. One question.

sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
zhenyao: That's a good question.  I tested on Mac, actually clear test with the rgb10_a2 also fails on Mac (and this 0.004 threshold will fix that).  So maybe we should set the threshold depending on format instead?

kenrussell: That sounds reasonable. If you can reproduce the failure and fix, would you put up a pull request and I'll review?

zhenyao: Sure.

zhenyao: https://github.com/KhronosGroup/WebGL/pull/1887

gyagp:
I did the strict check on purpose as the case is the only one failed on Linux Intel, and in this way, we may catch regression timely. And it's also my thought during the patch that we may loose the restriction once we have more failures. So I'm definitely good with #1887.

PR #1885, opened 2016-07-15
Increase the threshold on the dEQP compressed_cube tests.

vonture:
ANGLE passes this test on native dEQP so I looked into the test source and found it uses a threshold of 17!  A threshold of 6 was enough for Chrome to pass the test on Windows and Mac.


zhenyao:
Can you add a comment indication this is for software decoder numerical error on Mac?

vonture:
Done.

zhenyao:
Thanks.

kenrussell:
Nice work. LGTM after the fact.

PR #1887, opened 2016-07-15
Expand threshold for RGB10_A2 format tests.

zhenyao:


zhenyao:
@kenrussell PTAL
@gyagp FYI

kenrussell:
Great. LGTM.

PR #1888, opened 2016-07-15
Fix clear() conflicts with integer format images in shaderderivate.

zhenyao:
@kenrussell PTAL

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fShaderDerivateTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fShaderDerivateTests.js
@@ -28,6 +28,7 @@ goog.require('framework.opengl.gluPixelTransfer');
 goog.require('framework.opengl.gluShaderProgram');
 goog.require('framework.opengl.gluShaderUtil');
 goog.require('framework.opengl.gluTexture');
+goog.require('framework.opengl.gluTextureUtil');
 goog.require('framework.common.tcuInterval');
 goog.require('framework.common.tcuFloat');
 goog.require('framework.common.tcuLogImage');
@@ -51,6 +52,7 @@ goog.scope(function() {
     var gluShaderProgram = framework.opengl.gluShaderProgram;
     var gluShaderUtil = framework.opengl.gluShaderUtil;
     var gluTexture = framework.opengl.gluTexture;
+    var gluTextureUtil = framework.opengl.gluTextureUtil;
     var tcuInterval = framework.common.tcuInterval;
     var tcuFloat = framework.common.tcuFloat;
     var tcuLogImage = framework.common.tcuLogImage;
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fShaderDerivateTests.js
```diff
@@ -647,7 +649,27 @@ goog.scope(function() {
         /** @type {Array<number>} */ var indices = [0, 2, 1, 2, 3, 1];
 
         gl.clearColor(0.125, 0.25, 0.5, 1.0);
-        gl.clear(gl.COLOR_BUFFER_BIT | gl.DEPTH_BUFFER_BIT | gl.STENCIL_BUFFER_BIT);
+        // We can't really call clear() on gl.COLOR_BUFFER_BIT here as in c++ deqp.
+        // The fbo format might be of integer type and WebGL2 requires an INVALID_OPERATION to be generated.
+        var formatObj = gluTextureUtil.mapGLInternalFormat(fboFormat);
+	var fmtClass = tcuTexture.getTextureChannelClass(formatObj.type);
```
kenrussell: Please unify the indentation before landing.

sdk/tests/deqp/functional/gles3/es3fShaderDerivateTests.js
```diff
@@ -647,7 +649,27 @@ goog.scope(function() {
...
+	switch (fmtClass) {
+            case tcuTexture.TextureChannelClass.FLOATING_POINT:
+            case tcuTexture.TextureChannelClass.SIGNED_FIXED_POINT:
+            case tcuTexture.TextureChannelClass.UNSIGNED_FIXED_POINT:
+                gl.clear(gl.COLOR_BUFFER_BIT | gl.DEPTH_BUFFER_BIT | gl.STENCIL_BUFFER_BIT);
+                break;
+            case tcuTexture.TextureChannelClass.UNSIGNED_INTEGER:
+                gl.clearBufferuiv(gl.COLOR, 0, new Uint32Array(4));
```
kenrussell: The value cleared to is zero instead of the clear color above. Is that the intent?

sdk/tests/deqp/functional/gles3/es3fShaderDerivateTests.js
```diff
@@ -647,7 +649,27 @@ goog.scope(function() {
...
+                gl.clear(gl.DEPTH_BUFFER_BIT | gl.STENCIL_BUFFER_BIT);
+                break;
+            case tcuTexture.TextureChannelClass.SIGNED_INTEGER:
+                gl.clearBufferiv(gl.COLOR, 0, new Int32Array(4));
```
kenrussell: Same problem here -- zeroing out the color buffer.

kenrussell:
LGTM overall. Couple of minor points. Please merge yourself once addressed.

sdk/tests/deqp/functional/gles3/es3fShaderDerivateTests.js
zhenyao: I don't know how to map to int and uint values.  The semantic meaning is unclear.  Also, I assume the C++ deqp behavior is leave the int and uint buffers untouched when call clear(), so its the same as clearing them to 0.  This is the least behavioral change - may not be the best behavior though.  Do you have suggestions?

kenrussell: The only question is whether the clear color is deliberately non-zero in order to put something that isn't solid black (or transparent black, in this case) into the back buffer.

No concrete suggestion. I might suggest multiplying the clear color above by 255 and putting those values into these Int32Arrays and Uint32Arrays, if the test still passes by doing so. Otherwise, your code as is looks fine.

zhenyao: I'll try that. Thanks.

zhenyao: Yeah they still pass

PR #1889, opened 2016-07-16
Add a test verifying clear and clearBuffer* func type should be

zhenyao:
compatible with fbo buffer types.

### DIFF
--- a/sdk/tests/conformance2/rendering/clear-func-buffer-type-match.html
+++ b/sdk/tests/conformance2/rendering/clear-func-buffer-type-match.html
@@ -1,0 +1,164 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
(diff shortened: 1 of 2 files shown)


zhenyao:
@kenrussell Can you review this? But leave it to me to merge. I want to land the Chromium fix first before rolling this in.

sdk/tests/conformance2/rendering/clear-func-buffer-type-match.html
```diff
@@ -0,0 +1,166 @@
...
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<canvas id="canvas" width="20" height="20"> </canvas>
+<script>
+"use strict";
+description("This tests WebGL2 specific contraint that clear or clearBuffer* functions have to be compatible to fbo's buffer format");
```
kenrussell: minor grammar: "tests WebGL2" -> "tests the WebGL2", "compatible to" -> "compatible with"
minor typo: "contraint" -> "constraint"

sdk/tests/conformance2/rendering/clear-func-buffer-type-match.html
```diff
@@ -0,0 +1,166 @@
...
+  debug("Float buffer");
+
+  colorbuffer = setupRenderbuffer(gl.COLOR_ATTACHMENT0, gl.RGBA8);
+  var ext = gl.getExtension("EXT_color_buffer_float");
+  if (ext) {
+    debug("EXT_color_buffer_float is available: testing RGBA32F");
+    colorbuffer1 = setupRenderbuffer(gl.COLOR_ATTACHMENT1, gl.RGBA32F);
+  } else {
+    debug("EXT_color_buffer_float is unavailable: testing RGBA8");
+    colorbuffer1 = setupRenderbuffer(gl.COLOR_ATTACHMENT1, gl.RGBA8);
```
kenrussell: I think that RGBA8 should always be tested, and RGBA32F conditionally tested only if EXT_color_buffer_float is available.

kenrussell:
LGTM with a couple of updates.

sdk/tests/conformance2/rendering/clear-func-buffer-type-match.html
zhenyao: Done

sdk/tests/conformance2/rendering/clear-func-buffer-type-match.html
zhenyao: It is.  I have two attachments. One is always RGBA8, the other is either RGBA32F or RGBA8, depending on the extension is available or not.

kenrussell: If RGBA8 is well tested already with these clearBuffer calls, then I would just make this entire test block conditional on the presence of the extension, rather than confusingly running the code with an RGBA8 attachment.

zhenyao: Done.

PR #1892, opened 2016-07-16
Output log to console only for failed tests

qkmiao:


zhenyao:
lgtm

kenrussell:
Thank you for fixing this!

PR #1894, opened 2016-07-16
More clearances in console log

qkmiao:


zhenyao:
LGTM
Thanks for cleaning up.

kenrussell:
Thanks for these cleanups -- LGTM after the fact.

PR #1895, opened 2016-07-16
Remove debug code in glsBuiltinPrecisionTests.js

qkmiao:
The console log should be clean now with this last patch.


qkmiao:
@zhenyao should we remove this debug code or output them when tests fail?

zhenyao:
Let's keep it.  We still need to work on these tests, at least to optimize running time.  This might come handy.

qkmiao:
OK. Close now.

PR #1896, opened 2016-07-16
Fix a typo in WebGL 2.0 spec

qkmiao:
@zhenyao I will try to add test on same image attach to two different attachment points in the same fbo and fix it for chrome.


zhenyao:
Thanks!

PR #1897, opened 2016-07-16
Add test to check framebuffer unsupported

qkmiao:
If an image is attached to more than one color attachment point in a
framebuffer, checkFramebufferStatus returns FRAMEBUFFER_UNSUPPORTED. See
Section 5.34 Framebuffer color attachments in WebGL 2.0 spec.


zhenyao:
lgtm

zhenyao:
I just realized this should also be part of the WebGL 1 WEBGL_draw_buffers behavior. We will need to update the extension spec and add this test case in WebGL 1 conformance.

PR #1898, opened 2016-07-17
Remove a few test cases from fbodepthbuffer due to NVidia Linux drive…

zhenyao:
…r bugs.

### DIFF
--- a/sdk/tests/deqp/framework/common/tcuSkipList.js
+++ b/sdk/tests/deqp/framework/common/tcuSkipList.js
@@ -142,7 +142,7 @@ goog.scope(function() {
         _skip("texture_functions.textureprojgradoffset.sampler2dshadow_vertex");
         _skip("texture_functions.textureprojgradoffset.sampler2dshadow_fragment");
 
-        _setReason("MacOSX drivers share namespaces where they shoudl not");
+        _setReason("MacOSX drivers share namespaces where they should not");
         // https://github.com/KhronosGroup/WebGL/issues/1890
         // deqp/data/gles3/shaders/scoping.html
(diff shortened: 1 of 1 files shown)


zhenyao:
@kenrussell @Kangz @jdashg FYI

Kangz:
LGTM after the fact, thanks for handling this!

jdashg:
Isn't finding these bugs what we have tests for?
Please revert.

@zhenyao @grorg @kenrussell @RafaelCintron

zhenyao:
We agreed not to block WebGL 2 release by non-critical driver bugs.  These tests are not removed permanently.  We should add a mechanism to track them and add them back as the driver bugs being fixed by drivers.

We can do 1) create issues to track them; 2) add a mechanism to let them still be run for future WebGL 2 versions, just like the --min-version flag.

zhenyao:
For what it's worth, tcuSkipList is a central place for all the skipped tests in deqp, so it's not like we will lose them in the stack of numerous test cases.

jdashg:
You cannot disable your way to victory over conformance. I stand against skipping tests, particularly ones as well-founded as this one.

Having failures in the conformance suite does not strictly-speaking block release. If it blocks yours, I recommend you reconsider that, rather than disable well-founded tests.

Let's discuss this on the WG call this Thursday.

zhenyao:
Let's discuss this on Thursday.  Just for your information, every entry we put into this skip list, we spent time studying and verifying it's indeed a driver bug, and isolating to the root cause of the wrong behavior. Then we file bugs against driver vendors. This is for the benefit of WebGL 2 in general.

PR #1899, opened 2016-07-18
Clarify output from #1889 when EXT_color_buffer_float isn't available.

kenrussell:


kenrussell:
Minor update - @zhenyao PTAL. Thanks.

zhenyao:
Thanks for fixing this. LGTM

PR #1893, opened 2016-07-16
Clear log in es3fFragmentOutputTests.js

qkmiao:

PR #1901, opened 2016-07-18
Fix a typo from https://github.com/KhronosGroup/WebGL/pull/1893

zhenyao:
@kenrussell Thank you for noticing this.
@qkmiao FYI.


kenrussell:
LGTM after the fact.

PR #1904, opened 2016-07-18
Fix the invariant(all) pragma in fragment shader

jchen10:
According to the section 4.6.1 of GLSL ES specification 3.00.4 and 4.8.1
of GLSL specification 4.40, the pragma invariant(all) is not allowed to
be used by a fragment shader.


jchen10:
@qkmiao Please take a look, Thanks!

zhenyao:
It's only illegal in GLSL ES 3, not 1.  So you should conditionally add this test for WebGL 1 only.

See https://github.com/KhronosGroup/WebGL/pull/1420 discussion.

jchen10:
@zhenyao Add the condition for webgl1/2，please take a look！

zhenyao:
lgtm

kenrussell:
Thanks for fixing this.

PR #1905, opened 2016-07-18
Fix conformance/context/context-lost.html to generate INVALID_OPERATI…

jdashg:
…ON for lost WebGLObjects.

The spec says that when a method on the context or an extension is called
that doesn't [WebGLHandlesContextLoss], first generate INVALID_OPERATION
if any object args have their 'invalidated flag' set. If either that, or
IsContextLost() is true, return defaults, etc.

The spec language does not appear to address what to do about methods
called on extensions which have been disabled/detached. This is not
covered in this test, though.

There's a couple things that need to be tightened up, really, but this will get the conversation started.

### DIFF
--- a/conformance-suites/1.0.3/conformance/context/context-lost.html
+++ b/conformance-suites/1.0.3/conformance/context/context-lost.html
@@ -138,9 +138,9 @@ function testValidContext()
     }
 }
 
-function testGLNOErrorFunctions(tests) {
+function testErrorFunctions(expectedError, tests) {
   tests.forEach(function(test) {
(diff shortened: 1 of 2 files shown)


kenrussell:
Thanks for submitting this for consideration.

The intent of the spec was that all methods should be short-circuited if the context is lost, and that the invalidated state of incoming WebGLObject arguments is only checked if the context is still live.

I've submitted pull request #1911 to clarify this. Please review it. The test should stand as is. The spec change could be backported to 1.0.3 if desired.

jdashg:
With the clarification, I agree that the test stands.

PR #1906, opened 2016-07-18
Deinit vertex attrib array state in InstancedRenderingCase

xhcao:

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fInstancedRenderingTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fInstancedRenderingTests.js
@@ -372,6 +372,15 @@ var deMath = framework.delibs.debase.deMath;
         }
     };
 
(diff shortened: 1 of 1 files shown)


Kangz:
LGTM too

zhenyao:
The fix looks good in itself.  I am curious, why this isn't an issue on some platforms but only on others?

xhcao:
@zhenyao 
In mesa+i965 platform:
case 1):     gl.getAttribLocation(program, 'a_position') return 12;
                 gl.getAttribLocation(program, 'a_instanceR')  return 0;
case 2):     gl.getAttribLocation(program, 'a_position') return 0;
But in nvidia platform, It seems that gl.getAttribLocation(program, 'a_position') are all return 0 in case 1) and case 2), so there is no overriding situation.
It seems that the distribution location values algorithm of mesa is different as nvidia.

zhenyao:
Thanks for the explanation.

PR #1907, opened 2016-07-18
Improve tricky loop conditions test

Oletus:
Using exact float comparisons makes some of the tests fail on D3D once
the other bugs are fixed, so add a little bit of tolerance to the
float comparisons.

WebGL has also made it illegal to use the sequence operator on arrays,
so remove the incorrect test for that.

One test description is also clarified.

### DIFF
--- a/sdk/tests/conformance2/glsl3/tricky-loop-conditions.html
+++ b/sdk/tests/conformance2/glsl3/tricky-loop-conditions.html
@@ -127,7 +127,7 @@ var doWhileLoopConditionTemplate = [
 
 var testDataList = [
 {
-    description: 'indexing a variable assignment',
+    description: 'indexing an array assignment',
     globalScopePrefix: '',
     mainPrefix: [
             'float a[2] = float[2](0.0, 0.0);',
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/glsl3/tricky-loop-conditions.html
```diff
@@ -138,7 +138,7 @@
     loopCondition: 'bool((c = (a = b)[0]) + 1.0)',
     loopContents: 'b[0] += 1.0;',
     loopIterations: 3,
-    passCondition: 'c == 4.0'
+    passCondition: 'abs(c - 4.0) < 0.01'
```
zhenyao: Maybe you can define a var tol = '0.01' and set it as passCondition: 'abs(c - 4.0) < ' + tol, something like that, here and below. So if we need to, we can easily adjust tolerance in the future.

zhenyao:
LGTM with one minor suggestion, but let's merge for now.

kenrussell:
Very nice. Thanks Olli for the fix.

PR #1891, opened 2016-07-16
Move a few non-critical driver bug tests to 2.0.1

zhenyao:

PR #1909, opened 2016-07-18
Split the NV driver bug from general vector-dynamic-indexing.html

zhenyao:
So we can enable most of the test cases in WebGL 2.0.0


zhenyao:
Addressing @kenrussell  concern in https://github.com/KhronosGroup/WebGL/pull/1891

kenrussell:
LGTM. Thanks for splitting this off.

PR #1903, opened 2016-07-18
Fix PixelStore tests according to the new WebGL2 constraints.

zhenyao:
This is @jdashg 's proposal. @jdashg made two valid points: 1) this is the most strict constraints, and it's always easier to loose constraints than to tighten them in the future; 2) unconstrained skip parameters usage always can be achieved through adjusting the data store offset.

Also, I verified deqp tests are all within this constraints, i.e., anything violating these constraints are not tested, therefore, bug prone.

Let's go with this for now.

@kenrussell @jdashg Please take a look.

All tests are modified according to the bots from this CL (https://codereview.chromium.org/2154153002/)

@Oletus @qkmiao @Richard-Yunchao @Kangz @null77 @vonture @grorg @RafaelCintron FYI

PR #1910, opened 2016-07-18
Fixed tiny typo from #1903 .

kenrussell:


kenrussell:
Trivial; merging myself.

PR #1911, opened 2016-07-18
Clarify context lost behavior of methods taking WebGLObjects.

kenrussell:
The intent of the specification was to short-circuit the
implementation of all methods when the context was lost. The
INVALID_OPERATION error for previously-invalidated WebGLObject
arguments is only supposed to be generated if the context is not lost.


kenrussell:

@RafaelCintron @grorg and others, comments welcome.

zhenyao:
Looks good to me, but let @jdashg take a look before merging.

jdashg:
Perfect.

PR #1912, opened 2016-07-18
Skip WebGL 2 blit tests mark as having issues in GLES3.

null77:
Some of these tricky tests were failing on various drivers and Hw,
so they were removed from the Android CTS. The tests themselves
might not be buggy, but they are not required to pass on Android.

PTAL, this should fix some failures on Windows, probably on other platforms.


zhenyao:
LGTM

@kenrussell , are you OK with this?  These are not part of the ToT native deqp anymore.

kenrussell:
Yes, LGTM too. Thanks for investigating the exact commits which introduced these suppressions.

PR #1913, opened 2016-07-18
Remove suppressions for preprocessor tests.

Kangz:
These have been fixed in ANGLE's preprocessor so they can now be removed.


kenrussell:
Excellent. Thanks for following up on this. LGTM.

PR #1908, opened 2016-07-18
Fix a bug in tex-input-validation.js

zhenyao:

PR #1914, opened 2016-07-18
Revert https://github.com/KhronosGroup/WebGL/pull/1908

zhenyao:
Sorry for the mistake, but if the destination format is unsized,
then we don't require component size to match, just that requiured
components are available.


zhenyao:
@kenrussell @null77 Please review.

This is a bug ANGLE needs to fix.

kenrussell:
OK. Sounds good though I didn't review the spec text and this particular case in detail.

PR #1915, opened 2016-07-18
Test the case in TexImage{2|3}D where no source data is provided.

zhenyao:
Unpack param constraints should not apply.


zhenyao:
@kenrussell Please review.

kenrussell:
Is this the behavior that's really desired? In this case, since the texture's supposed to be zeroed out, the browser has to allocate zero-filled memory and (probably) reset these unpack parameters. Would it be more consistent to have these cases fail?

zhenyao:
Browsers are controlling unpack params anyway for TexImage calls from dom elements, etc. So it should be easy to implement.

The current spec says it should not fail.  So the test is to make sure of that.  But it's no big deal.  Let's not merge and leave it untested.

kenrussell:
Sorry, I didn't mean to shoot down this pull request. Let's merge this after all. It's better to have the behavior defined and tested.

PR #1919, opened 2016-07-19
Fix framebuffer unsupported error for getFramebufferAttachmentParameter testing

qkmiao:


zhenyao:
LGTM
But let's wait until the Windows situation is clarified.  From the framebuffer-unsupported.html test you added, it seems like d3d does not have this limitation.  So we might be able to lift it from WebGL2 spec.

qkmiao:
Thanks. I see.  I will also hold off the chrome fix.

zhenyao:
Cleared with ANGLE folks.  The rule has to stay. Merging this.

PR #1920, opened 2016-07-19
Fix wrong message in unform-buffers.html

qkmiao:


zhenyao:
lgtm
Thank you for fixing this

PR #1921, opened 2016-07-19
Fix WebGL 2 skips for fbo.blit.

null77:
These were prefixed with 'fbo' when it seems they should start
directly at 'blit'.

PTAL, I think this is the correct syntax.


kenrussell:
LGTM after the fact. It looks like the grouping is a little different than the native dEQP (see FboTests in es3fFunctionalTests.cpp, which puts the FramebufferBlitTests under the "fbo" group), but you could also change the 'blit' to 'fbo.blit' in es3fFramebufferBlitTests.FramebufferBlitTests in es3fFramebufferBlitTests.js if you wanted to use the same names as the native suppression list.

PR #1922, opened 2016-07-19
Update ImageBitmap tests

xidachen:
Please refer to the description in this bug:
https://bugs.chromium.org/p/chromium/issues/detail?id=628954

Basically I found that without this change, an ImageBitmap can be used for texImage2D and then the same ImageBitmap will be used for texSubImage2D. This causes test to fail in the case where texImage2D does GPU-GPU texture copy, and texSubImage2D does GPU readback.

In the long run, we should find out why GPU readback fails after the GPU-GPU texture copy. In the short term, a simple fix is to create a new copy of that ImageBitmap for texSubImage2D.

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-with-image-bitmap-utils.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-with-image-bitmap-utils.js
@@ -163,14 +163,14 @@ function runTestOnBindingTargetImageBitmap(bindingTarget, program, bitmaps, opti
     internalFormat, pixelFormat, pixelType, gl, tiu, wtu)
 {
     var cases = [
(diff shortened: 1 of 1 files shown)


xidachen:
Please review. @zhenyao @kenrussell

zhenyao:
Xida, we've been practicing this rule that we don't change the test to make Chrome pass if the test is correct.

zhenyao:
If you want to make this change, at least you will need to add a separate small test case to catch this bug. So we can isolate the failures down to one test.

kenrussell:
Mo's correct on all points. It's a hard rule that changes are not made to the WebGL conformance tests to work around browser bugs.

xidachen:
Closing this pull request since we have a fix here: https://codereview.chromium.org/2165913003/. Thanks.

PR #1923, opened 2016-07-19
Skip some tests on which the HLSL compiler generates invalid code.

vonture:
Corentin (@Kangz) diagnosed these failures in ANGLE and determined that they are due to the HLSL compiler generating invalid bytecode.

His findings, along with a shader that reproduces them, can be found here: https://gist.github.com/Kangz/a0624ca46fb1da5e7804ee80900bf291


Kangz:
LGTM

kenrussell:
LGTM but will let @zhenyao merge.

Please make a plan to get these re-enabled; perhaps file a follow-on issue in this issue tracker for re-enabling them for WebGL 2.0.1?

PR #1925, opened 2016-07-20
Large image requests can always receive OOM.

jdashg:
For sizes <MAX_SIZE, they must not generate INVALID_VALUE, but OOM is fine. (and reasonable in many cases)

### DIFF
--- a/conformance-suites/1.0.3/conformance/textures/texture-size-limit.html
+++ b/conformance-suites/1.0.3/conformance/textures/texture-size-limit.html
@@ -155,9 +155,9 @@ function testFormatType(t, test) {
       var badOtherDimension = t.target == gl.TEXTURE_2D ? 1 : badSize;
       var pixels = new test.dataType(badSize * badOtherDimension * test.size);
       gl.texImage2D(target, level, test.format, size, otherDimension, 0, test.format, test.type, pixels);
-      wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no error for level: " + level + " " + size + "x" + otherDimension);
+      wtu.glErrorShouldBe(gl, [gl.NO_ERROR, gl.OUT_OF_MEMORY], "there should be no error for level: " + level + " " + size + "x" + otherDimension);
       gl.texImage2D(target, level, test.format, otherDimension, size, 0, test.format, test.type, pixels);
-      wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no error for level: " + level + " " + otherDimension + "x" + size);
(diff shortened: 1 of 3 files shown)


zhenyao:
lgtm

kenrussell:
This is problematic. The OpenGL state is undefined after a GL_OUT_OF_MEMORY error is generated, and for this reason, Chrome forces a context loss if such an error is generated.

Chrome also occasionally produces GL_OUT_OF_MEMORY errors running this test, and fails because all subsequent WebGL calls fail.

Instead of allowing GL_OUT_OF_MEMORY, can you think of another restructuring of the test that would avoid stressing the GPU's memory so much?

jdashg:
The crux of the issue you mention with OOM is that there are theoretically two OOMs: With and without state being preserved safely.

We technically have no insight as clients to the API, because GL lacks a mechanism for GL_EVERYTHING_IS_TERRIBLE, and overloads GL_OOM with that meaning.

Firefox relies on this OOM being issued in good faith during allocation operations.

Within WebGL we should probably make it a rule that state is always preserved, except for context loss. Then, force of context loss handles our EVERTHING_IS_TERRIBLE case, and OOM can be well-behaved. Firefox does not quite do this. There are a couple cases where we still generate OOM on internal error and safely null things out.


In short, spec-wise we need to accept OOM here. Whether we choose to lose the context after driver-generated OOMs is an implementation detail.

kenrussell:
I understand what you're saying.

The Chrome team has occasionally seen OUT_OF_MEMORY errors running this test. These are still flagged as errors in our testing setup. I'm concerned that if OUT_OF_MEMORY is allowed as a legal error condition for these tests -- even though the tests are stressful -- that it will admit poor quality WebGL implementations that can't even allocate one-dimensional textures up to MAX_TEXTURE_SIZE.

What if we instead changed the tests so that they re-allocate each mipmap level back to 1x1 after seeing whether they can allocate it at 1xN (where N is the legal, large, texture size)? That would reduce the memory pressure of these tests and still provide a strong requirement that implementations be able to allocate textures of that size.

kenrussell:
@jdashg can you comment on whether making the change I suggested above will allow these tests to be more strict, rather than always allowing OUT_OF_MEMORY errors as an acceptable alternative? It seems to me that the latter will allow poor quality WebGL implementations to ship, whether accidentally or deliberately.

dmikis:
A little bit off topic, but `GL_OUT_OF_MEMORY` can be a good signal for an application to dial down it's resource consumption (and, actually, try to wait for `webglcontextrestore` event). There is no such signal right now, AFAIK.

jdashg:
I can't know until we test it. That *should* fit, but I don't know as we
can expect this from all drivers.

jdashg:
Maybe we don't need this after #2970.

PR #1926, opened 2016-07-20
Add negative tests for invariance in ESSL 3.00 shaders.

kenrussell:
Follow-up to #1924 .


kenrussell:
@zhenyao PTAL.

@jdashg @qkmiao @jchen10 @null77 @Kangz @vonture FYI. I don't anticipate these will cause any problems. They already pass in Chrome and I'm quite sure ANGLE's shader translator is catching the errors, so they should pass in Firefox too.

kenrussell:
Hold off merging this for a second -- forgot the title.

kenrussell:
Ready for review.

zhenyao:
LGTM

PR #1927, opened 2016-07-20
invariant(all) is not allowed in essl3 frag shaders.

jdashg:
Maybe this should go in conformance2, but it's useful to keep it in one place.

### DIFF
--- a/sdk/tests/conformance/glsl/misc/shaders-with-invariance.html
+++ b/sdk/tests/conformance/glsl/misc/shaders-with-invariance.html
@@ -48,6 +48,16 @@ void main()
     gl_Position = v_varying;
 }
 </script>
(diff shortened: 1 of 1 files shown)


kenrussell:
Thanks for adding these tests. I just uploaded https://github.com/KhronosGroup/WebGL/pull/1926 which tests this case, as well as attempted uses of "invariant" with input variables in ESSL 3.00 fragment shaders, which is also disallowed.

Let's keep the ESSL 3.00 cases together. Would you review my other pull request instead?

kenrussell:
#1926 was just merged, so let's close this. Sorry for the collision.

jdashg:
I have some more tests to add, so I'll submit a rebased PR.

PR #1928, opened 2016-07-20
fix the bug that caused by difference between webgl and gles

Richard-Yunchao:
This patch delete one line from https://github.com/KhronosGroup/WebGL/pull/1906. 

Calling to vertexAttribPointer for all attribs in native deqp during deinit is valid. However, It is not valid in webgl according to the latest spec: https://www.khronos.org/registry/webgl/specs/latest/2.0/#5.31. 

PTAL. Thanks a lot!

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fInstancedRenderingTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fInstancedRenderingTests.js
@@ -376,7 +376,6 @@ var deMath = framework.delibs.debase.deMath;
         var numVertexAttribArrays = /** @type{number} */ (gl.getParameter(gl.MAX_VERTEX_ATTRIBS));
         for (var idx = 0; idx < numVertexAttribArrays; idx++) {
             gl.disableVertexAttribArray(idx);
-            gl.vertexAttribPointer(idx, 4, gl.FLOAT, gl.FALSE, 0, null);
             gl.vertexAttribDivisor(idx, 0);
         }
     };


sdk/tests/deqp/functional/gles3/es3fInstancedRenderingTests.js
```diff
@@ -376,7 +376,6 @@ var deMath = framework.delibs.debase.deMath;
         var numVertexAttribArrays = /** @type{number} */ (gl.getParameter(gl.MAX_VERTEX_ATTRIBS));
         for (var idx = 0; idx < numVertexAttribArrays; idx++) {
             gl.disableVertexAttribArray(idx);
-            gl.vertexAttribPointer(idx, 4, gl.FLOAT, gl.FALSE, 0, null);
```
Richard-Yunchao: Here we set all attributes format to GL_FLOAT by vertexAttribPointer during deinit. 
If some attributes are INT or UINT in shader, and one of them is not used at all (no vertexAttrib API to specify data), so according to the spec, it will fail during drawArray/drawElements and their variants. Because vertexAttrib's type doesn't not match with attrib type in shader. (They don't know vertexAttribPointer call sites are used to deinit variables). 
If no vertexAttribPointer is called during deinit, It is OK. 

Besides, the API vertexAttribPointer here is different from that in native deqp(https://android.googlesource.com/platform/external/deqp/+/master/framework/opengl/gluStateReset.cpp#64). Because in WebGL, the last argument is always an offset, it will specify buffer data to variable in shader. This is not a de-initiation at all.  
However, in native deqp, the last argument in vertexAttribPointer means a NULL pointer because ARRAY_BUFFER is bound to 0 during deinit (https://android.googlesource.com/platform/external/deqp/+/master/framework/opengl/gluStateReset.cpp#52), so the data specified to the variable in shader will be unbind. This is really a de-initiation.

Richard-Yunchao:
@zhenyao and @kenrussell , please help to review. Thanks a lot!
@xhcao , FYI.

zhenyao:
I don't understand this.  Since we also call disableVertexAttribArray in deinit, vertexAttribPointer should not matter in the next test, unless it enable it again? If it's enabled, then you have to bind vertex array to it at that time, and the vertexAttribPointer call at deinit is overwritten.

Richard-Yunchao:
Seems that the spec don't state this clearly if vertexAttribArray is disabled then call to vertexAttribPointer. The chromium gpu bots of Mac differs from the behavior of that on Windows/Linux. I will take a look into it.

kenrussell:
I don't understand what's going on either, but this call doesn't do what's intended. The WebGL IDL for vertexAttribPointer takes an integer offset as its last argument, not an object (or WebGLBuffer), and it doesn't allow calling the function when no WebGLBuffer is bound to the ARRAY_BUFFER binding point. Essentially, unless there's a buffer bound, this call in deinit() will generate an INVALID_OPERATION.

Perhaps it should be removed, but it should be understood what's going on first.

zhenyao:
Agreed with kbr.  This doesn't do what it does in native code.  Let's remove this. However, your CL failing the test is still a bug, which I already pointed out in the CL.

PR #1929, opened 2016-07-20
Add tests for CUBE and 3D texture in framebuffer-unsupported.html

qkmiao:
@zhenyao Please review. This adds tests as your comments in https://codereview.chromium.org/2159183002/#msg5.


zhenyao:
lgtm
Thanks

kenrussell:
Thanks for adding these tests. LGTM too after the fact.

PR #1930, opened 2016-07-20
Only log to the console in the image bitmap tests if the test fails.

kenrussell:


kenrussell:
@zhenyao please review.

@xidachen FYI.

zhenyao:
lgtm

PR #1931, opened 2016-07-21
The issue related with interpolation value range of dfdx

xhcao:


zhenyao:
lgtm
Thank you for tracking this down.

kenrussell:
Very nice. Thank you too for tracking this down.

PR #1932, opened 2016-07-21
Add more tests for sequence operator corner cases

Oletus:
1. Check that all short-circuiting operators work as the second
operand of a sequence operator.

2. Check that a sequence operator with indexing an array constructor
inside works inside a loop condition or loop expression.

3. Check that dynamic indexing of vectors as an l-value works inside
a sequence operator.


zhenyao:
LGTM

Thank you for testing all these cases.

Oletus:
Thanks for merging! A fix for one issue exposed by these tests also already landed in ANGLE.

PR #1933, opened 2016-07-22
error to reattach COLOR_ATTACHEMENT0 in fbo completeness …

xhcao:
Although seems that it does not resolve a issue.

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsFboCompletenessTests.js
+++ b/sdk/tests/deqp/modules/shared/glsFboCompletenessTests.js
@@ -927,7 +927,7 @@ goog.scope(function() {
                 throw new Error('Multiple attachments not supported');
             var maxAttachments = gl.getParameter(gl.MAX_COLOR_ATTACHMENTS);
 
-            for (var i = 0; i < maxAttachments; ++i) {
+            for (var i = 1; i < maxAttachments; ++i) {
                 this.attachTargetToNew(gl.COLOR_ATTACHMENT0 + i, this.m_params.colornKind,
                                        glsFboUtil.ImageFormat.none(), 64, 64, builder, gl);
             }


kenrussell:
LGTM comparing to the C++ code but let's let @zhenyao review and merge.

zhenyao:
LGTM

But why is it a problem to reattach the image? It seems to me harmless. 

I am merging now because we should track c++ deqp, but please help me understand.

xhcao:
@zhenyao It only attaches COLOR_ATTACHEMENT0 repeatedly, I also think it is not a problem.

zhenyao:
This caused fbocompleteness.html to fail on MacOSX.

Can you dig a little and see what's going on there?

xhcao:
@zhenyao , we will investigate the root cause.    
FYI @Richard-Yunchao

Richard-Yunchao:
Without this change, The test can pass on Mac. After applied this change. it failed. I can reproduce this issue on Mac Intel. 

I will take a look.

Richard-Yunchao:
Got the root cause now. This change is correct. It conforms to the native one, and conform to what it want to test. 
The root cause is the same with framebuffer-completeness-unaffected.html. In MacOS, when the target of  readBuffer / drawBuffers have no image attached, it will report FRAMEBUFFER_MISSING_ATTACHMENT. This is a specific issue for MacOS.

sdk/tests/deqp/modules/shared/glsFboCompletenessTests.js
```diff
@@ -927,7 +927,7 @@ goog.scope(function() {
                 throw new Error('Multiple attachments not supported');
             var maxAttachments = gl.getParameter(gl.MAX_COLOR_ATTACHMENTS);
 
-            for (var i = 0; i < maxAttachments; ++i) {
+            for (var i = 1; i < maxAttachments; ++i) {
                 this.attachTargetToNew(gl.COLOR_ATTACHMENT0 + i, this.m_params.colornKind,
```
Richard-Yunchao: The failed cases are none_rbo_none_none and none_tex_none_none. For these two test cases, parameter this.m_params.colornKind would be RENDERBUFFER / TEXTURE, but this.m_params.color0Kind (called by another attachTargetToNew on line 922#) would be 0, it will not attach any image to COLOR_ATTACHMENT0 at line 922#. So if the iteration begins from 1, instead of 0, the COLOR_ATTACHEMNT0 would have no image to attach. 

But readBuffer's target is COLOR_ATTACHMENT0 by default, So it fails on MacOS, because readBuffer's target has no image attached. 

If you simply add one line code: gl.readBuffer(gl.COLOR_ATTACHMENT1), it can pass.

zhenyao:
Thank you Yunchao for getting the root cause of this.

kenrussell:
What will the resolution be for this issue on Mac OS? Does Chrome need to add a driver bug workaround? What is the general form of this workaround?

zhenyao:
Yes, Chrome will need to work around this driver bug (and we should file the bug to Apple as soon as we can).

The workaround is we will have to mask the READ_BUFFER or DRAW_BUFFERs to NONE if the images are null.

PR #1934, opened 2016-07-22
Moved WEBGL_compressed_texture_s3tc_srgb to draft after discussion in…

kenrussell:
… working group.


kenrussell:
Working group members: I re-read the spec and confirmed there's no need to establish any dependencies between this and the EXT_sRGB extension wrapper in WebGL. If this extension is available, it can be enabled.

Merging now.

PR #1935, opened 2016-07-22
Relax EXT_color_buffer_float extension to allow multisample renderbuf…

kenrussell:
…fers.

Per discussion in WebGL working group. Firefox already allows this. The conformance tests need to be updated. An issue will be filed about doing so.


kenrussell:
As discussed in the working group. It's much simpler to simply allow this as optional functionality, like the native extension does, as opposed to adding yet another extension.

Please review. @grorg @jdashg @msc- @zhenyao @RafaelCintron

jdashg:
Cool, thanks.

PR #1937, opened 2016-07-22
Add getBufferSubDataAsync.

kenrussell:
Per discussion in the WebGL working group, this asynchronous variant
will impose significantly less overhead in WebGL implementations with
deep graphics pipelines, since it does not require the pipeline to be
drained.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -705,6 +705,9 @@ interface <dfn id="WebGL2RenderingContextBase">WebGL2RenderingContextBase</dfn>
   // can not be exposed safely to JavaScript. GetBufferSubData
   // replaces it for the purpose of fetching data back from the GPU.
   void getBufferSubData(GLenum target, GLintptr offset, ArrayBuffer? returnedData);
+  // Asynchronous version of getBufferSubData which fulfills the returned promise when the data becomes available.
+  Promise&lt;ArrayBuffer&gt; getBufferSubDataAsync(GLenum target, GLintptr offset, ArrayBuffer returnedData,
+                                             unsigned long returnedOffset = 0, unsigned long returnedSize = 0);
 
   /* Framebuffer objects */
   void blitFramebuffer(GLint srcX0, GLint srcY0, GLint srcX1, GLint srcY1, GLint dstX0, GLint dstY0,
@@ -1083,7 +1086,7 @@ WebGL2RenderingContext implements WebGL2RenderingContextBase;
         beyond the end of the buffer an <code>INVALID_VALUE</code> error is generated.
         If <code>returnedData</code> is null then an <code>INVALID_VALUE</code> error is generated.
         If <code>offset</code> is less than zero, an <code>INVALID_VALUE</code> error is generated.
-        If zero is bound to <code>target</code>, an <code>INVALID_OPERATION</code> error is generated.
+        If no WebGLBuffer is bound to <code>target</code>, an <code>INVALID_OPERATION</code> error is generated.
         If <code>target</code> is <code>TRANSFORM_FEEDBACK_BUFFER</code>, and any transform feedback object
         is currently active, an <code>INVALID_OPERATION</code> error is generated.
         If any error is generated, no data is written to <code>returnedData</code>.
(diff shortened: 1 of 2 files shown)


kenrussell:
Working group members, please review: @grorg @jdashg @zhenyao @RafaelCintron in particular.

@junov this is the first WebGL method which returns a Promise. Any feedback on the design is appreciated.

Community feedback welcome as well.

This version returns its errors in "DOM style" by rejecting the Promise, rather than in OpenGL or WebGL style. For this reason the passed ArrayBuffer is mandated to be non-null.

I would ideally like to add conformance tests for this new entry point before merging this, but per discussion, here is the strawman proposal.

kenrussell:
Note: had to fix the whitespace errors to avoid breaking the new lint script. Also, the removed comments in webgl2.idl occurred simply because it wasn't regenerated during the most recent spec updates.

Kangz:
The way the extension is specified, there will be JS garbage created. Can we have a version of it with an explicit offset and length for where to return the data in the return array? This will be consistent with our effort to add such entry points to the data upload functions.

specs/latest/2.0/index.html
```diff
@@ -1092,6 +1094,52 @@ <h2 class="no-toc">Table of contents</h2>
...
+        <ul>
+
+        <br><br>
+
+        then the returned Promise is rejected with an "InvalidStateError" DOMException, and no data
+        is written to <code>returnedData</code>.
+
+        <br><br>
+
+        Otherwise, the Promise is fulfilled when the data becomes available,
```
junov: I think there possible asynchronous failures that need to be covered here, which should probably result in the promise being rejected (timeout, context lost)?

specs/latest/2.0/index.html
```diff
@@ -1092,6 +1094,52 @@ <h2 class="no-toc">Table of contents</h2>
         it is the responsibility of the WebGL API to ensure that data are accessed consistently. This applies
         even if the buffer is currently bound to a transform feedback binding point.
       </dd>
+      <dt>
+        <p class="idl-code">Promise&lt;ArrayBuffer&gt; getBufferSubDataAsync(GLenum target, GLintptr offset, ArrayBuffer returnedData)
```
junov: I am not sure that passing in "returnedData" is the best idea for the async version of this API. This can easily lead to programming errors such as race conditions (apps touching returnedData before the promise has resolved) and memory leaks (apps not handling promise rejection adequately).  What are the downsides of having this API allocate its own ArrayBuffer?

specs/latest/2.0/index.html
```diff
@@ -1092,6 +1094,52 @@ <h2 class="no-toc">Table of contents</h2>
...
+        bytes from the buffer starting at byte offset <code>offset</code> and write them
+        to <code>returnedData</code>. This command executes serially with respect to others in the
+        command stream, but the data is only written to <code>returnedData</code> when the promise
+        is fulfilled.
+
+        <br><br>
+        If any of the following error conditions occur:
+
+        <ul>
+          <li><code>offset + returnedData.byteLength</code> would extend beyond the end of the buffer</li>
```
junov: "the buffer" is not clear.   Perhaps you should formally define an identifier somewhere above.  Something like ` Let <var>buffer</var> be the WebGLBuffer bound to <code>target</code>.`

specs/latest/2.0/index.html
kenrussell: Good point. Will handle context lost in the next push of this pull request. Let me know if you find other failure modes that need to be covered. (Or should there be a catch-all?)

specs/latest/2.0/index.html
kenrussell: I think it's important for the application to allocate its own destination ArrayBuffer. It's been found that garbage generation is prohibitive for some applications, and there's already work in progress to eliminate garbage production for sophisticated applications (see #1940 ).

Race conditions are solved by only performing the write to returnedData when resolving the Promise. When the Promise callback is called, the data is available, but not before.

specs/latest/2.0/index.html
kenrussell: Will clarify this in the next update.

kenrussell:
@Kangz good point about the garbage generation. Will address this in the next update. This pull request should be considered to follow @jdashg 's #1940 .

specs/latest/2.0/index.html
junov: Catch-all and specific at the same time, perhaps? Something like: If a failure occurs that prevents the promise from being fullfilled, such as a context loss, ...

kenrussell: Thanks, good suggestion. Will incorporate this in the next version.

kenrussell:
Thanks for previous feedback. Updated the description; please re-review. Thanks.

specs/latest/2.0/index.html
```diff
@@ -1092,6 +1095,63 @@ <h2 class="no-toc">Table of contents</h2>
...
+          <li><code>offset + numBytes</code> would extend beyond the end of <code>buffer</code></li>
+          <li><code>returnedOffset + numBytes</code> would extend beyond the end of <code>returnedData</code>
+          <li><code>target</code> is <code>TRANSFORM_FEEDBACK_BUFFER</code>, and any transform feedback object
+            is currently active</li>
+          <li>the context is lost before the Promise is fulfilled</li>
+        <ul>
+
+        <br><br>
+
+        then the returned Promise is rejected with an "InvalidStateError" DOMException, and no data
```
junov: In this sentence, it is not clear whether the Promise is already rejected when it is returned, or whether the rejection happens later (async).  IMHO, the rejection should be synchronous for all conditions, except for context loss.  This should be clear in the spec.

specs/latest/2.0/index.html
```diff
@@ -1092,6 +1095,63 @@ <h2 class="no-toc">Table of contents</h2>
...
+
+        <ul>
+        <li> If <code>returnedSize</code> is zero, <code>numBytes = returnedData.byteLength - returnedOffset</code></li>
+        <li> Otherwise, <code>numBytes = returnedSize</code>.
+        </ul>
+
+        This command executes serially with respect to others in the command stream, but the data is
+        only written to <code>returnedData</code> when the promise is fulfilled; it is available in
+        the Promise's success callback, but not earlier.
+
```
zhenyao: Should we specify if the data should be when the call is made, or when the promis is fulfilled, or anything in between? Since it's an async call, after this call, we can modify the content of the buffer.

zhenyao: Never mind.  I see you clarified this below.

zhenyao:
LGTM

specs/latest/2.0/index.html
```diff
@@ -1092,6 +1095,63 @@ <h2 class="no-toc">Table of contents</h2>
...
+        Otherwise, the Promise is fulfilled when the data becomes available,
+        and <code>returnedData</code>, originally passed to <code>getBufferSubDataAsync</code>, is
+        passed to the success callback.
+
+        <br><br>
+        If the buffer is written and read sequentially by other operations
+        and <code>getBufferSubDataAsync</code>, it is the responsibility of the WebGL API to ensure
+        that data are accessed consistently. This applies even if the buffer is currently bound to a
+        transform feedback binding point.
+
```
RafaelCintron: Implementation question: Suppose a web developer gives an ArrayBuffer to this API and then turns around and reads and writes to the buffer.  When the GPU thread is fulfilling the request, I presume it is going to fulfill it into its own separate buffer and then copy the separate buffer into the JavaScript buffer right before fulfilling the promise?  Otherwise, how else are we going to avoid partially available data before the promise is fulfilled?  I presume this is what you mean by "it is the responsibility of the WebGL API to ensure the data are accessed consistently"?  

Along a similar line, if the WebGL developer uses the same buffer for two calls to the async API, it is guaranteed that the promise for the first call will be fulfilled before the promise for the second call?

kenrussell: Yes, the WebGL implementation is responsible for ensuring that the returned data isn't visible until just before the promise is fulfilled. This does probably mean reading back into a separate buffer, and then copying it into this destination buffer on the main thread just before resolving the promise.

It's a good question about the guarantees if doing multiple asynchronous readbacks into the same buffer. For sanity's sake, I think that yes, the first Promise must be resolved before the second one. Do you foresee any problems with providing this guarantee?

specs/latest/2.0/index.html
kenrussell: Added a clarification. Let me know what you think.

specs/latest/2.0/index.html
```diff
@@ -1092,6 +1095,64 @@ <h2 class="no-toc">Table of contents</h2>
...
+          <li><code>returnedOffset + numBytes</code> would extend beyond the end of <code>returnedData</code>
+          <li><code>target</code> is <code>TRANSFORM_FEEDBACK_BUFFER</code>, and any transform feedback object
+            is currently active</li>
+          <li>the context is lost before the Promise is fulfilled</li>
+        <ul>
+
+        <br><br>
+
+        then the returned Promise is rejected with an "InvalidStateError" DOMException, and no data
+        is written to <code>returnedData</code>. The Promise is rejected synchronously for all kinds
```
junov: Nit: "kinds of" is superfluous IMHO

junov:
lgtm

specs/latest/2.0/index.html
kenrussell: Removed.

kainino0x:
This should closed in favor of #2079. The new spec text and the tests are there.

kenrussell:
Thanks @kainino0x for picking this up. Closing in favor of #2079 .

PR #1941, opened 2016-07-25
add some test cases to test generic attrib

Richard-Yunchao:
Add some test cases to cover generic attribs when check the vertex type by vertexAttrib API with the corresponding type in shader. 

Fortunately, add these new added test cases can pass after the Chromium CL https://codereview.chromium.org/2148723004/ applied. 

@zhenyao , PTAL.

### DIFF
--- a/sdk/tests/conformance2/rendering/attrib-type-match.html
+++ b/sdk/tests/conformance2/rendering/attrib-type-match.html
@@ -63,6 +63,42 @@ void main() {
 }
 </script>
 
+<script id='vshader_inactive_attrib' type='x-shader/x-vertex'>#version 300 es
+in ivec4 p;
+in ivec4 a;
+void main()
+{
+    gl_Position = vec4(p);
+}
+</script>
+<script id='vshader_active_attrib_int' type='x-shader/x-vertex'>#version 300 es
+in ivec4 p;
+in ivec4 a;
+in uvec4 b;
+void main()
+{
+    gl_Position = vec4(p) + vec4(a) + vec4(b);
+}
+</script>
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/rendering/attrib-type-match.html
```diff
@@ -115,11 +216,8 @@
     }
     wtu.glErrorShouldBe(gl, gl.NO_ERROR, "No GL error from set up");
 
-    debug("");
-    debug("Test vertexAttrib with drawArrays and drawArraysInstanced");
     wtu.setupUnitQuad(gl, 0);
 
-    debug("Correct setup");
```
zhenyao: why do you remove this line?  I think this indicates the following test is a positive test, whereas all other cases are negative test.  This line makes the test case clearer.

zhenyao:
LGTM
Nice test cases

sdk/tests/conformance2/rendering/attrib-type-match.html
Richard-Yunchao: Ah... I just moved the code 118 ~ 119 to line 201 ~ 203, to show that here begins the other part, in addition to my new added part (generic attrib). 

So the "correct setup“ seems to be unnecessary because there is a "No GL error from set up" on line 116/217...

zhenyao: Maybe "correct setup" isn't clear enough.  It probably should say "draw call should succeed with correct setup."

Richard-Yunchao: Agreed!

PR #1918, opened 2016-07-19
Enable Lint tool on Travis CI

Honry:

PR #1876, opened 2016-07-13
Add a lint tool for catching common mistakes in test files

renchenglei:
@kenrussell @zhenyao @Honry I modified the lint tool to make it more suitable for WebGL. In WebGL conformance tests, most of lint errors are TRAILING WHITESPACE, INDENT TABS and CR AT EOL. So we removed non-existent lint error scansion and ran the tool without installing libraries. Since there are many \*.js, \*.frag and \*.vert files in WebGL tests, we also scanned this three file formats in our new tool. You can try the tool from the root of your local WebGL/sdk/tests working directory as follow:
**If you want to check your pull request:**
`./py/lint/lint.py -p`
**If check specified folder, the specified folder must be relative path of WebGL/sdk/tests:**
`./py/lint/lint.py -d conformance/attribs`
**If check other repos, you can specify the repo name with `-r`, default is WebGL/sdk/tests:**
`./py/lint/lint.py -r demo-express`
Looking forward to your suggestion, thanks a lot!

PR #1942, opened 2016-07-25
add class Alternative to glsBuiltinPrecision.js according to native deqp

JiangYizhou:
We implement "Alternatives" class according to native deqp.

### DIFF
--- a/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
+++ b/sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
@@ -479,6 +479,8 @@ var setParentClass = function(child, parent) {
         /** type{tcuInterval.Interval} */ this.iVal;
     };
 
+    setParentClass(glsBuiltinPrecisionTests.ScalarTraits, glsBuiltinPrecisionTests.Traits);
+
     glsBuiltinPrecisionTests.ScalarTraits.prototype = Object.create(glsBuiltinPrecisionTests.Traits.prototype);
     glsBuiltinPrecisionTests.ScalarTraits.prototype.constructor = glsBuiltinPrecisionTests.ScalarTraits;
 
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
LGTM. This change implemented "Alternative" class according to the patches from native deqp.

@zhenyao, PTAL. Thanks a lot!

JiangYizhou:
@zhenyao @kenrussell Why is the 'failing check' after I pulling request?

qkmiao:
"failing check" means there are some lint errors in your pr. Please check https://github.com/KhronosGroup/WebGL/pull/1876 on how to find what are the errors.  The check was enabled in https://github.com/KhronosGroup/WebGL/pull/1918.

JiangYizhou:
@qkmiao @Jiawei-Shao Thank you, the check can pass now!

sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
```diff
@@ -4006,13 +4004,37 @@ var setParentClass = function(child, parent) {
...
         }
     };
 
+    /**
+     * @constructor
+     * @extends {glsBuiltinPrecisionTests.Func}
+     * @param {glsBuiltinPrecisionTests.Signature} Sig_ template <typename Sig_>
+     */
+    glsBuiltinPrecisionTests.Alternatives = function(Sig_) {
+         glsBuiltinPrecisionTests.Func.call(this, Sig_);
```
zhenyao: I am surprised this pass the newly checked in lint tools because indent here is wrong

zhenyao:
lgtm

sdk/tests/deqp/modules/shared/glsBuiltinPrecisionTests.js
kenrussell: Not sure, but the dEQP files are all problematic. They use tabs instead of spaces. I'm not sure it is a good idea to reformat them because doing so will lose "git blame" information.

PR #1943, opened 2016-07-26
Turn skipped multisample tests into warnings.

kenrussell:
This was the only dEQP test using "testSkippedOptions", causing the
pass rate of the conformance suite to be less than 100%.


kenrussell:
@zhenyao @null77 @vonture @jdashg please review.

This is the pattern used by the other dEQP tests when a format or configuration isn't supported, and is the source of the skipped tests causing the pass rate to be less than 100% on some configs.

zhenyao:
lgtm
thanks for looking into this

PR #1945, opened 2016-07-28
Add DEPTH_STENCIL to depth-renderable list.

JerryShih:
In webgl2 3.7.5 Renderbuffer objects:
To be backward compatible with WebGL 1, also accepts internal format
DEPTH_STENCIL, which should be mapped to DEPTH24_STENCIL8 by implementations.


jdashg:
This matches my understanding of the intent of getInternalformatParameter.
@kenrussell @grorg

zhenyao:
I think this is a reasonable behavior for getinternalformatparameter and renderbufferstorage to be consistent. We should be explicitly say it in the spec in getinternalformatparameter to be clearer.

PR #1946, opened 2016-07-29
Multi-sample floating-point color buffers can be optional supported.

Richard-Yunchao:
Enable the tests and move related files to 2.0.1

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fFboMultisampleTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fFboMultisampleTests.js
@@ -79,12 +79,13 @@ var DE_ASSERT = function(x) {
 
     es3fFboMultisampleTests.BasicFboMultisampleCase.prototype.preCheck = function() {
         this.checkFormatSupport(this.m_colorFormat);
-        this.checkSampleCount(this.m_colorFormat, this.m_numSamples);
+        var sampleCountSupported = this.checkSampleCount(this.m_colorFormat, this.m_numSamples);
 
         if (this.m_depthStencilFormat != gl.NONE) {
             this.checkFormatSupport(this.m_depthStencilFormat);
-            this.checkSampleCount(this.m_depthStencilFormat, this.m_numSamples);
+            sampleCountSupported = this.checkSampleCount(this.m_depthStencilFormat, this.m_numSamples);
         }
+        return sampleCountSupported;
     };
 
     /**
@@ -305,14 +306,14 @@ var DE_ASSERT = function(x) {
             gl.R8,
 
             // gl.EXT_color_buffer_float
-            // Multi-sample floating-point color buffers are not supported, see https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float/
-            // gl.RGBA32F,
-            // gl.RGBA16F,
-            // gl.R11F_G11F_B10F,
-            // gl.RG32F,
-            // gl.RG16F,
-            // gl.R32F,
-            // gl.R16F
+            // Multi-sample floating-point color buffers can be optional supported, see https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float/
+            gl.RGBA32F,
+            gl.RGBA16F,
+            gl.R11F_G11F_B10F,
+            gl.RG32F,
+            gl.RG16F,
+            gl.R32F,
+            gl.R16F
         ];
 
         /** @const {Array<number>} */ var depthStencilFormats = [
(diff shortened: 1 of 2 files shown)


Richard-Yunchao:
@kenrussell and @zhenyao , now that we have decide to add multi-sample floating point color buffer support into WebGL 2 after discussions within community (https://github.com/KhronosGroup/WebGL/pull/1935), this small change enables the tests for this feature. I prefer to add it into 2.0.1 and implement it later.  

PTAL. Thanks.

@jchen10 , FYI. 
@jchen10 will fix the bug in Chromium.

sdk/tests/deqp/functional/gles3/00_test_list.txt
```diff
@@ -27,9 +27,9 @@ clipping.html
 defaultvertexattribute.html
 fbocompleteness.html
 fbodepthbuffer.html
-fbomultisample.2_samples.html
-fbomultisample.4_samples.html
-fbomultisample.8_samples.html
+--min-version 2.0.1 fbomultisample.2_samples.html
```
zhenyao: Let's not take this out to 2.0.1.

Instead, we should add checkFramebufferStatus() to all these optional multisampled float formats.  If it returns something other than FRAMEBUFFER_COMPLETE, skip further testing but still allow the test to pass.

Richard-Yunchao:
@zhenyao, @kenrussell , Thanks for your review. I have some concerns about the WebGL extension spec: https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float/. 

1. multisample float-point color formats: I think R11F_G11F_B10F, RG16F, R16F can be supported by multisample d renderbuffer. 
According to the native spec: https://www.khronos.org/registry/gles/extensions/EXT/EXT_color_buffer_float.txt. It says that "and the RGBA16F, R32F, RG32F and RGBA32F formats, for which they are only required to support creation of single sample renderbuffers.", so I think R11F_G11F_B10F, RG16F, R16F should be supported by multi-sample renderbuffers if it claims to support this extension. 
Seems that the native deqp also obey to this, see the code: https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/gles3/functional/es3fFboTestCase.cpp#313. 

2. How to know that these color formats can be supported by multisampled renderbuffer.  
In the WebGL extension spec, you say that "check for framebuffer completeness after attaching such a renderbuffer." I wonder why don't we let the user check the sample count by getInternalFormat() API. I think it is direct and clear. 

Maybe I am wrong, any insights about my concerns? 

I have updated the test case according to my suggestions.

sdk/tests/deqp/functional/gles3/es3fFboMultisampleTests.js
```diff
@@ -305,14 +306,14 @@ var DE_ASSERT = function(x) {
...
             // gl.EXT_color_buffer_float
-            // Multi-sample floating-point color buffers are not supported, see https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float/
-            // gl.RGBA32F,
-            // gl.RGBA16F,
-            // gl.R11F_G11F_B10F,
-            // gl.RG32F,
-            // gl.RG16F,
-            // gl.R32F,
-            // gl.R16F
+            // Multi-sample floating-point color buffers can be optional supported, see https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float/
```
kenrussell: optional -> optionally

kenrussell:
Yunchao, thanks for these updates. To reply to your questions:

1. Yes, R11F_G11F_B10F, RG16F, and R16F are all supported by the WebGL version of the EXT_color_buffer_float extension. This is documented in https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float/ . It's fine to include these formats in the WebGL 2.0 conformance tests.

2. I was confused. Looking back at the ES 3.0 spec (section 4.4.4, subsection "Whole Framebuffer Completeness"), it's required that all renderbuffer attachments have the same number of samples, so yes, as long as that's obeyed, the wording about checking framebuffer completeness is unnecessary. I'm removing it in https://github.com/KhronosGroup/WebGL/pull/1949 .

kenrussell:
Thanks @Richard-Yunchao for this patch. Per your questions above, it looks good. LGTM. I trust you've tested this on a conformant implementation and it passes. Merging now. Feel free to ignore the minor typo above.

FYI: @zhenyao @Kangz @null77 @vonture @jdashg , who are focusing on the WebGL 2.0 suite.

PR #1947, opened 2016-08-02
Update tests for recent ECMAScript changes

bakkot:
Several tests are making outdated assertions about various TypedArray and DataView related functions. Their behavior was changed by TC39 [recently](https://github.com/tc39/ecma262/pull/410).

This PR amends those tests so that either the old or new behavior is accepted. I'm not sure if there's a way to officially mark one behavior as deprecated. This PR doesn't update the snapshots, just the tests under sdk/.

Ping @kenrussell; we discussed this briefly over email a while ago. @littledan @leobalter

### DIFF
--- a/sdk/tests/conformance/typedarrays/array-unit-tests.html
+++ b/sdk/tests/conformance/typedarrays/array-unit-tests.html
@@ -641,21 +641,27 @@ function testConstructionWithNullBuffer(type, name) {
     var array;
     try {
         array = new type(null);
-        testFailed("Construction of " + name + " with null buffer should throw exception");
+        assertEq("Length of " + name + " constructed with null", 0, array.length);
+        testPassed("Construction of " + name + " with null produced a " + name + " of length 0");
     } catch (e) {
-        testPassed("Construction of " + name + " with null buffer threw exception");
+        // This used to be correct, but TC39 has changed the behavior of these constructors.
+        testPassed("Construction of " + name + " with null threw exception");
+    }
+    try {
+        array = new type(null, 0);
+        assertEq("Length of " + name + " constructed with null", 0, array.length);
+        testPassed("Construction of " + name + " with null produced a " + name + " of length 0");
+    } catch (e) {
+        // This used to be correct, but TC39 has changed the behavior of these constructors.
+        testPassed("Construction of " + name + " with (null, 0) threw exception");
     }
     try {
         array = new type(null, 0, 0);
-        testFailed("Construction of " + name + " with (null buffer, 0) should throw exception");
+        assertEq("Length of " + name + " constructed with null", 0, array.length);
+        testPassed("Construction of " + name + " with null produced a " + name + " of length 0");
     } catch (e) {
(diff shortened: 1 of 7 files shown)


kenrussell:
Thanks, this looks good. Could you please:

1. Apply these changes to the 1.0.1, 1.0.2 and 1.0.3 snapshots in the conformance-suites/ top-level directory, run the associated tests, and make sure they pass.

2. Squash commits and re-upload this pull request, probably using the --force option to "git push" as described in https://www.khronos.org/webgl/wiki/Using_Github_To_Contribute . We'd like to merge a single commit with these changes.

Thanks.

CC'ing @jdashg @RafaelCintron @grorg @zhenyao . The old behavior should be removed in the top-of-tree conformance suite, once all browsers implement the new behavior. I'll file a follow-on bug about this once it lands.

sdk/tests/conformance/typedarrays/array-unit-tests.html
```diff
@@ -641,21 +641,27 @@
     var array;
     try {
         array = new type(null);
-        testFailed("Construction of " + name + " with null buffer should throw exception");
+        assertEq("Length of " + name + " constructed with null", 0, array.length);
+        testPassed("Construction of " + name + " with null produced a " + name + "of length 0");
```
bterlson: missing a space at the beginning of the last literal in this line

sdk/tests/conformance/typedarrays/array-unit-tests.html
```diff
@@ -641,21 +641,27 @@
...
     }
     try {
         array = new type(null, 0, 0);
-        testFailed("Construction of " + name + " with (null buffer, 0) should throw exception");
+        assertEq("Length of " + name + " constructed with null", 0, array.length);
+        testPassed("Construction of " + name + " with null produced a " + name + "of length 0");
     } catch (e) {
-        testPassed("Construction of " + name + " with (null buffer, 0) threw exception");
+        // This used to be correct, but TC39 has changed the behavior of these constructors.
+        testPassed("Construction of " + name + " with (null, 0) threw exception");
```
bterlson: `with (null, 0)` doesn't seem to match line 651?

bakkot: Ah, good catch, will fix - that appears to be [an existing bug](https://github.com/KhronosGroup/WebGL/blob/df8a264ebd99b2c4b7f788b81bb9ff6611381213/sdk/tests/conformance/typedarrays/array-unit-tests.html#L649-L655) I blindly copied over.

bakkot:
Done. Both Chrome and Firefox pass on 1.0.1, 1.0.2, and 1.0.3, and Chrome continues to pass (which it did not previously) when updated to use the now-current semantics.

I can't easily test other browsers, but I could do so if necessary.

kenrussell:
Thank you very much for updating the older snapshots (so they'll continue to pass on older browsers) and for testing with multiple browsers. The testing you did is great. LGTM. Merging now.

PR #1949, opened 2016-08-03
Simplified wording about optional multi-sampled floating-point color …

kenrussell:
…renderbuffers.

Thanks to @Richard-Yunchao for this feedback in #1946 pointing out
that the wording was unnecessary and technically incorrect.


kenrussell:
@jdashg @zhenyao @grorg @RafaelCintron please review. I shouldn't have added this sentence in the first place.

zhenyao:
Looks good.

PR #1950, opened 2016-08-05
Add sample count check on GL_MAX_SMAPLES

jchen10:
GLES3 spec, section 6.1.15
"Since multisampling is not supported for signed and unsigned integer internal formats, the value of
NUM_SAMPLE_COUNTS will be zero for such formats. For every other accepted internalformat, the value of NUM_SAMPLE_COUNTS is guaranteed to be at least one, and the maximum value in SAMPLES is guaranteed to be at least the value of MAX_SAMPLES."

@Richard-Yunchao please have a look, 3x!

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fFboTestCase.js
+++ b/sdk/tests/deqp/functional/gles3/es3fFboTestCase.js
@@ -135,7 +135,7 @@ var DE_ASSERT = function(x) {
             /** @const @type {goog.NumberArray} */ var supportedSampleCounts = es3fFboTestCase.querySampleCounts(sizedFormat);
             var supported = Array.prototype.slice.call(supportedSampleCounts);
             if (supported.indexOf(numSamples) == -1) {
-                if (numSamples >= 2) {
+                if (minSampleCount == 0 || numSamples > gl.getParameter(gl.MAX_SAMPLES)) {
                     testPassed("Sample count not supported, but it is allowed.");
                     return false;
                 } else {


sdk/tests/deqp/functional/gles3/es3fFboTestCase.js
```diff
@@ -135,7 +135,7 @@ var DE_ASSERT = function(x) {
             /** @const @type {goog.NumberArray} */ var supportedSampleCounts = es3fFboTestCase.querySampleCounts(sizedFormat);
             var supported = Array.prototype.slice.call(supportedSampleCounts);
             if (supported.indexOf(numSamples) == -1) {
-                if (numSamples >= 2) {
+                if (minSampleCount == 0 || numSamples > gl.getParameter(gl.MAX_SAMPLES)) {
```
kenrussell: Why doesn't this check exist in the native dEQP at:
https://android.googlesource.com/platform/external/deqp/+/deqp-dev/modules/gles3/functional/es3fFboTestCase.cpp#360
?

Richard-Yunchao: I think the native deqp is wrong. If these floating point formats can not be multisampled, the native deqp will fail. 
This is not correct. You know, these floating point formats are not required to be multisampled. It is only optional.

kenrussell: Reviewing this change and studying the spec it looks correct to me, but I'd like @zhenyao to review it too and ask @phaulos to please take a look too.

zhenyao: The c++ code throws a NotSupportedError if the numSamples isn't supported.  Is that a test failure?  I thought it's actually not a failure.  In that case, we don't really need to do this if-else here, we can just do "testPassed('Sample count not supported'); return false;"

jchen10: Per my understanding, R11F_G11F_B10F, RG16F, and R16F are totally optional regarding the supported samples. So for these 4 formats, it's okay to pass the test if numSamples isn't supported. But RGBA32F, RGBA16F and RG32F are partially optional. They must support those samples which are no more than MAX_SAMPLES. So for these 3 formats, we must fail the test if this's not obeyed. Furthermore if numSamples is more than MAX_SMAPLES, the support is optional, so we can pass the test with "Sample count not supported".

zhenyao: OK, I found the spec text to support what you said.

"Implementations must support creation of renderbuffers in these required formats with up to the value of MAX_SAMPLES multisamples, with the exception of signed and unsigned integer formats."

With EXT_color_buffer_float, the last part became "with the exception of signed and unsigned integer, RGBA16F, R32F, RG32F and RGBA32F formats."

zhenyao:
lgtm

PR #1951, opened 2016-08-05
Add test for GenerateMipmap with sRGB formated texture

qkmiao:


qkmiao:
@zhenyao is this the test case you mentioned in https://bugs.chromium.org/p/chromium/issues/detail?id=634519?

zhenyao:
It's more complicated than that. I don't think the spec is clear about sRGB behavior in GenerateMipmap.

In common sense, when we generate mipmap, for example, we do the bi-linear interpolation to get the next mip level, but we should decode sRGB to linear before the interpolation, and then re-encode it to sRGB.  However, most drivers just simply do interpolation.

Your test won't catch this bug, because you test with a solid color.

If you come up with a test case, please do it in a separate test file and add it to 2.0.1 because it's not clear in the spec and let's not block our WebGL2 release by this.

qkmiao:
Thanks for review. Sorry, I didn't get the reason why the test wouldn't catch the bug. @zhenyao can you explain it a little more about "solid color"? Thank you!

qkmiao:
@zhenyao Thanks for detailed explanations. I updated the pr. PTAL.

sdk/tests/conformance2/textures/misc/tex-srgb-mipmap.html
```diff
@@ -0,0 +1,184 @@
...
+    for (var i = 0; i < 7 * 7; i++) {
+        buf[4 * i + 0]  = 255;
+        buf[4 * i + 1] = 0;
+        buf[4 * i + 2] = 0;
+        buf[4 * i + 3] = 255;
+    }
+    gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, 7, 7, gl.RGBA, gl.UNSIGNED_BYTE, buf);
+    gl.generateMipmap(gl.TEXTURE_2D);
+
+    // Decode the srgba texture to a linear texture which will be used as reference.
```
zhenyao: I think here you should compute the decoding manually and set up the linear texture like you set up the sRGB texture. Just to minimize the code paths being tested here. (Here you are testing rendering with sRGB texture generates correct results)

zhenyao:
The test is smart.
With two suggestions:
1) I already mentioned in the code, to minimize the tested code paths.
2) Can you find a way to visualize the two rendered images (one from sRGB, one from linear)? So it's easy to see the difference and the bug.

zhenyao:
@kenrussell @jdashg @RafaelCintron @grorg FYI.  Unfortunately, the GenerateMipmap spec isn't clear and there are no deqp tests covering this, so it looks like all drivers got this wrong. I am not sure what's the best way to move forward, but it's a real world issue - we discovered this because Skia team is suffering from it.

sdk/tests/conformance2/textures/misc/tex-srgb-mipmap.html
```diff
@@ -0,0 +1,184 @@
...
+    // 7 is selected to make sure bi-linear interpolation operates on different colors.
+    var buf = new Uint8Array(7 * 7 * 4);
+    for (var i = 0; i < 7 * 7; i++) {
+        buf[4 * i + 0]  = 255;
+        buf[4 * i + 1] = 0;
+        buf[4 * i + 2] = 0;
+        buf[4 * i + 3] = 255;
+    }
+    gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, 7, 7, gl.RGBA, gl.UNSIGNED_BYTE, buf);
+    gl.generateMipmap(gl.TEXTURE_2D);
```
kenrussell: At this point do you think it would be a good idea to change the filtering mode of this texture to something like NEAREST_MIPMAP_NEAREST, so that later when the texture is drawn on the unit quad, it's essentially copying the expected mipmap level directly into the framebuffer with no extra filtering?

sdk/tests/conformance2/textures/misc/tex-srgb-mipmap.html
```diff
@@ -0,0 +1,184 @@
...
+    var linearTex = gl.createTexture();
+    gl.bindTexture(gl.TEXTURE_2D, linearTex);
+    wtu.fillTexture(gl, linearTex, width, height, colors['blank'], 0, gl.RGBA, gl.UNSIGNED_BYTE);
+    var fbo = gl.createFramebuffer();
+    gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, linearTex, 0);
+    gl.bindTexture(gl.TEXTURE_2D, srgbTex);
+    wtu.clearAndDrawUnitQuad(gl);
+    gl.bindTexture(gl.TEXTURE_2D, linearTex);
+    gl.generateMipmap(gl.TEXTURE_2D);
```
kenrussell: Same question about changing the filtering mode of linearTex here, after its mipmaps have been generated.

kenrussell:
Very nice test @qkmiao and thanks @zhenyao for tracking down this bug so carefully.

This LGTM to me whenever @zhenyao thinks the test is ready. I agree with adding this to 2.0.1 since it's a complex corner case unlikely to affect most applications and will need more thought to figure out a workaround.

sdk/tests/conformance2/textures/misc/tex-srgb-mipmap.html
qkmiao: I set  the minification filter as NEAREST_MIPMAP_NEAREST here and below. Thanks for reminder, @kenrussell .

qkmiao:
@zhenyao @kenrussell thanks for review. I do srgb texture decoding manually and display the result, reference, and difference image if the test fails. PTAL.

zhenyao:
Great. LGTM

kenrussell:
Very nice. LGTM after the fact.

jdashg:
I suspect that we would have to explicitly spec this behavior.
GenerateMipmap is spec'd very loosely, and I don't think we should aim to
improve it in WebGL.
The best mipmaps will always be custom.
It's useful to have a test, but I suspect we shouldn't spec anything
additional here, thus we shouldn't take this test in the conformance suite.

kenrussell:
I think this test should be in the conformance suite. sRGB textures are an important new feature of ES 3.0, and it's important to make sure vendors implement key functionality identically.

This test is deferred to 2.0.1. If we find that it's infeasible to enforce at that time, we can consider pushing it out further.

PR #1952, opened 2016-08-05
Add test to check invariant qualifier doesn't leak state

qkmiao:


kenrussell:
Aren't these cases already covered in sdk/tests/conformance/glsl/misc/shaders-with-invariance.html ?

kenrussell:
Sorry for my confusion; I just saw your https://chromium-review.googlesource.com/366720 which fixes bugs in this area.

I continue to wonder whether sdk/tests/conformance/glsl/misc/shaders-with-invariance.html already covers this case.

qkmiao:
Actually, these tests in this pr are from sdk/tests/conformance/glsl/misc/shaders-with-invariance.html, but in different orders. This pr moves the test which has invariant qualifier varying before the test which doesn't have invariant qualifier. So, the bug happens.

sdk/tests/conformance/glsl/bugs/invariant-does-not-leak-across-shaders.html
```diff
@@ -0,0 +1,90 @@
...
+varying vec4 v_varying;
+
+void main() {
+    gl_Position = v_varying;
+}
+</script>
+<script type="text/javascript">
+"use strict";
+description("The use of the invariant qualifier in one shader must not affect other shaders.");
+
```
kenrussell: Could you add a debug() call here, mentioning that this is a deliberate subset of glsl/misc/shaders-with-invariance.html, and linking to http://crbug.com/634813 ? Thanks.

kenrussell:
One minor request. Could you please add it and comment on this pull request when done? Thanks.

qkmiao:
Thanks for review. debug message was added in the new pr.

kenrussell:
Thank you. LGTM.

PR #1953, opened 2016-08-05
Check getProgramInfoLog does not return correct empty string

qkmiao:
Chrome fails this test both on nVidia and Intel. Firefox passes this test.

### DIFF
--- a/sdk/tests/conformance/programs/00_test_list.txt
+++ b/sdk/tests/conformance/programs/00_test_list.txt
@@ -6,5 +6,6 @@ gl-get-active-uniform.html
 gl-getshadersource.html
 gl-shader-test.html
 invalid-UTF-16.html
+--min-version 1.0.4 program-infolog.html
 program-test.html
 --min-version 1.0.2 use-program-crash-with-discard-in-fragment-shader.html
--- a/sdk/tests/conformance/programs/program-infolog.html
+++ b/sdk/tests/conformance/programs/program-infolog.html
@@ -1,0 +1,83 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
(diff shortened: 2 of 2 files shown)


qkmiao:
@kenrussell PTAL.
There are some discussions in https://github.com/mrdoob/three.js/pull/9456.

zhenyao:
We had this discussion before - this behavior is not supported by spec, so we should not verify it.  In theory, anything can be returned in the log, whether the compile/link succeeds or not.

sdk/tests/conformance/programs/program-infolog.html
```diff
@@ -0,0 +1,85 @@
...
+  testPassed("context exists");
+}
+
+debug("");
+
+var program = wtu.loadProgramFromScript(gl, 'vertexShader', 'fragmentShader');
+var infolog = gl.getProgramInfoLog(program);
+if (infolog === '\0') {
+  testFailed("getProgramInfoLog should not return '\\0'");
+} else if (infolog === '') {
```
kenrussell: I think this test would be OK if this was:

  } else { testPassed("getProgramInfoLog did not return '\\0'"); }

it would be more of a quality of implementation issue then, as opposed to requiring that a successful compilation has an empty program info log -- which is not guaranteed.

kenrussell:
@zhenyao what do you think about my comment on the pull request? Would it be OK then?

For the record: this test passes on Chrome 54.0.2816.0 on my Linux workstation. about:gpu is here:

https://gist.github.com/kenrussell/75b47c5680ba68a35e9ffbeb0a5bfb8d

qkmiao:
I updated this pr to follow @kenrussell 's comments. This test fails on Mesa 11.2.0 and 12.0.1, NVIDIA 349.16. It passed on NVIDIA 364.19.

sdk/tests/conformance/programs/00_test_list.txt
```diff
@@ -6,5 +6,6 @@ gl-get-active-uniform.html
 gl-getshadersource.html
 gl-shader-test.html
 invalid-UTF-16.html
+program-infolog.html
```
kenrussell: Add "--min-version 1.0.4".

kenrussell:
Thanks for re-testing.

This LGTM for what it's worth, aside from restricting the test to 1.0.4. Let's let @zhenyao re-review and decide what to do.

zhenyao:
OK, this requirement is reasonable. LGTM

PR #1956, opened 2016-08-08
Add multsampling testcases for float-point color renderbuffers

jchen10:
Jointly with https://github.com/KhronosGroup/WebGL/pull/1946 to cover the multisample support test for float-point color renderbuffers.

### DIFF
--- a/sdk/tests/conformance2/extensions/ext-color-buffer-float.html
+++ b/sdk/tests/conformance2/extensions/ext-color-buffer-float.html
@@ -193,37 +193,66 @@ function runFloatTextureRenderTargetTest(enabled, internalFormat, format, testPr
 function runFloatRenderbufferRenderTargetTest(enabled, internalFormat, testProgram, numberOfChannels, subtractor)
 {
     var formatString = wtu.glEnumToString(gl, internalFormat);
-    debug("");
-    debug("testing floating-point " + formatString + " renderbuffer render target");
-
-    var colorbuffer = gl.createRenderbuffer();
-    var width = 2;
-    var height = 2;
-    gl.bindRenderbuffer(gl.RENDERBUFFER, colorbuffer);
-    gl.renderbufferStorage(gl.RENDERBUFFER, internalFormat, width, height);
-    if (!enabled) {
-        wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "floating-point renderbuffer allocation should fail if EXT_color_buffer_float is not enabled");
-        return;
-    } else {
-        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "floating-point renderbuffer allocation should succeed if EXT_color_buffer_float is enabled");
+    var samples = [0];
+    if (enabled) {
+        samples = Array.prototype.slice.call(gl.getInternalformatParameter(gl.RENDERBUFFER, internalFormat, gl.SAMPLES));
+        samples.push(0);
     }
+    for (var ndx = 0; ndx < samples.length; ++ndx) {
+        debug("");
+        debug("testing floating-point " + formatString + " renderbuffer render target with number of samples " + samples[ndx]);
 
-    // Try to use this texture as a render target.
-    var fbo = gl.createFramebuffer();
-    gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
-    gl.framebufferRenderbuffer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.RENDERBUFFER, colorbuffer, 0);
-    gl.bindRenderbuffer(gl.RENDERBUFFER, null);
+        var colorbuffer = gl.createRenderbuffer();
+        var width = 2;
+        var height = 2;
+        gl.bindRenderbuffer(gl.RENDERBUFFER, colorbuffer);
+        if (samples[ndx] == 0)
+            gl.renderbufferStorage(gl.RENDERBUFFER, internalFormat, width, height);
+        else
+            gl.renderbufferStorageMultisample(gl.RENDERBUFFER, samples[ndx], internalFormat, width, height);
+        if (!enabled) {
+            wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "floating-point renderbuffer allocation should fail if EXT_color_buffer_float is not enabled");
+            return;
+        } else {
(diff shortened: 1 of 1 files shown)


jchen10:
@Richard-Yunchao Please take a look first! Thanks!

sdk/tests/conformance2/extensions/ext-color-buffer-float.html
```diff
@@ -193,37 +193,63 @@
...
     }
+    for (var ndx = 0; ndx < samples.length; ++ndx) {
+        debug("");
+        debug("testing floating-point " + formatString + " renderbuffer render target with number of samples " + samples[ndx]);
+
+        var colorbuffer = gl.createRenderbuffer();
+        var width = 2;
+        var height = 2;
+        gl.bindRenderbuffer(gl.RENDERBUFFER, colorbuffer);
+        gl.renderbufferStorageMultisample(gl.RENDERBUFFER, samples[ndx], internalFormat, width, height);
```
zhenyao: Although according to the spec, renderbufferStorageMultisample(samples == 0) is the same as renderbufferStorage, but we still should verify the implementation follows the spec.  I suggest we test both renderbufferStorageMultisample(samples == 0) and renderbufferStorage

sdk/tests/conformance2/extensions/ext-color-buffer-float.html
```diff
@@ -193,37 +193,63 @@
...
-    gl.bindRenderbuffer(gl.RENDERBUFFER, colorbuffer);
-    gl.renderbufferStorage(gl.RENDERBUFFER, internalFormat, width, height);
-    if (!enabled) {
-        wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "floating-point renderbuffer allocation should fail if EXT_color_buffer_float is not enabled");
-        return;
-    } else {
-        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "floating-point renderbuffer allocation should succeed if EXT_color_buffer_float is enabled");
+    var samples = [0];
+    if (enabled) {
+        samples = Array.prototype.slice.call(gl.getInternalformatParameter(gl.RENDERBUFFER, internalFormat, gl.SAMPLES));
```
zhenyao: This is incorrect.  If extension is not enabled, the float format isn't color renderable and calling getInternalformatParameter on it will generate INVALID_ENUM.

sdk/tests/conformance2/extensions/ext-color-buffer-float.html
```diff
@@ -193,37 +193,63 @@
...
+        } else {
+            wtu.glErrorShouldBe(gl, gl.NO_ERROR, "floating-point renderbuffer allocation should succeed if EXT_color_buffer_float is enabled");
+        }
 
-    // Try to use this texture as a render target.
-    var fbo = gl.createFramebuffer();
-    gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
-    gl.framebufferRenderbuffer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.RENDERBUFFER, colorbuffer, 0);
-    gl.bindRenderbuffer(gl.RENDERBUFFER, null);
+        // Try to use this texture as a render target.
```
zhenyao: not your fault, but can you fix this comment? It's not a texture, but a renderbuffer.

sdk/tests/conformance2/extensions/ext-color-buffer-float.html
```diff
@@ -193,37 +193,63 @@
...
+        if (completeStatus != gl.FRAMEBUFFER_COMPLETE) {
+            testFailed("floating-point " + formatString + " render target not supported");
+            return;
+        }
+        var resolveColorRbo = null;
+        var resolveFbo = null;
+        if (samples[ndx] > 0) {
+            resolveColorRbo = gl.createRenderbuffer();
+            gl.bindRenderbuffer(gl.RENDERBUFFER, resolveColorRbo);
+            gl.renderbufferStorageMultisample(gl.RENDERBUFFER, 0, internalFormat, width, height);
```
zhenyao: Can we just use renderbufferStorage here?

sdk/tests/conformance2/extensions/ext-color-buffer-float.html
```diff
@@ -270,6 +296,41 @@
...
+        if (fmt >= firstMultiOnlyFormat && (samples.length == 0 || samples[0] < maxSamples)) {
+            testFailed("the maximum value in SAMPLES should be at least " + maxSamples);
+            return;
+        }
+
+        var prevSampleCount = 0;
+        var sampleCount;
+        for (var ndx = 0; ndx < samples.length; ++ndx, prevSampleCount = sampleCount) {
+            sampleCount = samples[ndx];
+            // sample count must be > 0
```
zhenyao: Is that true? For the first four formats, I think it can be 0.

sdk/tests/conformance2/extensions/ext-color-buffer-float.html
jchen10: It's already under the "enabled" branch to call getInternalformatParameter. Did you mean we'd better test "enabled" and "disabled" both?

sdk/tests/conformance2/extensions/ext-color-buffer-float.html
jchen10: "If internalformat is GL_RGBA16F, GL_R32F, GL_RG32F, or GL_RGBA32F, the value of **GL_NUM_SAMPLE_COUNTS may be zero**, or else the maximum value in GL_SAMPLES may be less than the value of GL_MAX_SAMPLES"

I think this mostly indicates that the length of returned array, not its elements may zero. Actually the elements should be positive according to the below sentence.

"if pname is SAMPLES, the sample counts supported for internalformat and target are written into params, in descending numeric order. **Only positive values are returned.**"

sdk/tests/conformance2/extensions/ext-color-buffer-float.html
zhenyao: You are correct.

sdk/tests/conformance2/extensions/ext-color-buffer-float.html
zhenyao: You are correct.  Apparently I was half sleeping when I made these comments.  My apology.

Issue #1954, opened 2016-08-06
No tests for no attribute shaders in WebGL 1.0 (and 2.0?) conformance tests


greggman:
I made a shader like this

    void main() {
      gl_PointSize = 100.0;
      gl_Position = vec4(0, 0, 0, 1);
    }

Then with no buffers and no attributes I called `gl.drawArrays(gl.POINTS, 0, 1)`.

This works in Firefox and Chrome but fails in Safari. It probably needs a test?

Interestingly Safari succeeds if I add an attribute but don't actually set one up. In other words

    attribute vec4 position;
    void main() {
      gl_PointSize = 100.0;
      gl_Position = position;
    }

But I don't create any buffers nor do I enable any attributes

Note [vertexshaderart.com](http://vertexshaderart.com) is entirely based on no attribute shaders. You can think of it as

    void main() {
      // use gl_VertexId to generate values for gl_Position
      ...
    }

    ...
    gl.drawArrays(mode, 0, count);

No attributes. 

Currently vertexshaderart.com is using an attribute to supply the vertex id because it's WebGL 1.0 which doesn't support `gl_VertexId` but WebGL 2.0 does and so it should function with no attributes.

Here's a live example: https://jsfiddle.net/greggman/0qtcc2v0/

Safari gets 

    [Warning] WebGL: INVALID_OPERATION: drawArrays: attempt to access out of bounds arrays

kenrussell:
Thanks for reporting this. I'm embarrassed to say I found out about the problems with no-attribute programs a few years ago when the WebGL Programming Guide was written, but it seemed like such a corner case that I didn't file a bug or write a conformance test for it.

Agree that in WebGL 2.0 this is a much more important case to make work. Do you think you could provide a conformance test for this, perhaps under sdk/tests/conformance2/glsl3/ ? Or at least a suggestion on how this test should be written? Could be a little tricky since the only browsers with WebGL 2.0 implementations right now are Firefox and Chrome.

A WebGL 1.0 conformance test would be welcome, too, but I'd like to defer making all the browsers work around this bug in WebGL 1.0 until WebGL 2.0 is out the door -- which shouldn't take long now.

zhenyao:
Should we close this since we do have a conformance test for this (Thanks Gregg to locate it)?

kenrussell:
[...]

PR #1957, opened 2016-08-08
add a no-attribute test

greggman:
RE #1954 

I'm not sure this test is needed actually

There's a test here

https://www.khronos.org/registry/webgl/sdk/tests/conformance/rendering/point-no-attributes.html?webglVersion=1&quiet=0

That already tests no attributes and Safari fails it. 

Please push Apple to fix that. Also iOS has never supported `preserveDrawingbuffer: true` (which given Canvas2D works just fine it seems like it can't be that much work)

### DIFF
--- a/sdk/tests/conformance2/glsl3/00_test_list.txt
+++ b/sdk/tests/conformance2/glsl3/00_test_list.txt
@@ -16,6 +16,7 @@ invalid-default-precision.html
 invalid-invariant.html
 loops-with-side-effects.html
 misplaced-version-directive.html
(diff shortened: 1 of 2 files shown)


zhenyao:
@grorg For your attention.

kenrussell:
Thanks Gregg for writing the WebGL 2.0 version of this test. I think it's worth adding because it uses ESSL 3.00 features that are expected to work in this scenario. I'm pretty sure Chrome and Firefox (the first WebGL 2.0 implementations) will pass this test, so we can remove the 2.0.1 minimum version requirement if you like.

Merging now.

PR #1958, opened 2016-08-09
TexSubImage3D should not modify the data outside of the subregion in 3D/2D_ARRAY texture

Richard-Yunchao:
@zhenyao , @kenrussell , PTAL. Thanks a lot!


Richard-Yunchao:
The code is not ready, close this pr first.

PR #1959, opened 2016-08-09
Fix regression on preCheck

qkmiao:
In https://github.com/KhronosGroup/WebGL/pull/1946, preCheck's behavior
is only changed for fbomultisample case. This breaks behavior of
preCheck for other cases. This patch makes all the appeared preChecks
return true if no exception or early return.

This also changes test status from pass to a warning for tests which
returns early.


kenrussell:
LGTM. I haven't tested this locally but assume you have. Thanks for the fix; let's get this in right away.

PR #1960, opened 2016-08-10
Add test case to function copyTexSubImage3D

JiangYizhou:
copyTexSubImage3D should mark a specific layer/level to be cleared, but it is not.
so textures is wrongly initialized after copyTexSubImage3D by framebuffertexturelayer.

### DIFF
--- a/sdk/tests/conformance2/textures/misc/copy-texture-image-webgl-specific.html
+++ b/sdk/tests/conformance2/textures/misc/copy-texture-image-webgl-specific.html
@@ -130,6 +130,50 @@ function copytexsubimage3d_valid_operation_diff_layer() {
     gl.deleteTexture(texture);
 }
 
+function copytexsubimage3d_texture_wrongly_initialized() {
+    debug("");
+    debug("Testing copytexsubimage3d_texture_wrongly_initialized");
+    var texture = [];
+    texture[0] = gl.createTexture();
+    texture[1] = gl.createTexture();
+    var layer = 0;
+    var width = 2;
+    var height = 2;
+    var depth = 2;
+    gl.bindTexture(gl.TEXTURE_2D, texture[0]);
+    var uint = new Uint8Array([0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F, 0x10]);
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, width, height, 0, gl.RGBA, gl.UNSIGNED_BYTE, uint);
+
+    var fbo = gl.createFramebuffer();
+    gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture[0], 0);
+    if (gl.checkFramebufferStatus(gl.FRAMEBUFFER) == gl.FRAMEBUFFER_COMPLETE) {
+        gl.bindTexture(gl.TEXTURE_3D, texture[1]);
+        gl.texStorage3D(gl.TEXTURE_3D, 1, gl.RGBA8, width, height, depth);
+        gl.copyTexSubImage3D(gl.TEXTURE_3D, 0, 0, 0, layer, 0, 0, width, height);
+        gl.framebufferTextureLayer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, texture[1], 0, layer);
+        var bytes = new Uint8Array(width*height*4);
+        gl.readPixels(0, 0, width, height, gl.RGBA, gl.UNSIGNED_BYTE, bytes);
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "readpixel should succeed.");
+        for (var x = 0; x < width * height * 4; x++) {
+            if (bytes[x] != uint[x]) {
+                testFailed("byte comparison failure, byte at "+ x + " was " + bytes[x] +
+                ", should be " + uint[x]);
+                break;
+            }
+        }
+    } else {
+        testFailed("framebuffer not complete");
+    }
+
+    gl.bindTexture(gl.TEXTURE_2D, null);
+    gl.bindTexture(gl.TEXTURE_3D, null);
+    gl.bindFramebuffer(gl.FRAMEBUFFER, null);
+    gl.deleteFramebuffer(fbo);
+    gl.deleteTexture(texture[0]);
+    gl.deleteTexture(texture[1]);
+};
+
 if (!gl) {
     testFailed("WebGL context does not exist");
 } else {
@@ -137,6 +181,7 @@ if (!gl) {
     copytexsubimage3d_invalid_operation_feedbackloops();
     copytexsubimage3d_valid_operation_diff_level();
     copytexsubimage3d_valid_operation_diff_layer();
+    copytexsubimage3d_texture_wrongly_initialized();
 }
 
 var successfullyParsed = true;


JiangYizhou:
@Richard-Yunchao PTAL

Richard-Yunchao:
Good work, Yizhou. 
@zhenyao , @kenrussell , PTAL. I have discussed with Yizhou offline, I think it is OK now. 
@zhenyao , This pr implemented your suggestion in my previous patch about CopyTexSubImage3D at https://codereview.chromium.org/2182443003/. This pr can expose such an error in current Chromium. I will fix the error in Chromium.

sdk/tests/conformance2/textures/misc/copy-texture-image-webgl-specific.html
```diff
@@ -130,13 +130,57 @@
...
+
+    var fbo = gl.createFramebuffer();
+    gl.bindFramebuffer(gl.FRAMEBUFFER,fbo);
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture[0], 0);
+    if (gl.checkFramebufferStatus(gl.FRAMEBUFFER) == gl.FRAMEBUFFER_COMPLETE) {
+        gl.bindTexture(gl.TEXTURE_3D, texture[1]);
+        gl.texStorage3D(gl.TEXTURE_3D, 1, gl.RGBA8, width, height, depth);
+        gl.copyTexSubImage3D(gl.TEXTURE_3D, 0, 0, 0, layer, 0, 0, width, height);
+        gl.framebufferTextureLayer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, texture[1], 0, layer);
+        var pixels = new Uint8Array(width*height*4);
```
zhenyao: Should you check fbo completeness here also? Or you can check gl error after readPixels.

sdk/tests/conformance2/textures/misc/copy-texture-image-webgl-specific.html
```diff
@@ -130,13 +130,57 @@
...
+        gl.readPixels(0, 0, width, height, gl.RGBA, gl.UNSIGNED_BYTE, pixels);
+        for (var x = 0; x < width * height * 4; x++) {
+            if (pixels[x] != uint[x]) {
+                testFailed("pixel comparison failure, pixel at "+ x + " was " + pixels[x] +
```
zhenyao: This is not a intuitive message.  You can either say "byte" instead of "pixel", or compare rgba instead of bytes.

JiangYizhou:
@zhenyao I have revised the code, PTAL.

zhenyao:
Thanks. LGTM.

sdk/tests/conformance2/textures/misc/copy-texture-image-webgl-specific.html
```diff
@@ -130,13 +130,58 @@
...
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, width, height, 0, gl.RGBA, gl.UNSIGNED_BYTE, uint);
+
+    var fbo = gl.createFramebuffer();
+    gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture[0], 0);
+    if (gl.checkFramebufferStatus(gl.FRAMEBUFFER) == gl.FRAMEBUFFER_COMPLETE) {
+        gl.bindTexture(gl.TEXTURE_3D, texture[1]);
+        gl.texStorage3D(gl.TEXTURE_3D, 1, gl.RGBA8, width, height, depth);
+        gl.copyTexSubImage3D(gl.TEXTURE_3D, 0, 0, 0, layer, 0, 0, width, height);
+        gl.framebufferTextureLayer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, texture[1], 0, layer);
```
kenrussell: Is there any issue where the fact that the 3D texture is bound to the TEXTURE_3D target at the same time that its 0th layer is bound to the framebuffer could cause a rendering feedback loop? I realize you aren't issuing a draw call here, but theoretically, if one was made here, that would be illegal.

kenrussell:
LGTM after the fact. This is a nice test and I'm very interested to see what's wrong in the texture tracking that's causing it to fail in Chrome.

sdk/tests/conformance2/textures/misc/copy-texture-image-webgl-specific.html
zhenyao: Yes, it is a feedback loop.  I have to double check to see if the current detection code covers 3D texture case as well.

Richard-Yunchao: That's true. should we generates Feedback Loop error here too?  I can do this in Chromium.

kenrussell: No. It has to be possible to get to this state. It's just draw calls that should fail in this situation. ReadPixels is OK. I'm just wondering whether we should call gl.bindTexture(gl.TEXTURE_3D, null) before ReadPixels, for clarity.

Richard-Yunchao: Yes. I have add the code for CopyTexSubImage3D to detect feedback loop. And it is a 'draw' call too. 

BTW, In GLES 3 spec, I didn't see (but not very sure) the spec says that framebufferTextureLayer should clear the texture to 0 if the level/layer of the bound texture is not cleared. But Chromium will do this (So this test case failed). Is this a particular behavior in WebGL? In this situation, we should update the WebGL 2 spec about this feature. @kenrussell , @zhenyao , any comments about this?

kenrussell: I don't think the call to FramebufferTextureLayer itself should clear the target texture. Instead, just before that texture is drawn to or copied from, if it isn't cleared, it should be cleared.

zhenyao: It's not framebufferTextureLayer, but anytime we use a texture, we can't leak the uninitialized vram, so we have to initialize it to 0 (this is webgl specific).

In the case here, texStorage3D left the texture uninitialized, and copyTexSubImage3D() wrote to part of the texture, so we have to initialize the rest.

Richard-Yunchao: Ken and Zhenyao, thanks for your explanation. But I still confused by the Spec and the implementation in Chromium. Let me state it clearly. 

As you said, the texStorage3D doesn't initiate the texture, CopyTexSubImage3D should do this. This feature is done by my patch at https://codereview.chromium.org/2232743002/. But that patch has not been merged yet. However, when the test reads pixels (on line 156) from the texture, it found that the texture has been initialized to 0 (So the test fails in latest Chromium). So I think it is framebufferTextureLayer do this initialization work. 
But the WebGL spec doesn't say that framebufferTextureLayer should clear the uninitialized texture.  

However, the spec about TexStorage3D in WebGL 2 says that: "The image contents are set as if a buffer of sufficient size initialized to 0 would be passed to each texImage3D call in the pseudocode in The OpenGL ES 3.0 specification section 3.8.4", So I think it should be initialized to 0 at the very beginning texStorage3D allocate the memory for texture. This is webgl specific, because in gles, it just allocate the memory, but don't initialized the memory. 

So It seems to me that the implementation about TexStorage3D and framebufferTextureLayer conflicts with the WebGL 2 spec about these two APIs(Maybe I misunderstand you and the code).

zhenyao: No, because without your CL, when we do CopyTexSubImage3D, we fail to initialize the texture, so the texture is still marked as uninitialized.  When call readPixels, of course we can't return the random vram, so we initialized all to zero, and therefore incorrectly cleared the entire texture, including the sub region specified by CopyTexSubImage3D.

zhenyao: Also, to conform to the WebGL spec, we don't initialize it right away, we only initialize it when we have to, "delayed" clearing to zero.

Richard-Yunchao: Got it. Thanks for your explanation, Zhenyao. So for the texStorage3D, we only initialize it when we have to. So Chromium implementation conforms to the WebGL spec too. 
Regarding to the actual initialization, it is done by readPixels, instead of by framebufferTextureLayer. 
Thanks a lot!

PR #1961, opened 2016-08-11
Add a test for getFragDataLocation.

zhenyao:


zhenyao:
@kenrussell Please review.
@Richard-Yunchao @qkmiao Can you verify the behavior of Chrome on Intel mesa with this test?
@jdashg @grorg FYI: Mac driver core profile doesn't handle getFragDataLocation correctly when output variable is an array.  Say "FragData[2]", getFragDataLocation("FragData[0]") and getFragDataLocation("FragData[1]") both return -1, but getFragDataLocation("FragData") return valid value.

kenrussell:
Great test. LGTM.

@jdashg : this test is important to add in 2.0.0 because Chrome ran into this particular driver bug in one of its non-WebGL tests, while enabling the Core Profile on macOS. It's easy to work around in the browser; the browser can synthesize the frag data locations for output arrays from the location of the base variable (in this case, synthesizing the locations of "fragColor[0]" and "fragColor[1]" from the location of "fragColor").

gyagp:
I tested the combinations mesa_version*platform as (11.2, 12.0, ToT-12.1)*(Broadwell, Skylake) on Linux, and all passed.

zhenyao:
@gyagp Thank you.

PR #1964, opened 2016-08-12
Add a new test to expose issues in texture-offset-out-of-range

Jiawei-Shao:
In texture-offset-out-of-range the shader uses a uniform variable to
represent the texture coordination, which is not a common usage and will
lead to chromium WebGL2 context crash on Windows using Intel GPU on
Broadwell machines.

This patch uses attribute instead of uniform variables to work around this
issue.

### DIFF
--- a/sdk/tests/conformance2/glsl3/00_test_list.txt
+++ b/sdk/tests/conformance2/glsl3/00_test_list.txt
@@ -28,6 +28,7 @@ shader-with-invalid-characters.html
 shader-with-mis-matching-uniform-block.html
 short-circuiting-in-loop-condition.html
 texture-offset-out-of-range.html
+--min-version 2.0.1 texture-offset-uniform-texture-coordinate.html
 --min-version 2.0.1 tricky-loop-conditions.html
 uniform-location-length-limits.html
 vector-dynamic-indexing.html
--- a/sdk/tests/conformance2/glsl3/texture-offset-out-of-range.html
+++ b/sdk/tests/conformance2/glsl3/texture-offset-out-of-range.html
@@ -38,16 +38,27 @@
 <body>
 <div id="description"></div>
 <div id="console"></div>
+<script id="vshaderInvalidOffset" type="x-shader/x-vertex">#version 300 es
+in vec4 a_position;
+in vec2 a_in0;
+out vec2 v_texCoord;
+
+void main()
+{
+    gl_Position = a_position;
+    v_texCoord = a_in0;
+}
+</script>
 <script id="fshaderInvalidOffset" type="x-shader/x-fragment">#version 300 es
 precision mediump float;
 
+in vec2 v_texCoord;
 out vec4 my_FragColor;
 uniform sampler2D u_sampler;
-uniform vec2 u_texCoord;
 uniform int x;
 
 void main() {
-    my_FragColor = textureOffset(u_sampler, u_texCoord, ivec2(0, $(yoffset)));
+    my_FragColor = textureOffset(u_sampler, v_texCoord, ivec2(0, $(yoffset)));
 }
 </script>
 <script type="application/javascript">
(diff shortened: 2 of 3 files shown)


Jiawei-Shao:
@qkmiao PTAL. Thanks!

qkmiao:
LGTM

sdk/tests/conformance2/glsl3/texture-offset-out-of-range.html
```diff
@@ -38,16 +38,27 @@
...
+    v_texCoord = a_in0;
+}
+</script>
 <script id="fshaderInvalidOffset" type="x-shader/x-fragment">#version 300 es
 precision mediump float;
 
+in vec2 v_texCoord;
 out vec4 my_FragColor;
 uniform sampler2D u_sampler;
-uniform vec2 u_texCoord;
```
Jiawei-Shao: As I know, texture coordinates are seldom declared as uniform in shader functions.
On Windows with Intel GPU on Broadwell, WebGL 2 context will crash after the program finishes linking. The crash occurs in the driver (illegal memory access) and does not happen on a specific line of code in chromium.

msc-:
I thought it was our policy to not change valid tests to work around driver bugs.

Jiawei-Shao:
This case is mainly intended to test the solution when the offset of the function textureOffset exceeds the required range, not to test if this function can use a variable as its argument. You can see that the driver can deal with this situation correctly with this patch and other dEQP tests such as deqp/functional/gles3/shadertexturefunction/textureoffset.html.
I think, if you agree, we can create another test to explicitly expose the bug that when textureOffset using a uniform texture location may cause the crash of WebGL context at linking.

msc-:
I can see the proposed change does not change the purpose of this test so I agree to it provided we create another test to expose the crash bug.

sdk/tests/conformance2/glsl3/texture-offset-uniform-texture-coordination.html
```diff
@@ -0,0 +1,201 @@
...
+	fShaderSuccess: true,
+	linkSuccess: true,
+	passMsg: 'textureLodOffset with uniform texture coordination should compile'
+  },
+  {
+    fShaderSource: shaderGradSrcValid,
+    fShaderSuccess: true,
+	linkSuccess: true,
+	passMsg: 'textureGradOffset with uniform texture coordination should compile'
+  },
```
Jiawei-Shao: Texture functions textureOffset, textureLodOffset, textureGradOffset will cause WebGL context crash on Windows with Intel GPU on BroadWell machines.

sdk/tests/conformance2/glsl3/texture-offset-uniform-texture-coordination.html
```diff
@@ -0,0 +1,201 @@
...
+    fShaderSuccess: true,
+    linkSuccess: true,
+    passMsg: 'textureLodOffset with uniform texture coordination should compile'
+  },
+  {
+    fShaderSource: shaderGradSrcValid,
+    fShaderSuccess: true,
+    linkSuccess: true,
+    passMsg: 'textureGradOffset with uniform texture coordination should compile'
+  },
```
Jiawei-Shao: Texture functions textureOffset, textureLodOffset, textureGradOffset will cause WebGL context crash on Windows with Intel GPU on Broadwell machines.

sdk/tests/conformance2/glsl3/texture-offset-uniform-texture-coordination.html
```diff
@@ -0,0 +1,201 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
```
qkmiao: 2015 -> 2016

sdk/tests/conformance2/glsl3/texture-offset-uniform-texture-coordination.html
```diff
@@ -0,0 +1,191 @@
...
+uniform vec2 u_texCoord;
+uniform vec2 u_lod;
+
+void main() {
+    my_FragColor = texelFetchOffset(u_sampler, ivec2(u_texCoord), int(u_lod), ivec2(0, 1));
+}
+</script>
+<script type="application/javascript">
+"use strict";
+description("Out-of-range texture offset should not compile.");
```
zhenyao: This isn't the right description.

sdk/tests/conformance2/glsl3/00_test_list.txt
```diff
@@ -28,6 +28,7 @@ shader-with-invalid-characters.html
 shader-with-mis-matching-uniform-block.html
 short-circuiting-in-loop-condition.html
 texture-offset-out-of-range.html
+texture-offset-uniform-texture-coordination.html
```
kenrussell: Throughout this patch, please rename "coordination" to "coordinate".

I think it'd also be OK to move this new test to 2.0.1 if you think fixing the driver bug is not feasible in the short term.

sdk/tests/conformance2/glsl3/texture-offset-uniform-texture-coordination.html
kenrussell: Right. It should be something like "Out-of-range texture offset expressed as uniform variable should not crash."

sdk/tests/conformance2/glsl3/texture-offset-uniform-texture-coordination.html
```diff
@@ -0,0 +1,191 @@
...
+
+if (!gl) {
+    testFailed("Unable to initialize WebGL 2.0 context.");
+} else {
+    GLSLConformanceTester.runTests([
+    {
+        fShaderSource: shaderTextureOffsetSrc,
+        fShaderSuccess: true,
+        linkSuccess: true,
+        passMsg: 'textureOffset with uniform texture coordination should compile'
```
kenrussell: Should this be "...should not crash" here and throughout?

kenrussell:
With @zhenyao 's comments addressed this looks good to me.

@msc- is exactly right that there is a strict policy to not work around driver bugs in the WebGL conformance tests. However, as he points out, it's OK to move tests that accidentally uncovered a driver bug from one file to another.

sdk/tests/conformance2/glsl3/texture-offset-uniform-texture-coordination.html
Jiawei-Shao: Sorry for my carelessness......

kenrussell:
Thanks for updating this pull request. LGTM, but @zhenyao should re-review and merge.

zhenyao:
lgtm

PR #1965, opened 2016-08-12
CopyTexSubImage3D should leave pixels outside the framebuffer untouched.

kenrussell:
This change was made for CopyTexSubImage2D some time ago, but the
change was never propagated to the WebGL 2.0 specification.


kenrussell:
Thanks @Richard-Yunchao for tracking this down in http://crbug.com/636987 .

@jdashg @zhenyao @grorg @RafaelCintron please review.

zhenyao:
lgtm

PR #1967, opened 2016-08-15
Fix the qualifiers mismatch error in differing_interpolation_2

jchen10:
The differing_interpolation_2 case fails in my MAC. According to the ELSL 3.0 spec, it seems MAC is right on this. See the below texts at the end of section 4.3.4 and 4.3.9.

4.3.4
"The output of the vertex shader and the input of the fragment shader form an interface.  For this interface, vertex shader output variables and fragment shader input variables of the same name must match in type and qualification (other than out matching to in)."
4.3.9
"The type and presence of the interpolation qualifiers and storage qualifiers and invariant qualifiers of 
variables with the same name declared in all linked shaders must match, otherwise the link command will 
fail."

### DIFF
--- a/sdk/tests/deqp/data/gles3/shaders/linkage.test
+++ b/sdk/tests/deqp/data/gles3/shaders/linkage.test
@@ -506,7 +506,7 @@ group varying "Varying linkage"
 			vertex ""
 				#version 300 es
 				${VERTEX_DECLARATIONS}
-				smooth out mediump float var;
+				smooth centroid out mediump float var;
 				void main()
 				{
 					var = in0;


jchen10:
@Richard-Yunchao @zhenyao Please have a look, Thanks!

qkmiao:
ANGLE doesn't follow ES 3.0 spec here intentionally. Please see https://cs.chromium.org/chromium/src/third_party/angle/src/compiler/translator/ShaderVars.cpp?q=ShaderVars&sq=package:chromium&l=26.

jchen10:
@qkmiao Thanks for the clarification!

Richard-Yunchao:
@jchen10 , you need to check the OGL spec and find out the difference between OGL / GLES.

jchen10:
The GL version of my MAC is 4.1.
The corresponding texts in GLSL4.1 is basically same as ELSL3.0.
"The fragment shader inputs form an interface with the last active shader in the vertex processing pipeline.For this interface, the last active shader stage  output variables and fragment shader input variables of the same name must match in type and qualification (other than out matching to in)."

"The type and presence of the interpolation qualifiers and storage qualifiers and invariant qualifiers of
variables with the same name declared in all linked shaders must match, otherwise the link command will fail. "

jchen10:
So should we fix the case or walk around it in ANGLE for MAC?

qkmiao:
I think you should workaround it in ANGLE.
@Kangz @null77 @zhenyao @kenrussell what's your opinion?

jchen10:
In sdk/tests/deqp/data/gles3/shaders/qualification_order.test, there are a few similar failures on MAC.
The cause is that an "invariant smooth centroid out lowp float x0;" in vextex shader VS "smooth centroid in lowp float x0;" in fragment shader doesn't follow the "qualifiers must be same" rule. So the MAC driver complains "Input of fragment shader ‘x0’ differs in type/qualifiers to that written by vertex shader". if add "invariant" to x0 in fragment shader as well, the qualifiers are then exactly same, and should pass on MAC according to the below text in the section of 4.6.1 in GLSL 4.1:
“Only variables output from a shader (including those that are then input to a subsequent shader) can be candidates for invariance. This includes user-defined output variables and the built-in output variables. For variables leaving one shader and coming into another shader, **the invariant keyword has to be used in both shaders, or a link error will result.**”.
But actually it fails too saying "Only output variables can be invariant". It seems MAC is messy regarding these rules.

With regard to GLSL4.4/4.5 and ELSL3.0 specs, there should be no need to add 'invariant' for fragment input. 
GLSL 4.4/4.5
“Only variables output from a shader (including those that are then input to a subsequent shader) can be candidates for invariance. This includes user-defined output variables and the built-in output variables. As only outputs need be declared with invariant, an output from one shader stage will still match an input of a subsequent stage without the input being declared as invariant.”

ELSL 3.0
“Only variables output from a shader can be candidates for invariance. This includes user-defined output variables and the built-in output variables. As only outputs can be declared as invariant, an invariant output from one shader stage will still match an input of a subsequent stage without the input being declared as invariant.”

zhenyao:
According to the ESSL 3.0 spec, it's quite clear that storage qualifiers are required to match.  The fact 3.1 relaxes on it should not change the behavior on 3.0 - I vote we follow the 3.0 requirement in WebGL2. We do the validation in Chrome, making sure the same behavior everywhere.

kenrussell:
The statement in the GLSL ES 3.00.4 specification, section 4.3.9 "Interpolation", page 45:
"The type and presence of the interpolation qualifiers and storage qualifiers and invariant qualifiers of variables with the same name declared in all linked shaders must match, otherwise the link command will fail."

seems to be incorrect. In ESSL 3.00 shaders, the invariant qualifier can only be applied to output variables, not input variables (section 4.6.1, "The Invariant Qualifier", page 52: "Only variables output from a shader can be candidates for invariance."). Therefore the statement in 4.3.9 is suspect. We need to evaluate it carefully and see what is actually implemented by the OpenGL drivers. (Note that the invariance rules are different between ESSL 1.00 and ESSL 3.00, and, I think, between GLSL 1.00 and GLSL 3.00 shaders. I think that macOS' Core Profile implements invariance matching rules which are compatible with those in ESSL 3.00.)

Probably the best rule to apply for WebGL 2.0 is that the centroid qualifier must match between the vertex and fragment shaders. Is that compatible with the behavior that macOS enforces?

jchen10:
Yes, The case with this patch passed on MacBook and MacMini per my tests. Moreover, it seems Mac rigidly and blindly enforce that all qualifiers must strictly match.

zhenyao:
Further discussion with @kenrussell , we agreed the best way to move forward is to drop "centroid" keywords on MacOSX in ANGLE translator as a driver bug workaround.

kenrussell:
To expand upon this decision: the dEQP tests currently enforce that the centroid qualifiers don't need to match between the vertex and fragment shaders. The GLSLang 4.1 spec is very poor quality in its description of these restrictions; some of its paragraphs are not even self-consistent. Example from section 4.6.1 "The Invariant Qualifier", p. 57 in GLSLangSpec.4.10.6.clean.pdf:

"Only variables output from a shader (including those that are then input to a subsequent shader) can be candidates for invariance...For variables leaving one shader and coming into another shader, the invariant keyword has to be used in both shaders, or a link error will result."

These two sentences are directly contradictory.

For this reason we are not going to enforce the poorly informed restrictions from macOS and earlier versions of the GLSLang spec on WebGL 2.0. Instead, we will do whatever is needed to the shaders to make them compile. I suspect that dropping the centroid qualifier on macOS will have only a limited impact on applications and we will try that as the first approach.

I'm closing this pull request to indicate that we do not plan to modify this test.

Thanks for discovering and raising this issue.

msc-:
> "Only variables output from a shader (including those that are then input to a subsequent shader) can be candidates for invariance...For variables leaving one shader and coming into another shader, the invariant keyword has to be used in both shaders, or a link error will result."

>These two sentences are directly contradictory.

I don't see the contradiction. The parenthetical expression explicitly extends "candidates for invariance." I agree with @zhenyao that WebGL 2 should follow the ESSL 3.0 spec. Upthread it seemed to be invariant that was causing the problem yet it is suddenly announced that centroid will be dropped. Typo?

zhenyao:
The thing is ES3 DEQP verifies otherwise. ES3 DEQP tests requires "centroid out" and "in" to match and link to succeed.  We consider DEQP part of the extended spec, so it is clear without ambiguity what the expected driver behavior is.

kenrussell:
It's the centroid qualifier that's causing the problem here. macOS' GLSL compiler and linker correctly applies the invariance rules for #version 150 shaders (which is what ANGLE translates ESSL 3.00 shaders to).

Silently dropping the centroid qualifier when translating ESSL 3.00 shaders on macOS will work around this problem.

PR #1969, opened 2016-08-15
Increase the timeout delay for ext-disjoint-timer-query.html

qkmiao:
These tests will be timeout on some slow platforms.

### DIFF
--- a/sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
+++ b/sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
@@ -65,6 +65,11 @@ if (!gl) {
         testPassed("No EXT_disjoint_timer_query support -- this is legal");
         finishTest();
     } else {
+        // On some platforms, tests in this file will consume more than the default 20 seconds.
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
```diff
@@ -78,6 +83,9 @@
         verifyQueryResultsNotAvailable();
 
         window.requestAnimationFrame(checkQueryResults);
+
+        if (window.parent.webglTestHarness)
+            window.parent.webglTestHarness.setTimeoutDelay(20000);
```
zhenyao: This is very flaky. In case we decide to change the global timeout delay, we most likely won't remember this place.

Is there a way to have a general mechanism to mark the test as "slow" and automatically double the timeout (and set it back afterwards)?

kenrussell:
Sorry, but this change does not look good to me.

There is a pretty obvious bug in the availability check at the bottom of this test which I'm addressing in #1971 . Can you test that fix on your machines, and does it fix this problem?

kenrussell:
Let's close this in favor of #1971 .

PR #1970, opened 2016-08-15
Fix a typo in deqp

qkmiao:


zhenyao:
lgtm

kenrussell:
Thanks for fixing this. It's been annoying to look at. We'll have to update the associated suppression on Mac AMD in webgl2_conformance_expectations.py when rolling this change into Chrome.

PR #1971, opened 2016-08-15
Fix query availability check in EXT_disjoint_timer_query test.

kenrussell:
This should address the issue that @qkmiao reported in #1969 .

### DIFF
--- a/sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
+++ b/sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
@@ -241,18 +241,21 @@ function verifyQueryResultsNotAvailable() {
 }
 
 function checkQueryResults() {
-    if (availability_retry > 0 && (timestamp_counter_bits == 0 ||
-        !ext.getQueryObjectEXT(timestamp_query2, ext.QUERY_RESULT_AVAILABLE_EXT))) {
-        var error = gl.getError();
-        if (error != gl.NO_ERROR) {
-            testFailed("getQueryObjectEXT should have no errors: " + wtu.glEnumToString(gl, error));
-            debug("");
-            finishTest();
+    if (availability_retry > 0) {
+        // Make a reasonable attempt to wait for the queries' results to become available.
+        if (!ext.getQueryObjectEXT(elapsed_query, ext.QUERY_RESULT_AVAILABLE_EXT) ||
+            (timestamp_counter_bits > 0 && !ext.getQueryObjectEXT(timestamp_query2, ext.QUERY_RESULT_AVAILABLE_EXT))) {
+            var error = gl.getError();
+            if (error != gl.NO_ERROR) {
+                testFailed("getQueryObjectEXT should have no errors: " + wtu.glEnumToString(gl, error));
+                debug("");
+                finishTest();
+                return;
+            }
+            availability_retry--;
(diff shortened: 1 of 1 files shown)


kenrussell:
@zhenyao in particular, please review.

sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
```diff
@@ -241,18 +241,22 @@
...
+            (timestamp_counter_bits > 0 && !ext.getQueryObjectEXT(timestamp_query2, ext.QUERY_RESULT_AVAILABLE_EXT))) {
+            var error = gl.getError();
+            if (error != gl.NO_ERROR) {
+                testFailed("getQueryObjectEXT should have no errors: " + wtu.glEnumToString(gl, error));
+                debug("");
+                finishTest();
+                return;
+            }
+            availability_retry--;
+            gl.finish();
```
zhenyao: Just try to fully understand, so this is what makes a difference in timing? Is it better if we just call it right before the first requestAnimationFrame() in the main function?

sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
```diff
@@ -241,18 +241,22 @@
...
-        !ext.getQueryObjectEXT(timestamp_query2, ext.QUERY_RESULT_AVAILABLE_EXT))) {
-        var error = gl.getError();
-        if (error != gl.NO_ERROR) {
-            testFailed("getQueryObjectEXT should have no errors: " + wtu.glEnumToString(gl, error));
-            debug("");
-            finishTest();
+    if (availability_retry > 0) {
+        // Make a reasonable attempt to wait for the queries' results to become available.
+        if (!ext.getQueryObjectEXT(elapsed_query, ext.QUERY_RESULT_AVAILABLE_EXT) ||
+            (timestamp_counter_bits > 0 && !ext.getQueryObjectEXT(timestamp_query2, ext.QUERY_RESULT_AVAILABLE_EXT))) {
```
qkmiao: Per my testing, this line makes a difference in timing. Total time of this test file reduces to ~2 seconds from 20+ seconds with this pr at my side.

sdk/tests/conformance/extensions/ext-disjoint-timer-query.html
kenrussell: Actually, the gl.finish() is not really relevant. I can remove it if it is causing confusion.

The reason the test runs much faster is that the if-test above was wrong. timestamp_counter_bits is always 0 on macOS, and I think, actually 0 on all platforms Chrome runs on. This was causing this part of the test to always count down availability_retry from 512 to 0 before proceeding.

Please let me know if you'd rather I remove the gl.finish() call.

zhenyao: I see. Thanks for the explanation.  Could you please remove gl.finish() if it's not needed? It does confuse me (and probably others who look at conformance tests for guidance on how to use a feature).

zhenyao: Please merge yourself after the line is removed.

kenrussell: Done. Merging.

PR #1763, opened 2016-06-10
Add a no alpha context test.

pavolas:
Add a no alpha context test.

The test makes a no alpha context, and a framebuffer/texture with an alpha channel. The test checks that working with the framebuffer/texture doesn't affect the back buffer.

The newly added test fails without https://codereview.chromium.org/2053983002/#, but succeeds with the CL.

PR #1973, opened 2016-08-18
Fixes to no-alpha context test.

kenrussell:
These fixes make the test from #1763 pass on top of tree Chromium. In
particular it was necessary to use a tolerance of 1.


kenrussell:
@zhenyao please review.

@pavolas FYI.

zhenyao:
lgtm

PR #1974, opened 2016-08-20
Add test cases for buffer{Sub}Data with ArrayBufferView sub source.

zhenyao:
@jdashg @kenrussell Please review

### DIFF
--- a/sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
+++ b/sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
@@ -1,0 +1,204 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+
+<script>
+"use strict";
+description("Test bufferData/bufferSubData with ArrayBufferView sub source input");
+
+var wtu = WebGLTestUtils;
+var gl = wtu.create3DContext(undefined, undefined, 2);
+
+function verifyBufferData(testCase, sourceByteOffset, size, data) {
+    var readbackBuffer = new ArrayBuffer(testCase.size * size);
+    var readbackView = new window[testCase.type](readbackBuffer);
+    gl.getBufferSubData(gl.ARRAY_BUFFER, sourceByteOffset, readbackView);
+    var pass = true;
+    var offset = sourceByteOffset / testCase.size;
+    for (var ii = 0; ii < size; ++ii) {
+        if (readbackView[ii] != data[ii]) {
+            testFailed("expected data at " + ii + ": " + data[iit] + ", got " + readbackView[ii]);
+            pass = false;
+        }
+    }
+    if (pass) {
+        testPassed("buffer data uploaded correctly");
+    }
+}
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
```diff
@@ -0,0 +1,207 @@
...
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+
+<script>
+"use strict";
+description("Test bufferData/bufferSubData with ArrayBufferView sub source input");
+
+var testCases = [
```
kenrussell: I'd like to suggest moving the testCases variable into the "else" branch just before calling bufferDataTest and bufferSubDataTest, and passing it as argument to each of those calls. That will reduce interdependencies in the code.

sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
```diff
@@ -0,0 +1,207 @@
...
+
+var wtu = WebGLTestUtils;
+var gl = wtu.create3DContext(undefined, undefined, 2);
+if (!gl) {
+    testFailed("WebGL context does not exist");
+} else {
+    testPassed("WebGL context exists");
+
+    bufferDataTest();
+    bufferSubDataTest();
```
kenrussell: It'd be easier to understand if the dependent functions were declared before this one, so that this block of code was the last code in this function.

sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
```diff
@@ -0,0 +1,207 @@
...
+            testFailed("expected data at " + ii + ": " + data[iit] + ", got " + readbackView[ii]);
+            pass = false;
+        }
+    }
+    if (pass) {
+        testPassed("buffer data uploaded correctly");
+    }
+}
+
+function bufferDataTest() {
```
kenrussell: I think it'd be cleaner if this took "testCases" as an argument instead of referencing it as a global variable. Would you make that change? Here and for bufferSubDataTest.

sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
```diff
@@ -0,0 +1,207 @@
...
+    for (var test = 0; test < testCases.length; ++test) {
+        debug("");
+        debug("Test bufferData with " + testCases[test].type);
```
kenrussell: Renaming "test" to "idx" and making a variable like "var test = testCases[idx];" would reduce a fair amount of duplicated code.

sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
```diff
@@ -0,0 +1,207 @@
...
+        var sourceByteOffset = testCases[test].size * 2;
+
+        gl.bufferSubData(gl.ARRAY_BUFFER, sourceByteOffset, view, 0, count + 1);
+        wtu.glErrorShouldBe(gl, gl.INVALID_VALUE,
+            "calling bufferSubData when srcOffset + length is larger than source size");
+
+        gl.bufferSubData(gl.ARRAY_BUFFER, sourceByteOffset, view, count + 1);
+        wtu.glErrorShouldBe(gl, gl.INVALID_VALUE,
+            "calling bufferSubData when srcOffset + length is larger than source size");
+        gl.bufferSubData(gl.ARRAY_BUFFER, sourceByteOffset, view, count + 1, 0);
```
kenrussell: Since the "optional GLuint length" argument defaults to 0, isn't this actually the same as the call above? It doesn't matter whether 0 is explicitly provided; it always means to upload the "rest" of the given view.

sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        gl.bufferData(gl.ARRAY_BUFFER, view, gl.STATIC_DRAW, 0, count + 1);
+        wtu.glErrorShouldBe(gl, gl.INVALID_VALUE,
+            "calling bufferData when srcOffset + length is larger than source size");
+
+        gl.bufferData(gl.ARRAY_BUFFER, view, gl.STATIC_DRAW, count + 1);
+        wtu.glErrorShouldBe(gl, gl.INVALID_VALUE,
+            "calling bufferData when srcOffset + length is larger than source size");
+        gl.bufferData(gl.ARRAY_BUFFER, view, gl.STATIC_DRAW, count + 1, 0);
```
kenrussell: Same comment as for the bufferSubData version of this test, below.

kenrussell:
Thanks @zhenyao for putting these tests together. They look good overall. A few minor comments.

The Travis CI failure is a random transient failure as far as I can tell. Don't worry about it.

sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
zhenyao: Done

sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
zhenyao: Done

sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
zhenyao: Done

sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
zhenyao: Done

sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
zhenyao: Removed, here and below

sdk/tests/conformance2/buffers/buffer-data-and-buffer-sub-data-sub-source.html
zhenyao: Done

zhenyao:
Revised.  Please take another look.

kenrussell:
Thanks Mo for putting together these tests. LGTM. Merging now.

PR #1976, opened 2016-08-22
Add testcase to test luminance alpha format for copytexsubimage3d.

JiangYizhou:


sdk/tests/conformance2/textures/misc/00_test_list.txt
```diff
@@ -1,6 +1,7 @@
 compressed-tex-image.html
 copy-texture-image.html
 copy-texture-image-webgl-specific.html
+copy-texture-image-luma-format.html
```
zhenyao: move this one up - alphabetic order

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,241 @@
...
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL vertex_array_object Conformance Tests</title>
```
zhenyao: wrong title

JiangYizhou:
@zhenyao @Richard-Yunchao Thank you for your review, I have revised the code. PTAL.

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,240 @@
...
+        for (var i = 0; i < uint2.length; ++i) {
+            uint2[i] = 0xFF;
+        }
+
+        gl.bindTexture(gl.TEXTURE_2D, texture[2]);
+        gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, width, height, 0, gl.RGBA, gl.UNSIGNED_BYTE, uint);
+
+        for (var i = 0; i < 2; ++i) {
+            if (i == 0) {
+                var program = wtu.setupProgram(gl, ["vshader", "fshader_rgba_" + testcase.name], ["a_position", "a_coord"], [0, 1]);
```
zhenyao: This var is local to the if block

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,240 @@
...
+    o_color0 = vec4(0, 0, 0, alpha);
+}
+</script>
+<script id="fshader_rgba_3d_luminance" type="x-shader/x-fragment">#version 300 es
+in highp vec2 v_coord;
+uniform highp sampler3D u_sampler0;
+uniform highp float u_depth;
+out highp vec4 o_color0;
+void main (void) {
+    highp float r = texture(u_sampler0, vec3(v_coord, u_depth)).r;
```
zhenyao: For this test setup it's fine, because r==g==b, but it's better (and easier to read) to actually calculate luminance from rgb.

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,240 @@
...
+    o_color0 = vec4(r, r, r, 1);
+}
+</script>
+<script id="fshader_rgba_3d_luminance_alpha" type="x-shader/x-fragment">#version 300 es
+in highp vec2 v_coord;
+uniform highp sampler3D u_sampler0;
+uniform highp float u_depth;
+out highp vec4 o_color0;
+void main (void) {
+    highp float r = texture(u_sampler0, vec3(v_coord, u_depth)).r;
```
zhenyao: Same here.

JiangYizhou:
@zhenyao  I'm a little confused, do you mean lumiance = (r + g + b) / 3 ?

zhenyao:
Luminance =0.2126R+0.7152G+0.0722B

zhenyao:
If it's not too much work, it's better testing luminance with R!=G!=B

Richard-Yunchao:
@zhenyao , I am confused by the Chromium implementation about emulation luminance color format in gles2_cmd_copy_tex_image.cc|.h. Because I didn't find any code to calculate of this equation: 
Luminance =0.2126R+0.7152G+0.0722B. 
How could the copyTex{Sub}Image{2D|3D} to copy a color buffer with RGB related color format from framebuffer to a texture with Luminance format? It seems to me the shader in gles2_cmd_copy_tex_image just copy RED channel from color buffer to the target texture (Maybe I missed something in Chromium Implementation).

zhenyao:
@vonture Can you comment on @Richard-Yunchao 's question?

vonture:
For RGB->L conversion, the ES spec says that the L channel should just copy the red channel.

See pg 169 of the ES 3.2 spec (https://www.khronos.org/registry/gles/specs/3.2/es_spec_3.2.pdf):
> When matching formats that involve a luminance component, a luminance component is considered to match with a red component

zhenyao:
My reading of the spec is slightly different.  That part is talking about channel matching mechanism, but not color conversion.  Say, you have a RGB color (0, 255, 255), and luminance ends up as 0, that seems wrong to me.

@kenrussell Can you provide a second opinion?

vonture:
I agree that it doesn't seem like a very useful conversion.  There are dEQP tests that cover this and I recently updated ANGLE's logic from an averaging function to an R->L copy.

kenrussell:
My understanding of the spec matches @vonture 's. I think it should only be copying the red channel into the luminance texture, not averaging the color channels.

JiangYizhou:
@zhenyao @kenrussell @Richard-Yunchao Thank you for your review, I have updated the code. PTAL.

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,240 @@
...
+                    gl.activeTexture(gl.TEXTURE0);
+                    setUpTexStatus();
+                    gl.uniform1i(sampler, 0);
+                    gl.clear(gl.COLOR_BUFFER_BIT);
+                }
+                wtu.drawUnitQuad(gl);
+
+                bytes[i] = new Uint8Array(width * height * 4);
+                gl.readPixels(0, 0, width, height, gl.RGBA, gl.UNSIGNED_BYTE, bytes[i]);
+                wtu.glErrorShouldBe(gl, gl.NO_ERROR, "readpixel should succeed.");
```
kenrussell: I think this should use wtu.checkCanvasRect rather than doing readPixels directly. That way a tolerance can be easily added. Is there a reason to not do this?

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,240 @@
...
+        var layer = 0;
+        var width = testcase.width;
+        var height = testcase.height;
+        var depth = testcase.depth;
+        var w = testcase.width;
+        var h = testcase.height;
+        var format;
+        var uint = new Uint8Array(width * height * 4);
+        for (var i = 0; i < uint.length; ++i) {
+            uint[i] = Math.random() * 0xFF;
```
kenrussell: It's problematic to use fully random numbers in these tests. This can cause hard-to-reproduce bugs. Please figure out a way to make this test deterministic.

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,240 @@
...
+            }
+
+            gl.bindFramebuffer(gl.FRAMEBUFFER, null);
+            gl.deleteFramebuffer(fbo);
+            gl.deleteTexture(texture[i]);
+            gl.deleteProgram(program);
+        }
+
+        for(var i = 0; i < width * height * 4; ++i) {
+            if(bytes[0][i] != bytes[1][i]) {
```
kenrussell: The strict equality here is problematic. I think the test should allow a tolerance. See above.

JiangYizhou:
@kenrussell   luminance is read from  red channel, so I wonder why does it need a tolerance here?

kenrussell:
Since we're rendering from the texture, sampling from it, I assume that a small error might plausibly be introduced in the rendering pipeline. I'd suggest a tolerance of 1.

JiangYizhou:
@kenrussell Thank you, I have revised the code. PTAL.

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,245 @@
...
+out highp vec4 o_color0;
+void main (void) {
+    highp float r = texture(u_sampler0, vec3(v_coord, u_depth)).r;
+    highp float alpha = texture(u_sampler0, vec3(v_coord, u_depth)).a;
+    o_color0 = vec4(r, r, r, alpha);
+}
+</script>
+<script>
+"use strict";
+description("This test verifies the functionality of the coptexsubimage3d.");
```
kenrussell: typo: copyTexSubImage3D

I'd suggest the following rewrite:

"This test verifies the behavior of copyTexSubImage3D with luminance textures."

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,245 @@
...
+                format = testcase.format;
+            }
+            gl.bindTexture(gl.TEXTURE_2D, texture[2]);
+            gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, width, height, 0, gl.RGBA, gl.UNSIGNED_BYTE, uint);
+            var fbo = gl.createFramebuffer();
+            gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+            gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture[2], 0);
+            if (gl.checkFramebufferStatus(gl.FRAMEBUFFER) == gl.FRAMEBUFFER_COMPLETE) {
+                gl.bindTexture(gl.TEXTURE_3D, texture[i]);
+                gl.texImage3D(gl.TEXTURE_3D, 0, format, width, height, depth, 0, format, gl.UNSIGNED_BYTE, uint2);
```
kenrussell: Why is "uint2" passed in here? All that's needed is for the 3D texture to be allocated with the correct width, height, and depth, correct? Why not pass in null? It would simplify the test.

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,245 @@
...
+        var bytes;
+        var layer = 0;
+        var width = testcase.width;
+        var height = testcase.height;
+        var depth = testcase.depth;
+        var w = testcase.width;
+        var h = testcase.height;
+        var format;
+        var msg;
+        var uint = new Uint8Array(width * height * 4);
```
kenrussell: It would be clearer if this were named "uint1" or "array1" or something similar. "uint" looks like a built-in symbol to me.

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,245 @@
...
+                    gl.activeTexture(gl.TEXTURE0);
+                    setUpTexStatus();
+                    gl.uniform1i(sampler, 0);
+                    gl.clear(gl.COLOR_BUFFER_BIT);
+                }
+                wtu.drawUnitQuad(gl);
+
+                if(i == 0) {
+                    bytes = new Uint8Array(width * height * 4);
+                    gl.readPixels(0, 0, width, height, gl.RGBA, gl.UNSIGNED_BYTE, bytes);
```
kenrussell: Why is it necessary to do a readPixels here? Isn't the intent that these bytes be equivalent to "uint"? Reading back this data from the framebuffer increases the likelihood that multiple bugs in the WebGL implementation may cause the test to pass.

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,245 @@
...
+                var sampler = gl.getUniformLocation(program, 'u_sampler0');
+                if(sampler) {
+                    gl.activeTexture(gl.TEXTURE0);
+                    setUpTexStatus();
+                    gl.uniform1i(sampler, 0);
+                    gl.clear(gl.COLOR_BUFFER_BIT);
+                }
+                wtu.drawUnitQuad(gl);
+
+                if(i == 0) {
```
kenrussell: Why are no checks done on iteration 0? I find the intent of this test difficult to follow. Please at least add some comments about what it's doing on what iteration.

kenrussell:
Jiang, thanks for continuing to work on this test but I think it needs more work. Comments are inline in the code.

JiangYizhou:
@kenrussell Thanks for your review, I have updated the code. Could you have another look?

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,187 @@
...
+            var program = wtu.setupProgram(gl, ["vshader", "fshader_luminance_alpha"]);
+            wtu.setupUnitQuad(gl, 0, 1);
+            wtu.drawUnitQuad(gl);
+
+            for (var y = 0; y < height; ++y) {
+                for (var x = 0; x < width; ++x) {
+                    var cur = y * width * 4 + x * 4;
+                    if (testcase.format == gl.ALPHA) {
+                        wtu.checkCanvasRect(gl, x, y, 1, 1, [ 0, 0,
+                        0, uint1[cur + 3]], msg, [1, 1, 1, 1]);
```
kenrussell: The formatting is weird here and below. Please look into re-indenting this later. You can align to the open parenthesis "(" or at least use four space indent.

sdk/tests/conformance2/textures/misc/copy-texture-image-luma-format.html
```diff
@@ -0,0 +1,187 @@
...
+in highp vec2 v_coord;
+uniform highp sampler3D u_sampler0;
+out highp vec4 o_color0;
+void main (void) {
+    o_color0 = vec4(texture(u_sampler0,vec3(v_coord, 0)));
+}
+</script>
+<script>
+"use strict";
+description("This test verifies the behavior of copTexSubImage3D with luminance textures.");
```
kenrussell: typo: copyTexSubImage3D

kenrussell:
Jiang, thanks, the revised test looks excellent. Merging now. A couple of minor cleanups for later.

PR #1977, opened 2016-08-22
Some anti-alias algorithms may not effect on diagonal line pixels, but must preserve the behavior on the non-edge pixel.

ds-hwang:
Some anti-alias algorithms may not effect on 2x2 canvas.

A postprocessing AA cannot handle 2x2 canvas because the following example cannot
be interpreted as triangle.
```
    _____
    |X| |
    -----
    | | |
    ¯¯¯¯¯
```
So enlarge the canvas to 3x3, because the following example can be interpreted
as triangle.
```
    _______
    |X|X|X|
    -------
    |X|X| |
    -------
    |X| | |
    ¯¯¯¯¯¯¯
```

Follow-up of PR #1730
BUG=crbug.com/634945

### DIFF
--- a/sdk/tests/conformance/context/context-attributes-alpha-depth-stencil-antialias.html
+++ b/sdk/tests/conformance/context/context-attributes-alpha-depth-stencil-antialias.html
@@ -62,8 +62,7 @@ void main()
 var wtu = WebGLTestUtils;
 var gl;
 var contextAttribs = null;
-var pixel_1 = [0, 0, 0, 1];
-var pixel_2 = [0, 0, 0, 1];
+var redChannels = [0, 0, 0];
 var correctColor = null;
 var framebuffer;
 var fbHasColor;
@@ -303,11 +302,15 @@ function testAntialias(antialias)
 {
     debug("Testing antialias = " + antialias);
     if (antialias)
-        shouldBeNonNull("gl = getWebGL(2, 2, { depth: false, stencil: false, alpha: false, antialias: true }, [ 0, 0, 0, 1 ], 1, 0)");
+        shouldBeNonNull("gl = getWebGL(3, 3, { depth: false, stencil: false, alpha: false, antialias: true }, [ 0, 0, 0, 1 ], 1, 0)");
     else
-        shouldBeNonNull("gl = getWebGL(2, 2, { depth: false, stencil: false, alpha: false, antialias: false }, [ 0, 0, 0, 1 ], 1, 0)");
+        shouldBeNonNull("gl = getWebGL(3, 3, { depth: false, stencil: false, alpha: false, antialias: false }, [ 0, 0, 0, 1 ], 1, 0)");
     shouldBeNonNull("contextAttribs = gl.getContextAttributes()");
 
+    // Draw a triangle on 3x3 canvas.
+    // ----
+    // | /
+    // |/
     var vertices = new Float32Array([
          1.0, 1.0, 0.0,
         -1.0, 1.0, 0.0,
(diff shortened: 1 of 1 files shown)


ds-hwang:
@zhenyao  could you review for https://bugs.chromium.org/p/chromium/issues/detail?id=634945 ?
PR #1730 changes the test to pass CMAA wrong behavior.
As [the CMAA bug is fixed](https://chromium.googlesource.com/chromium/src/+/9df37c48526a8f6b350687e34e75b2eca1961736), this test needs to be more reasonable. See the patch description. 


@xhcao @gyagp Could you check?

kenrussell:
This seems OK to me but @zhenyao should review too.

I edited the PR description to get the ASCII diagrams to look better. Please double-check and see if they still match your intent.

sdk/tests/conformance/context/context-attributes-alpha-depth-stencil-antialias.html
```diff
@@ -323,11 +323,14 @@
...
-    //    - If it is 0, expect that the red channel of the pixel at (0, 1) is not 0 and not 255.
-    shouldBe("pixel_1[0] != 255 && pixel_1[0] != 0 || pixel_1[0] == 0 && pixel_2[0] != 255 && pixel_2[0] != 0",
-        "contextAttribs.antialias");
+    if (!contextAttribs.antialias) {
+        shouldBeTrue("pixel_1[0] == 0 && pixel_2[0] == 255");
+    } else {
+        // For some anti-alias algorithms, effects may be not on diagonal line pixels, so that either:
+        //    - The red channel of the pixel at (0, 0) is not 0 and not 255, or,
+        //    - If it is 0, expect that the red channel of the pixel at (0, 1) is 255.
+        shouldBeTrue("(pixel_1[0] != 255 && pixel_1[0] != 0) || (pixel_1[0] == 0 && pixel_2[0] == 255)");
```
zhenyao: So for the second case, how do you verify it is antialiased? My concern is now any implementation could just pass this test without doing any antialiasing.

ds-hwang: Exactly. I thought same concern when I was lying on the bed last night.
The new patch set resolves this issue.

ds-hwang:
@zhenyao could you review again?

There are 2 changes;
 1. enlarge the triangle to ensure the triangle to include the sampling point in the edge pixels.
 2. extend canvas from 2x2 to 3x3

For 2nd change, a postprocessing AA cannot handle 2x2 canvas because the following example cannot be interpreted as triangle.
```
    _____
    |X| |
    -----
    | | |
    ¯¯¯¯¯
```
So enlarge the canvas to 3x3, because the following example can be interprested as triangle.
```
    _______
    |X|X|X|
    -------
    |X|X| |
    -------
    |X| | |
    ¯¯¯¯¯¯¯
```

gyagp:
It seems to me a reasonable limitation that CMAA couldn't handle 2*2 canvas as in this case (Thanks DS for the clarification). 
But I don't think we need to do first change (enlarge the triangle to ensure ...) as general application wouldn't do this. 
As a simple solution, how about just changing the canvas size to 3*3, if we all agree 2*2 is an acceptable exception?

ds-hwang:
@gyagp you are right. It's not necessary the enlargement change. I removed the code.
It's because both following two cases apply AA on the middle pixel anyway.
```
    _______
    |X|X| |
    -------
    |X| | |
    -------
    | | | |
    ¯¯¯¯¯¯¯
    _______
    |X|X|X|
    -------
    |X|X| |
    -------
    |X| | |
    ¯¯¯¯¯¯¯
```

zhenyao:
lgtm

kenrussell:
LGTM after the fact. Agree with the analysis to increase the canvas's size to 3x3.

gyagp:
Sorry, I still think the new method to check the red channel from left top to right bottom is not good, as it couldn't apply to 4x4 canvas because all the colors from left top to right bottom are either 0 or 255 for MSAA. In general, for canvas size NxN, current method can only cover the condition when N is an odd number, but fails with an even N. 
The previous method doesn't have this problem. It always check a pixel on the diagonal and a pixel above it. AA works well if one of the pixels has red channel value between 0 and 255 (not including 0 and 255).

ds-hwang:
@gyagp this method works on both odd and even sized canvas. At even case, the value of two edge pixels is between 0 and 255.
For example, in 4x4 canvas, (1, 2) and (2, 1) pixels have value between 0 and 255, while (0, 3) and (3, 0) pixels have value either 0 or 255.

gyagp:
@ds-hwang I mean MSAA, not CMAA. For MSAA (we still need to take care of this as it's more common than CMAA), red channel of (1, 2) is 0, while it's 255 for (2, 1). In general for MSAA, only the pixels on diagonal from left bottom to right top are 127, and others are either 0 or 255. The two diagonals (left bottom to right top and right bottom to left top) have no matched pixel for NxN canvas if N is even.

ds-hwang:
@gyagp I have to say disagree. In MSAA, (1, 2) and (2, 1) pixels have value between 0 and 255.
Let me give a counter example. This test before @xhcao's change creates 2x2 canvas and check (0,0) pixel is value between 0 and 255 in MSAA. If your explanation is right, (0,0) should be 0 or 255.

ds-hwang:
@gyagp I was wrong. In 4x4 canvas, (1,2) and (2,1) pixels have 0 or 255 in MSAA, but CMAA sets value between 0 and 255.

And I understand what you want. You want to preserve the removed ASSERT.
```
-    shouldBe("pixel_1[0] != 255 && pixel_1[0] != 0 || pixel_1[0] == 0 && pixel_2[0] != 255 && pixel_2[0] != 0",
-        "contextAttribs.antialias");
```
Because it is true in 4x4 case as follows
```
CMAA
           255 255 196 0
           255 196  58 0
pixel_2 -> 196  58  23 0
pixel_1 -> 0    23   0 0
```
```
MSAA
           255 255 255 128
           255 255 128   0
pixel_2 -> 255 128   0   0
pixel_1 -> 128   0   0   0
```

However, I think `pixel_1[0] == 0 && pixel_2[0] != 255 && pixel_2[0] != 0` is bad condition. `pixel_1[0]` can be non-zero in other post processing AA case. It very depends on AA algorithm as well as hardware sampling point position.

I think new condition is more intuitive, while it is not generic for even x even canvas.

kenrussell:
I think the new code is clearer. The upper left pixel should be fully within the triangle; the lower right pixel should be fully outside; and the middle pixel will be either fully filled or unfilled if antialiasing is off, but otherwise it will be some intermediate value.

gyagp:
The problem is for canvas with an even N, the middle pixel (actually there are two) is not on the diagonal from left bottom to right top, so it always has fully filled/unfilled value with MSAA. I had a offline sync with @ds-hwang, and we both agree to check pixel at (1, 1). I created a PR at https://github.com/KhronosGroup/WebGL/pull/1988, please review!
Hope it looks better:) And if you're OK with new PR, I will change the copies in other versions samely.

PR #1978, opened 2016-08-22
Add EXT_clip_cull_distance extension proposal.

kenrussell:
This fairly new OpenGL ES 3.0 / WebGL 2.0 extension has already been
requested by a couple of partners.

### DIFF
--- a/extensions/extension.xsl
+++ b/extensions/extension.xsl
@@ -343,9 +343,14 @@
               function.
             </li>
           </xsl:for-each>
+          <xsl:for-each select="input">
+            <li>
+                <code><xsl:call-template name="shader_variable"/></code> is a built-in input.
+            </li>
+          </xsl:for-each>
           <xsl:for-each select="output">
             <li>
-                <code><xsl:call-template name="shader_output"/></code> is a built-in output.
+                <code><xsl:call-template name="shader_variable"/></code> is a built-in output.
             </li>
           </xsl:for-each>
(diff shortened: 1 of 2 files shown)


kenrussell:
@zhenyao @jdashg @RafaelCintron @grorg in particular, please review. Thanks.

zhenyao:
So this doesn't exist on Mac (at least not my Mac with AMD GPU). Will this be a concern?

kenrussell:
On desktop platforms it's part of the OpenGL 4.5 core and was previously exposed as GL_ARB_clip_distance. It does look like macOS doesn't expose these on either the compatibility or core profile, but I'm not sure that should hold back exposing this as an extension.

jdashg:
Seems fine.

RafaelCintron:
I am fine with the features this extension enables.

I see that the backing Khronos extension requires ES 3.0, which means WebGL 2.0.  If the extension will greatly help WebVR and other scenarios, we should strongly consider exposing it to WebGL 1.0 developers on platforms that support it like D3D on Windows.

zhenyao:
Seems no objection so far. Let's merge now.

kenrussell:
Thanks all for your reviews. @RafaelCintron , yes, we can consider making the extension work with ES 2.0 and WebGL 1.0. First though let's prototype it on WebGL 2.0, where the underlying extension is targeted.

PR #1982, opened 2016-08-24
Buffer

zhenyao:
@kenrussell Please review.
@jdashg FYI.


kenrussell:
The texture tests look superb. I can't think of a single improvement. LGTM.

It looks like your branch needs to be rebased before committing this, since it contains a previously-merged commit. I'll work with you on this offline.

PR #1983, opened 2016-08-25
Add COPY_READ_BUFFER and COPY_WRITE_BUFFER to targets for bindBuffer.

kenrussell:
These were added elsewhere in the spec after issues were resolved some
time ago, but were not added to this detail table for this method.


kenrussell:
Thanks @kainino0x for pointing out this omission.

@zhenyao @jdashg PTAL.

zhenyao:
LGTM

jdashg:
LGTM

kenrussell:
Thanks for your reviews. It looks like Travis is down or busted right now so I'm merging this trivial change without waiting for it to complete.

PR #1984, opened 2016-08-25
Skip the texture_functions.texture.sampler2darrayshadow_vertex test.

vonture:
It falls into the same bucket as the other missing D3D11 shadow sampling
functions but was missed on the first pass of supressions.

See KhronosGroup#1870


zhenyao:
lgtm

PR #1986, opened 2016-08-26
Add a test to verify RGB8 renderbuffer/texture works

zhenyao:
(and they don't on MacOSX)

### DIFF
--- a/sdk/tests/conformance2/rendering/rgb-format-support.html
+++ b/sdk/tests/conformance2/rendering/rgb-format-support.html
@@ -1,0 +1,132 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/rendering/rgb-format-support.html
```diff
@@ -0,0 +1,132 @@
...
+description("Verify RGB/RGB8 textures and renderbuffers support");
+
+var gl = wtu.create3DContext(undefined, undefined, 2);
+
+function testRenderbuffer(width, height) {
+    var samples = gl.getInternalformatParameter(gl.RENDERBUFFER, gl.RGB8, gl.SAMPLES);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be no errors from getInternalformatParameter()");
+
+    if (!samples || samples.length == 0) {
+        testFailed("getInternalformatParameter on RGB8 fails to return valid sampels");
```
kenrussell: typo: samples

kenrussell:
Very nice. Thanks @zhenyao for putting this together.

@jdashg : FYI, this targeted test exposes driver bugs on macOS across all GPU types. We feel that these will be easy to work around in the browser which is why this test is being targeted for WebGL 2.0.0 instead of 2.0.1. Having consistent RGB8 renderbuffer and texture behavior will be beneficial to developers.

Merging this now. Travis is running slowly for Khronos pull requests due to interference from other projects. I've alerted the webmaster.

jdashg:
Haha, I figured someone would decide not to allow RGB8 to be renderable,
but I didn't expect it to be Apple!
This will be a larger pain to work around than you expect though, I think!
(copyTexImage, multisampling blits, and unpack formats)

I'd actually prefer to break compat for WebGL 2 here, and not guarantee
RGB8 renderability in WebGL2, adding this to the porting errata. It's
really quite easy for app devs to fix, and it saves implementations at
least some complexity. (that is, we only need to have workarounds
sufficient for WebGL 1)

kenrussell:
After discussion on the last WebGL working group conference call, perhaps the best thing to do is require that allocation of an RGB8 multisampled renderbuffer (with samples > 0) generate an INVALID_OPERATION. That's the minimum change to the spec that will prevent the creation of non-portable content.

PR #1987, opened 2016-08-26
Fix type.

zhenyao:
Add a few missed R11F_G11F_B10F test cases to skip list. (we have another test to expose the bug)


zhenyao:
Simple patch. Merging myself.

kenrussell:
LGTM after the fact.

PR #1988, opened 2016-08-26
Check the result from AA in a general way

gyagp:
When drawing a square with 2 different color triangles, some AA may
affect all the pixels on diagonal, while some others, such as CMAA, may
keep the pixels on the egde intact. Given this, (1, 1) can be always a
good candidate for the check.

### DIFF
--- a/sdk/tests/conformance/context/context-attributes-alpha-depth-stencil-antialias.html
+++ b/sdk/tests/conformance/context/context-attributes-alpha-depth-stencil-antialias.html
@@ -301,16 +301,15 @@ function testStencilAndDepth(stencil, depth)
 function testAntialias(antialias)
 {
     debug("Testing antialias = " + antialias);
+    // Both the width and height of canvas are N.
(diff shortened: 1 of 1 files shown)


ds-hwang:
lgtm

You may run the anti alias test 2 times on 3x3 and 4x4 canvas to justify this general change.

zhenyao:
This looks to me, but @kenrussell should also take a look.

kenrussell:
Thanks for this contribution. It looks good. Let's be careful to keep the tests readable, and not over-generalize, but this change looks positive. Please feel free to backport this to the earlier conformance suite snapshots.

gyagp:
Thank you for the review! 
@ds-hwang During development, I ran the tests with N from 3 to 100. I think current implementation isn't specific to odd or even N, so I didn't test both 3x3 and 4x4 in code.
I have applied the similar changes to versions from 1.0.0 to 1.0.3. Please review it at https://github.com/KhronosGroup/WebGL/pull/1996.

PR #1990, opened 2016-08-26
Add a texture mipmap case

jchen10:
To catch a bug on MacOSX. glTexStorage* don't work as expected if
the current GL_TEXTURE_BASE_LEVEL is not 0. As a result, the shader
function textureSize returns wrong size.

### DIFF
--- a/sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
+++ b/sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
@@ -59,6 +59,41 @@ void main()
     gl_FragColor = texture2D(tex, texCoord);
 }
 </script>
+
+<script id="vshader_texsize" type="x-shader/x-vertex">#version 300 es
+in vec4 vPosition;
+void main()
+{
+    gl_Position = vPosition;
+}
+</script>
+
+<script id="fshader_texsize_2d" type="x-shader/x-fragment">#version 300 es
+
+precision mediump float;
+uniform sampler2D tex;
+uniform int lod;
+uniform ivec2 texSize;
+out vec4 fragColor;
+void main()
+{
+    fragColor = (textureSize(tex, lod) == texSize ? vec4(255, 0, 0, 255) : vec4(0, 0, 0, 255));
+}
+</script>
+
+<script id="fshader_texsize_3d" type="x-shader/x-fragment">#version 300 es
+precision mediump float;
+uniform highp sampler3D tex;
+uniform int lod;
+uniform ivec3 texSize;
+out vec4 fragColor;
+void main()
+{
+    fragColor = (textureSize(tex, lod) == texSize ? vec4(255, 0, 0, 255) : vec4(0, 0, 0, 255));
+}
+</script>
+
+
 <script>
 "use strict";
 description(document.title);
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
```diff
@@ -59,6 +59,28 @@
     gl_FragColor = texture2D(tex, texCoord);
 }
 </script>
+
+<script id="vshader_texsize" type="x-shader/x-vertex">#version 300 es
+uniform vec4 uMult;
```
zhenyao: What's the point of uMult here? Looks irrelevant to me.

sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
```diff
@@ -189,6 +211,30 @@
...
+
+  // Test textureSize should work correctly with non-zero base level
+  var program = wtu.setupProgram(
+      gl, ['vshader_texsize', 'fshader_texsize'], ['vPosition'], [0]);
+
+  gl.uniform1i(gl.getUniformLocation(program, "tex"), 0);
+  gl.uniform4f(gl.getUniformLocation(program, "uMult"), 1, 1, 1, 1);
+  gl.uniform1i(gl.getUniformLocation(program, "lod"), 1);
+  gl.uniform2i(gl.getUniformLocation(program, "texSize"), 7, 4);
+  tex = gl.createTexture();
```
zhenyao: My understanding is lod0 is (31, 17), lod1 is (15, 8).

sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
```diff
@@ -189,6 +211,30 @@
...
+  gl.bindTexture(gl.TEXTURE_2D, tex);
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texParameter(TEXTURE_MAG_FILTER) should succeed");
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texParameter(TEXTURE_MIN_FILTER) should succeed");
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_BASE_LEVEL, 1);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texParameter(TEXTURE_BASE_LEVEL) should succeed");
+  gl.texStorage2D(gl.TEXTURE_2D, 4, gl.RGBA8, 31, 17);
```
zhenyao: Can you also test texStorage3D?

sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
kenrussell: Right, but the base level is being set to 1 below, so the call to textureSize for level 1 actually fetches lod 1+1 = 2, which would be (7, 4) -- if I'm understanding the test correctly.

sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
kenrussell: Is there a reason to use these odd numbers? Why not use a size that at least is evenly divisible by 4 in both dimensions?

kenrussell:
Thanks for this test. On the whole it looks good, but please address @zhenyao 's comments and let him merge it.

sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
zhenyao: You are right.

sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
jchen10: Yes, it can be omitted. It was just because of the existing code around. Fixed anyway!

sdk/tests/conformance2/textures/misc/tex-mipmap-levels.html
jchen10: Covered in the newly added texStorage3D case!

zhenyao:
lgtm
but @kenrussell should take another look

kenrussell:
LGTM too. Thanks for the expanded tests.

PR #1993, opened 2016-08-29
Add a test with a tricky nested sequence operator

Oletus:
Covers https://bugs.chromium.org/p/chromium/issues/detail?id=638313


zhenyao:
LGTM

Thanks for the test.

kenrussell:
LGTM too after the fact. Thanks for creating this nice reduced test case.

PR #1994, opened 2016-08-29
Add test: getExtension while PBO is bound should not crash

kainino0x:
Regression test for [Chromium Issue 641643](https://bugs.chromium.org/p/chromium/issues/detail?id=641643)


kenrussell:
Looks good, but could you please add the new test to conformance2/misc/00_test_list.txt as well? Please squash commits when doing so, and use the --force argument to git push if necessary to re-upload. Thanks.

kainino0x:
Okay, done.

kenrussell:
Thanks, looks great. Merging.

zhenyao:
Good job!

PR #1995, opened 2016-08-30
Add more detail about which iteration failed in transform feedback te…

kenrussell:
…sts.


kenrussell:
Merging myself. FYI @zhenyao @jdashg . For debugging and bug reporting purposes.

PR #1996, opened 2016-08-30
Check the result from AA in a general way in versions from 1.0.0 to 1…

gyagp:
….0.4

When drawing a square with 2 different color triangles, some AA may
affect all the pixels on diagonal, while some others, such as CMAA, may
keep the pixels on the egde intact. Given this, (1, 1) can be always a
good candidate for the check.

This is a follow-up of https://github.com/KhronosGroup/WebGL/pull/1988,
to apply same rule to versions from 1.0.0 to 1.0.4.


zhenyao:
LGTM

kenrussell:
Thanks @gyagp for backporting this fix.

Issue #1998, opened 2016-08-30
Broken link in WebGL 2 ToC


lexaknyazev:
Link to "Pixel store parameters for uploads from `TexImageSource`" (5.35) in WebGL 2 [Table of Contents](https://www.khronos.org/registry/webgl/specs/latest/2.0/#toc) doesn't work.

PR #1999, opened 2016-08-30
Fix link in table of contents

qkmiao:
&lt;code&gt; in &lt;a&gt; will make the link in tables of contents be unavailable.
Other similar usage doesn't use &lt;code&gt; tag too.

This fixes https://github.com/KhronosGroup/WebGL/issues/1998.


kenrussell:
Thanks for tracking this down and fixing it.

zhenyao:
lgtm

Issue #1997, opened 2016-08-30
LUMINANCE format is missing from WebGL 2 spec


lexaknyazev:
`LUMINANCE` format is missing from of _internalformat_, _format_, and _type_ combinations table:
https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.7.6

zhenyao:
Thanks for reporting.  It seems like a bug.  We'll add it to the table.

PR #2000, opened 2016-08-30
Add missing LUMINANCE format to the tex format table in WebGL2 spec.

zhenyao:
This is reported on https://github.com/KhronosGroup/WebGL/issues/1997

@kenrussell PTAL


kenrussell:
LGTM

PR #2001, opened 2016-08-31
Conformance test for the nv driver bug

Kangz:
PTAL, this adds a test for the NVIDIA driver bug related to cubemap specification order.


kenrussell:
Excellent. Do you have a link to any more detail, such as a Chromium bug? Do you want to include it in the debug output of this test, or should we merge this as is?

Kangz:
Unfortunately we don't have a Chromium bug for this, the commit only had BUG=(bug ID for the COmmand buffer upgrade to ES 3.0)

zhenyao:
Usually I file a bug in crbug.com to describe the driver bug issue, and add a link to the test, and maybe block crbug.com/429053

kenrussell:
If the associated Chromium commit was far in the past and wasn't well documented, that's fine. Thanks very much for going back and adding this thorough test.

Issue #1991, opened 2016-08-26
Should trying to bind a deleted renderbuffer clear the current binding or silently ignore it?


emilio:
At Servo we've had this discussion in https://github.com/servo/servo/pull/12934, so we looked what other engines do, and found conflicting implementations.

On one hand, Gecko silently ignores them: http://searchfox.org/mozilla-central/source/dom/canvas/WebGLContextGL.cpp#173

On the other hand, Chromium sets the buffer pointer to null, clearing the binding: https://cs.chromium.org/chromium/src/third_party/WebKit/Source/modules/webgl/WebGLRenderingContextBase.cpp?sq=package:chromium&rcl=1472224659&l=1551

What should be the adequate behavior? The current spec only specifies that is null is given, the binding should be cleared, but it leaves this case unspecified.

kenrussell:
Very good question.

I would have hoped that one of the WebGL conformance tests would have covered this -- perhaps sdk/tests/conformance/misc/object-deletion-behaviour.html . Unfortunately, given Firefox's and Chrome's differing behavior, it seems clear this isn't tested.

Chrome treats all incoming deleted objects to bind calls, not just renderbuffers, as though 0 were being bound to the target.

It's difficult to argue for one behavior or the other. At the native API level, deleted objects can be kept alive by objects that reference them. See appendix C.1.3 of the ES 2.0.25 specification. In general, since binding an unused name in OpenGL ES creates a new object, doing this at the native API level usually creates a new object of the same name -- behavior that is not compatible with the way WebGL works, since it requires calling the create* methods to create new objects.

I somewhat like Chrome's current behavior since clearing the current binding will probably make it more obvious to the programmer that they have made an error, rather than silently leaving the binding in its previous state. Regardless, the behavior must be defined and a test written.

Comments @zhenyao @jdashg @grorg @RafaelCintron ?

zhenyao:
Or another option is to generate an INVALID_OP, making it clear it's not a valid op.

emilio:
Heh, indeed, blaming Gecko's implementation it seems they originally generated an INVALID_OP, but removed that to pass conformance tests: https://bugzilla.mozilla.org/show_bug.cgi?id=681835#c1

kenrussell:
Argh.

[...]

PR #2002, opened 2016-09-01
Yield an INVALID_OPERATION error when trying to bind a deleted object

emilio:
cc #1991

### DIFF
--- a/conformance-suites/1.0.0/conformance/object-deletion-behaviour.html
+++ b/conformance-suites/1.0.0/conformance/object-deletion-behaviour.html
@@ -67,7 +67,7 @@ shouldGenerateGLError(gl, gl.NO_ERROR, "gl.deleteTexture(tex)");
 shouldBeNull("gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)");
 shouldBeFalse("gl.isTexture(tex)");
 shouldBeNull("gl.getParameter(gl.TEXTURE_BINDING_2D)");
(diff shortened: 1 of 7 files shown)


zhenyao:
Looks fine to me.  Thanks for the effort.  We will make conformant implementation non-conformant due to the back port, but this is trivia corner case and a fix should be easy, so probably OK.

@kenrussell @RafaelCintron @grorg @kenrussell @jdashg any objections we merge this?

kenrussell:
Thanks @emilio for looking into this and making these changes.

May I please ask you to also do the following:

1) Update the top-of-tree version of the conformance test under sdk/tests/conformance/misc/object-deletion-behaviour.html

2) Update the spec under specs/latest/1.0/index.html defining this behavior for all of the bind* operations

Please squash commits when doing this (you will probably have to use the --force argument to git push to re-upload per https://www.khronos.org/webgl/wiki/Using_Github_To_Contribute )

Thanks very much.

emilio:
No problem! :)

I think that should be pretty much it, I'm double-checking that Firefox doesn't have any expected failure that could affect this.

kenrussell:
Thank you @emilio . Looks great. I hate to ask -- would it be much trouble for you to update specs/latest/2.0/index.html too?

emilio:
Sure, just did it, and the tests should be ok, just checked  and there isn't any test that exercises this disabled in Gecko.

kenrussell:
Excellent. Thank you very much for this thorough fix!

kenrussell:
I realized too late that the earlier conformance snapshots need to admit both behaviors. Only the top of tree one should enforce the new behavior. If it's possible for you to revise the snapshots I'd appreciate it...

emilio:
Sure, I'll do, no problem :-)

PR #2003, opened 2016-09-01
Fix wrong link which causes timeout

qkmiao:
This is introduced in https://github.com/KhronosGroup/WebGL/pull/1994.


kenrussell:
Thanks. @kainino0x FYI.

PR #2004, opened 2016-09-01
Add a test to verify clipping works fine with wide points.

zhenyao:
This test is set to 2.0.1, to unblock WebGL2 release.

Move the failing test cases from clipping.html (they are covered in the newly added test)

### DIFF
--- a/sdk/tests/conformance/rendering/clipping-wide-points.html
+++ b/sdk/tests/conformance/rendering/clipping-wide-points.html
@@ -1,0 +1,47 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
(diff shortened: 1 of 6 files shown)


zhenyao:
@kenrussell Please take a look. (Tested locally: fails on mac AMD and passes on Linux NVidia).
@jdashg @grorg FYI, another driver bug.

sdk/tests/conformance/rendering/clipping-wide-points.html
```diff
@@ -0,0 +1,46 @@
...
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
```
kenrussell: Add title?

sdk/tests/conformance2/rendering/clipping-wide-points.html
```diff
@@ -0,0 +1,46 @@
...
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
```
kenrussell: Same here - add title?

kenrussell:
Looks good - please merge yourself when addressed. Thanks.

PR #2005, opened 2016-09-02
Allow gl.NO_ERROR when attempting to bind a deleted object in old test suites.

emilio:
Too bad I started fixing from 1.0.0 to 1.0.3 and not in the opposite order, because otherwise I would have just copy-pasted the version of `shouldGenerateGLError` that accepts an array instead of re-implementing it myself.

If it's not too much of a churn I'd like to leave it as is, but I can also take more time to do that.

r? @kenrussell


zhenyao:
@emilio Thanks for dealing this issue.  Still, I feel it's better to align the code between versions, in case we need to update the code across versions again, it will be a much easier job if they stay the same.  So could you spend a bit time to do that? Appreciate your input.

emilio:
@zhenyao fair enough, I'll take some time to update it.

emilio:
@zhenyao done, sorry for the delay :)

zhenyao:
lgtm
Thank you for cleaning this up

kenrussell:
Thanks @emilio for cleaning this up and @zhenyao for pushing for the cleanup and reviewing it.

Looks good after the fact.

kenrussell:
Note that this is a follow-on from #2002 .

PR #2007, opened 2016-09-02
Add a test to verify int(bool) works correctly.

zhenyao:
It is not working properly on Mac AMD drivers.


zhenyao:
@toji

toji:
LGTM

kenrussell:
LGTM after the fact.

PR #2009, opened 2016-09-02
Expand int(bool) cast tests to also cover float(bool) and uint(bool)

zhenyao:
int(true) and uint(true) are incorrect on MacOSX AMD


zhenyao:
This pull request renames the filename and expand to also test float(bool) and uint(bool) on top of int(bool).  uint(bool) is WebGL2 only.

(For the record, int(true) and uint(true) are buggy on Mac AMD whereas float(bool) works fine.

zhenyao:
@kenrussell @kainino0x @qkmiao whoever gets to this first, please review.

sdk/tests/conformance/glsl/bugs/bool-type-cast-bug.html
```diff
@@ -0,0 +1,317 @@
...
+        gl.deleteProgram(program);
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "Should be no errors from testing");
+    }
+
+    debug("");
+    program = wtu.setupProgram(gl, ["vshader-int2", "fshader-int2"], ["aPosition"]);
+    if (!program) {
+        testFailed("Fail to set up program");
+    } else {
+        var uniformLoc = gl.getUniformLocation(program, 'bvalue');
```
kainino0x: JSHint has trained me against redefinition of variables in the same scope (`var uniformLoc; var uniformLoc;`). I don't think it really matters, I think it's mainly to guard against accidental variable collisions.

sdk/tests/conformance/glsl/bugs/bool-type-cast-bug.html
```diff
@@ -0,0 +1,317 @@
...
+        wtu.checkCanvas(gl, [0, 255, 0, 255]);
+        debug("Testing fragment shader float(true)");
+        gl.uniform1i(uniformLoc, 1);
+        wtu.drawUnitQuad(gl);
+        wtu.checkCanvas(gl, [255, 0, 0, 255]);
+        gl.deleteProgram(program);
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "Should be no errors from testing");
+    }
+
+    if (wtu.getDefault3DContextVersion() > 1) {
```
kainino0x: It seems a little weird to me to put a WebGL2-specific test in conformance/, but I don't see any problem with it. @kenrussell, can you comment?

kainino0x:
LGTM. Some very minor things in line comments.

sdk/tests/conformance/glsl/bugs/bool-type-cast-bug.html
```diff
@@ -0,0 +1,317 @@
...
+    } else {
+        fvalue = -1.0;
+    }
+}
+</script>
+<script id="fshader-int1" type="x-shader/x-fragment">
+varying mediump float fvalue;
+
+void main() {
+    if (fvalue == 1.0)
```
qkmiao: There may be some problems use "==" for float number.

sdk/tests/conformance/glsl/bugs/bool-type-cast-bug.html
```diff
@@ -0,0 +1,317 @@
...
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Verify int(bool) uint(bool) float(bool) works correctly (Mac AMD driver bug)</title>
```
qkmiao: works -> work

sdk/tests/conformance/glsl/bugs/bool-type-cast-bug.html
```diff
@@ -0,0 +1,317 @@
...
+    else if (uvalue == 0u)
+        myFragColor = vec4(0.0, 1.0, 0.0, 1.0);
+    else
+        myFragColor = vec4(0.0, 0.0, 0.0, 1.0);
+}
+</script>
+
+<script type="application/javascript">
+"use strict";
+description("Verify int(bool) float(bool) uint(bool) works correctly");
```
qkmiao: works -> work

qkmiao:
LGTM with a comment on float number comparison.

sdk/tests/conformance/glsl/bugs/bool-type-cast-bug.html
kenrussell: I agree that this is confusing and would recommend that the JavaScript from this HTML file be factored out into a separate file, and included by two .html files, one under conformance/ and one under conformance2/. This has been done for other tests and should help our bookkeeping in the future.

sdk/tests/conformance/glsl/bugs/bool-type-cast-bug.html
kenrussell: I think this is fine. 1.0 is precisely representable as a floating-point number. Using an epsilon comparison will just complicate things.

kenrussell:
Thanks for tracking this down. LGTM overall. One comment on refactoring the test's code.

zhenyao:
Thank you all for the feedback.

After some consideration, I decided to spit the test into two html files rather than using one js file. It's because then we have to represent shader sources in an ugly way in js.

@kenrussell Please take another look.

kenrussell:
Thanks for thinking that through. LGTM.

PR #2010, opened 2016-09-05
test blitframebuffer when bliting pixels outside of read framebuffer

Richard-Yunchao:
PTAL. Thanks a lot!


Richard-Yunchao:
This one is not correct, sampling from a srgb image to a linear draw buffer would convert srgb to linear during sampling. So we should read pixels after blitFramebuffer, instead of sampling and drawing and reading pixels finally. 

Close this one.  See this one https://github.com/KhronosGroup/WebGL/pull/2011

PR #2011, opened 2016-09-06
Test blitframebuffer when bliting pixels outside of read framebuffer

Richard-Yunchao:
PTAL. Thanks!

https://github.com/KhronosGroup/WebGL/pull/2010 is not correct.

### DIFF
--- a/sdk/tests/conformance2/rendering/blitframebuffer-outside-readbuffer.html
+++ b/sdk/tests/conformance2/rendering/blitframebuffer-outside-readbuffer.html
@@ -1,0 +1,299 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/rendering/blitframebuffer-outside-readbuffer.html
```diff
@@ -0,0 +1,299 @@
...
+    // Initiate data to read framebuffer
+    var width_read = 3;
+    var height_read = 3;
+    var uint_read = new Uint8Array(width_read * height_read * 4);
+    var start = 0x20;
+    for (var ii = 0; ii < width_read * height_read * 4; ii += 4) {
+        for (var jj = 0; jj < 3; ++jj) {
+          uint_read[ii + jj] = start;
+        }
+        uint_read[ii + jj] = 0xff;
```
zhenyao: jj here is out of the scope.

zhenyao:
Looks great with one bug fixed

Richard-Yunchao:
Thanks for your review, code has been updated to fix the minor bug, @zhenyao . PTAL.

zhenyao:
Looks good.

sdk/tests/conformance2/rendering/blitframebuffer-outside-readbuffer.html
```diff
@@ -0,0 +1,299 @@
...
+                    var expectedColor = ref[loc];
+                    // We may need to covert the color space for pixels in blit region
+                    if ((readbufferHasSRGBImage ^ drawbufferHasSRGBImage)) {
+                        if (drawbufferHasSRGBImage) {
+                            expectedColor = linearToSRGB(expectedColor);
+                        } else {
+                            expectedColor = sRGBToLinear(expectedColor);
+                        }
+                    }
+                    if (checkPixel(color, expectedColor) == true) {
```
kenrussell: In a future CL, it might be better to try to use wtu.checkCanvasRect for each individual pixel rather than hand-writing the comparison function, if that would reduce the amount of code.

kenrussell:
Excellent test @Richard-Yunchao and thanks @zhenyao for reviewing and merging. Looks great after the fact.

sdk/tests/conformance2/rendering/blitframebuffer-outside-readbuffer.html
Richard-Yunchao: Thanks for your suggestion, @kenrussell , I will use that function in my other patches. 

For this one, I saw that checkCanvasRect will readPixels again and again during each iteration. So I choose to readPixels in one time, and compare these pixels in a for loop by a hand-writing function to save lots of readPixels call sites to save time. 

Hope that the difference is not a big deal. :)

PR #2012, opened 2016-09-07
Add test cases to verify ivec(bvec) and uvec(bvec) works correctly.

zhenyao:
(and they fail on Mac AMD)


zhenyao:
@kenrussell Please review.

kenrussell:
Excellent work tracking this down. LGTM

PR #2013, opened 2016-09-07
The src region and/or dst region may be reversed during bliting, add tests to cover this feature

Richard-Yunchao:
PTAL. @zhenyao  and @kenrussell


zhenyao:
lgtm

PR #2016, opened 2016-09-07
roundEven should not distinguish -0/+0

zhenyao:
Mac AMD drivers do not, so this caused some cases to fail
in shadercommonfunction.html

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fShaderCommonFunctionTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fShaderCommonFunctionTests.js
@@ -618,7 +618,7 @@ goog.scope(function() {
         /** @type {number} */ var rounded = (q > 0.5) ? (truncated + 1) : // Rounded up
             (q == 0.5 && (truncated % 2 != 0)) ? (truncated + 1) : // Round to nearest even at 0.5
             truncated; // Rounded down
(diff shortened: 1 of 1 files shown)


zhenyao:
@kenrussell @jdashg

kenrussell:
Is this the fix we discussed? I thought we were going to change the implementation of es3fShaderCommonFunctionTests.roundEven so that it doesn't return -0. That is a difference compared to the C++ code that should be eliminated.

kenrussell:
(Maybe we should make this change, too, but I think the change to roundEven should also be made.)

zhenyao:
Just always returning the +0 in js roundEven() emulation won't make the test pass - most drivers do return -0, so we have to fix the comparison function also. I did both, although I feel changing the roundEven() emulation is unnecessary - returning -0/+0 or always returning +0 are both correct behavior.

kenrussell:
I'd like the roundEven function's code to be as close to the C++ version as possible. I think your change there does that, but if you look it over again and think that change should be undone, I trust your judgment. Please merge yourself after giving this one last thought.

zhenyao:
The latest CL is fine. I agree with you that we should track c++ code.  The emulation of roundEven() can either return +0 only or +0/-0, both are valid code (and both are observed on real drivers).

PR #2023, opened 2016-09-12
Add test cases for blitFramebuffer to filter srgb. This means that the width/height of src region and dst region are different.

Richard-Yunchao:
PTAL. Thanks a lot!

This test case will fail on MacOSX (and will fail on other platforms which is desktop OGL and the version < 4.4). But with this patch: https://codereview.chromium.org/2286593002/. It can pass!

### DIFF
--- a/sdk/tests/conformance2/rendering/blitframebuffer-filter-srgb.html
+++ b/sdk/tests/conformance2/rendering/blitframebuffer-filter-srgb.html
@@ -1,0 +1,220 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL BlitFramebuffer Tests</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/rendering/blitframebuffer-filter-srgb.html
```diff
@@ -0,0 +1,220 @@
...
+}
+
+var tex_read = gl.createTexture();
+var tex_draw = gl.createTexture();
+var fbo_read = gl.createFramebuffer();
+var fbo_draw = gl.createFramebuffer();
+var size_read = 4;
+var size_draw = 0;
+
+function blitframebuffer_helper(readbufferFormat, drawbufferFormat, filter, data, size_draw) {
```
zhenyao: You pass size_draw, which overwrites the global variable of the same name; and then you use the global size_read.

Can we be consistent in handling these parameters? If we go down the path of passing down the params, can they use different names from the global variables?

sdk/tests/conformance2/rendering/blitframebuffer-filter-srgb.html
```diff
@@ -0,0 +1,220 @@
...
+    var pixels = new Uint8Array(size_draw * size_draw * 4);
+    gl.bindFramebuffer(gl.READ_FRAMEBUFFER, fbo_draw);
+    gl.readPixels(0, 0, size_draw, size_draw, gl.RGBA, gl.UNSIGNED_BYTE, pixels);
+    return pixels;
+}
+
+function blitframebuffer_filter_srgb(readbufferFormat, drawbufferFormat, filter, minified) {
+    debug("");
+    debug("Test srgb filtering for blitFramebuffer, the current filter is: " + wtu.glEnumToString(gl, filter));
+    var min_mag = minified ? "minified 2 times." : "magnified 2 times.";
```
zhenyao: I got confused a bit until I read through the code.

Maybe "minified to half the size" and "magnified to double the size"?

zhenyao:
This is a great test, very thorough and clear.  It would have been better if we can generate the ref image without calling blitFramebuffer, but that would require more code. Let's just land it as is and trust blitFramebuffer works for RGBA -> RGBA.

Some nits, please fix and then I am happy to merge.

sdk/tests/conformance2/rendering/blitframebuffer-filter-srgb.html
```diff
@@ -0,0 +1,220 @@
...
+<body>
+<canvas id="example" width="8" height="8"></canvas>
+<div id="description"></div>
+<div id="console"></div>
+
+<script>
+"use strict";
+
+var wtu = WebGLTestUtils;
+description("This test verifies the functionality of blitFramebuffer.");
```
kenrussell: Should this read something like "blitFramebuffer with sRGB framebuffers"?

kenrussell:
Thanks @Richard-Yunchao for the thorough test and @zhenyao for carefully reviewing it. I defer to @zhenyao 's review.

Richard-Yunchao:
Thanks for your review, @zhenyao  and @kenrussell . Nits have been fixed, please take another look. Thanks a lot!

zhenyao:
lgtm

zhenyao:
By the way, in the future, can we not use git commit -a and upload with --force? After each review, the change should be in a separate CL, so reviewers can see the change since last review instead of the entire CL. (We discovered there is an option to squash all commits into one at merging)

kenrussell:
@zhenyao yes, good point. @Richard-Yunchao , others on your team: no longer necessary to squash all commits manually; Github added an option to do this upon merge which I hadn't known about.

Richard-Yunchao:
@zhenyao and @kenrussell, no problem! Thanks for your kind reminder. I will tell this to the whole team if they don't know. In fact, we always tell new-comers to squash the commits to avoid multiple commits for one patch. Haha... Now we don't need to do this any longer.

Richard-Yunchao:
BTW, how to do that? squash all commits when merging?

zhenyao:
Next to the Merge button, there is a drop down list button.  Click that and we will have two options: merge as multiple commits, or merge as one commit. We will also have the opportunity to update the commit log if we choose to merge as multiple commits.

Richard-Yunchao:
Got it. Thank you very much! @zhenyao and @kenrussell

PR #2026, opened 2016-09-12
Add a test to catch the Mac AMD readPixels from RGB buffer into PBO bug

zhenyao:
Also, skip the two related failing cases from pixelbufferobject.html.


zhenyao:
@kenrussell Please review.

kenrussell:
Excellent. LGTM.

PR #2029, opened 2016-09-14
Add commit to ignoredMethods

xidachen:
Right now chromium is implementing WebGL's commit API, to make sure all the conformance tests will be passing, this API has to be added to ignoredMethods first.

PTAL.
@kenrussell

### DIFF
--- a/sdk/tests/conformance/context/methods.html
+++ b/sdk/tests/conformance/context/methods.html
@@ -183,6 +183,8 @@ var methods = [
 // Properties to be ignored because they were added in versions of the
 // spec that are backward-compatible with this version
 var ignoredMethods = [
+  // There is no official spec for the commit API yet, the proposal link is:
+  // https://wiki.whatwg.org/wiki/OffscreenCanvas
   "commit"
 ];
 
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/context/methods.html
```diff
@@ -183,6 +183,7 @@
 // Properties to be ignored because they were added in versions of the
 // spec that are backward-compatible with this version
 var ignoredMethods = [
+  "commit"
```
kenrussell: Please add a comment saying from where this method comes: i.e., a link to the appropriate spec.

sdk/tests/conformance2/context/methods-2.html
```diff
@@ -273,6 +273,7 @@
 // Properties to be ignored because they were added in versions of the
 // spec that are backward-compatible with this version
 var ignoredMethods = [
+  "commit"
```
kenrussell: Same here.

xidachen:
I have updated with comments. PTAL. @kenrussell

kenrussell:
Thanks Xida. Looks good. Merging now.

Issue #2030, opened 2016-09-15
Remove ETC2/EAC compressed texture formats from WebGL 2.0 core


kenrussell:
On the public_webgl list it has been noted by developers that decompressing ETC2/EAC compressed texture formats in order to claim conformance on devices which don't support them natively is counter-productive; it increases the VRAM consumption by a factor of six. See https://www.khronos.org/webgl/public-mailing-list/archives/1609/threads.php#00000 .

After discussion, it sounds like the best option is to remove these enums from the core WebGL 2.0 specification, and leave them in the WEBGL_compressed_texture_es3 extension:
https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_es3/

So, for concrete steps:

1) Chrome should implement this draft extension, so there are two implementations (Firefox being the other): http://crbug.com/455568
2) It should be promoted to community approved.
3) The enums should be removed from the WebGL 2.0 spec, and the conformance tests updated to only run these tests if the extension is available.

msc-:
Unless you can guarantee that you will never expose the extension when the native OGL advertises ETC2 support but is actually decompressing in software, this proposal is not going to solve the problem.

Also, _if_ the extension is to be carried forward, it should should be renamed from `es3` to `etc`. The current name will cause confusion in the future as ASTC is a required part of OpenGL ES 3.2, as well as OpenGL 4.3+, and I am sure we'll want to expose it to WebGL soon.

kenrussell:
Thanks for your feedback Mark. I think this will address the issue for most problematic cases; Chrome contains ETC texture decompression code for the purpose of advertising universal support for these texture formats, and stopping using it will solve at least most, if not all, of the problem.

It's a good point about the naming of the extension. I'll talk directly with the extension author about renaming it.

Maksims:
Thank you for addressing this issue.

Regarding naming, there is currently ETC1, and ETC2 being backwards-compatible with ETC1 although EAC being very similar, is used for 1/2 channel compression. And is part of ES3 extension.
ETC1 compression (`ETC1_RGB8`) can be decoded using `RGB8_ETC2`, so feels like `WEBGL_compressed_texture_etc` would be more appropriate name for this extension.

kenrussell:
CC'ing @kainino0x who will be helping at least with Chrome's implementation if not also the tests.

pyalot:
Please investigate if you can use:

[...]

PR #2031, opened 2016-09-16
Renamed WEBGL_compressed_texture_es3 to _etc to future-proof it.

kenrussell:
Part of work for #2030.


kenrussell:
@jdashg please review.

kenrussell:
Renamed to WEBGL_compressed_texture_es3_0 instead per discussion on the earlier commit. Please take another look.

PR #2032, opened 2016-09-16
Move srgb converter related apis into wtu

Richard-Yunchao:
They are common functions, move them to webgl-test-utils to reuse these functions among tests. 


kenrussell:
Thanks @Richard-Yunchao . Nice refactoring.

PR #2034, opened 2016-09-19
Add test cases to test blitframebuffer against multisampled srgb image

Richard-Yunchao:

Per the discussion in this Chromium CL: https://codereview.chromium.org/2286593002/.  We have no tests to blit multisampled srgb to a srgb image. So this test comes, it can expose some errors in Mac (OGL 4.1).

### DIFF
--- a/sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
+++ b/sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
@@ -128,7 +128,7 @@ function blitframebuffer_helper(readbufferFormat, drawbufferFormat, filter) {
 }
 
 function checkPixel(color, expectedColor) {
-  var tolerance = 63;
+  var tolerance = 15;
   return (Math.abs(color[0] - expectedColor[0]) <= tolerance &&
           Math.abs(color[1] - expectedColor[1]) <= tolerance &&
           Math.abs(color[2] - expectedColor[2]) <= tolerance &&
@@ -140,13 +140,21 @@ function draw_triangles(conversion) {
     gl.enableVertexAttribArray(positionLoc);
     // Draw two triangles. They are almost right-triangles: one is actually an obtuse triangle. The other is actually an acute triangle.
     // This means that some egdes of the triangles almost parallel with x-axis and y-axis. So they can show the alias.
-    var positions = new Float32Array([
+    /* var positions = new Float32Array([
          0.85, -0.90,
         -0.90,  0.85,
         -0.87, -0.87,
          0.90, -0.85,
          0.93,  0.93,
         -0.85,  0.90,
+    ]);*/
+    var positions = new Float32Array([
+         1.0, -1.0,
+        -1.0,  1.0,
+        -1.0, -1.0,
+         1.0, -1.0,
+         1.0,  1.0,
+        -1.0,  1.0,
     ]);
     var positionBuffer = gl.createBuffer();
     gl.bindBuffer(gl.ARRAY_BUFFER, positionBuffer);
@@ -154,13 +162,21 @@ function draw_triangles(conversion) {
     gl.vertexAttribPointer(positionLoc, 2, gl.FLOAT, false, 0, 0);
 
     gl.enableVertexAttribArray(colorLoc);
-    var colors = [
+    /* var colors = [
         [255, 0, 0, 255],
         [0, 255, 0, 255],
         [0, 0, 255, 255],
         [253, 122, 5, 255],
         [6, 252, 120, 255],
         [118, 4, 254, 255]
+    ];*/
+    var colors = [
+        [255, 0, 0, 255],
+        [0, 255, 0, 255],
+        [0, 0, 255, 255],
+        [255, 0, 0, 255],
+        [0, 0, 255, 255],
+        [0, 255, 0, 255],
     ];
     // We need to convert srgb color space to linear color space for reference image.
     var vertexColors = new Uint8Array(24);
--- a/sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
+++ b/sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
@@ -1,0 +1,301 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL BlitFramebuffer Tests</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+<script src="../../js/glsl-conformance-test.js"></script>
+</head>
+<body>
+<canvas id="canvas" width="8" height="8"></canvas>
+<div id="description"></div>
+<div id="console"></div>
+
+<script id="vshader" type="x-shader/x-vertex">#version 300 es
+in highp vec2 aPosition;
+in highp vec4 aColor;
+out mediump vec4 vColor;
+void main() {
+    gl_Position = vec4(aPosition, 0.0, 1.0);
+    vColor = aColor;
+}
+</script>
+
+<script id="fshader" type="x-shader/x-fragment">#version 300 es
+precision mediump float;
+out vec4 oColor;
+in vec4 vColor;
+void main() {
+    oColor = vColor;
+}
+</script>
+
+<script>
+"use strict";
+
+var wtu = WebGLTestUtils;
+description("This test verifies the functionality of blitFramebuffer with multisampled readbuffer and sRGB color buffer.");
+
+var gl = wtu.create3DContext("canvas", undefined, 2);
+
+var tex_draw = gl.createTexture();
+var fb0 = gl.createFramebuffer();
+var fb1 = gl.createFramebuffer();
+var rb0 = gl.createRenderbuffer();
+var rb1 = gl.createRenderbuffer();
+var fbo_draw = gl.createFramebuffer();
+var size = 256;
+var vao;
+var program;
+var positionLoc;
+var colorLoc;
+
+if (!gl) {
+    testFailed("WebGL context does not exist");
+} else {
+    testPassed("WebGL context exists");
+
+    init();
+
+    var filters = [gl.LINEAR, gl.NEAREST];
+    for (var ii = 0; ii < filters.length; ++ii) {
+        blitframebuffer_multisampled_readbuffer(gl.SRGB8_ALPHA8, gl.SRGB8_ALPHA8, filters[ii]);
+    }
+}
+
+function init() {
+    program = wtu.setupProgram(gl, ['vshader', 'fshader']);
+    positionLoc = gl.getAttribLocation(program, "aPosition");
+    colorLoc = gl.getAttribLocation(program, "aColor");
+    if (!program || positionLoc < 0 || colorLoc < 0) {
+        testFailed("Set up program failed");
+        return;
+    }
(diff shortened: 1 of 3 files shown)


sdk/tests/js/webgl-test-utils.js
```diff
@@ -2971,18 +2971,18 @@ function sRGBChannelToLinear(value) {
...
+    } else if (value < 0.0031308) {
+        value = value * 12.92;
+    } else if (value < 1) {
+        value = Math.pow(value, 0.41666) * 1.055 - 0.055;
     } else {
-        color = 1.0;
+        value = 1.0;
     }
-    return Math.trunc(color * 255 + 0.5);
+    return Math.trunc(value * 255 + 0.5);
```
Richard-Yunchao: Just to keep the naming in linearToSRGB() and linearChannelToSRGB() aligned with that in sRGBToLinear() and sRGBChannelToLinear.

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
```diff
@@ -0,0 +1,301 @@
...
+    // Read pixels for comparision
+    var pixels = new Uint8Array(size * size * 4);
+    gl.bindFramebuffer(gl.READ_FRAMEBUFFER, fbo_draw);
+    gl.readPixels(0, 0, size, size, gl.RGBA, gl.UNSIGNED_BYTE, pixels);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "readpixels should succeed");
+    return pixels;
+}
+
+function checkPixel(color, expectedColor) {
+  var tolerance = 15;
```
Richard-Yunchao: Not quite sure about the tolerance. If I set the tolerance to 7(In some other tests, we set the tolerance to 7). It will fail on my Linux desktop with OGL 4.5. And the biggest diff is actually 8. 

So I set the tolerance to 15. I suppose the implementation of OGL 4.5 can resolve multi-sampled srgb and draw to a srgb image. Then the test can pass on Linux with OGL 4.5. 

BTW, it seems to be reasonable that the diff would become bigger if we enabled msaa.

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
```diff
@@ -0,0 +1,300 @@
...
+    // Read pixels for comparision
+    var pixels = new Uint8Array(size * size * 4);
+    gl.bindFramebuffer(gl.READ_FRAMEBUFFER, fbo_draw);
+    gl.readPixels(0, 0, size, size, gl.RGBA, gl.UNSIGNED_BYTE, pixels);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "readpixels should succeed");
+    return pixels;
+}
+
+function checkPixel(color, expectedColor) {
+  var tolerance = 63;
```
Richard-Yunchao: Set tolerance to 15 is still too small. Set it to 63...

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
```diff
@@ -140,27 +140,43 @@
...
         [6, 252, 120, 255],
         [118, 4, 254, 255]
+    ];*/
+    var colors = [
+        [255, 0, 0, 255],
+        [0, 255, 0, 255],
+        [0, 0, 255, 255],
+        [255, 0, 0, 255],
+        [0, 0, 255, 255],
+        [0, 255, 0, 255],
```
Richard-Yunchao: If we draw two triangles, and they can well show alias like patch set 2. I suspect that the multisampling algorithm would make pixels vary a lot, especially the pixels around the edge. I mean when we use the way to compare in this test case:
blit multisampled srgb to srgb directly 
vs
 multisampled srgb - (decode) - multisampled linear - (blit) - linear - (encode) - srgb. 

However, there are so many different pixels inside triangles between these two images by the two methods when running on MacOSX (OGL 4.1), which is still out of my expectation. But on my Linux desktop, just some pixels around the edge vary a lot if I compare these two results. 
 
But If we draw a rectangle like the commit 3, the case can pass on MacOSX too (OGL 4.1) . 

So, I am not sure which test is correct, commit 2 or commit 3?

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
```diff
@@ -0,0 +1,316 @@
...
+
+    gl.enableVertexAttribArray(colorLoc);
+    /* var colors = [
+        [255, 0, 0, 255],
+        [0, 255, 0, 255],
+        [0, 0, 255, 255],
+        [253, 122, 5, 255],
+        [6, 252, 120, 255],
+        [118, 4, 254, 255]
+    ];*/
```
kenrussell: Please remove commented-out code.

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
kenrussell: The tolerance should be set as low as possible to pass on the known platforms. If the test will pass with a tolerance of 8 then let's set it to that value.

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
kenrussell: Let's keep the test simple and draw a rectangle, so as to avoid antialiasing issues as much as possible.

Something seems wrong with this test. If I change all of the per-vertex colors to (200, 0, 0) -- i.e., solid red, which I assume should fall within the sRGB color gamut (not sure whether the fully-saturated colors do) -- then the test fails badly, and the colors are visibly different on my MacBook Pro with NVIDIA GPU, on a Chromium continuous build containing your fix https://codereview.chromium.org/2286593002/ .

Can you double-check the logic throughout? With the above change, does the test fail on your Linux workstation with Intel GPU?

I assume that if the conversions between linear and sRGB are all correct in this test, that it shouldn't even need a tolerance of 7 or 8.

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
```diff
@@ -0,0 +1,316 @@
...
+        testFailed("Framebuffer incomplete.");
+        return;
+    }
+    draw_triangles(false);
+    gl.bindFramebuffer(gl.READ_FRAMEBUFFER, fb0);
+    var pixels = blitframebuffer_helper(readbufferFormat, drawbufferFormat, filter);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "Blit from a multi-sampled srgb image to a srgb image should succeed");
+
+    // Draw to a multi-sampled linear image, and blit to a linear image.
+    // This result is usded as reference. We suppose that this is correct.
```
kenrussell: Would it be better to write a simpler test for this case which just draws a solid color, and thereby knows the reference color ahead of time? It would be more robust than performing a render and reading back the results for the reference.

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
```diff
@@ -0,0 +1,316 @@
...
+        [0, 255, 0, 255],
+        [0, 0, 255, 255],
+        [255, 0, 0, 255],
+        [0, 0, 255, 255],
+        [0, 255, 0, 255],
+    ];
+    // We need to convert srgb color space to linear color space for reference image.
+    var vertexColors = new Uint8Array(24);
+    for (var i = 0; i < 6; ++i) {
+        var color = conversion ? wtu.sRGBToLinear(colors[i]) : colors[i];
```
kenrussell: Are you sure that you're supposed to manually convert these colors from the sRGB color space to the linear color space? I thought the shaders were supposed to assume they are working on data in the linear color space, and it's during writes to the multisampled sRGB renderbuffer that they are converted from linear to sRGB. Maybe that discrepancy is why the test fails so badly when the colors aren't fully saturated.

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
```diff
@@ -0,0 +1,316 @@
...
+    gl.bindTexture(gl.TEXTURE_2D, tex_draw);
+    gl.texImage2D(gl.TEXTURE_2D, 0, drawbufferFormat, size, size, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+    gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, fbo_draw);
+    gl.framebufferTexture2D(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, tex_draw, 0);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "setup draw framebuffer should succeed");
+
+    gl.blitFramebuffer(0, 0, size, size, 0, 0, size, size, gl.COLOR_BUFFER_BIT, filter);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "blitframebuffer should succeed");
+
+    // Read pixels for comparision
```
kenrussell: typo: comparison

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
```diff
@@ -0,0 +1,316 @@
...
+    // Draw two triangles. They are almost right-triangles: one is actually an obtuse triangle. The other is actually an acute triangle.
+    // This means that some egdes of the triangles almost parallel with x-axis and y-axis. So they can show the alias.
+    /* var positions = new Float32Array([
+         0.85, -0.90,
+        -0.90,  0.85,
+        -0.87, -0.87,
+         0.90, -0.85,
+         0.93,  0.93,
+        -0.85,  0.90,
+    ]);*/
```
kenrussell: Remove commented-out code.

sdk/tests/js/webgl-test-utils.js
kenrussell: This seems fine.

kenrussell:
Thanks @Richard-Yunchao for putting together this test. I experimented with it locally and think there are some problems. Please see comments inline.

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
Richard-Yunchao: Good suggestion, let's fill a rectangle with solid color and draw to a multisampled srgb draw buffer. It is simple and direct.

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
Richard-Yunchao: That's true.

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
Richard-Yunchao: Done

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
Richard-Yunchao: The latest code have no this problem now.

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
Richard-Yunchao: done

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
Richard-Yunchao: done

Richard-Yunchao:
Thanks for your suggestion, @kenrussell . The latest commit will fill a rect with solid color, and draw it to a multisampled srgb image. Then blit to a singlesampled srgb image. 

The code can pass both on Linux desktop and MacOSX. PTAL. Thanks a lot!

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
```diff
@@ -0,0 +1,215 @@
...
+
+function checkPixel(color, expectedColor) {
+  var tolerance = 7;
+  return (Math.abs(color[0] - expectedColor[0]) <= tolerance &&
+          Math.abs(color[1] - expectedColor[1]) <= tolerance &&
+          Math.abs(color[2] - expectedColor[2]) <= tolerance &&
+          Math.abs(color[3] - expectedColor[3]) <= tolerance);;
+}
+
+function draw_rect(color) {
```
kenrussell: At this point, draw_rect and the custom shaders, etc. can be replaced with WebGLTestUtils.setupColorQuad / drawFloatColorQuad. Could you make this simplification in a follow-on pull request?

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
```diff
@@ -0,0 +1,215 @@
...
+    // Read pixels for comparison
+    var pixels = new Uint8Array(size * size * 4);
+    gl.bindFramebuffer(gl.READ_FRAMEBUFFER, fbo_blit);
+    gl.readPixels(0, 0, size, size, gl.RGBA, gl.UNSIGNED_BYTE, pixels);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "readpixels should succeed");
+    return pixels;
+}
+
+function checkPixel(color, expectedColor) {
+  var tolerance = 7;
```
kenrussell: Does the tolerance still need to be this high in order to pass the test on your system? On my MacBook Pro Retina I can set it to 0 and the test still passes.

Richard-Yunchao: If the code do wrong decode/encode for srgb image, the diff would much more than 7. So 7 is OK. But you are correct, I re-test it on my MacOS. And it can still pass even the tolerance is 0. I suppose that the only factor to cause pixel diff may come from precision lost during srgb conversion. But it can be very very small. I will set the tolerance to 1, in case we indeed have precision lost on some devices.

sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
Richard-Yunchao: Good point. This can make the test simple. Please see the pr at https://github.com/KhronosGroup/WebGL/pull/2050

Issue #1981, opened 2016-08-24
get.webgl2.org


greggman:
Should there be a `get.webgl2.org` (or `get.webgl.org/webgl2`) to tell people to download firefox or chrome (or is there any word that Safari and Edge will support WebGL2 anytime soon?)

zhenyao:
From what I heard, khronos is working on it. @kenrussell should know the details.

kenrussell:
Good idea. Actually Khronos isn't working on this at the moment to the best of my knowledge. Any contributions to a get.webgl.org/webgl2/ page would be welcome.

There's no public announcement to make about WebGL 2.0 support in Safari and Edge yet, but all WebGL implementers have been supportive of the direction.

PR #2036, opened 2016-09-19
code for get.webgl.org/webgl2

greggman:
will need to update the upgradeUrls once each browser ships enabled to stable

re #1981


kenrussell:
Thanks Gregg for putting this together! The "2" cube looks nice. :)

What would you think about linking to https://www.khronos.org/webgl/wiki/Getting_a_WebGL_Implementation as a general landing page?

Merging now; thanks for this contribution. Will be glad for any changes/additions you would like to make.

PR #2037, opened 2016-09-19
Add test to expose unary minus bug on Windows Intel

Jiawei-Shao:
On some Intel D3D drivers, evaluating unary minus operator
on an integer variable may get wrong answer in vertex shader.
One of such situation is when attempting to read the answer
in a dynamic loop.

This patch adds test to expose this Intel D3D driver bug.


Jiawei-Shao:
This test also exposes the root cause of the failure in deqp/functional/gles3/shaderstruct.html on Windows Intel on Broadwell platforms.

kenrussell:
Excellent work Jiawei tracking this down and creating the reduced test cases. Looks perfect.

PR #2038, opened 2016-09-19
getBufferSubData should take ArrayBufferView.

jdashg:


kenrussell:
While there was some discussion earlier about whether getBufferSubData should continue to accept ArrayBuffer, the current spec states that it only takes ArrayBufferView, and from that standpoint these changes look correct.

LGTM. Thanks for updating these tests. Merging now.

PR #2039, opened 2016-09-19
Apply fix from @Elainejly for textureprojoffset test.

### LINKED ISSUE #2033: [webGL2.0] Textureprojoffset failed in AMD graphics for OpenGL rendering
OS: w10
Graphics: AMD asic
browser: firefox/chrome
config: enable webGL2.0, enable OpenGL rendering

steps: 
1) run chrome with "--enable-unsafe-es3-apis --enable-experimental-canvas-features"
1) visit "https://www.khronos.org/registry/webgl/sdk/tests/webgl-conformance-tests.html?version=2.0.0"
2) find test "all/deqp/functional/gles3/shadertexturefunction/textureprojoffset" and run

expected: test pass
actually: failed: texture_function.textureprojoffset.sample2dshadow_vertex: Fail
          there are two pixels less in the first column in the "Result" frame dump compared with "Reference"

Root cause: 
    the number of 23, 24, 25, 26, 28 points in the draw, are trying to sample the texture with coordinate (u=0, v=xxx), which value is not defined for filter GL_REPEAT. you need to add an offset 0.5/256 (256 is texture width) to u if you want to get the exact value.   
    file: WebGL-master\sdk\tests\deqp\functional\gles3\es3fShaderTextureFunctionTests.js line: 2384
     es3fShaderTextureFunctionTests.getCaseSpec('sampler2dshadow', es3fShaderTextureFunctionTests.TexFunction.TEXTUREPROJ, [0.2, 0.6, 0.0, 1.5], [-2.25, -3.45, 1.5, 1.5],    false,    0.0,    0.0,    true, [-8, 7, 0],    tex2DShadow, es3fShaderTextureFunctionTests.evalTexture2DShadowProjOffset, es3fShaderTextureFunctionTests.CaseFlags.VERTEX),
               
Suggestion solution: 
    Change the offset to avoid sampling with u=0. [-8, 7, 0] -> [7,-8,0]

Attach file include the first 36 attributes of webgl_b735f978b01bb186 in the test draw.
[vertex and texture.xlsx](https://github.com/KhronosGroup/WebGL/files/479174/vertex.and.texture.xlsx)

kenrussell:
The offset was causing sampling of a boundary condition. Change it to be more
robust.

Verified on Chrome with --use-angle=gl as well as ANGLE's D3D11 backend.

Fixes #2033 .


kenrussell:

zhenyao:
lgtm

PR #2040, opened 2016-09-20
When the command fails, don't expect state to update.

jdashg:


kenrussell:
Yes, that's a definite bug in the top-of-tree conformance suite. Thanks for catching and fixing it. LGTM

PR #2043, opened 2016-09-20
Add skip entries for three es3fShaderTextureFunction tests.

kenrussell:
These tests are in the skip lists for the native dEQP suite. The most
immediate reason to skip them is that they are failing on macOS with
AMD GPUs, but they are skipped because they will fail on other GPU
types as well.


kenrussell:

zhenyao:
I agree we should not include tests that are bypassed on native deqp. LGTM.

PR #2044, opened 2016-09-20
Add test cases to verify blitFramebuffer with multiple drawbuffers

Richard-Yunchao:

Add conformance test for this TODO: https://cs.chromium.org/chromium/src/gpu/command_buffer/service/gles2_cmd_decoder.cc?q=gles2_cmd_decoder&sq=package:chromium&l=7734.

### DIFF
--- a/sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
+++ b/sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
@@ -137,7 +137,7 @@ function blitframebuffer_srgb_and_linear_drawbuffers(readbufferFormat, drawbuffe
     }
 
     // Blit to multiple draw buffers with srgb images and linear images
-    var dstSize = (readbufferFormat == gl.SRGB8_ALPHA8) ? 5 : size;
+    var dstSize = size - 1;
     gl.blitFramebuffer(0, 0, size, size, 0, 0, dstSize, dstSize, gl.COLOR_BUFFER_BIT, filter);
     wtu.glErrorShouldBe(gl, gl.NO_ERROR, "blitframebuffer should succeed");
 
--- a/sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
+++ b/sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
@@ -1,0 +1,228 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL BlitFramebuffer Tests</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+<script src="../../js/glsl-conformance-test.js"></script>
+</head>
+<body>
+<canvas id="canvas" width="8" height="8"></canvas>
+<div id="description"></div>
+<div id="console"></div>
+
+<script>
+"use strict";
+
+var wtu = WebGLTestUtils;
+description("This test verifies the functionality of blitFramebuffer with multiple draw buffers (srgb image and linear image).");
+
+var gl = wtu.create3DContext("canvas", undefined, 2);
+var linearMask = 1;
+var srgbMask = 2;
+
+if (!gl) {
+    testFailed("WebGL context does not exist");
+} else {
+    testPassed("WebGL context exists");
+
+    var filters = [gl.LINEAR, gl.NEAREST];
+    var drawbuffersFormats = [linearMask, srgbMask, linearMask | srgbMask];
+    for (var ii = 0; ii < filters.length; ++ii) {
+        for (var jj = 0; jj < drawbuffersFormats.length; ++jj) {
+            blitframebuffer_srgb_and_linear_drawbuffers(gl.SRGB8_ALPHA8, drawbuffersFormats[jj], filters[ii]);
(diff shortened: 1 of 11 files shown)


kenrussell:
Thanks @Richard-Yunchao for this test. Are you sure it is correct? I see it fail both on a MacBook Pro, which might be expected because it's running OpenGL 4.1 and may need more workarounds -- but also on Linux with an NVIDIA GPU running OpenGL 4.5. It also fails on Linux with ANGLE's OpenGL backend.

Richard-Yunchao:
Ah... I made a mistake. I should change the srgb to linear as expected linear color if the src image is srgb and the dst image is linear. 
Now the code is OK. It can pass on Linux desktop with OGL 4.5. But it fails on MacOSX with OGL 4.1.

PTAL. @kenrussell . 

(I updated the code at local, then revise the code and re-submit, I found that the commit will contain many commits those have been merged... So I used "git pull --force" to remove those commits...)

sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
```diff
@@ -0,0 +1,228 @@
...
+    testFailed("WebGL context does not exist");
+} else {
+    testPassed("WebGL context exists");
+
+    var filters = [gl.LINEAR, gl.NEAREST];
+    var drawbuffersFormats = [linearMask, srgbMask, linearMask | srgbMask];
+    for (var ii = 0; ii < filters.length; ++ii) {
+        for (var jj = 0; jj < drawbuffersFormats.length; ++jj) {
+            blitframebuffer_srgb_and_linear_drawbuffers(gl.SRGB8_ALPHA8, drawbuffersFormats[jj], filters[ii]);
+            blitframebuffer_srgb_and_linear_drawbuffers(gl.RGBA, drawbuffersFormats[jj], filters[ii]);
```
zhenyao: nit: maybe use sized format gl.RGBA8 to be consistent?

sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
```diff
@@ -0,0 +1,228 @@
...
+        gl.bindTexture(gl.TEXTURE_2D, tex_srgb1);
+        gl.texImage2D(gl.TEXTURE_2D, 0, gl.SRGB8_ALPHA8, size, size, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+        gl.framebufferTexture2D(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT2, gl.TEXTURE_2D, tex_srgb1, 0);
+        drawbuffers[0] = gl.COLOR_ATTACHMENT0;
+        drawbuffers[2] = gl.COLOR_ATTACHMENT2;
+    }
+
+    if (drawbuffersFormatMask & linearMask) {
+        gl.bindTexture(gl.TEXTURE_2D, tex_linear0);
+        gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, size, size, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
```
kenrussell: Use sized internal format here too?

sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
```diff
@@ -0,0 +1,228 @@
...
+        gl.framebufferTexture2D(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT1, gl.TEXTURE_2D, tex_linear0, 0);
+        gl.bindTexture(gl.TEXTURE_2D, tex_linear1);
+        gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, size, size, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
```
kenrussell: Here too?

sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
```diff
@@ -0,0 +1,228 @@
...
+    gl.deleteTexture(tex_linear0);
+    gl.deleteTexture(tex_srgb1);
+    gl.deleteTexture(tex_linear1);
+    gl.deleteTexture(tex_read);
+    gl.deleteFramebuffer(fbo_read);
+    gl.deleteFramebuffer(fbo_draw);
+}
+
+function checkPixel(color, expectedColor) {
+  var tolerance = 7;
```
kenrussell: Does the tolerance need to be this high? This test passes on my MacBook Pro Retina with NVIDIA GPU with a tolerance of 2.

zhenyao: This seems very large.  Maybe 3?

sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
Richard-Yunchao: done

sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
Richard-Yunchao: done

sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
Richard-Yunchao: I set it to 3 in the new commit.

sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
Richard-Yunchao: done

sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
```diff
@@ -0,0 +1,228 @@
...
+        gl.framebufferTexture2D(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT3, gl.TEXTURE_2D, tex_linear1, 0);
+        drawbuffers[1] = gl.COLOR_ATTACHMENT1;
+        drawbuffers[3] = gl.COLOR_ATTACHMENT3;
+    }
+
+    gl.drawBuffers(drawbuffers);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "setup draw framebuffer should succeed");
+
+    var status = gl.checkFramebufferStatus(gl.DRAW_FRAMEBUFFER) && gl.checkFramebufferStatus(gl.READ_FRAMEBUFFER);
+    if (status != gl.FRAMEBUFFER_COMPLETE) {
```
Richard-Yunchao: Ah... My mistake... It is weird to me too. The js works...

sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
Richard-Yunchao: Done

zhenyao:
What about kbr's question where the updated test fails?

Richard-Yunchao:
@zhenyao and @kenrussell , the blitframebuffer doesn't honor the drawBuffers() API at my side (MacOSX with Intel GPU, OGL 4.1). Only the COLOR_ATTACHMENT0 is rendered during blitFramebuffer. 

I will ask driver team whether they have latest driver to update.

@kenrussell , I do a slightly revision, where the blitframebuffer must filter the image: src region and dst region is different now. The previous commits have the same bounds for src region and dst region, the driver may can do some optimization without filtering. So, could you retest the lastest commit on Mac Nvidia, @kenrussell ?

kenrussell:
LGTM

All of the tests pass on my Retina MacBook Pro with NVIDIA GPU and 10.11.6.

Unfortunately, you'll need to resolve conflicts before this can be merged. Thanks for the new test.

sdk/tests/conformance2/rendering/blitframebuffer-srgb-and-linear-drawbuffers.html
Richard-Yunchao: done

zhenyao:
LGTM

Richard-Yunchao:
Thanks for your review, @kenrussell and @zhenyao, I fixed the conflicts now. But It seems that the pr is in a mess... I updated my local master to upstream, then rebase master to this branch locally, fix the conflict and then push this branch.

PR #2049, opened 2016-09-21
Add test cases for blitframebuffer from/to a multisampled srgb image

Richard-Yunchao:
@kenrussell , per our discussion about blitframebuffer from a multi-sampled srgb image to a single-sampled image, I add some tests to verify all restrictions: src/dst resource should have the same region, the same format/type, and the dst resource should not have multisampled image. 

PTAL. It can pass on both MacOSX. and Linux Desktop.

### DIFF
--- a/sdk/tests/conformance2/rendering/blitframebuffer-test.html
+++ b/sdk/tests/conformance2/rendering/blitframebuffer-test.html
@@ -295,7 +295,7 @@ function blit_framebuffer_multisampling_srgb() {
     gl.blitFramebuffer(0, 0, 2, 2, 0, 0, 2, 2, gl.COLOR_BUFFER_BIT, gl.LINEAR);
     wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "blitFramebuffer to a multisampled srgb image should generate INVALID_OPERATION.");
 
-    // BlitFramebuffer from a multisampled srgb image, the src region and the dst region should be exactly the same.
+    // BlitFramebuffer from a multisampled srgb image, the src region and the dst region must be exactly the same.
     gl.bindRenderbuffer(gl.RENDERBUFFER, rb1);
     gl.renderbufferStorage(gl.RENDERBUFFER, gl.SRGB8_ALPHA8, width, height);
     gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, fb1);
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/rendering/blitframebuffer-test.html
```diff
@@ -245,11 +256,81 @@
...
+    debug("");
+    debug("This test vefify the functionality of blitframebuffer from or to a multisampled srgb image.");
+
+    // Read buffer can have multisampled srgb image, but draw buffers can not.
+    var rb0 = gl.createRenderbuffer();
+    var fb0 = gl.createFramebuffer();
+    var rb1 = gl.createRenderbuffer();
+    var fb1 = gl.createFramebuffer();
+    gl.bindRenderbuffer(gl.RENDERBUFFER, rb0);
+    gl.renderbufferStorageMultisample(gl.RENDERBUFFER, 4, gl.SRGB8_ALPHA8, width, height);
```
kenrussell: This really should query gl.MAX_SAMPLES properly, rather than assuming that 4 samples will work. Or are we already making this assumption throughout the conformance suite? + @zhenyao

sdk/tests/conformance2/rendering/blitframebuffer-test.html
```diff
@@ -245,11 +256,81 @@
...
+    gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, fb1);
+    gl.framebufferRenderbuffer(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.RENDERBUFFER, rb1);
+    if (gl.checkFramebufferStatus(gl.DRAW_FRAMEBUFFER) != gl.FRAMEBUFFER_COMPLETE) {
+        testFailed("Framebuffer incomplete.");
+        return;
+    }
+    gl.blitFramebuffer(0, 0, 2, 2, 0, 0, 2, 2, gl.COLOR_BUFFER_BIT, gl.LINEAR);
+    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "blitFramebuffer to a multisampled srgb image should generate INVALID_OPERATION.");
+
+    // BlitFramebuffer from a multisampled srgb image, the src region and the dst region should be exactly the same.
```
kenrussell: Here and throughout: "should be" -> "must be"

sdk/tests/conformance2/rendering/blitframebuffer-test.html
zhenyao: No we didn't make that assumption throughout the conformance suite.

Richard-Yunchao: Done

sdk/tests/conformance2/rendering/blitframebuffer-test.html
Richard-Yunchao: Done.

Richard-Yunchao:
Thanks for your review, @kenrussell  and @zhenyao , the code has been updated accordingly. Please take another look. Thanks a lot!

kenrussell:
Thanks for revising. The new test LGTM.

PR #2050, opened 2016-09-21
Use existed function in wtu js library to simplify the test

Richard-Yunchao:
@kenrussell , per your suggestions at https://github.com/KhronosGroup/WebGL/pull/2034, this small change use functions in WebGLTestUtils to simplify the test. PTAL. Thanks a lot!

### DIFF
--- a/sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
+++ b/sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
@@ -93,7 +93,7 @@ function blitframebuffer_helper(readbufferFormat, drawbufferFormat, filter) {
 }
 
 function checkPixel(color, expectedColor) {
-  var tolerance = 1;
+  var tolerance = 3;
   return (Math.abs(color[0] - expectedColor[0]) <= tolerance &&
           Math.abs(color[1] - expectedColor[1]) <= tolerance &&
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/rendering/blitframebuffer-multisampled-readbuffer.html
```diff
@@ -126,35 +93,13 @@
 }
 
 function checkPixel(color, expectedColor) {
-  var tolerance = 7;
+  var tolerance = 1;
```
zhenyao: This seems pretty tight to me, maybe 3 or 5? which is enough to catch any bugs in the functionality.

Richard-Yunchao: On my MacBook, it can pass even the tolerance is 0. Because we draw the rect with a solid color, so the only pixel diff comes from precision lost during srgb conversion, if it can do correct srgb conversion. precision lost would be very very small color diff.
 
However, set it to 7 or even more bigger, it can catch the error too. Because the diff would be very big if srgb conversion itself is wrong or not done.

I set the tolerance to 3 in the new commit. Please take another look.   @zhenyao and @kenrussell . (Zhenyao, welcome back to office, :) )

kenrussell:
Very nice. Thanks for simplifying the code. LGTM

PR #2051, opened 2016-09-21
Fix the failure of pixelbufferobject.html

Richard-Yunchao:
Run the webgl2 conformance test on MacOSX and fount the pixelbufferobject.html failed. It turned out to be a typo. 

@kenrussell , @zhenyao , PTAL. Thanks.


zhenyao:
LGTM

Is it a recent regression?

Richard-Yunchao:
Yes, it is a recent regression, @zhenyao . Jeff pushed a patch two days ago to handle ArrayBufferView issue, which caused this regression.

kenrussell:
Thanks for catching and fixing this.

PR #2052, opened 2016-09-21
Remove tcuSkipList entries for shaderloop*

Kangz:
They were caused by a D3D compiler bug that Microsoft diagnosed. It is now worked around in ANGLE's shader translator.

See https://github.com/google/angle/commit/509e4560b3c897311a101d4d41bb27a51a6c439a


zhenyao:
This is great! LGTM

kenrussell:
Awesome work!!! Thank you for enabling these tests to run everywhere!

PR #2056, opened 2016-09-22
Replaced tabs with spaces so lint checks touching this file pass.

kenrussell:


kenrussell:
Simple change, but can't be reviewed in the tool because it's too large. tabs -> 8 spaces.

PR #2057, opened 2016-09-22
Fixes to ported dEQP multisample test.

kenrussell:
- Fixed bug in MultisampleCase.init where it was not returning a
  return value. Callers were only executing if it returned true. I'm
  not sure how any of these tests were ever running since it should
  have been implicitly returning "undefined" which should evaluate to
  false. Fixing this bug caused several more tests to start to run.

- Eliminated unnecessary typed array allocations for rendered images.
  Roughly twice the needed memory was being allocated.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
@@ -469,8 +469,6 @@ goog.scope(function() {
             '    o_color = v_color;\n' +
             '}\n';
 
-
-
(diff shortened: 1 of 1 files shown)


kenrussell:
@zhenyao @jdashg I'm going to merge this myself as I need to reference this (fixed) test case in a Radar I'm filing with Apple.

kenrussell:
FYI, I tested this on a MacBook Pro with NVIDIA GPU before merging. All tests passed.

zhenyao:
LGTM after the fact. Thanks for cleaning this up.

Issue #2058, opened 2016-09-22
 WebGL Conformance Test 1.0.4 (beta) does not work in IE11.


cx20:
The following errors will occur in IE11.

https://www.khronos.org/registry/webgl/sdk/tests/conformance/attribs/gl-bindAttribLocation-aliasing.html

> SCRIPT1046: Multiple definitions of a property not allowed in strict mode
> 
> webgl-test-utils.js (3116,28)

 I think the following lines is that it is unnecessary.

https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/js/webgl-test-utils.js#L3116

```javascript
Object.defineProperties(API, {
  noTexCoordTextureVertexShader: { value: noTexCoordTextureVertexShader, writable: false },
  simpleTextureVertexShader: { value: simpleTextureVertexShader, writable: false },
  simpleColorFragmentShader: { value: simpleColorFragmentShader, writable: false },
  simpleVertexShader: { value: simpleVertexShader, writable: false },
  simpleTextureFragmentShader: { value: simpleTextureFragmentShader, writable: false },
  simpleCubeMapTextureFragmentShader: { value: simpleCubeMapTextureFragmentShader, writable: false },
  //simpleTextureVertexShader: { value: simpleTextureVertexShader, writable: false }, // code has been duplicate.
  simpleVertexColorFragmentShader: { value: simpleVertexColorFragmentShader, writable: false },
  simpleVertexColorVertexShader: { value: simpleVertexColorVertexShader, writable: false }
});
```

kenrussell:
Thanks for your fix above.

PR #2059, opened 2016-09-22
Fix duplicate definition.

cx20:
Duplicate definition line was removed.

See: https://github.com/KhronosGroup/WebGL/issues/2058


kenrussell:
Thanks for this fix.

PR #2061, opened 2016-09-23
Make uninitialized-test-2.html create a WebGL 2.0 context by default.

kenrussell:
This avoids needing the ?webglVersion=2 query arg.

Thanks to jkummerow@ for pointing this out while investigating
https://bugs.chromium.org/p/v8/issues/detail?id=5386 .


kenrussell:
Trivial fix. Merging myself.

FYI @zhenyao @jdashg .

PR #2063, opened 2016-09-26
Clarify rationale note for texImage2D taking TexImageSource.

kenrussell:
Explicitly mention that it applies to ImageBitmap and ImageData as
well.


kenrussell:
Trivial change to clear up confusion that came up on https://codereview.chromium.org/2370703003/ . Merging myself.

PR #2064, opened 2016-09-27
Add test for TexImage with different data source

qkmiao:
@kenrussell PTAL. More tests will be added to conformance2/textures/[canvas|video|image|image_data...] for invalid internalformat/format/type combinations.

### DIFF
--- a/sdk/tests/conformance2/textures/misc/tex-image-with-different-data-source.html
+++ b/sdk/tests/conformance2/textures/misc/tex-image-with-different-data-source.html
@@ -1,0 +1,72 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/textures/misc/tex-image-with-different-data-source.html
```diff
@@ -0,0 +1,70 @@
...
+
+// Do TexImage2D taking a canvas source first.
+gl.bindTexture(gl.TEXTURE_2D, tex);
+gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA8, gl.RGBA, gl.UNSIGNED_BYTE, c);
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "TexImage2D taking a canvas source should succeed");
+gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGB10_A2, 16, 16, 0, gl.RGBA, gl.UNSIGNED_INT_2_10_10_10_REV, null);
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "Teximage2D taking a null array buffer should succeed");
+gl.deleteTexture(tex);
+
+// Do TexImage2D taking an array buffer first.
```
kenrussell: The second part of this test would need to be done with a newly created context in order to catch the bug that was almost introduced in Chrome, wouldn't it? If so, please update the test to do the two parts in separate contexts. Thanks.

qkmiao: PR has been updated, PTAL.

kenrussell: Thanks. Can you confirm that both portions of the test fail with your CL that was found to be incorrect?

qkmiao: With the incorrect CL:
First part failed.
Second part passed because RGB10_A2 format will cause image unpack step INVALID_VALUE error. With my latest chromium CL: second part will generate INVALID_VALUE at validation time before image unpack step.

kenrussell: OK. Thanks @qkmiao for verifying. Merging now.

PR #2066, opened 2016-09-27
Skip tests of bliting pixels out-of-bounds

Richard-Yunchao:
@zhenyao and @kenrussell , PTAL. Thanks a lot!

Please see the comment at https://bugs.chromium.org/p/chromium/issues/detail?id=644740.


zhenyao:
Let's talk about this in the working group this Thursday.
@kenrussell @jdashg FYI

kenrussell:
Agreed. This is on the agenda for tomorrow's working group meeting.

Richard-Yunchao:
I suppose that we should not skip these LINEAR related tests, instead, we need to fix them in Chromium. 

So I will close this one.

PR #2067, opened 2016-09-28
Fix a small bug for blitFramebuffer conversion test

Richard-Yunchao:
Basically a typo. PTAL. Thanks a lot!

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fFramebufferBlitTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fFramebufferBlitTests.js
@@ -726,7 +726,7 @@ goog.scope(function() {
         for (var ndx = 0; ndx < 2; ndx++) {
             ctx.bindFramebuffer(gl.FRAMEBUFFER, ndx ? dstFbo : srcFbo);
             if (ndx) {
-                gradientToDstShader.setGradient(ctx, gradShaderDstID, dstRangeInfo.valueMax, dstRangeInfo.valueMin);
+                gradientToDstShader.setGradient(ctx, gradShaderDstID, dstRangeInfo.valueMin, dstRangeInfo.valueMax);
                 rrUtil.drawQuad(ctx, gradShaderDstID, [-1, -1, 0], [1, 1, 0]);
             } else {
                 gradientToSrcShader.setGradient(ctx, gradShaderSrcID, srcRangeInfo.valueMin, srcRangeInfo.valueMax);


sdk/tests/deqp/functional/gles3/es3fFramebufferBlitTests.js
```diff
@@ -726,7 +726,7 @@ goog.scope(function() {
         for (var ndx = 0; ndx < 2; ndx++) {
             ctx.bindFramebuffer(gl.FRAMEBUFFER, ndx ? dstFbo : srcFbo);
             if (ndx) {
-                gradientToDstShader.setGradient(ctx, gradShaderDstID, dstRangeInfo.valueMax, dstRangeInfo.valueMin);
```
Richard-Yunchao: That's true. But the C++ code have issue too, I think. 
See the callee where it use valueMax and valueMin: https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/deqp/functional/gles3/es3fFboTestUtil.js#L251. 

And the shader where it use the uniform value: https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/deqp/functional/gles3/es3fFboTestUtil.js#L235. 

I didn't see why we need to reverse the valueMin and valueMax.

sdk/tests/deqp/functional/gles3/es3fFramebufferBlitTests.js
```diff
@@ -726,7 +726,7 @@ goog.scope(function() {
...
+                gradientToDstShader.setGradient(ctx, gradShaderDstID, dstRangeInfo.valueMin, dstRangeInfo.valueMax);
                 rrUtil.drawQuad(ctx, gradShaderDstID, [-1, -1, 0], [1, 1, 0]);
             } else {
                 gradientToSrcShader.setGradient(ctx, gradShaderSrcID, srcRangeInfo.valueMin, srcRangeInfo.valueMax);
```
qkmiao: This line isn't consistent with C++ code either.  Don't know the reason. dstRangeInfo.valueMax and srcRangeInfo.valueMax have the same value in this case.
https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fFramebufferBlitTests.cpp#409

Don't know why C++ code reverse valueMin and valueMax order. I think initialized data in dst  framebuffer doesn't take effect to the blit result.

Richard-Yunchao: That's true, the C++ code is not correct here. It should be srcRangeInfo.valueMin for the last parameter on that line @qkmiao pointed out. Because the dstRangeInfo.valueMax is different from srcRangleInfo.valueMax, they are defined here: https://android.googlesource.com/platform/external/deqp/+/master/framework/common/tcuTextureUtil.cpp#355 

Seems that the c++ code is buggy at this code snippet. @zhenyao

zhenyao:
I agree with you and agree the fix in this pull request is correct.  I just emailed dEQP folks and let's wait for them to confirm before merging, in case we misunderstood some design.

zhenyao:
From Pyry: "I think flipping the endpoints do have value for coverage. we need dstFbo have reversed gradient from srcFbo."

Richard-Yunchao:
OK. Then I will close this one. @zhenyao , thanks for your double check with c++ deqp guys.

PR #2070, opened 2016-09-29
update the spec for blitting pixels outside the read buffer

Richard-Yunchao:
@zhenyao , @kenrussell , PTAL. Thanks a lot!

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -1274,6 +1274,9 @@ WebGL2RenderingContext implements WebGL2RenderingContextBase;
         framebuffer. If the value of SAMPLE_BUFFERS for the read framebuffer is one and the value of
         SAMPLE_BUFFERS for the draw framebuffer is zero, the samples corresponding to each pixel location in
         the source are converted to a single sample before being written to the destination.
+
+        For any source pixel lying outside the read framebuffer, the corresponding destination pixel remains
+        untouched.
       </dd>
       <dt>
         <p class="idl-code">void framebufferTextureLayer(GLenum target, GLenum attachment, WebGLTexture? texture, GLint level, GLint layer)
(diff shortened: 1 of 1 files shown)


kenrussell:
Thank you Yunchao for writing up this clarification.

specs/latest/2.0/index.html
```diff
@@ -1274,6 +1274,9 @@ <h2 class="no-toc">Table of contents</h2>
         framebuffer. If the value of SAMPLE_BUFFERS for the read framebuffer is one and the value of
         SAMPLE_BUFFERS for the draw framebuffer is zero, the samples corresponding to each pixel location in
         the source are converted to a single sample before being written to the destination.
+
+        For any source pixel lying outside the read framebuffer, the corresponding destination pixel remains
+        untouched.
```
kenrussell: It's a good question. Looking back at the old Khronos bug https://cvs.khronos.org/bugzilla/show_bug.cgi?id=3576 (sorry, non-public) linked from http://crbug.com/644740 , defining the math for the coordinate system transformations seemed difficult, and the OpenGL ES working group themselves opted for simpler spec language.

@Kangz do you have any suggestion about how to phrase this?

There does seem to be some ambiguity about what happens with pixels on the border of the source rectangle. For a LINEAR blit, CLAMP_TO_EDGE filtering should presumably be used, so that even if the source rectangle is fully inside the source framebuffer, pixels on the other side aren't referenced at all.

Richard-Yunchao: For pixels lying on the border, if the majority of the pixels is in the blit region, I think it should be blitted for NEAREST filter. And it should be always blitted for pixels lying on the border for LINEAR filter. WDYT? @zhenyao , @kenrussell , @Kangz ? 

This Chromium revision will follow this criteria: https://codereview.chromium.org/2409523002/.

zhenyao: I think for NEAREST or LINEAR, the destination pixels should be fully within the adjusted region (due to src being not fully within the read buffer bound).

Kangz: @zhenyao the spec includes special language to say that GL_LINEAR filtering might cause reads outside of the src region in that case.
What we want is for a dest pixel to be written iff it is in the intersection of the the draw framebuffer bounds and the dest region, and its corresponding sample point is in the intersection of the read framebuffer and the source region ([x0, x1] x [y0, y1]). sampling at x0 + y0 will access source pixels at position (x0 - 0.5, y0 - 0.5) or clamp to edge.

My take on the clarification:
A destination pixel is written only if its corresponding sample point is included both in the source region and the source framebuffer bounds.

Richard-Yunchao: I agree with @zhenyao. Let's discuss this issue at https://github.com/KhronosGroup/WebGL/pull/2098.

PR #2071, opened 2016-09-29
Clean up some text that's outdated in copyBufferSubData.

zhenyao:
@kenrussell PTAL


kenrussell:
LGTM. From re-reading this section it's clear that this sentence was accidentally left over from an earlier revision of the spec.

PR #2072, opened 2016-09-30
Fix bugs in blitFramebuffer-filter-outofbounds.html

Richard-Yunchao:
Sorry, stay too late yesterday... 

PTAL. Thanks! @zhenyao , @kenrussell .


zhenyao:
lgtm

Richard-Yunchao:
I am not 100% sure whether I understand this issue correctly: 
For pixels lying outside the read fb, the corresponding dst pixels should keep untouched. But the outside pixels should be considered for scaling when blitting. 
For pixels lying outside the draw fb, the pxiels are simply ignored (dst region intersects with draw buffers), the outside pixels would not be considered for scaling when blitting.

Richard-Yunchao:
@zhenyao and @kenrussell .

zhenyao:
What I am saying is (this is the wrong place because the discussion was at another pull request), for a pixel on the boundary of draw fb, if the filter is LINEAR, then its value could be determined by more than one pixel from the read fb, and these pixels could be partially in bound, partially out of bound.

PR #2073, opened 2016-09-30
Fix more bugs in blitframebuffer-outside-readbuffer.html

Richard-Yunchao:
Update the test cases. Any pixel lying outside of the read buffer, the corresponding dst pixel should remain untouched, instead of clamp_to_edge. 

I will merge this one with blitframebuffer-filter-outofbounds.html later (Or just remove this file if deqp has thoroughly tested the feature). This one test out-of-bounds too, but have tests to reversed src/dst, and pixels outside every edge the readbuffer, and so forth. 

PTAL. @zhenyao and @kenrussell . 

It turns out that this test can pass in Linux now, but fail in MacOSX.


zhenyao:
Great. LGTM.

PR #2074, opened 2016-10-01
Move a test to the right folder

zhenyao:
@kenrussell @RafaelCintron FYI


zhenyao:
Merging myself

kenrussell:
Could you also update the 00_test_list.txt files in the two folders?

PR #2076, opened 2016-10-01
Split WebGL 2.0 version of EXT_disjoint_timer_query into its own file.

kenrussell:
It was too difficult to specialize the IDL and documentation for WebGL
2.0 and keep it in the same file. Document that the extension is still
fetched via getExtension('EXT_disjoint_timer_query').

### DIFF
--- a/extensions/EXT_disjoint_timer_query_webgl2/extension.xml
+++ b/extensions/EXT_disjoint_timer_query_webgl2/extension.xml
@@ -32,9 +32,9 @@
     <features>
       <feature>
         This extension provides the same functionality as <a
-        href="../EXT_disjoint_timer_query/">EXT_disjoint_timer_query</a>, and is queried from the
-        context under the same name. This IDL and description are specialized for WebGL 2.0, which
-        incorporates query objects into the core specification.
+        href="../EXT_disjoint_timer_query/">EXT_disjoint_timer_query</a>. The IDL, description, and
+        extension name are specialized for WebGL 2.0, which incorporates query objects into the core
+        specification.
       </feature>
     </features>
   </overview>
@@ -42,8 +42,6 @@
   <idl xml:space="preserve">
 typedef unsigned long long GLuint64EXT;
 
-// Note that this extension is still queried via
-// getExtension('EXT_disjoint_timer_query').
 [NoInterfaceObject]
 interface EXT_disjoint_timer_query_webgl2 {
   const GLenum QUERY_COUNTER_BITS_EXT      = 0x8864;
@@ -111,7 +109,7 @@ interface EXT_disjoint_timer_query_webgl2 {
   <samplecode xml:space="preserve">
     <pre>
         // Example (1) -- uses beginQuery/endQuery.
-        var ext = gl.getExtension('EXT_disjoint_timer_query');
+        var ext = gl.getExtension('EXT_disjoint_timer_query_webgl2');
         var query = gl.createQuery();
         gl.beginQuery(ext.TIME_ELAPSED_EXT, query);
 
@@ -137,7 +135,7 @@ interface EXT_disjoint_timer_query_webgl2 {
         //----------------------------------------------------------------------
 
         // Example (2) -- same as the example above, but uses queryCounterEXT instead.
-        var ext = gl.getExtension('EXT_disjoint_timer_query');
+        var ext = gl.getExtension('EXT_disjoint_timer_query_webgl2');
         var startQuery = gl.createQuery();
         var endQuery = gl.createQuery();
         gl.queryCounterEXT(startQuery, ext.TIMESTAMP_EXT);
@@ -166,7 +164,7 @@ interface EXT_disjoint_timer_query_webgl2 {
 
         // Example (3) -- check the number of timestamp bits to determine how to best
         // measure elapsed time.
-        var ext = gl.getExtension('EXT_disjoint_timer_query');
+        var ext = gl.getExtension('EXT_disjoint_timer_query_webgl2');
         var timeElapsedQuery;
         var startQuery;
         var endQuery;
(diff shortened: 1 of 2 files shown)


kenrussell:
Please review. This addresses feedback raised on the public_webgl mailing list here:
https://www.khronos.org/webgl/public-mailing-list/archives/1609/msg00093.php

and is structured to address the working group's feedback (reuse the WebGL 2.0 core context functionality; eliminate redundant enums and methods).

Once this is approved, I'll update the conformance test.

CC @jdashg @zhenyao @pyalot @juj

extensions/EXT_disjoint_timer_query_webgl2/extension.xml
```diff
@@ -0,0 +1,240 @@
...
+// Note that this extension is still queried via
+// getExtension('EXT_disjoint_timer_query').
+[NoInterfaceObject]
+interface EXT_disjoint_timer_query_webgl2 {
+  const GLenum QUERY_COUNTER_BITS_EXT      = 0x8864;
+  const GLenum TIME_ELAPSED_EXT            = 0x88BF;
+  const GLenum TIMESTAMP_EXT               = 0x8E28;
+  const GLenum GPU_DISJOINT_EXT            = 0x8FBB;
+
+  void queryCounterEXT(WebGLQuery? query, enum target);
```
jdashg: No point in leaving this nullable if we're just going to generate INVALID_OP.

extensions/EXT_disjoint_timer_query_webgl2/extension.xml
```diff
@@ -0,0 +1,240 @@
...
+        This extension provides the same functionality as <a
+        href="../EXT_disjoint_timer_query/">EXT_disjoint_timer_query</a>, and is queried from the
+        context under the same name. This IDL and description are specialized for WebGL 2.0, which
+        incorporates query objects into the core specification.
+      </feature>
+    </features>
+  </overview>
+
+  <idl xml:space="preserve">
+typedef unsigned long long GLuint64EXT;
```
jdashg: Don't we already have a GLuint64 type in the webgl2 webidl?

kenrussell: No, only GLint64. GLuint64 isn't that useful -- not all of the possible values are representable in JavaScript -- but I left this typedef alone for parity with the WebGL 1.0 version of the extension.

extensions/EXT_disjoint_timer_query_webgl2/extension.xml
kenrussell: Good point; removed in the current commit. (We should use "squash and merge" when merging this.)

kenrussell:
Thanks for your review. Any feedback on whether this should be queried via:
  getExtension('EXT_disjoint_timer_query')
or
  getExtension('EXT_disjoint_timer_query_webgl2')
?

It would be more consistent to use EXT_disjoint_timer_query_webgl2. If you agree, I'll update this pull request.

kenrussell:
Since there was no feedback I've made the decision to query the WebGL 2.0 version of the extension under the name "EXT_disjoint_timer_query_webgl2" for consistency. I'm merging this now and preparing an update to the conformance tests to handle the name change.

extensions/EXT_disjoint_timer_query_webgl2/extension.xml
```diff
@@ -0,0 +1,238 @@
...
+          adjustObjectLODBasedOnDrawTime(timeElapsed);
+        }
+
+        //----------------------------------------------------------------------
+
+        // Example (2) -- same as the example above, but uses queryCounterEXT instead.
+        var ext = gl.getExtension('EXT_disjoint_timer_query_webgl2');
+        var startQuery = gl.createQuery();
+        var endQuery = gl.createQuery();
+        gl.queryCounterEXT(startQuery, ext.TIMESTAMP_EXT);
```
juj: This looks odd, should it be `ext.queryCounterEXT()` instead? Or does the new `_webgl2` extension introduce an extension function `queryCounterEXT` on the core GL context? This would be a first, other extensions don't add new functions with extension string suffixes ("...EXT" here) on the core GL object, but they add functions on the extension objects. (Just want to confirm that this was the intent and not a typo, I'm not sure I have a strong preference either way)

extensions/EXT_disjoint_timer_query_webgl2/extension.xml
```diff
@@ -0,0 +1,238 @@
...
+  const GLenum QUERY_COUNTER_BITS_EXT      = 0x8864;
+  const GLenum TIME_ELAPSED_EXT            = 0x88BF;
+  const GLenum TIMESTAMP_EXT               = 0x8E28;
+  const GLenum GPU_DISJOINT_EXT            = 0x8FBB;
+
+  void queryCounterEXT(WebGLQuery query, enum target);
+};
+  </idl>
+
+  <newtok>
```
juj: Sorry, I'm unfamiliar with reading these in text form: Does this `<newtok> ... </newtok>` section mean that the core GL context object function `GLcontext.beginQuery()` gains a new ability to accept `TIME_ELAPSED_EXT`? (and same way for the other newtoks below?)

extensions/EXT_disjoint_timer_query_webgl2/extension.xml
```diff
@@ -0,0 +1,238 @@
...
+        } else {
+          timeElapsedQuery = gl.createQuery();
+          gl.beginQuery(ext.TIME_ELAPSED_EXT, timeElapsedQuery);
+        }
+
+        // Draw object
+        gl.drawElements(...);
+
+        if (useTimestamps) {
+          gl.queryCounterEXT(endQuery, ext.TIMESTAMP_EXT);
```
kenrussell: Fixing here and throughout in #2088 .

PR #2078, opened 2016-10-06
clarify and test corner cases of getBufferSubData

kainino0x:
@zhenyao PTAL

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -1200,7 +1200,8 @@ WebGL2RenderingContext implements WebGL2RenderingContextBase;
         <code>dstBuffer.length - dstOffset</code>; otherwise, let
         <code>copyLength</code> be <code>length</code>.
         <br><br>
-        Copy <code>copyLength</code> typed elements (each of size <code>dstBuffer.BYTES_PER_ELEMENT</code>)
+        If <code>copyLength</code> is greater than zero,
+        copy <code>copyLength</code> typed elements (each of size <code>dstBuffer.BYTES_PER_ELEMENT</code>)
         from <code>buf</code> into <code>dstBuffer</code>,
         reading <code>buf</code> starting at byte index <code>srcByteOffset</code> and
         writing into <code>dstBuffer</code> starting at element index <code>dstOffset</code>.
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -1202,13 +1202,15 @@ WebGL2RenderingContext implements WebGL2RenderingContextBase;
               an <code>INVALID_VALUE</code> error is generated.
           <li>If <code>srcByteOffset</code> is less than zero, an <code>INVALID_VALUE</code>
               error is generated.
-          <li>If zero is bound to <code>target</code>, an <code>INVALID_OPERATION</code>
+          <li>If no WebGLBuffer is bound to <code>target</code>, an <code>INVALID_OPERATION</code>
               error is generated.
           <li>If <code>target</code> is <code>TRANSFORM_FEEDBACK_BUFFER</code>,
               and any transform feedback object is currently active, an <code>INVALID_OPERATION</code>
               error is generated.
           <li>If <code>srcByteOffset + length*ElementSize(dstBuffer)</code> is larger
               than the length of the GL buffer, generate INVALID_OPERATION.
+          <li>If <code>dstBuffer.length</code> equals <code>dstOffset</code>, no data is written to
+            <code>dstBuffer</code>.
         </ul>
         If any error is generated, no data is written to <code>dstBuffer</code>.
         <br><br>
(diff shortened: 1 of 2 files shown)


kainino0x:
This technically includes a regression test for Chrome, but I don't think there is a crbug for it.

kainino0x:
Updated.

specs/latest/2.0/index.html
```diff
@@ -1202,13 +1202,15 @@ <h2 class="no-toc">Table of contents</h2>
...
-          <li>If zero is bound to <code>target</code>, an <code>INVALID_OPERATION</code>
+          <li>If no WebGLBuffer is bound to <code>target</code>, an <code>INVALID_OPERATION</code>
               error is generated.
           <li>If <code>target</code> is <code>TRANSFORM_FEEDBACK_BUFFER</code>,
               and any transform feedback object is currently active, an <code>INVALID_OPERATION</code>
               error is generated.
           <li>If <code>srcByteOffset + length*ElementSize(dstBuffer)</code> is larger
               than the length of the GL buffer, generate INVALID_OPERATION.
+          <li>If <code>dstBuffer.length</code> equals <code>dstOffset</code>, no data is written to
+            <code>dstBuffer</code>.
```
kenrussell: Is the implication that length == 0 in this case since this check is listed after "if dstOffset+length is greater than dstBuffer.length"? That's not obvious. It would be possible for the user to specify both the dstOffset and length arguments, instead of taking the default computed value, and then both the first and last rules in this list could apply.

Should this instead state something like: "if the computed length equals zero, no data is written to dstBuffer"?

zhenyao:
LGTM once kbr's suggestion is addressed.

specs/latest/2.0/index.html
zhenyao: I think it's OK to update line 1196 to max(0, dstBuffer.length - dstOffset)

kainino0x:
I've rewritten the spec text; please review.

specs/latest/2.0/index.html
```diff
@@ -1193,22 +1193,33 @@ <h2 class="no-toc">Table of contents</h2>
...
+        Let <code>buf</code> be the buffer bound to <code>target</code>.
+        If <code>length</code> is 0, let <code>copyLength</code> be
+        <code>dstBuffer.length - dstOffset</code>; otherwise, let
+        <code>copyLength</code> be <code>length</code>.
+        <br><br>
+        Copy <code>copyLength</code> typed elements (each of size <code>dstBuffer.BYTES_PER_ELEMENT</code>)
+        from <code>buf</code> into <code>dstBuffer</code>,
+        reading <code>buf</code> starting at byte index <code>srcByteOffset</code> and
+        writing into <code>dstBuffer</code> starting at element index <code>dstOffset</code>.
+        If <code>copyLength</code> is 0, no data is written to <code>dstBuffer</code>.
```
kenrussell: Does this fully specify all of the corner cases you ran into? Do we need to specify the case when copyLength is less than zero (though there are some signed/unsigned arithmetic issues in the expression above since dstOffset and length are both unsigned ints)?

kainino0x: The entire paragraph has issues if the error checks aren't done first: `buf` may not be defined, `copyLength` may be negative/underflow. But this error condition should catch it: "If <code>dstOffset</code> is greater than <code>dstBuffer.length</code>"

kainino0x:
I just added [more tests](https://github.com/KhronosGroup/WebGL/pull/2078/commits/939f82d5ab66f20108baa92b2651b836cde1687e) based on @Kangz's [comment on getBufferSubDataAsync](https://github.com/KhronosGroup/WebGL/pull/2079#discussion_r82403797). PTAL.

Once again, there's a regression test here but no crbug for it.

specs/latest/2.0/index.html
zhenyao: Worth mentioning here that "but generates no GL error because of this", something like that.

specs/latest/2.0/index.html
```diff
@@ -1193,22 +1193,33 @@ <h2 class="no-toc">Table of contents</h2>
...
-        elements into <code>dstBuffer</code>, reading starting at <code>srcByteOffset</code>
-        and writing starting at <code>dstOffset</code> into <code>dstBuffer</code>.
+        Reads back data from the bound WebGLBuffer into <code>dstBuffer</code>.
+        <br><br>
+        Let <code>buf</code> be the buffer bound to <code>target</code>.
+        If <code>length</code> is 0, let <code>copyLength</code> be
+        <code>dstBuffer.length - dstOffset</code>; otherwise, let
+        <code>copyLength</code> be <code>length</code>.
+        <br><br>
+        Copy <code>copyLength</code> typed elements (each of size <code>dstBuffer.BYTES_PER_ELEMENT</code>)
```
zhenyao: Here worth mentioning, if copyLength > 0.

I feel it's OK to leave copyLength < 0 behavior out here because it's defined later.

kainino0x: That sounds like a good resolution to me, thanks.

zhenyao:
LGTM, but @kenrussell should take a look.

kenrussell:
LGTM. Thanks for thinking this through so carefully. Merging now.

PR #2079, opened 2016-10-06
WEBGL_get_buffer_sub_data_async extension

### LINKED ISSUE #1938: Add conformance tests for getBufferSubDataAsync
In #1937 a new getBufferSubDataAsync method is being proposed. The current pull request doesn't contain conformance tests so this issue is being filed to ensure that they are added before WebGL 2.0 is shipped, assuming that this pull request is merged.

kainino0x:
I took @kenrussell's spec text from #1937 but significantly rewrote it text to exactly match the new text for getBufferSubData where appropriate. I also changed the IDL to match getBufferSubData.

This entry point solves the problem of synchronous, blocking GPU-to-CPU memory readback on the main thread.

Some performance results in real-world usage (screen picking):
https://github.com/kainino0x/getBufferSubDataAsync-Demo
- [x] Copy rewording from #2078 once it's settled
- [x] Don't test ordering: https://github.com/KhronosGroup/WebGL/pull/2079#discussion_r82401676
- [x] Add test for the transform feedback error case to the test added in #2097 (it should be added to `verifyGetBufferSubData`)
- [x] Move to an extension
- [x] Remove unnecessary manual tests in `sdk/tests/extra/`.

This PR fixes #1938.


sdk/tests/conformance2/buffers/getBufferSubDataAsync-stress.html
```diff
@@ -0,0 +1,98 @@
...
+
+var TEST_COUNT = 300;
+var last_counter = -1;
+var promises = [];
+promises.length = TEST_COUNT;
+debug("kicking off " + TEST_COUNT + " getBufferSubDataAsync calls");
+for (var i = 0; i < TEST_COUNT; i++) {
+  (function() {
+    var counter = i;
+    var color = [counter & 0xff, (counter << 8) & 0xff, (counter << 16) & 0xff, 255];
```
kenrussell: Did you mean to right shift instead of left shift these? They'll always shift in zero bits as written.

sdk/tests/conformance2/buffers/getBufferSubDataAsync.html
```diff
@@ -0,0 +1,249 @@
...
+  var code = "promise = gl.getBufferSubDataAsync(gl.ARRAY_BUFFER, -1, dstData)";
+  wtu.shouldGenerateGLError(gl, gl.INVALID_VALUE, code);
+  return expectRejected(promise, code);
+});
+
+allPromises.then(function() {
+  debug("");
+  finishTest();
+}).catch(function(e) {
+  testFailed(e.toString());
```
kenrussell: Should this also call finishTest so the test harness (webgl-conformance-test.html) doesn't report this as a timeout if an exception is thrown?

sdk/tests/conformance2/buffers/getBufferSubDataAsync-stress.html
kenrussell: Still looking for an update in this area.

sdk/tests/conformance2/buffers/getBufferSubDataAsync-stress.html
```diff
@@ -0,0 +1,98 @@
...
+          buf[1] == color[1] &&
+          buf[2] == color[2] &&
+          buf[3] == color[3]) {
+        testPassed("color readback #" + counter + " was correct");
+      } else {
+        testFailed("color readback #" + counter + "was incorrect: was " +
+            buf + " but expected " + color);
+      }
+      if (counter <= last_counter) {
+        testFailed("getBufferSubDataAsync calls #" + last_counter + " and #" + counter + " out of order");
```
Kangz: The getBufferSubDataAsync promises being called in order doesn't appear in the spec (or I didn't find it) and seems to be an implementation detail. How about having a vector of bools and in Promise.all(promises).then() check that it is all true?

sdk/tests/conformance2/buffers/getBufferSubDataAsync.html
```diff
@@ -0,0 +1,249 @@
...
+// For each test in this chain, if it performs test checks in a promise,
+// it MUST return that promise so that it completes before the next test.
+var allPromises = Promise.resolve();
+
+allPromises = allPromises.then(function() {
+  debug("Testing promise rejection");
+});
+
+allPromises = allPromises.then(function() {
+  var code = "promise = gl.getBufferSubDataAsync(gl.ARRAY_BUFFER, 0, new ArrayBuffer(4))";
```
Kangz: Why should these be rejected? Is it because no buffer is bound to gl.ARRAY_BUFFER (I think some might be)?

sdk/tests/conformance2/buffers/getBufferSubDataAsync.html
```diff
@@ -0,0 +1,249 @@
...
+  return expectRejected(promise, code);
+});
+
+allPromises = allPromises.then(function() {
+  debug("Test that getBufferSubDataAsync fails when offset is less than 0");
+  var code = "promise = gl.getBufferSubDataAsync(gl.ARRAY_BUFFER, -1, dstData)";
+  wtu.shouldGenerateGLError(gl, gl.INVALID_VALUE, code);
+  return expectRejected(promise, code);
+});
+
```
Kangz: A couple more tests that might be useful:
 - Check that getBufferSubDataAsync on an uninitialized (gl.bufferData with null) buffer returns zeros.
 - Check that getBufferSubDataAsync immediately followed by a buffer resize to be too small doesn't crash.

sdk/tests/conformance2/buffers/getBufferSubDataAsync.html
kainino0x: Those are type checks, which I didn't make clear at all. I added debug prints to explain.

```
PASS correctly rejected with TypeError: Failed to execute 'getBufferSubDataAsync' on 'WebGL2RenderingContext': parameter 3 is not of type 'ArrayBufferView'.: promise = gl.getBufferSubDataAsync(gl.ARRAY_BUFFER, 0, new ArrayBuffer(4))
```

```
PASS correctly rejected with TypeError: Failed to execute 'getBufferSubDataAsync' on 'WebGL2RenderingContext': parameter 3 is not of type 'ArrayBufferView'.: promise = gl.getBufferSubDataAsync(gl.ARRAY_BUFFER, 0, null)
```

sdk/tests/conformance2/buffers/getBufferSubDataAsync-stress.html
kenrussell: I'm not sure how the guarantees about what data is seen in the ArrayBufferView would be maintained if the promises resolved out of order. The same ArrayBufferView might be used as the target to getBufferSubData multiple times, and the promises' "then" callbacks would assume the data would be observed sequentially.

Should we say something in the spec about sequential calls A and B to getBufferSubDataAsync, and that the promise for call A must be resolved before the promise to call B, assuming both are not rejected early?

It'd be fine by me to change the test. Is there a simple way to express success? I would assume that Promise.all's resolve callback would only be called upon success of all of the individual promises and their resolve callbacks -- so throwing an exception out of one of them would cause Promise.all to fail. Could you investigate?

kainino0x: Since the ArrayBufferView isn't written until promise resolution, I don't see a problem there. We don't make any guarantees about the contents of the ArrayBufferView after the promise callback is executed; only during its execution.

Yes, failing one would cause Promise.all to fail. But there can be `testFailed` messages inside the individual callbacks as well so I think that would be okay.

specs/latest/2.0/index.html
```diff
@@ -1216,6 +1219,77 @@ <h2 class="no-toc">Table of contents</h2>
...
+        only written to <code>dstData</code> when the promise is fulfilled; it is available in
+        the Promise's success callback, but not earlier.
+
+        <br><br>
+        If a failure occurs that prevents the promise from being fulfilled, then the returned
+        Promise is rejected with an "InvalidStateError" DOMException, and no data is written to
+        <code>dstData</code>. If any of the above errors occur, the Promise is already rejected upon
+        return. If the context is lost before the Promise is fulfilled, or if <code>dstData</code>
+        is neutered at the time the Promise is fulfilled, then the Promise will be rejected later.
+        In this case, no WebGL error is generated. If the promise rejects, no data is written to
```
kainino0x: @kenrussell Just want to make sure that this is the semantic we want. Right now no WebGL error will ever be generated upon resolution of the promise.

sdk/tests/conformance2/buffers/getBufferSubDataAsync-stress.html
Kangz: Like Kai, I don't think we need to add ordering to the spec, the fact that all the promises are resolved on the borwsing context event loop gives the guarantee that inside the promise.then, the ArrayBufferView content is correct.

kenrussell: OK. Given your analyses I think it's fine to leave the spec as is.

specs/latest/2.0/index.html
kenrussell: Yes, that sounds like the best semantic to me.

kainino0x:
I merged master into this branch, updated the spec language to match getBufferSubData, etc. I think I've addressed all of the outstanding comments and issues.

Kangz:
Still LGTM

kainino0x:
Added a test for the transform feedback error case (finally).

Also added in getBufferSubData tests in #2097

kainino0x:
> Added a test for the transform feedback error case (finally).

Removed, will add it back to this PR once that one has landed and this PR's fate has stabilized.

kenrussell:
Excellent work. For a little more background for other reviewers: the new form of the spec text came from a suggestion from Elliott Sprehn from the Blink team, pointing us at other web specs that return promises.

LGTM.

kainino0x:
I've finally moved this entire spec into an extension. The extension is implemented and tested on my local Chromium build. Aside from the new specification text, everything is as it was before. Ready for review.

sdk/tests/conformance2/extensions/webgl-get-buffer-sub-data-async-stress.html
```diff
@@ -0,0 +1,111 @@
...
+  gl.bufferData(gl.PIXEL_PACK_BUFFER, 4, gl.DYNAMIC_READ);
+  gl.bindBuffer(gl.PIXEL_PACK_BUFFER, null);
+  var readbackBuffer = new Uint8Array(4);
+
+  var TEST_COUNT = 300;
+  var lastCounter = -1;
+  var promises = [];
+  debug("kicking off " + TEST_COUNT + " getBufferSubDataAsync calls");
+  for (var i = 0; i < TEST_COUNT; i++) {
+    (function() {
```
zhenyao: Curious: why make the loop body a function? why regular code will not do?

sdk/tests/conformance2/extensions/webgl-get-buffer-sub-data-async.html
```diff
@@ -0,0 +1,330 @@
...
+    gl.bindBuffer(gl.ARRAY_BUFFER, null);
+    return expectRejected(promise, code);
+  });
+
+  allPromises = allPromises.then(function() {
+    debug("");
+  });
+
+  for (var i = 0; i < types.length; i++) {
+    (function() {
```
zhenyao: Same question here.

sdk/tests/extra/getBufferSubDataAsync-dom-suspension-manual.html
```diff
@@ -0,0 +1,125 @@
...
+  }, 0);
+}
+
+function render() {
+  var TEST_COUNT = 100;
+  var promises = [];
+  debug("kicking off " + TEST_COUNT + " getBufferSubDataAsync calls");
+  for (var i = 0; i < TEST_COUNT; i++) {
+    insertTimeoutAssertion();
+    (function() {
```
zhenyao: Same question here?

sdk/tests/conformance2/extensions/webgl-get-buffer-sub-data-async-stress.html
kainino0x: This is a javascript programming pattern that guards against a weirdness of closures, which is that they always capture variables, not values. So if you capture `i`, then all of the closures (promise callbacks) will see `i == TEST_COUNT` instead of the loop value.

After writing this, though, I learned about `let`, which actually will solve this. I'm going to go ahead and clean this up using `let`.

sdk/tests/extra/getBufferSubDataAsync-dom-suspension-manual.html
kainino0x: I actually forgot about these tests and haven't updated them. I need to do this before I merge.

zhenyao: What about leaving these manual tests out of this pull request?

kainino0x: Sure, good call. Done.

zhenyao:
lgtm

kenrussell:
Very nice @kainino0x and thanks @zhenyao for carrying forward the code review. LGTM after the fact.

PR #2080, opened 2016-10-06
Add a test to cover an unreferenced varyings bug in ANGLE.

null77:
See http://crbug.com/651493

PTAL @kenrussell @zhenyao


zhenyao:
I am curious: what's so specific about unreferenced variables to be of number 2? If we have only one unreferenced variable, without the ANGLE fix, it runs OK?

null77:
Yes, there's an early-exit of a for loop. Works fine for one variable, 2+ fails.

zhenyao:
LGTM

kenrussell:
Excellent! Thank you Jamie for the conformance test.

PR #2081, opened 2016-10-06
Add a test for an ANGLE bug with UBO sizing.

null77:
When using a UBO with a size larger than the D3D11 maximum
constant buffer size, ANGLE would incorrectly try to memcpy
with the unclamped size.

See http://crbug.com/651493

PTAL @zhenyao @kenrussell

### DIFF
--- a/sdk/tests/conformance2/buffers/one-large-uniform-buffer.html
+++ b/sdk/tests/conformance2/buffers/one-large-uniform-buffer.html
@@ -1,0 +1,152 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/buffers/one-large-uniform-buffer.html
```diff
@@ -0,0 +1,152 @@
...
+void main()
+{
+    fragColor = color;
+}
+</script>
+<script>
+"use strict";
+description("This test covers an ANGLE bug when using a large uniform block data store. ANGLE would confuse an internal clipped uniform buffer size and produce an assert or error.");
+
+debug("");
```
kenrussell: In a follow-on pull request, could you add a link like:
debug('Regression test for &lt;a href="http://crbug.com/651493"&gt;http://crbug.com/651493&lt;/a&gt;');

to this test and your other recent one? It'll make it a little easier to trace the test back to the underlying bug.

PR #2084, opened 2016-10-08
Fix infinite loop errors in es3fTextureFormatTests.js

Jiawei-Shao:
The last argument of function assertMsgOptions() should be set
to false in es3fTextureFormatTests.js or this function will throw
an exception when the test fails, which will destroy the current
function stack and prevent the updates on some important
variables, such as this.m_curFace.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fTextureFormatTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fTextureFormatTests.js
@@ -176,7 +176,7 @@ es3fTextureFormatTests.Texture2DFormatCase.prototype.iterate = function() {
     // Compare and log.
     var isOk = glsTextureTestUtil.compareImages(referenceFrame, renderedFrame, threshold);
 
-    assertMsgOptions(isOk, es3fTextureFormatTests.testDescription(), true, true);
+    assertMsgOptions(isOk, es3fTextureFormatTests.testDescription(), true, false);
     return tcuTestCase.IterateResult.STOP;
 };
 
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/functional/gles3/es3fTextureFormatTests.js
```diff
@@ -176,7 +176,7 @@ es3fTextureFormatTests.Texture2DFormatCase.prototype.iterate = function() {
     // Compare and log.
     var isOk = glsTextureTestUtil.compareImages(referenceFrame, renderedFrame, threshold);
 
-    assertMsgOptions(isOk, es3fTextureFormatTests.testDescription(), true, true);
+    assertMsgOptions(isOk, es3fTextureFormatTests.testDescription(), true, false);
```
Jiawei-Shao: Function assertMsgOptions() is defined here:
https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/js/js-test-pre.js#L679
If test fails, assertMsgOptions() will call testFailedOptions():
https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/js/js-test-pre.js#L311
If exthrow is true, then testFailedOptions() will throw an exception and destroy current function stack.
For example, this.m_curFace will not be updated when an exception is thrown in this.testFace():
https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/deqp/functional/gles3/es3fTextureFormatTests.js#L328

qkmiao:
lgtm

zhenyao:
LGTM

PR #2085, opened 2016-10-09
Add test for TexImage with bad internalformat/format/type combination

qkmiao:
Generate INVLID_OPERATION error for test where each separate argument
is valid but their combination is invalid.


zhenyao:
lgtm

PR #2086, opened 2016-10-10
Fix bug when blitting pixels outside read framebuffer

Richard-Yunchao:
After holiday, I carefully revisited the behavior of blitting from pixels outside read framebuffer TO PIXELS OUTSIDE OF  DRAW FRAMEBUFFER, I found that I misunderstand the behavior suggested in this issue: https://cvs.khronos.org/bugzilla/show_bug.cgi?id=3576. The out-of-bounds pixels in dst region should be considered for scaling during blitting, although these outofbounds pixels in dst region will not be drew. But they should not be ignored. 

This small pr will correct the behavior.

After this revision applied, this test can pass in Linux and Windows (But still fail on MacOSX as expected). PTAL. @zhenyao and @kenrussell .


Kangz:
LGTM.

zhenyao:
lgtm

PR #2088, opened 2016-10-11
Fixed errors in sample code from #2076 pointed out by @juj.

kenrussell:


kenrussell:
Trivial fix. Merging myself.

zhenyao:
lgtm after the fact

PR #2089, opened 2016-10-11
Test transformFeedbackVarying() behaviors.

zhenyao:


zhenyao:
@kenrussell @Kangz @kainino0x Can one of you review?

sdk/tests/conformance2/transform_feedback/transform_feedback.html
```diff
@@ -315,6 +315,22 @@
     runVaryingsTest();
 }
 
+function verifyTransformFeedbackVarying(progrom, index, valid, name) {
```
kainino0x: > progrom

I'm sure this works because program is a global variable - maybe rename it so the local doesn't shadow the global

kenrussell:
LGTM with @kainino0x 's review feedback addressed.

sdk/tests/conformance2/transform_feedback/transform_feedback.html
zhenyao: Done

kenrussell:
LGTM. Merging.

PR #2090, opened 2016-10-12
add WebGL 1 test for uninitialized buffers

kainino0x:


kainino0x:
@zhenyao PTAL

zhenyao:
The test looks great!  Can you get a follow up patch to test bufferData(null) and bufferSubData() only fills a subset, and verifies the range bufferSubData doesn't touch is initialized to zero.

zhenyao:
Ah, just realized you forgot to update the 00_test_list.txt file

kainino0x:
> follow up patch

Sure, I'll do that.

> 00_test_list.txt

Ugh, I keep forgetting to do that. Thanks for fixing it.

PR #2091, opened 2016-10-12
Add missing assertOK() in texImage2D.html

Jiawei-Shao:
Function assertOK() should be called in testTexImage2DNull, or the result
of the test will not be displayed on the webpage.


Jiawei-Shao:
@qkmiao Please help to view. Thanks!

zhenyao:
lgtm

PR #2094, opened 2016-10-12
Add test for active 3d texture bug on Mac OSX

qkmiao:
This will catch crash on Mac OSX 10.12 with Intel HD Graphics 515
graphics card.

### DIFF
--- a/sdk/tests/conformance2/textures/misc/active-3d-texture-bug.html
+++ b/sdk/tests/conformance2/textures/misc/active-3d-texture-bug.html
@@ -1,0 +1,140 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL Active TEXTURE1 Bug Tests</title>
+<link rel="stylesheet" href="../../../resources/js-test-style.css"/>
+<script src="../../../js/js-test-pre.js"></script>
+<script src="../../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/textures/misc/active-3d-texture-bug.html
```diff
@@ -0,0 +1,140 @@
...
+in vec2 v_coord;
+uniform mediump sampler3D u_sampler;
+out vec4 o_color;
+void main () {
+    o_color = texture(u_sampler, vec3(v_coord, 0.0));
+}
+</script>
+<script>
+"use strict";
+description("Test for a Mac OSX driver bug activing TEXTURE1 for 3d texture");
```
zhenyao: activating

sdk/tests/conformance2/textures/misc/active-3d-texture-bug.html
```diff
@@ -0,0 +1,140 @@
...
+
+    // Render using sampler
+    render(0, width, height, black, "Render from TEXTURE0 should be black");
+
+    gl.bindSampler(0, null);
+    gl.deleteSampler(sampler);
+
+    // Render using texture
+    render(0, width, height, black, "Render from TEXTURE0 should be black");
+    render(1, width, height, green, "Render from TEXTURE1 should be green");
```
zhenyao: Maybe I misunderstood something, but why sometimes it's black and other times it's green? Whether TEXTURE0 or TEXTURE1, it's the same texture, regardless of sampling method.

sdk/tests/conformance2/textures/misc/active-3d-texture-bug.html
```diff
@@ -0,0 +1,140 @@
...
+
+    gl.bindTexture(gl.TEXTURE_3D, null);
+    gl.deleteTexture(texture);
+    gl.deleteProgram(program);
+}
```
zhenyao: Can you add comment either in the beginning of the code or where it crashed the detailed nature of this driver bug? Is it absolutely necessary for three passes of rendering to catch this crash?

qkmiao: I added comments where driver crashed. The first two render passes are required. The last render pass isn't required to catch the crash. I just use the last render pass to show the correct drawing from TEXTURE1.

zhenyao: So basically TEXTURE1 isn't necessary to reproduce the crash? Is the 3d texture also necessary?

It seems to me the relevant path to crash is:

bindSampler
render with default 3D texture + sample
deleteSampler
render with default 3D texture without sample (crash)

If that's the case, let's minimize it so it's easier for driver vendors to debug.

qkmiao: Active TEXTURE1 and 3d texture are necessary to reproduce the crash.
I removed the third render pass to minimize the code.
Thanks for review. PTAL.

sdk/tests/conformance2/textures/misc/active-3d-texture-bug.html
```diff
@@ -0,0 +1,146 @@
...
+in vec2 v_coord;
+uniform mediump sampler3D u_sampler;
+out vec4 o_color;
+void main () {
+    o_color = texture(u_sampler, vec3(v_coord, 0.0));
+}
+</script>
+<script>
+"use strict";
+description("Test for a Mac OSX driver crash bug activating TEXTURE1 for 3d texture");
```
zhenyao: Sorry I didn't include this in my previous review, but can you add a comment to say this bug is on MacOSX 10.12 with Intel GPUs?

zhenyao:
Thanks for minimizing this.  LGTM with one suggestion to add a comment.

sdk/tests/conformance2/textures/misc/active-3d-texture-bug.html
```diff
@@ -0,0 +1,146 @@
...
+    wtu.checkCanvasRect(gl, 0, 0, width, height, expectedColor, msg);
+}
+
+function activeTextureTest() {
+    var texture = gl.createTexture();
+    var sampler = gl.createSampler();
+    var width = 64;
+    var height = 64;
+    var depth = 4;
+    var green = [0, 255, 0, 255];
```
zhenyao: Also, could you remove green? It's not used now.

sdk/tests/conformance2/textures/misc/active-3d-texture-bug.html
qkmiao: Done

zhenyao:
Thank you!

PR #2095, opened 2016-10-12
buffer-uninitialized: repeat 50 times, and test bufferSubData

kainino0x:
@zhenyao does this cover the case you mentioned?

I also repeated the test 50 times, because when I came in this morning the test was flaky on my machine.

### DIFF
--- a/sdk/tests/conformance/buffers/buffer-uninitialized.html
+++ b/sdk/tests/conformance/buffers/buffer-uninitialized.html
@@ -62,7 +62,7 @@ shouldBeTrue("program != null");
 
 var TEST_LENGTH = 1024;
 var TEST_BUFSIZE = TEST_LENGTH * 4;
-var data = new Float64Array(TEST_LENGTH / 4);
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/buffers/buffer-uninitialized.html
```diff
@@ -62,31 +62,57 @@
...
+debug("");
+debug("testing bufferData(..., size, ...)");
+for (j = 0; j < REPETITIONS; j++) {
+  test(function() {
+    gl.bufferData(gl.ARRAY_BUFFER, TEST_BUFSIZE, gl.STATIC_DRAW);
+  });
+}
+
+debug("");
+debug("testing bufferSubData(..., size, ...) of uninitialized buffer");
```
zhenyao: nit: bufferSubData doesn't really take a size arg

kainino0x: Oops, fixed

zhenyao:
great. lgtm

PR #2096, opened 2016-10-12
Rename getBufferSubData.html to get-buffer-sub-data.html

zhenyao:
to be consistent with other tests under conformance2


zhenyao:
Simple change, merging myself

PR #2097, opened 2016-10-12
add test for getBufferSubData when transform feedback is active

kainino0x:
This error case from the spec was previously untested.

@zhenyao yet another PR for you to review :)

### DIFF
--- a/sdk/tests/conformance2/buffers/get-buffer-sub-data.html
+++ b/sdk/tests/conformance2/buffers/get-buffer-sub-data.html
@@ -168,20 +168,6 @@ wtu.shouldGenerateGLError(gl, gl.NO_ERROR,
 gl.deleteBuffer(buffer);
 shouldBeTrue("areArraysEqual(retArray, floatArray)");
 
-debug("");
-debug("Test that getBufferSubData...");
-var transformFeedbackBuffer = gl.createBuffer();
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/buffers/get-buffer-sub-data.html
```diff
@@ -168,6 +168,20 @@
...
 
+debug("");
+debug("Test that getBufferSubData...");
+var transformFeedbackBuffer = gl.createBuffer();
+gl.bindBuffer(gl.TRANSFORM_FEEDBACK_BUFFER, transformFeedbackBuffer);
+gl.bufferData(gl.TRANSFORM_FEEDBACK_BUFFER, floatArray, gl.STATIC_DRAW);
+var transformFeedback = gl.createTransformFeedback();
+debug("... passes when a transform feedback object is not bound");
+wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.getBufferSubData(gl.TRANSFORM_FEEDBACK_BUFFER, 0, retArray, retArray.length)");
+gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, transformFeedback);
```
zhenyao: It should be beginTransformFeedback() that causes dual usage of the buffer with getBufferSubData. Per our offline discussion, it's easier to add this to the transform_feedback.html test.

zhenyao:
lgtm

PR #2098, opened 2016-10-13
blitting pixels lying exactly on the effective border

Richard-Yunchao:
As we have discussed several days ago, any pixels lying outside the read framebuffer, the corresponding pixels in dst region should be untouched during blitting. But there are still some corner cases. I want to make it clear for these two issues below when blitting pixels out-of-bounds. I read the discussions at https://cvs.khronos.org/bugzilla/show_bug.cgi?id=3576, but there is no answer either. 

The blitFramebuffer API (https://www.khronos.org/opengles/sdk/docs/man3/html/glBlitFramebuffer.xhtml):  
blitFramebuffer(srcX0, srcY0, srcX1, srcY1, dstX0, dstY0, dstX1, dstY1, mask, filter).

**Issue 1: Which is the lower bounds of src/dst region during blitframebuffer? The small value between srcX0/srcX1, srcY0/srcY1, dstX0/dstX1, dstY0/dstY1, or defined by srcX0, srcY0, dstX0, and dstY0?** 

I would like to choose the small value, here are some reasons: 
a. From the gles3 man page (https://www.khronos.org/opengles/sdk/docs/man3/html/glBlitFramebuffer.xhtml), it says that "The lower bounds of the rectangle are inclusive, while the upper bounds are exclusive". I think "the lower bounds of the rectangle" should be the small value. 
b. In our common sense, if both src region and dst region are reversed, it should be identical to no reversion at all. But if we define the lower bounds to srcX0, srcY0, dstX0, dstY0, then the assumption will be incorrect. For example, blitting [0, 0, 2, 2] in read buffer to [0, 0, 2, 2] in draw buffers vs blitting from [2, 2, 0, 0] to [2, 2, 0, 0]. The blitted pixels are different if we define lower bounds by srcX0, srcY0, dstX0, dstY0: [0, 0] and [2, 2] are included/excluded in different way. Moreover, [2, 2] would be out-of-bounds if the read/draw buffers' width/height are 2.
c. The current implementations follow this way. I test this pr, It can pass on Linux/Windows directly.
 
**Issue 2: How to handle the pixels lying exactly on the effective border for LINEAR/NEAREST?** For example, when blitting from [-2, -2, 4, 4] to [0, 0, 8, 8] or [0, 0, 7, 7], assuming that both the read/draw buffer are bigger than [0, 0, 8, 8], the effective src region would be [0, 0, 4, 4]. When mapping the effective src region to dst region, it would be [2.666667, 2.666667, 8, 8] and [2.333333, 2.333333, 7, 7] respectively. So, should we blit the dst pixels whose x/y is 2 for these two situations? 
I would like to not blit pixels whose x/y=2 for the former, but blit the pixels whose x/y=2 for the latter. Some reasons: 
a. From the gles3 spec on page 197 (https://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-4.3.3), it says that "Only pixels whose
centers lie within the destination rectangle are written by BlitFramebuffer" for NEAREST filter. I think we should do this for LINEAR filter too. 
b. The current implementations follow this way for both NEAREST/LINEAR filter on Linux/Windows.  You can try to run this pr to see whether it can pass on your Linux/Windows.

Anyway, we can choose the different one for these two corner cases, but we need to define them clearly. 
 
PTAL. Thanks a lot!


Richard-Yunchao:
Could you take a look? @kenrussell , @zhenyao , @Kangz , @jdashg?

sdk/tests/conformance2/rendering/blitframebuffer-filter-outofbounds.html
```diff
@@ -111,6 +115,10 @@
             [4, 4, 8, 8],
             [8, 8, 8, 8],
             [6, 6, 8, 8],
+            [5, 5, 8, 8],
```
Kangz: I believe this should start at 4 instead of 5, as (4, 4) will sample from (0.25, 0.25) which is valid.

zhenyao:
To clarify, I actually think the spec is clear for both issues now that I re-read the spec.

1) Issue 1, the spec says the pixel (x, y) 's center is (x+0.5, y+0.5).  And it says all pixels whose center are between the two integer bounds (low, high) are included.  So for pixel low, it's center is low+0.5, which is between (low, high), for pixel high, it's center is high + 0.5, so it's outside (low, high).  So basically the smaller bound pixel is included, but the larger bound pixel is excluded.  No ambiguity here.

2) Issue 2, on ES spec 3.0.4, page 197, "Only pixels whose centers lie within the destination rectangle are written by BlitFramebuffer. Linear filter sampling (see below) may result in pixels outside the source rectangle being read." As you can see, the first sentence doesn't limit it to Linear filter.  It applies to all filters.  The second sentence is talking about the source rectangle, and it has nothing to do with destination pixels.

Richard-Yunchao:
OK, @zhenyao , Then this pr follow the spec.

sdk/tests/conformance2/rendering/blitframebuffer-filter-outofbounds.html
Richard-Yunchao: And I carefully designed another test, it is the next one in this pr: blitting from [-2, -2, 4, 4] to [3, 3, 10, 10]. 
The scaleFactor = 7/6. 
The untouched src region width/height = 2. The untouched dst region would be 2 * 7/6 = 7/3. 
So the effective dst region [3+7/3, 10] = [5.333333, 10]. Then the pixel (5, 5)'s center (5.5, 5.5) lies inside the effective dst region, so the blitted area (intersects with draw buffer) would be [5, 5, 8, 8].

sdk/tests/conformance2/rendering/blitframebuffer-filter-outofbounds.html
```diff
@@ -95,13 +95,17 @@
...
+            [0, 0, 6, 6, 7, 7, 10, 10], // only dst region is out-of-bounds, dst region has different width/height as src region after dst region is clipped to the bounds of draw buffer.
+            [0, 0, 6, 6, 4, 4, 10, 10], // only dst region is out-of-bounds, dst region has the same width/height as src region after dst region is clipped to the bounds of draw buffer.
+            [-2, -2, 4, 4, 7, 7, 10, 10], // both src and dst region are out-of-bounds, dst region has different width/height as src region after dst region is clipped to the bounds of draw buffer.
+            [-2, -2, 4, 4, 4, 4, 10, 10], // both src and dst region are out-of-bounds, dst region has the same width/height as src region after dst region is clipped to the bounds of draw buffer.
+            [-2, -2, 4, 4, 2, 2, 10, 10], // both src and dst region are out-of-bounds. There are some dst pixels (x and y are within [4, 8] , and x or y equals to 4) whose corresponding src pixels are partially inside the effective border and partially outside the effective border. But the centers of such src pixels are lying outside the effective border (the effective border is [0, 0, 4, 4]).
```
Richard-Yunchao: That's true. Thanks for your suggestion. I update the description to "real sampling area", which is from the perspective of sampling as you suggested. @Kangz

Richard-Yunchao:
@zhenyao , Now that the spec is clear, so we don't need to update the WebGL spec. We have discussed about the issue 2 I listed in this pr at #2070.

zhenyao:
LGTM, but @Kangz should take another look.

sdk/tests/conformance2/rendering/blitframebuffer-filter-outofbounds.html
Kangz: Thanks we have the same intrepretation, I just messed my computation :)

zhenyao:
OK, let's merge then.

PR #2099, opened 2016-10-15
Non nullable

jdashg:
This does not include test updates yet.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -1608,8 +1608,8 @@ interface <dfn id="WebGLRenderingContextBase">WebGLRenderingContextBase</dfn>
     object? getExtension(DOMString name);
 
     void activeTexture(GLenum texture);
-    void attachShader(WebGLProgram? program, WebGLShader? shader);
-    void bindAttribLocation(WebGLProgram? program, GLuint index, DOMString name);
+    void attachShader(WebGLProgram program, WebGLShader shader);
+    void bindAttribLocation(WebGLProgram program, GLuint index, DOMString name);
     void bindBuffer(GLenum target, WebGLBuffer? buffer);
     void bindFramebuffer(GLenum target, WebGLFramebuffer? framebuffer);
     void bindRenderbuffer(GLenum target, WebGLRenderbuffer? renderbuffer);
@@ -1622,8 +1622,8 @@ interface <dfn id="WebGLRenderingContextBase">WebGLRenderingContextBase</dfn>
                            GLenum srcAlpha, GLenum dstAlpha);
 
     void bufferData(GLenum target, GLsizeiptr size, GLenum usage);
-    void bufferData(GLenum target, BufferDataSource? data, GLenum usage);
-    void bufferSubData(GLenum target, GLintptr offset, BufferDataSource? data);
+    void bufferData(GLenum target, BufferDataSource data, GLenum usage);
+    void bufferSubData(GLenum target, GLintptr offset, BufferDataSource data);
 
     [WebGLHandlesContextLoss] GLenum checkFramebufferStatus(GLenum target);
     void clear(GLbitfield mask);
@@ -1631,7 +1631,7 @@ interface <dfn id="WebGLRenderingContextBase">WebGLRenderingContextBase</dfn>
     void clearDepth(GLclampf depth);
     void clearStencil(GLint s);
     void colorMask(GLboolean red, GLboolean green, GLboolean blue, GLboolean alpha);
-    void compileShader(WebGLShader? shader);
+    void compileShader(WebGLShader shader);
 
     void compressedTexImage2D(GLenum target, GLint level, GLenum internalformat,
                               GLsizei width, GLsizei height, GLint border,
@@ -1666,7 +1666,7 @@ interface <dfn id="WebGLRenderingContextBase">WebGLRenderingContextBase</dfn>
     void depthFunc(GLenum func);
     void depthMask(GLboolean flag);
     void depthRange(GLclampf zNear, GLclampf zFar);
-    void detachShader(WebGLProgram? program, WebGLShader? shader);
+    void detachShader(WebGLProgram program, WebGLShader shader);
     void disable(GLenum cap);
     void disableVertexAttribArray(GLuint index);
     void drawArrays(GLenum mode, GLint first, GLsizei count);
@@ -1685,11 +1685,11 @@ interface <dfn id="WebGLRenderingContextBase">WebGLRenderingContextBase</dfn>
 
     void generateMipmap(GLenum target);
 
-    WebGLActiveInfo? getActiveAttrib(WebGLProgram? program, GLuint index);
-    WebGLActiveInfo? getActiveUniform(WebGLProgram? program, GLuint index);
-    sequence&lt;WebGLShader&gt;? getAttachedShaders(WebGLProgram? program);
+    WebGLActiveInfo? getActiveAttrib(WebGLProgram program, GLuint index);
+    WebGLActiveInfo? getActiveUniform(WebGLProgram program, GLuint index);
+    sequence&lt;WebGLShader&gt;? getAttachedShaders(WebGLProgram program);
 
-    [WebGLHandlesContextLoss] GLint getAttribLocation(WebGLProgram? program, DOMString name);
+    [WebGLHandlesContextLoss] GLint getAttribLocation(WebGLProgram program, DOMString name);
 
     any getBufferParameter(GLenum target, GLenum pname);
     any getParameter(GLenum pname);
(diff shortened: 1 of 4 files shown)


zhenyao:
Looks fine in general.
One question for you to consider: do we wanna still allow delete*() with nullable object? Because in GL, delete(0) is allowed and it is a no op. Now we will get an exception, so it seems like an unnecessary behavior deviation.

specs/latest/1.0/index.html
```diff
@@ -1751,57 +1751,57 @@ <h2 class="no-toc">Table of contents</h2>
...
-    void uniform4fv(WebGLUniformLocation? location, Float32Array v);
-    void uniform4fv(WebGLUniformLocation? location, sequence&lt;GLfloat&gt; v);
-    void uniform4i(WebGLUniformLocation? location, GLint x, GLint y, GLint z, GLint w);
-    void uniform4iv(WebGLUniformLocation? location, Int32Array v);
-    void uniform4iv(WebGLUniformLocation? location, sequence&lt;long&gt; v);
-
-    void uniformMatrix2fv(WebGLUniformLocation? location, GLboolean transpose,
+                       GLenum format, GLenum type, TexImageSource source); // May throw DOMException
+
+    void uniform1f(WebGLUniformLocation location, GLfloat x);
```
kenrussell: These entry points must continue to accept a nullable WebGLUniformLocation. We discovered in earlier iterations of the WebGL spec that OpenGL ES applications rely on the ability to query a uniform's location, get back -1 (the invalid value -- mapped to null in WebGL), and pass that location back into the API, assuming that it will be ignored. A *lot* of content will be broken if this is changed.

specs/latest/1.0/webgl.idl
```diff
@@ -602,47 +602,47 @@ interface WebGLRenderingContextBase
...
-    DOMString? getShaderInfoLog(WebGLShader? shader);
+    DOMString? getShaderInfoLog(WebGLShader shader);
 
-    DOMString? getShaderSource(WebGLShader? shader);
+    DOMString? getShaderSource(WebGLShader shader);
 
     any getTexParameter(GLenum target, GLenum pname);
 
-    any getUniform(WebGLProgram? program, WebGLUniformLocation? location);
+    any getUniform(WebGLProgram program, WebGLUniformLocation location);
```
kenrussell: It should be OK to make this WebGLUniformLocation argument non-nullable. We should check the conformance tests to make sure the case of passing in a null WebGLUniformLocation is tested.

specs/latest/2.0/index.html
```diff
@@ -807,60 +807,60 @@ <h2 class="no-toc">Table of contents</h2>
...
   /* Programs and shaders */
-  [WebGLHandlesContextLoss] GLint getFragDataLocation(WebGLProgram? program, DOMString name);
+  [WebGLHandlesContextLoss] GLint getFragDataLocation(WebGLProgram program, DOMString name);
 
   /* Uniforms */
-  void uniform1ui(WebGLUniformLocation? location, GLuint v0);
-  void uniform2ui(WebGLUniformLocation? location, GLuint v0, GLuint v1);
-  void uniform3ui(WebGLUniformLocation? location, GLuint v0, GLuint v1, GLuint v2);
-  void uniform4ui(WebGLUniformLocation? location, GLuint v0, GLuint v1, GLuint v2, GLuint v3);
+  void uniform1ui(WebGLUniformLocation location, GLuint v0);
```
kenrussell: Same here; these APIs must allow the WebGLUniformLocation to be nullable, and continue to be a no-op if it's null.

specs/latest/2.0/webgl2.idl
```diff
@@ -423,60 +423,60 @@ interface WebGL2RenderingContextBase
...
   /* Programs and shaders */
-  [WebGLHandlesContextLoss] GLint getFragDataLocation(WebGLProgram? program, DOMString name);
+  [WebGLHandlesContextLoss] GLint getFragDataLocation(WebGLProgram program, DOMString name);
 
   /* Uniforms */
-  void uniform1ui(WebGLUniformLocation? location, GLuint v0);
-  void uniform2ui(WebGLUniformLocation? location, GLuint v0, GLuint v1);
-  void uniform3ui(WebGLUniformLocation? location, GLuint v0, GLuint v1, GLuint v2);
-  void uniform4ui(WebGLUniformLocation? location, GLuint v0, GLuint v1, GLuint v2, GLuint v3);
+  void uniform1ui(WebGLUniformLocation location, GLuint v0);
```
kenrussell: Same here.

zhenyao:
Good catch Ken.  In general, we should only change nullable to non-nullable if passing nullable generates a GL error - changing GL error to an exception seems OK.

jdashg:
"only change nullable to non-nullable if passing nullable generates a GL error" is good rationale. Will update.

specs/latest/1.0/webgl.idl
jdashg: GetUniform generates INVALID_OP if program is invalid or location is null, so I'll keep this non-nullable.

jdashg:
@kenrussell @zhenyao

specs/latest/1.0/index.html
```diff
@@ -1698,47 +1698,47 @@ <h2 class="no-toc">Table of contents</h2>
...
-    [WebGLHandlesContextLoss] GLboolean isFramebuffer(WebGLFramebuffer? framebuffer);
-    [WebGLHandlesContextLoss] GLboolean isProgram(WebGLProgram? program);
-    [WebGLHandlesContextLoss] GLboolean isRenderbuffer(WebGLRenderbuffer? renderbuffer);
-    [WebGLHandlesContextLoss] GLboolean isShader(WebGLShader? shader);
-    [WebGLHandlesContextLoss] GLboolean isTexture(WebGLTexture? texture);
+    [WebGLHandlesContextLoss] GLboolean isFramebuffer(WebGLFramebuffer framebuffer);
+    [WebGLHandlesContextLoss] GLboolean isProgram(WebGLProgram program);
+    [WebGLHandlesContextLoss] GLboolean isRenderbuffer(WebGLRenderbuffer renderbuffer);
+    [WebGLHandlesContextLoss] GLboolean isShader(WebGLShader shader);
+    [WebGLHandlesContextLoss] GLboolean isTexture(WebGLTexture texture);
```
kenrussell: These queries didn't generate an error if null was passed (at least in Chrome); they just returned false. By that criterion, should they remain nullable?

specs/latest/2.0/index.html
```diff
@@ -901,56 +901,56 @@ <h2 class="no-toc">Table of contents</h2>
...
-  void samplerParameterf(WebGLSampler? sampler, GLenum pname, GLfloat param);
-  any getSamplerParameter(WebGLSampler? sampler, GLenum pname);
+  void samplerParameteri(WebGLSampler sampler, GLenum pname, GLint param);
+  void samplerParameterf(WebGLSampler sampler, GLenum pname, GLfloat param);
+  any getSamplerParameter(WebGLSampler sampler, GLenum pname);
 
   /* Sync objects */
   WebGLSync? fenceSync(GLenum condition, GLbitfield flags);
-  [WebGLHandlesContextLoss] GLboolean isSync(WebGLSync? sync);
+  [WebGLHandlesContextLoss] GLboolean isSync(WebGLSync sync);
```
kenrussell: Same question here about the query function.

specs/latest/2.0/index.html
```diff
@@ -901,56 +901,56 @@ <h2 class="no-toc">Table of contents</h2>
...
+  void waitSync(WebGLSync sync, GLbitfield flags, GLint64 timeout);
+  any getSyncParameter(WebGLSync sync, GLenum pname);
 
   /* Transform Feedback */
   WebGLTransformFeedback? createTransformFeedback();
-  void deleteTransformFeedback(WebGLTransformFeedback?);
-  [WebGLHandlesContextLoss] GLboolean isTransformFeedback(WebGLTransformFeedback?);
-  void bindTransformFeedback (GLenum target, WebGLTransformFeedback? id);
+  void deleteTransformFeedback(WebGLTransformFeedback? tf);
+  [WebGLHandlesContextLoss] GLboolean isTransformFeedback(WebGLTransformFeedback tf);
```
kenrussell: Same question here.

specs/latest/2.0/index.html
```diff
@@ -901,56 +901,56 @@ <h2 class="no-toc">Table of contents</h2>
...
   void endQuery(GLenum target);
   WebGLQuery? getQuery(GLenum target, GLenum pname);
-  any getQueryParameter(WebGLQuery? query, GLenum pname);
+  any getQueryParameter(WebGLQuery query, GLenum pname);
 
   /* Sampler Objects */
   WebGLSampler? createSampler();
   void deleteSampler(WebGLSampler? sampler);
-  [WebGLHandlesContextLoss] GLboolean isSampler(WebGLSampler? sampler);
+  [WebGLHandlesContextLoss] GLboolean isSampler(WebGLSampler sampler);
```
kenrussell: Same question here.

specs/latest/2.0/index.html
```diff
@@ -901,56 +901,56 @@ <h2 class="no-toc">Table of contents</h2>
   /* Query Objects */
   WebGLQuery? createQuery();
   void deleteQuery(WebGLQuery? query);
-  [WebGLHandlesContextLoss] GLboolean isQuery(WebGLQuery? query);
-  void beginQuery(GLenum target, WebGLQuery? query);
+  [WebGLHandlesContextLoss] GLboolean isQuery(WebGLQuery query);
```
kenrussell: Same here.

jdashg:
@kenrussell Fixed. Also useProgram should not have been made non-nullable.

kenrussell:
Thanks, this looks good to me.

The only question in my mind is whether this should be merged now (after updating the associated tests) or after taking the WebGL 2.0 snapshot. Were you planning to update Firefox's implementation now?

jdashg:
It would be pretty rough if the snapshot didn't include this, though we could potentially update the snapshot.
I already de-nullabled the texture entrypoints for Firefox, and the others will follow shortly.

kenrussell:
OK, in that case I can commit to updating Chromium's implementation to match. Can you include any test updates in this pull request?

specs/latest/2.0/index.html
```diff
@@ -688,8 +688,8 @@ <h2 class="no-toc">Table of contents</h2>
   /* Buffer objects */
   // WebGL1:
   void bufferData(GLenum target, GLsizeiptr size, GLenum usage);
-  void bufferData(GLenum target, BufferDataSource? srcData, GLenum usage);
```
zhenyao: @jdashg This actually doesn't fit with WebIDL. 

Looking at https://heycam.github.io/webidl/#idl-overloading the matching table, numeric types / interface is no longer distinguishable because interface is no longer nullable.

zhenyao: @kenrussell @RafaelCintron @grorg @kainino0x FYI

zhenyao:
Let's merge this now and I'll upload a pull request to fix the bufferData issue. I didn't find any other issues while implementing this new behavior in Chrome. Tests update will also be uploaded soon after.

PR #2100, opened 2016-10-17
add tests for blitframebuffer when width or height of src/dst region overflow

Richard-Yunchao:
Add some test cases per the suggestion at https://codereview.chromium.org/2409523002/#msg43 by Zhenyao. 

PTAL. Thanks a lot! @zhenyao , @kenrussell .

### DIFF
--- a/sdk/tests/conformance2/rendering/00_test_list.txt
+++ b/sdk/tests/conformance2/rendering/00_test_list.txt
@@ -1,6 +1,7 @@
 attrib-type-match.html
 blitframebuffer-filter-outofbounds.html
 blitframebuffer-filter-srgb.html
+blitframebuffer-misc.html
 blitframebuffer-multisampled-readbuffer.html
 --min-version 2.0.1 blitframebuffer-outside-readbuffer.html
 blitframebuffer-scissor-enabled.html
--- a/sdk/tests/conformance2/rendering/00_test_list.txt
+++ b/sdk/tests/conformance2/rendering/00_test_list.txt
@@ -1,10 +1,10 @@
 attrib-type-match.html
 blitframebuffer-filter-outofbounds.html
 blitframebuffer-filter-srgb.html
-blitframebuffer-misc.html
 blitframebuffer-multisampled-readbuffer.html
 --min-version 2.0.1 blitframebuffer-outside-readbuffer.html
 blitframebuffer-scissor-enabled.html
+blitframebuffer-size-overflow.html
 --min-version 2.0.1 blitframebuffer-srgb-and-linear-drawbuffers.html
 --min-version 2.0.1 blitframebuffer-stencil-only.html
 blitframebuffer-test.html
--- a/sdk/tests/conformance2/rendering/blitframebuffer-misc.html
+++ b/sdk/tests/conformance2/rendering/blitframebuffer-misc.html
@@ -1,0 +1,104 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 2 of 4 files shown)


zhenyao:
Actually the working group consensus is not to enforce a GL error in such overflow cases.  Just call blitFramebuffer() and if the test is able to finish without crash, then it's OK.  I suggest to test that in a separate html (so if it crashes, it won't affect other test cases in the same html)

Richard-Yunchao:
That's true, generate INVALID_OPERATION is not a must. On many devices, the underlying drivers can handle this situation without any error. It can be NO_ERROR too. 
@zhenyao , move the test to a standalone file as you suggested too. Please take another look.

sdk/tests/conformance2/rendering/00_test_list.txt
```diff
@@ -1,6 +1,7 @@
 attrib-type-match.html
 blitframebuffer-filter-outofbounds.html
 blitframebuffer-filter-srgb.html
+blitframebuffer-misc.html
```
zhenyao: maybe blitframebuffer-size-overflow.html

sdk/tests/conformance2/rendering/blitframebuffer-misc.html
```diff
@@ -0,0 +1,104 @@
...
+</head>
+<body>
+<canvas id="example" width="8" height="8"></canvas>
+<div id="description"></div>
+<div id="console"></div>
+<script>
+"use strict";
+
+var wtu = WebGLTestUtils;
+description("This test verifies the functionality of blitFramebuffer for some corner cases.");
```
zhenyao: Maybe "This test verifies blitFramebuffer won't cause a crash when the computed sizes might overflow."

sdk/tests/conformance2/rendering/blitframebuffer-misc.html
```diff
@@ -0,0 +1,104 @@
...
+    gl.framebufferTexture2D(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, tex1, 0);
+    if ((gl.checkFramebufferStatus(gl.READ_FRAMEBUFFER) != gl.FRAMEBUFFER_COMPLETE) ||
+        (gl.checkFramebufferStatus(gl.DRAW_FRAMEBUFFER) != gl.FRAMEBUFFER_COMPLETE)) {
+        testFailed("Framebuffer incomplete.");
+        return;
+    }
+
+    var max = 0x7fffffff;
+    gl.blitFramebuffer(0, 0, max, max, 0, 0, width, height, gl.COLOR_BUFFER_BIT, gl.NEAREST);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "BlitFramebuffer should succeed.");
```
zhenyao: I feel these are extreme cases and there is no need for us to make sure it works.  Let's just not check GL errors.

zhenyao: Same below.

sdk/tests/conformance2/rendering/blitframebuffer-misc.html
```diff
@@ -0,0 +1,104 @@
...
+    gl.blitFramebuffer(0, 0, width, height, 0, 0, max, max, gl.COLOR_BUFFER_BIT, gl.NEAREST);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "BlitFramebuffer should succeed.");
+    gl.blitFramebuffer(0, 0, max, max, 0, 0, max, max, gl.COLOR_BUFFER_BIT, gl.NEAREST);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "BlitFramebuffer should succeed.");
+
+    gl.blitFramebuffer(-1, -1, max, max, 0, 0, width, height, gl.COLOR_BUFFER_BIT, gl.NEAREST);
+    wtu.glErrorShouldBe(gl, [gl.NO_ERROR, gl.INVALID_OPERATION], "The width and height of src region may overflow.");
```
zhenyao: Same here, I think if an implementation wants to generate INVALID_VALUE or OUT_OF_MEMORY, that's also fine.  Let's just not check GL errors here.

kenrussell:
Agree with Mo's observations; I defer review to him.

sdk/tests/conformance2/rendering/blitframebuffer-misc.html
Richard-Yunchao: Thanks for your review, Zhenyao. So, do you have some suggestions to do the check? I mean, what behavior is correct, what is false? @zhenyao.

sdk/tests/conformance2/rendering/00_test_list.txt
Richard-Yunchao: Done

sdk/tests/conformance2/rendering/blitframebuffer-misc.html
Richard-Yunchao: Make sense. I updated the code accordingly. @zhenyao. could you take another look?

zhenyao:
Thanks. LGTM.

Richard-Yunchao:
Thanks for merging. Zhenyao.

PR #2101, opened 2016-10-18
Add test for samplers in a ternary operator.

c0d1f1ed:


kenrussell:
Thanks for this test. LGTM. Out of curiosity, what bug motivated the addition of this test?

c0d1f1ed:
Some Android games have code like this, which causes issues in SwiftShader and ANGLE's shader compiler.

zhenyao:
Is there an ANGLE bug for this? We need to resolve this quickly or we will have to move this out of the current version of conformance for the upcoming WebGL2 release.

Oletus:
The shader used in the test is incorrect. ESSL 1.00 section 4.1.7:

"Except for parameters to texture lookup functions, array indexing, structure field selection, and parentheses, samplers are not allowed to be operands in expressions."

PR #2103, opened 2016-10-18
Allow only std140 layout in tests and make it the default

Oletus:
The spec is updated to mention that std140 layout is the initial
default for uniform blocks.

The tests are updated to expect compilation to fail when uniform block
layout is shared or packed. Some deqp tests that covered shared and
packed uniform buffers are removed from the suite. One linkage test
used to assume that the default layout would be shared, that one is
also removed.


zhenyao:
LGTM
Thanks Olli to putting this together quickly

Oletus:
Thanks for merging!

kenrussell:
Fantastic. Thank you Olli for picking this up.

PR #2104, opened 2016-10-19
Disallow unsized float and half-float texture formats in WebGL 2.0

qkmiao:


sdk/tests/conformance/textures/misc/tex-image-with-bad-args.html
```diff
@@ -0,0 +1,101 @@
...
+        extensionTests.forEach(doTexImage);
+    }
+} else {
+    var webgl2Tests = [
+        { internalformat: gl.RGBA, format: gl.RGBA, type: gl.FLOAT, errors: [gl.INVALID_OPERATION] },
+        { internalformat: gl.RGBA, format: gl.RGBA, type: gl.HALF_FLOAT, errors: [gl.INVALID_OPERATION] },
+    ];
+    webgl2Tests.concat(tests);
+    webgl2Tests.forEach(doTexImage);
+}
```
zhenyao: WebGL1 test coverage is already in the extension tests.

Why don't we only keep the WebGL2 test cases and move to conformance2/textures.misc?

Also, please test LUMINANCE as well because someone just reported using float luminance textures successfully in WebGL2 (which should be disallowed)

kenrussell: Agree with Mo's assessment; the WebGL 1.0 negative tests are already in conformance/extensions/ under oes-texture-float.html and oes-texture-half-float.html . Let's just add the new WebGL 2.0 tests under conformance2/textures/misc/ .

qkmiao: Previously, I put tests for WebGL 1.0 and 2.0 together to show the difference of the usage. It's OK to only keep negative tests for WebGL 2.0.
LUMINANCE was added to test cases.

kenrussell:
Thanks @qkmiao for revising this test. It looks perfect now.

Let's let @zhenyao re-review and merge this.

sdk/tests/conformance2/textures/misc/tex-image-with-bad-args.html
```diff
@@ -0,0 +1,77 @@
...
+<div id="description"></div>
+<div id="console"></div>
+<script>
+"use strict";
+description('Tests texImage2D with invalid internalformat/format/type combinations');
+debug("This is a regression test for <a href='https://bugs.chromium.org/p/chromium/issues/detail?id=656889'>Chromium Issue 656889</a>");
+
+var wtu = WebGLTestUtils;
+var contextVersion = wtu.getDefault3DContextVersion();
+var gl = wtu.create3DContext("c");
```
zhenyao: Let's just put 2 in the version here so the test can run without the extra url args

zhenyao:
LGTM with a minor suggestion on context creation version.

kenrussell:
LGTM

PR #2106, opened 2016-10-21
Fix a bug in transform_feedback.html

zhenyao:
1) getBufferSubData params are wrong
2) The last two test are async, so we need to run all the sync tests first


kenrussell:
Great job debugging this. LGTM

PR #2107, opened 2016-10-22
Fix typo which caused incorrect html rendering

kainino0x:


zhenyao:
lgtm
thanks for the fix

PR #2110, opened 2016-10-25
Change clientWaitSync to take GLuint64 for |timeout| (as vs GLint64).

zhenyao:
Also, remove texts that indicates special handling of TIMEOUT_IGNORED,
or -1 in WebGL2, which should not be for clientWaitSync.


zhenyao:
@kenrussell @jdashg Please take a look

kenrussell:
LGTM. Thanks for noticing this and clearing it up.

PR #2111, opened 2016-10-25
Disable dither in es3fPixelBufferObjectTest.js

qjia7:
Enabling GL_DITHER will cause dither on color components which
effets the comparing result of read pixel. Disable dither will
make sure that the diff is not because of dithering.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
+++ b/sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
@@ -287,6 +287,9 @@ var tcuImageCompare = framework.common.tcuImageCompare;
                 break;
         }
 
+        // GL_DITHER will cause the dither on color component which effects the result of rendering.
+        // Disable it to make sure that the diff of image is not because of dithering.
+        gl.disable(gl.DITHER);
         this.clearColor(this.m_colorScale * 0.4, this.m_colorScale * 1.0, this.m_colorScale * 0.5, this.m_colorScale * 1.0);
 
         if (this.m_useColorClears) {


qjia7:
@kenrussell  Please take a look.

sdk/tests/deqp/functional/gles3/es3fPixelBufferObjectTest.js
```diff
@@ -287,6 +287,9 @@ var tcuImageCompare = framework.common.tcuImageCompare;
                 break;
         }
 
+        // GL_DITHER will cause the dither on color component which effects the result of rendering.
```
kenrussell: Could you please indicate in this comment that this is a difference compared to the native dEQP?

kenrussell:
Thank you for putting this together.

It looks good to me. I'd like one other person to please review this too: @zhenyao @Kangz @phaulos , could you take a look?

phaulos:
I'm confused why this is needed on your implementation. What the tests do, is that they render once to a single RT, and then read it back using glReadPixels() twice, once using PBO, and once without. Tests expect identical results regardless of whether PBO was used, as it should not affect the data returned at all.

If the results don't match and disabling dithering fixes this, this sounds like an implementation bug, not test bug.

phaulos:
If this occurs on Mesa driver, you may want to take a look at https://chromium.googlesource.com/chromiumos/third_party/mesa/+/a882167417f0af365314a72c599d50dcc396e1c1

kenrussell:
Thanks @phaulos for your feedback and for preventing me from introducing a bug into the WebGL conformance tests. :)

@qjia7 , this sounds like a bug in Mesa. Could you please report it to the driver team? Let's close this so it isn't accidentally merged.

kenrussell:
Oh wait -- of course, this was a bug found on macOS. It sounds like the bug is probably similar to the one in Mesa. @phaulos thanks again for the pointer.

qjia7:
@phaulos @kenrussell  Originally, I just wanted to let the comparing result more clear. For example, test case pixel_buffer_object.renderbuffer.rgb5_a1_clears in pixelbufferobject.html is failed on Intel Mac. The default result is as below (with GL_DITHER enabled). The diff image is not completely red in error field.
enable_dither

In order to eliminate dither's effect, after disabling GL_DITHER, the result of that case will be as like:
disable_dither

From above two pictures, we can see that with or without dither, this case is always failed. But without dither, the left error filed in first picture disappeared.  It's hard to say that dither affects it since this error only happens on some special colors. I think we should firstly resolve this bug without dither. Then see that if there is still error with dither.

I am ok to abandon this patch. But maybe we need another case to test dither in WebGL.

phaulos:
My guess from those images is that RGB5_A1 is emulated using higher-bitdepth format and there is a difference in conversion / rounding when doing glReadPixels() with and without PBO.

kenrussell:
@qjia7 the problem with adding this patch is that it would allow this same bug to slip in undetected to other graphics drivers. I didn't understand the point of this test; it's a good thing that @phaulos caught my mistake.

There are probably other tests in deqp/functional/gles3 which exercise dithering; @phaulos do any come to mind?

phaulos:
Yep, we do have tests that specifically target dithering as well, at least on dEQP side: https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fDitheringTests.cpp

kenrussell:
@phaulos thanks for the pointer. It looks like we didn't port those into the WebGL conformance suite, probably because dithering wasn't a new feature in ES 3.0. However I'm sure the dEQP dithering tests would have added value to the WebGL suite.

PR #2112, opened 2016-10-25
Revert "Add test for samplers in a ternary operator."

Oletus:
The test added by the commit being reverted is incorrect. ESSL 1.00
section 4.1.7 Samplers:

"Except for parameters to texture lookup functions, array indexing,
structure field selection, and parentheses, samplers are not allowed
to be operands in expressions."

This reverts commit 499fd34102d683c10ee3c1e78ce9ea3e544543a4.


Kangz:
LGTM

kenrussell:
Thanks @Oletus for taking care of this and @Kangz for reviewing it.

@c0d1f1ed it looks like all the Android games that use this pattern in their shaders are incorrect. I wonder what percentage of them it is?

PR #2113, opened 2016-10-25
Add a test to verify WebGL2 specific sync object behaviors

zhenyao:
@kenrussell @kainino0x @jdashg PTAL


kenrussell:
Looks good.

PR #2114, opened 2016-10-25
Fix title of web page for pixel buffer object tests.

kenrussell:


kenrussell:
Trivial change - merging myself.

PR #2117, opened 2016-10-26
Fixed incorrect calculation of max sample count

qjia7:
gl.getInternalformatParameter(gl.RENDERBUFFER, gl.RGBA8, gl.SAMPLES) will return
an Int32Array containing sample counts supported for internalformat in descending order.
The old way incorrectly used Int32Array.length as the max sample number.
This patch fixes it by using Int32Array[0] as the max sample number.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
@@ -481,7 +481,13 @@ goog.scope(function() {
                 this.m_numSamples = this.m_fboParams.numSamples;
             else {
                 bufferedLogToConsole('Querying maximum number of samples for ' + gluStrUtil.getPixelFormatName(gl.RGBA8) + ' with gl.getInternalformatParameter()');
-                this.m_numSamples = /** @type {number} */ (gl.getInternalformatParameter(gl.RENDERBUFFER, gl.RGBA8, gl.SAMPLES).length);
+                var supportedSampleCountArray = /** @type {Int32Array} */ (gl.getInternalformatParameter(gl.RENDERBUFFER, gl.RGBA8, gl.SAMPLES));
+                if (supportedSampleCountArray.length == 0) {
+                    var msg = 'No supported sample counts';
+                    checkMessage(false, msg);
+                    return false;
+                }
+                this.m_numSamples = supportedSampleCountArray[0];
             }
 
             bufferedLogToConsole('Using FBO of size (' + this.m_renderWidth + ', ' + this.m_renderHeight + ') with ' + this.m_numSamples + ' samples');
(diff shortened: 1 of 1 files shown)


qjia7:
The corresponding native deqp case is in https://android.googlesource.com/platform/external/deqp/+/master/modules/gles3/functional/es3fMultisampleTests.cpp#452
@kenrussell @zhenyao @qkmiao Please take a look.

sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
```diff
@@ -481,7 +481,13 @@ goog.scope(function() {
                 this.m_numSamples = this.m_fboParams.numSamples;
             else {
                 bufferedLogToConsole('Querying maximum number of samples for ' + gluStrUtil.getPixelFormatName(gl.RGBA8) + ' with gl.getInternalformatParameter()');
-                this.m_numSamples = /** @type {number} */ (gl.getInternalformatParameter(gl.RENDERBUFFER, gl.RGBA8, gl.SAMPLES).length);
+                var supportedSampleCountArray = /** @type {number} */ (gl.getInternalformatParameter(gl.RENDERBUFFER, gl.RGBA8, gl.SAMPLES));
```
zhenyao: @type is wrong here

zhenyao:
Mostly looks good.  Does this fix the test on certain platforms?

qjia7:
>Does this fix the test on certain platforms?

Yes. On Intel Skylake with win10, all/deqp/functional/gles3/multisample.html has two cases failure because that the max sample number is 5 which is not in the supported array.

sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
qjia7: Done, thanks.

qjia7:
@zhenyao Should an error be generated if the sample number is not in the supported list but it is less than the max sample number?  It seems that the spec doesn't have a clear description about this.

zhenyao:
From ES 3.0.4 spec page 205:
"If samples is zero, then RENDERBUFFER_SAMPLES is set to zero. Otherwise
samples represents a request for a desired minimum number of samples. Since
different implementations may support different sample counts for multisample
rendering, the actual number of samples allocated for the renderbuffer image is
implementation-dependent. However, the resulting value for RENDERBUFFER_-
SAMPLES is guaranteed to be greater than or equal to samples and no more than the
next larger sample count supported by the implementation."

So it's clear if 5 is less than the max sample number, the driver should allow the call to succeed by upgrading to a samples > 5 internally.

I still think your fix to this test is valid because apparently the js code wasn't aligned with the c++. However, I do think the Win10 Intel Skylake has a bug in it.  Can you add a separate simple test to expose this bug?

qjia7:
>Can you add a separate simple test to expose this bug?

Sure. See #2118

PR #2118, opened 2016-10-27
Add test to check SAMPLE_COVERAGE capability with sample number from …

qjia7:
…1 to max sample number

The added tests will test sample coverage capbility with sample number from 1 to max sample number
which also includes the samples that may not be in the supported sample list by
getInternalformatParameter(gl.RENDERBUFFER, internalFormat, gl.SAMPLES)

### DIFF
--- a/sdk/tests/conformance2/renderbuffers/00_test_list.txt
+++ b/sdk/tests/conformance2/renderbuffers/00_test_list.txt
@@ -3,4 +3,5 @@ framebuffer-test.html
 framebuffer-texture-layer.html
 invalidate-framebuffer.html
 multisampled-renderbuffer-initialization.html
+--min-version 2.0.1 multisample-with-full-sample-counts.html
 readbuffer.html
--- a/sdk/tests/conformance2/renderbuffers/multisample-with-full-sample-counts.html
+++ b/sdk/tests/conformance2/renderbuffers/multisample-with-full-sample-counts.html
@@ -1,0 +1,119 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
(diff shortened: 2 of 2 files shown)


qjia7:
@zhenyao @kenrussell Please take a look.

sdk/tests/conformance2/renderbuffers/multisample-with-full-sample-counts.html
```diff
@@ -0,0 +1,141 @@
...
+    gl.blitFramebuffer(0, 0, size, size, 0, 0, size, size, gl.COLOR_BUFFER_BIT, gl.NEAREST);
+    gl.bindFramebuffer(gl.READ_FRAMEBUFFER, resolveFbo);
+
+    var pixels = new Uint8Array(size * size * 4);
+    gl.pixelStorei(gl.PACK_ALIGNMENT, 4);
+    gl.readPixels(0, 0, size, size, gl.RGBA, gl.UNSIGNED_BYTE, pixels);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "readpixels should succeed");
+
+    // Compare
+    var isOk = true;
```
qkmiao: Can you use wtu.checkCanvasRect to compare the result with expected color?

qjia7: Good suggestion. Done, thanks.

sdk/tests/conformance2/renderbuffers/multisample-with-full-sample-counts.html
```diff
@@ -0,0 +1,119 @@
...
+    wtu.drawFloatColorQuad(gl, quadColor);
+
+    gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, resolveFbo);
+    gl.blitFramebuffer(0, 0, size, size, 0, 0, size, size, gl.COLOR_BUFFER_BIT, gl.NEAREST);
+
+    gl.bindFramebuffer(gl.FRAMEBUFFER, resolveFbo);
+    wtu.checkCanvasRect(gl, 0, 0, size, size, [255, 0, 0, 255],
+                        "User buffer has been rendered to red with sample = "
+                        + sampleCount + ", coverageValue = " + coverageValue
+                        + " and isInverted = " + isInverted);
```
zhenyao: Can we add a tolerance of 2 or 3 here? That should still be able to verify the intended behavior.

sdk/tests/conformance2/renderbuffers/multisample-with-full-sample-counts.html
```diff
@@ -0,0 +1,119 @@
...
+        runTest(gl, i, false);
+        runTest(gl, i, true);
+    }
+}
+
+function runTest(gl, sampleCount, isInverted) {
+    // Setup multi-sample RBO
+    var msColorRbo = gl.createRenderbuffer();
+    gl.bindRenderbuffer(gl.RENDERBUFFER, msColorRbo);
+    gl.renderbufferStorageMultisample(gl.RENDERBUFFER, sampleCount, gl.RGBA8, size, size);
```
zhenyao: How does Intel fail on this? It generates a GL error? If yes, then you need to check the GL error here.  If it fails in other fashion, can you describe how it fails?

sdk/tests/conformance2/renderbuffers/00_test_list.txt
```diff
@@ -3,4 +3,5 @@ framebuffer-test.html
 framebuffer-texture-layer.html
 invalidate-framebuffer.html
 multisampled-renderbuffer-initialization.html
+multisample-with-full-sample-counts.html
```
zhenyao: Please put it to 2.0.1

sdk/tests/conformance2/renderbuffers/multisample-with-full-sample-counts.html
qjia7: It's not a GL error. Intel driver can process it which is not in the supported list. But the result is not completely red when the coverage is 1. See below result:
intel_error

qjia7:
@zhenyao  Please take another look

qkmiao:
I tested it passed on Linux Mesa, Linux NVIDIA, Mac Intel and Mac AMD.

zhenyao:
Thank you. LGTM

PR #2120, opened 2016-10-28
Add test for tex uploads with view offsets.

jdashg:
Also update idl for compressedTex uploads to allow an optional length.

Let me know if the readPixels tests are redundant, and I can remove them.

I think the uniform upload entrypoints still need tests, but we can do this a chunk at a time.

@kenrussell @zhenyao

### DIFF
--- a/sdk/tests/conformance/misc/views-with-offsets.html
+++ b/sdk/tests/conformance/misc/views-with-offsets.html
@@ -1,0 +1,353 @@
+<!--
+
+/*
+** Copyright (c) 2012 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/desktop-gl-constants.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+
+<script>
+"use strict";
+description("Tests texture uploads with ArrayBufferView+offsets");
+
+var wtu = WebGLTestUtils;
+var contextVersion = wtu.getDefault3DContextVersion();
+var gl = wtu.create3DContext();
+console.log(gl.getParameter(gl.VERSION));
+
+////
+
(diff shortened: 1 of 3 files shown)


zhenyao:
By the way, it would be nice to also test CompressedTex functions without the |srcLengthOverride| params and verify it works as intended.

sdk/tests/conformance/misc/views-with-offsets.html
```diff
@@ -0,0 +1,353 @@
...
+
+        gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, readPixelView);
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+        shouldBeWasArr(testView.slice(i, i+4), readPixelView);
+
+      } else if (effectiveViewLen >= 0) {
+        wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION);
+
+      } else {
+        wtu.glErrorShouldBe(gl, gl.INVALID_VALUE);
```
kenrussell: The "srcOffset" argument to texImage2D / texSubImage2D taking ArrayBufferView is GLuint, so INVALID_VALUE can't be generated for negative arguments; they look like large positive arguments, and generate INVALID_OPERATION.

sdk/tests/conformance/misc/views-with-offsets.html
```diff
@@ -0,0 +1,353 @@
...
+        gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, readPixelView);
+        debug(arrToStr(readPixelView));
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+        shouldBeWas(arr565[i], rgb888to565(readPixelView));
+
+      } else if (effectiveViewLen >= 0) {
+        wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION);
+
+      } else {
+        wtu.glErrorShouldBe(gl, gl.INVALID_VALUE);
```
kenrussell: Same issue here; this can only generate INVALID_OPERATION.

sdk/tests/conformance/misc/views-with-offsets.html
```diff
@@ -0,0 +1,353 @@
...
+    var viewLength = view.length;
+    var subViewLength = compressedByteCount;
+
+    for (var i = 0; i <= viewLength+1; i++) {
+      debug("offset=" + i);
+
+      fnToTest(i);
+      var effectiveViewLen = viewLength - i;
+
+      if (effectiveViewLen >= subViewLength) {
```
kenrussell: This is subtly tricky. All of the existing compressed texture tests assume that the data length is exactly equal to the required amount, rather than being greater than or equal. Section 3.8.6 "Compressed Texture Images" of the ES 3.0.4 spec says:

"If the imageSize parameter is not consistent with the format, dimensions, and contents of the compressed image (too little or too much data), an INVALID_VALUE error results."

I'd be very concerned with changing this requirement at this stage of WebGL 2.0.0's release; doing so will require all of the compressed texture tests to be updated. It's also not clear whether drivers will handle the case where "too much" data is supplied, rather than exactly the right amount. I think this should be changed to:
```
  if (effectiveViewLen == subViewLength)
```

specs/latest/2.0/index.html
```diff
@@ -785,27 +785,29 @@ <h2 class="no-toc">Table of contents</h2>
                             GLsizei height, GLint border, GLintptr offset);
   void compressedTexImage2D(GLenum target, GLint level, GLenum internalformat, GLsizei width,
                             GLsizei height, GLint border, ArrayBufferView srcData,
-                            optional GLuint srcOffset = 0);
+                            optional GLuint srcOffset = 0, optional GLuint srcLengthOverride = 0);
```
kenrussell: Could you document the behavior of srcLengthOverride? i.e., what it defaults to if it's 0?

kenrussell: Also, we need to define what happens if srcOffset + srcLengthOverride > srcData.length. Not sure whether that should be INVALID_OPERATION or INVALID_VALUE. The uniform[1234] entry points generate INVALID_VALUE for this sort of condition. The texImage entry points generate INVALID_OPERATION if srcOffset would mean reading beyond the end of the array.

Also, what happens if srcLengthOverride is longer than srcData.length - srcOffset? Should that be INVALID_OPERATION?

sdk/tests/conformance/misc/views-with-offsets.html
kenrussell: I didn't realize that you were already passing in the new "srcLengthOverride" argument. Given that, this is fine -- ignore my comment above.

specs/latest/2.0/index.html
kenrussell: Note: having just implemented this in Chromium -- both of these situations need to generate INVALID_VALUE in order to make the test above pass.

kenrussell:
With the two INVALID_OPERATION -> INVALID_VALUE changes above, these tests all pass in Chromium with https://codereview.chromium.org/2457363002/ applied. Thanks for putting them together.

kenrussell:
I'm going to merge this now and follow up with another pull request making this a WebGL 2.0-only test, and fixing the two errors above.

PR #2121, opened 2016-10-31
Detecte feedback loop between sampling textures and drawing framebuffer attachments

Richard-Yunchao:
PTAL. @zhenyao  and @kenrussell. Thanks a lot!

### DIFF
--- a/sdk/tests/conformance2/rendering/00_test_list.txt
+++ b/sdk/tests/conformance2/rendering/00_test_list.txt
@@ -18,6 +18,6 @@ framebuffer-unsupported.html
 instanced-arrays.html
 --min-version 2.0.1 instanced-rendering-bug.html
 out-of-bounds-index-buffers-after-copying.html
-rendering-sampling-feedback-loop.html
+--min-version 2.0.1 rendering-sampling-feedback-loop.html
 rgb-format-support.html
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/rendering/00_test_list.txt
```diff
@@ -18,5 +18,6 @@ framebuffer-unsupported.html
 instanced-arrays.html
 --min-version 2.0.1 instanced-rendering-bug.html
 out-of-bounds-index-buffers-after-copying.html
+rendering-sampling-feedback-loop.html
```
kenrussell: We're trying to wrap up version 2.0.0 of the conformance suite. At this point I think we should put pretty much all new tests except for those for the texture sub-source entry points into 2.0.1. What do you think @Richard-Yunchao @zhenyao @jdashg ?

zhenyao: I agree.  Let's put this to 2.0.1. Chrome will probably be able to enforce this in the initial release, so any contents violating this rule will not run.

Richard-Yunchao:
@kenrussell , @zhenyao , agreed. I have moved this into 2.0.1. 

Could you take another look?

zhenyao:
LGTM

PR #2123, opened 2016-11-02
Add a test to verify type mismatch

Richard-Yunchao:
@kenrussell and @zhenyao , PTAL. Thanks a lot! 

One more test case for undefined behavior in GLES3 spec, but we should generate INVALID_OPERATION in WebGL 2. 

Zhenyao has landed a patch to handle this issue at https://codereview.chromium.org/2142353002. But I didn't find a conformance test to cover this feature. So I wrote this one.

### DIFF
--- a/sdk/tests/conformance2/rendering/fs-color-type-mismatch-color-buffer-type.html
+++ b/sdk/tests/conformance2/rendering/fs-color-type-mismatch-color-buffer-type.html
@@ -29,7 +29,7 @@
 <html>
 <head>
 <meta charset="utf-8">
-<title>The Color Type of Fragment Shader's Output Should Match The Type of Color Buffers</title>
+<title>The Color Types of Fragment Shader's Outputs Should Match The Data Types of Color Buffers</title>
 <link rel="stylesheet" href="../../resources/js-test-style.css"/>
 <script src="../../js/js-test-pre.js"></script>
 <script src="../../js/webgl-test-utils.js"></script>
@@ -75,7 +75,7 @@ void main() {
 "use strict";
 
 var wtu = WebGLTestUtils;
-description("This test verifies that the color type of fragment shader's output should match color buffers' type.");
+description("This test verifies that the color types of fragment shader's outputs should match color buffers' types.");
 
 var gl = wtu.create3DContext("example", undefined, 2);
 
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/rendering/fs-color-type-mismatch-color-buffer-type.html
```diff
@@ -0,0 +1,168 @@
...
+void main() {
+    oColor = vec4(1.0, 0.0, 0.0, 0.0);
+}
+</script>
+
+<script id="fshaderMRT" type="x-shader/x-fragment">#version 300 es
+precision mediump float;
+out vec4 oColor[2];
+void main() {
+    oColor[0] = vec4(1.0, 0.0, 0.0, 0.0);
```
Richard-Yunchao: The similar question: If we have multiple render targets(MRTs) by calling drawBuffers() API, does all other MRTs will be written to (0, 0, 0, 0)? Or only the first one is written, the others are untouched? 
And does this feature differs between WEBGL_draw_buffers extension in WebGL 1 and drawBuffers API in WebGL 2 core feature?

sdk/tests/conformance2/rendering/fs-color-type-mismatch-color-buffer-type.html
```diff
@@ -0,0 +1,168 @@
...
+void main() {
+    gl_Position = aPosition;
+}
+</script>
+
+<script id="fshader" type="x-shader/x-fragment">#version 300 es
+precision mediump float;
+out vec4 oColor;
+void main() {
+    oColor = vec4(1.0, 0.0, 0.0, 0.0);
```
kenrussell: The broadcasting behavior does differ.

The only situation where writing to a single output variable broadcasts its value to all of the draw buffers is when using the  https://www.khronos.org/registry/webgl/extensions/WEBGL_draw_buffers/ extension and enabling it in the shader via
```
#extension GL_EXT_draw_buffers : enable
```

OpenGL ES 3.0 / WebGL 2.0 do not support broadcasting. The behavior is defined in section 3.9.2 "Shader Execution" of the ES 3.0.4 spec. It is necessary to define multiple output variables, one for each draw buffer, and write to each of them separately.

sdk/tests/conformance2/rendering/fs-color-type-mismatch-color-buffer-type.html
kenrussell: This is defined in https://www.khronos.org/registry/webgl/specs/latest/2.0/#5.3 . All user-defined output variables that are not written by the shader are initialized to zero.

kenrussell:
Thanks, this looks fine to me. @zhenyao , could you review too and "squash and merge"?

zhenyao:
lgtm
I am surprised we haven't had this covered already.

Richard-Yunchao:
Thanks for your review and merge. @kenrussell and @zhenyao . 

@qkmiao , FYI. 
The shader outputs can be float, int, or uint, although all tests are float output in webgl CTS. 
And the texture's data type (float, int or uint) should match the corresponding fragment output's type. 

This info might be helpful for your gpu-gpu texture uploading work.

PR #2124, opened 2016-11-03
Fix WebGL1 conformance tests to the API changes (null-able).

zhenyao:
Note this is ToT. Once this lands, I'll fix the earlier snapshots.


zhenyao:
@kenrussell @jdashg @kainino0x

zhenyao:
Oops. Thanks for catching this bug.

kenrussell:
These fixes look great -- are they complete at this point?

zhenyao:
For WebGL1, yes, it's complete.

kenrussell:
Excellent. Thank you for diligently taking care of these. Merging now.

slacka:
@zhenyao 
Can your changes be backported? I'm seeing the following failures on [1.0.3.](https://www.khronos.org/registry/webgl/conformance-suites/1.0.3/webgl-conformance-tests.html
)

> conformance/buffers/buffer-data-array-buffer.html
> conformance/context/context-lost.html
> conformance/misc/bad-arguments-test.html
> conformance/misc/error-reporting.html
> conformance/misc/null-object-behaviour.html
> conformance/programs/get-active-test.html
> conformance/state/gl-object-get-calls.html

With Chrome Canary. On 1.0.4 those tests are passing.

zhenyao:
I thought I did, but apparently not. I am on it.

PR #2125, opened 2016-11-03
bufferData can't take non-nullable ArrayBuffer{View} arg

zhenyao:
due to overload conflicts.

See discussions in https://github.com/KhronosGroup/WebGL/pull/2099


zhenyao:
@kenrussell @jdashg @RafaelCintron @grorg @kainino0x

kenrussell:
Looks great. Thanks for so carefully analyzing these changes and making the doc updates.

PR #2126, opened 2016-11-04
Fix a small bug in call sites to clearBuffer

Richard-Yunchao:
This is a typo: gl.DRAWBUFFER0 is meant to be gl.DRAW_BUFFER0. But we should use i, instead of DRAW_BUFFERi according to the spec. If we use the latter, it will report INVALID_ENUM. 

I felt weird why the original test can pass. So I dig a little into the code. It turned out that gl.DRAWBUFFER0 was undefined. So it was 0xNaN, which was thought to be 0 too... So it can pass by chance!

PTAL. Thanks a lot!


zhenyao:
Oops, thanks for catching this bug.

kenrussell:
Wow. Good catch Yunchao.

PR #2127, opened 2016-11-04
Add tests to verify that srgb conversion is done when it clears a srgb color buffer

Richard-Yunchao:
In GLES 3 spec, it explicitly says that the implementation should do srgb conversion when clear a srgb color buffer, see page 189 at section 4.2.3 <Clearing the Buffers>: https://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-4.2.3. 

However, the OGL spec may not. For example, in OGL 4.1 spec, the srgb conversion is not included by the per-fragment operations for clear buffer. See page 276 at section 4.2.3 <Clearing the Buffers>  : https://www.opengl.org/registry/doc/glspec41.core.20100725.pdf 

This behavior is consistent to other similar behaviors in OGL spec. For example, blitFramebuffer and generateMipmap against srgb color buffer may not do srgb conversion on OGL 4.3 profile or prior to OGL 4.3 profile.

I didn't find any test case covered this feature, so I wrote this test. 

However, the new test can pass on MacOS Intel (OGL 4.1). That's interesting...

PTAL. Thanks a lot!

### DIFF
--- a/sdk/tests/conformance2/rendering/clear-srgb-color-buffer.html
+++ b/sdk/tests/conformance2/rendering/clear-srgb-color-buffer.html
@@ -1,0 +1,131 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/rendering/clear-srgb-color-buffer.html
```diff
@@ -0,0 +1,131 @@
...
+
+    var color_ref = wtu.linearToSRGB(color);
+    for (var ii = 0; ii < size; ++ii) {
+        for (var jj = 0; jj < size; ++jj) {
+            var index = (ii * size + jj ) * 4;
+            var pix = [pixels[index], pixels[index + 1], pixels[index + 2], pixels[index + 3]];
+            if (checkPixel(pix, color_ref) == true) {
+                testPassed("pixel at [" + jj + ", " + ii + "] is (" + pix + "). It is correct!");
+            } else {
+                testFailed("pixel at [" + jj + ", " + ii + "] should be (" + color_ref + "), but the actual color is (" + pix + ")");
```
kenrussell: I'd really like to get into the habit of using WebGLUtils.checkCanvasRect instead of hand-writing these pixel comparisons into all of the tests. Could you please look into doing that refactoring?

zhenyao:
lgtm
I am also curious why it's passing on MacOSX

zhenyao:
Ah, I didn't see @kenrussell 's review comments. @Richard-Yunchao can you refactor according to the suggestion?

Richard-Yunchao:
Ah... I will refactor the code here, and all other similar cases. @kenrussell , @zhenyao .

zhenyao:
Thank you @Richard-Yunchao

PR #2130, opened 2016-11-07
Use existed function in WebGLTestUtil

Richard-Yunchao:
@kenrussell and @zhenyao  , per your request at #2127 , this small change use function in WebGLTestUtils (wtu.checkCanvasRect) to simplify the test. 

I also go through the similar usage, and found another one. The others still use a function to do pixels comparison, but they are not solid color, so they are not suitable to be replaced by wtu.checkCanvasRect. 

PTAL. Thanks a lot!


zhenyao:
lgtm

kenrussell:
Thanks for this cleanup @Richard-Yunchao .

PR #2133, opened 2016-11-08
Updated obsolete texSubImage3D signature in SVG image texture tests.

kenrussell:


kenrussell:
Trivial fix. Merging myself. @zhenyao @jdashg @kainino0x

PR #2136, opened 2016-11-09
move some driver bugs into skip list

Richard-Yunchao:
These failures should be fixed in MacOS driver, let's move them into skip list. Zhenyao's email also suggested to fix them in WebGL 2.0.1.

@zhenyao and @kenrussell , PTAL.

### DIFF
--- a/sdk/tests/deqp/framework/common/tcuSkipList.js
+++ b/sdk/tests/deqp/framework/common/tcuSkipList.js
@@ -118,6 +118,12 @@ goog.scope(function() {
         _skip("blit.rect.nearest_consistency_out_of_bounds_min_reverse_src_x");
         _skip("blit.rect.nearest_consistency_out_of_bounds_min_reverse_src_y");
 
+        // crbug.com/658724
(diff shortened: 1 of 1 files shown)


sdk/tests/deqp/framework/common/tcuSkipList.js
```diff
@@ -118,6 +118,12 @@ goog.scope(function() {
         _skip("blit.rect.nearest_consistency_out_of_bounds_min_reverse_src_x");
         _skip("blit.rect.nearest_consistency_out_of_bounds_min_reverse_src_y");
 
+        // crbug.com/658724
+        // deqp/functional/gles3/framebufferblit/rect_03.html
```
zhenyao: Can you add a _setReason?

Richard-Yunchao: Done. Thanks for your review, Zhenyao.

zhenyao:
Looks fine.  Thanks.

PR #2137, opened 2016-11-09
Add test cases to test the functionality when reading from or drawing to a color buffer without image attached

Richard-Yunchao:
I am investigating the framebuffer completeness issue, found that a corresponding feature is not covered when we read from or draw to a color buffer which has no image. 

In GLES 2.0/3.0 spec, it is not specified what happen in this situation. In WebGL 1.0, it explicitly says that it generates INVALID_OPERATION if reading from a color buffer without image, see https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.26. 
The corresponding APIs are readPixels, copyTexImage2D, copyTexSubImage2D, copyTexSubImage3D. 

When drawing to a color buffer without image, I propose to not generate any error, and do nothing. And the Chromium current implementation does follow this. 
The corresponding APIs are clear, clearBuffer*, drawElements, drawArrays. 

Another related API is blitFramebuffer, I propose to follow the same rule. So it generates INVALID_OPERATION if the read buffer has no image. It would not generate any error and do nothing if all draw buffers have no image. 

WDYT? @zhenyao  and @kenrussell ? PTAL. Thanks a lot!

### DIFF
--- a/sdk/tests/conformance2/rendering/read-draw-when-missing-image.html
+++ b/sdk/tests/conformance2/rendering/read-draw-when-missing-image.html
@@ -1,0 +1,270 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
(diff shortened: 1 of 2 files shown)


zhenyao:
Yunchao, thanks for bringing this up, but let's handle this after the Chrome's M56 branch point.

For BlitFramebuffer, if both src and dst images are none, it is allowed and it's simply a no-op.

Richard-Yunchao:
That's fine, Zhenyao. 
Let's revisit this feature after Chrome's M56 branch point.

Richard-Yunchao:
Update the conformance test according to #2137, please take a look when you have time, @zhenyao, @kenrussell. 

The conformance test is OK to land if It is correct, because we have already checked out the WebGL CTS 2.0.0 snapshot, although #2137 should be after WebGL Spec 2.0.0 snapshot.

kenrussell:
Sorry for the long delay @Richard-Yunchao . Your new test looks good to me, but I'd like to ask @zhenyao to do another review, and merge it. (The spec snapshot will be done in #2196 .)

sdk/tests/conformance2/rendering/read-draw-when-missing-image.html
```diff
@@ -0,0 +1,270 @@
...
+    checkTextureValue(fbo_draw, gl.COLOR_ATTACHMENT0, [0, 255, 0, 255]);
+
+    var data = new Float32Array(color);
+    gl.clearBufferfv(gl.COLOR, 1, data);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "Should be no error.");
+}
+
+function draw() {
+    debug("");
+    debug("draw a color buffer which has no image attached");
```
zhenyao: nit: draw to

zhenyao:
lgtm

PR #2138, opened 2016-11-10
add tests for upload sub-rectangles of ImageData

kainino0x:
Verified against Chromium's implementation in https://codereview.chromium.org/2489203002

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js
@@ -81,11 +81,19 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
         runTest();
     }
 
-    function runOneIteration(useTexSubImage2D, flipY, premultiplyAlpha, bindingTarget, program)
+    function runOneIteration(useTexSubImage2D, flipY, premultiplyAlpha,
+                             sourceSubRectangle, expected,
+                             bindingTarget, program)
     {
+        sourceSubRectangleString = '';
+        if (sourceSubRectangle) {
+            sourceSubRectangleString = ', sourceSubRectangle=' + sourceSubRectangle;
+        }
+        debug('');
         debug('Testing ' + (useTexSubImage2D ? 'texSubImage2D' : 'texImage2D') +
               ' with flipY=' + flipY + ' and premultiplyAlpha=' + premultiplyAlpha +
-              ', bindingTarget=' + (bindingTarget == gl.TEXTURE_2D ? 'TEXTURE_2D' : 'TEXTURE_CUBE_MAP'));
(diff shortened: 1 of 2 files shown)


sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js
```diff
@@ -107,30 +115,63 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
...
+            gl.pixelStorei(gl.UNPACK_SKIP_ROWS, sourceSubRectangle[1]);
+        }
         // Upload the image into the texture
         for (var tt = 0; tt < targets.length; ++tt) {
-            if (useTexSubImage2D) {
-                // Initialize the texture to black first
-                gl.texImage2D(targets[tt], 0, gl[internalFormat], imageData.width, imageData.height, 0,
-                              gl[pixelFormat], gl[pixelType], null);
-                gl.texSubImage2D(targets[tt], 0, 0, 0, gl[pixelFormat], gl[pixelType], imageData);
+            if (sourceSubRectangle) {
```
kenrussell: Looking at this again, could we define:
  var uploadWidth = imageData.width;
  var uploadHeight = imageData.height;
and override them if sourceSubRectangle is defined?

That way we could reference those variables in the tex{Sub}Image2D calls and avoid duplicated code.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js
```diff
@@ -181,19 +220,39 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
...
+            { expected: [r, k, g, k], flipY: false, premultiplyAlpha: true, sub: true },
+            { expected: [g, g, r, r], flipY: true, premultiplyAlpha: false, sub: false },
+            { expected: [g, g, r, r], flipY: true, premultiplyAlpha: false, sub: true },
+            { expected: [g, k, r, k], flipY: true, premultiplyAlpha: true, sub: false },
+            { expected: [g, k, r, k], flipY: true, premultiplyAlpha: true, sub: true },
         ];
 
+        if (wtu.getDefault3DContextVersion() > 1) {
+            var morecases = [];
+            // make 4 copies of the original case, one for each corner
```
kenrussell: Comment needs to be updated: "make 2 copies of the original case for the opposite corners" or similar.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-data.js
kenrussell: Sorry, I forgot that these are actually calling different overloads. These are fine as is.

kenrussell:
Thanks for putting these tests together. I'll merge them now and update the comment in a follow-on PR.

kenrussell:
Thanks. I fell asleep last night before merging this, so thanks for updating the comment. Merging now.

PR #2139, opened 2016-11-11
Change tests to match WebGL2's uniform block layout behaviors.

zhenyao:


zhenyao:
@kenrussell @kainino0x PTAL

zhenyao:
Also @jdashg: both linkage.html and the new test are able to pass with ToT ANGLE

kenrussell:
LGTM. It looks like there are conflicts; can you rebase?

PR #2140, opened 2016-11-11
Add a proposal for WEBGL_multiview extension

Oletus:
The extension document contains a proposal to amend the WebGL 1.0
specification to include optional stereo rendering capability. In case
the actual context parameter "stereo" is set to true, the default
framebuffer will have left and right buffers, and the user may
choose either left or right draw buffer of the default framebuffer as
a target of pixel color writes.

The intent is that the stereo default framebuffer can be used as a
VRLayer source in the WebVR API. The backing for stereo default
framebuffers should be possible to implement so that the framebuffer
may be submitted for display using a VR headset API with zero copies
in between. Since the exact layout of the framebuffer is not exposed
to the WebGL API, the WebGL implementation may choose a layout that
is suitable for the API provided by the connected headset. The case
where there are multiple connected headsets should be rare, but if it
happens, the implementation may also switch the layout of the default
framebuffer behind the scenes depending on which headset the
framebuffer was last displayed on.

The WEBGL_multiview extension itself is built on OVR_multiview to
enable more efficient stereo rendering.

WEBGL_multiview could be implemented either only inside the browser to
reduce overhead in high-level code, or all the way down to hardware
that supports native stereo acceleration.

### DIFF
--- a/extensions/proposals/WEBGL_multiview/extension.xml
+++ b/extensions/proposals/WEBGL_multiview/extension.xml
@@ -1,0 +1,228 @@
+<?xml version="1.0"?>
+
+<proposal href="proposals/WEBGL_multiview/">
+  <name>WEBGL_multiview</name>
+  <contact>
+    <a href="https://www.khronos.org/webgl/public-mailing-list/">WebGL working group</a> (public_webgl 'at' khronos.org)
+  </contact>
+  <contributors>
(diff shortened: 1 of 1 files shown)


kenrussell:
Thanks Olli for putting this together. Let's merge it now to make evaluation and discussion easier.

PR #2142, opened 2016-11-11
Add test cases to catch a chrome bug in the following path

zhenyao:
texStorage(compressed_format)
  texSubImage(compressed_format)


zhenyao:
@kenrussell

kenrussell:
Thanks for expanding this test to catch the bug. In the future, please reference the Chrome bug in the pull request's description; in this case, it's http://crbug.com/663960 .

kenrussell:
...and, to be clear, the bug was in the use of texStorage2D in conjunction with compressedTexSubImage2D. During a refactoring, some code in Chrome which handled this case was lost.

PR #2145, opened 2016-11-12
Add tests for uploading canvas sub-rectangles to 2D and 3D textures.

kenrussell:
Tested with Chromium's implementation in https://codereview.chromium.org/2495953002/ .


kenrussell:
This isn't ready for review yet; please don't merge it. (The 3D tests' code is bogus, though all the 2D tests are passing.) I'm uploading it as a pull request mainly to have a backup of the code. @zhenyao @kainino0x @jdashg

kenrussell:
Ready for review @zhenyao @kainino0x @jdashg .

These have been tested with https://codereview.chromium.org/2495953002/ (which I'm about to update with the code paths for handling 3D texture uploads).

I recommend starting the review at the bottom of the file list, since the files at the top of the review are all auto-generated. There are only ~4 hand-written files changed.

sdk/tests/conformance/textures/00_test_list.txt
```diff
@@ -1,5 +1,6 @@
 misc/00_test_list.txt
 canvas/00_test_list.txt
+--min-version 1.0.4 canvas_sub_rectangle/00_test_list.txt
```
zhenyao: Should this be 2.0.0?

kenrussell: I think 1.0.4 is the conceptually correct version, since these tests live in the "WebGL 1.0" sub-part of the tests' tree (conformance/ rather than conformance2/) -- this is because the texture formats that are tested in this directory are those which WebGL 1.0 supported.

Right now, the tests all skip themselves if they're run on WebGL 1.0 -- the expectation is that we *might* add this sub-source functionality to WebGL 1.0 later.

PR #2147, opened 2016-11-15
Fixed tests uploading canvases to textures via texSubImage3D.

kenrussell:
These were using an obsolete function signature.


kenrussell:
Found necessary while testing Chromium CL https://codereview.chromium.org/2495953002/ .

Merging myself. @zhenyao @kainino0x @jdashg

zhenyao:
LGTM after the fact

PR #2149, opened 2016-11-16
Clarify texture uploading from sub source spec.

zhenyao:
@jdashg Please review.

@kenrussell @kainino0x already took a look.


kenrussell:
I'm merging this now. It only clarifies behavior, is non-controversial, and #2158 updates the tests to follow this behavior.

PR #2150, opened 2016-11-16
test tex(Sub)Image2D/3D with HTMLVideoElement

kainino0x:
@kenrussell @zhenyao @jdashg 

These do NOT conform with the changes per the discussion we had today. Those changes will probably be made wholesale throughout the tests at once.


kenrussell:
Excellent work. These look perfect.

PR #2151, opened 2016-11-16
Fix gl uniform arrays

Richard-Yunchao:
The variable "positionLoc" is undefined. Chrome (and other browsers) doesn't report error because a variable with exactly the same name is defined in the following code by chance. 

PTAL when you have time.


zhenyao:
lgtm
thanks for catching this bug

PR #2153, opened 2016-11-16
Skip some tests due to driver bug

qkmiao:
Mac OSX with Intel GPU has bug on LINEAR filtering mode, see
crbug.com/656478. Skip the failed tests.


zhenyao:
Let's wait until we take a 2.0.0 snapshot, and disable them in the snapshot so we could pass conformance, but let the top of tree still runs them (at least on platforms where they can pass)

qkmiao:
Make sense:)

qkmiao:
Skip these tests in WebGL 2.0.0 snapshot now. Please review again. Thanks.

zhenyao:
@kainino0x Can you review this?  I know you are disabling all tests affected by Mac driver bugs, so this may conflict with your pull requests?

kainino0x:
ah, didn't realize this PR existed. I did the same thing in my PR (#2201) except I wrote tighter wildcard matches. So I would prefer to just close this and merge mine when it's ready.

qkmiao:
Thanks @kainino0x for fixing this.

PR #2154, opened 2016-11-16
Work around #1819 in conformance2/textures/image_data/tex-2d-*_integer-unsigned_byte.html

kainino0x:
Verified against Chrome on MacBook Air, Intel HD 4000, OS X 10.11.4.


zhenyao:
lgtm

kenrussell:
Excellent. LGTM after the fact.

PR #2156, opened 2016-11-17
Put missing image related tests to webgl 2.0.1

Richard-Yunchao:
Could we put them into WebGL 2.0.1?  I will fix them later by https://codereview.chromium.org/2496813002/ after Chromium M56. 

@zhenyao and @kenrussell .

### DIFF
--- a/sdk/tests/conformance2/renderbuffers/00_test_list.txt
+++ b/sdk/tests/conformance2/renderbuffers/00_test_list.txt
@@ -1,5 +1,5 @@
 framebuffer-object-attachment.html
-framebuffer-test.html
+--min-version 2.0.1 framebuffer-test.html
 framebuffer-texture-layer.html
 invalidate-framebuffer.html
 multisampled-renderbuffer-initialization.html
--- a/sdk/tests/conformance2/rendering/00_test_list.txt
+++ b/sdk/tests/conformance2/rendering/00_test_list.txt
@@ -23,7 +23,7 @@ draw-buffers.html
 --min-version 2.0.1 draw-with-integer-texture-base-level.html
 element-index-uint.html
 --min-version 2.0.1 framebuffer-completeness-draw-framebuffer.html
-framebuffer-completeness-unaffected.html
+--min-version 2.0.1 framebuffer-completeness-unaffected.html
 --min-version 2.0.1 framebuffer-texture-changing-base-level.html
 --min-version 2.0.1 framebuffer-texture-level1.html
 framebuffer-unsupported.html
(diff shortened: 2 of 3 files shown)


zhenyao:
Yes, that is the plan. Let's do it in the 2.0.0 snapshot though.  Currently Chromium bots run with top of tree 2.0.0, so if we take them out in the trunk and chromium does a roll, we lose these tests on all platforms.

kenrussell:
+1, please hold off touching the skip lists and 00_test_list.txt files until we've taken the 2.0.0 snapshot. It shouldn't be long. Thanks.

jdashg:
We took a 2.0.0 snapshot long ago. @Richard-Yunchao please rebase.

Richard-Yunchao:
@jdashg , done.

conformance-suites/2.0.0/conformance2/renderbuffers/00_test_list.txt
```diff
@@ -1,5 +1,5 @@
 framebuffer-object-attachment.html
-framebuffer-test.html
```
jdashg: Leave the 2.0.0 snapshot alone, actually.

Richard-Yunchao: Done.

Richard-Yunchao:
Shall we remove all "--min-version 2.0.1 TESTCASE" stuff in 2.0.0 snapshot. Per my understanding, that tests should not be tested in 2.0.0.

kenrussell:
Let's please not make lots of changes to the 2.0.0 snapshot. There is probably useful history there about when certain tests were added, which we might want to consult later.

jdashg:
Wait, this *adds* to the skip-list.
Since we took the snapshot with these tests left in, I think we should leave it this way.
Is there any reason to add these to the skip-list at this point?
@Richard-Yunchao @kenrussell

Richard-Yunchao:
This change was submitted long ago. Some tests fail because of driver issue and it is difficult to workaround in browser, so I add them into the skip-list. I revisited the chromium issue I reported at that time. That bug is not fixed yet.

jdashg:
Ok, let's see what @kenrussell would like to do here.
I suspect we just want to leave these tests in at this point, for both tip and 2.0.0.

kenrussell:
It turns out that Chromium has been suppressing failures of the entire deqp/functional/gles3/fbocompleteness.html file for some time:

https://chromium.googlesource.com/chromium/src/+/f86769a5f7318d66a0f87e3620f9e05459bd3747/content/test/gpu/gpu_tests/webgl2_conformance_expectations.py#501

so I think it is better to suppress just the two cases of this test which are failing due to driver bugs on macOS, and turn on the rest of the test on this and other platforms, as this pull request does.

On the whole it seems OK to make it more explicit that we intend to pass these tests on 2.0.1. If we can't motivate Apple to fix the driver bugs that continue to affect framebuffer-test.html and framebuffer-completeness-unaffected.html then we'll have to comment out the failing cases from these tests.

Sorry for the long delay looking at this again.

jdashg:
This also seems workaround-able to me. Do we still want this? @kenrussell @Richard-Yunchao

Richard-Yunchao:
@jdashg , @kenrussell , @jdarpinian , if you think that it is unnecessary, I am OK to close this pr, or you can close it directly.

jdashg:
Ok, I'll close this and we can reopen it if needed.

PR #2157, opened 2016-11-17
Add sub source test cases for texture uploading from ImageBitmap

zhenyao:
@kenrussell @kainino0x @jdashg


kenrussell:
Looks perfect. Thank you Mo for putting these together.

PR #2158, opened 2016-11-17
Fixed incorrect assumptions about behavior of UNPACK_IMAGE_HEIGHT

kenrussell:
pixel storage parameter in tests uploading HTML data sources to 3D
textures.

Added tests skipping some rows via UNPACK_IMAGE_HEIGHT for most data
sources. Left a TODO to do this for ImageData since the test was more
complex.

Tested with Chromium implementation in
https://codereview.chromium.org/2511153002 ; all tests pass.


kenrussell:

zhenyao:
lgtm

PR #2160, opened 2016-11-18
add test cases for webgl-draw-buffers-ext to detect feedback loop

Richard-Yunchao:
Copy the rendering_sampling_feedback_loop.html for drawBuffers API in WebGL 2 to WebGL_draw_buffers extension in WebGL 1, for the sake of this issue: https://bugs.chromium.org/p/angleproject/issues/detail?id=1619.

I have a fix for this issue in Chromium, but that fix need to do some optimization. I think I have time to do the optimization soon.   

PTAL when you have time, @zhenyao and @kenrussell .


zhenyao:
LGTM
I'll merge once @kenrussell takes the 2.0.0 snapshot

Also, I don't think we tested depth/stencil texture feedback loops and it would be nice to have it covered.

Richard-Yunchao:
Thanks for merging, Zhenyao. 

You are correct, I will add depth/stencil feedback loop detection in WebGL 1.0.

PR #2161, opened 2016-11-18
GREEN, BLUE, and NUM_SAMPLE_COUNTS enums were removed from WebGL2 core.

jdashg:
They were removed from the IDL a while ago, but not removed from the tests.


kenrussell:
Thanks for catching this. Looks good.

PR #2162, opened 2016-11-19
Snapshot version 2.0.0 of WebGL conformance suite.

kenrussell:
- Copy sdk/tests/ recursively into conformance-suites/2.0.0 .
 - Edit webgl-conformance-tests.html to pin it to version 2.0.0.


kenrussell:
@zhenyao @jdashg @kainino0x @Richard-Yunchao @qkmiao FYI.

There's not much point in reviewing this since it'll look like a ton of new files. The steps that were taken to make the snapshot are listed above.

We can always edit this snapshot after the fact to correct mistakes -- and we'll need to in order to disable tests currently affected by graphics driver bugs.

It looks like some whitespace fixes are needed, so I'll make those before merging.

kenrussell:
Lint is clean now; merging.

Richard-Yunchao:
Thanks for the info, Ken. I am really glad to see that you finally take the 2.0.0 snapshot!

PR #2164, opened 2016-11-21
add more types into wtu.getBytesPerComponent

Richard-Yunchao:
These types are webgl2 types for texture/renderbuffer. 

When I use wtu.fillTexture to fill data to texture, I found that the original code of wtu.getBytesPerComponent will report type not supported, which is not correct. 

PTAL, @zhenyao and @kenrussell . Thanks a lot!

### DIFF
--- a/sdk/tests/js/webgl-test-utils.js
+++ b/sdk/tests/js/webgl-test-utils.js
@@ -739,7 +739,16 @@ var glTypeToTypedArrayType = function(gl, type) {
     case gl.INT:
       return window.Int32Array;
     case gl.UNSIGNED_INT:
+    case gl.UNSIGNED_INT_5_9_9_9_REV:
+    case gl.UNSIGNED_INT_10F_11F_11F_REV:
+    case gl.UNSIGNED_INT_2_10_10_10_REV:
+    case gl.UNSIGNED_INT_24_8:
       return window.Uint32Array;
+    case gl.HALF_FLOAT:
+    case 0x8D61:  // HALF_FLOAT_OES
+      return window.Uint16Array;
+    case gl.FLOAT:
+      return window.Float32Array;
     default:
       throw 'unknown gl type ' + glEnumToString(gl, type);
   }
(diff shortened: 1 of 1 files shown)


sdk/tests/js/webgl-test-utils.js
```diff
@@ -739,7 +739,16 @@ var glTypeToTypedArrayType = function(gl, type) {
     case gl.INT:
       return window.Int32Array;
     case gl.UNSIGNED_INT:
+    case gl.UNSIGNED_INT_5_9_9_9_REV
+    case gl.UNSIGNED_INT_10F_11F_11F_REV:
+    case gl.UNSIGNED_INT_2_10_10_10_REV:
+    case gl.UNSIGNED_INT_24_8:
       return window.Uint32Array;
+    case gl.HALF_FLOAT:
```
Richard-Yunchao: We have no Float16Array, so I use Float32Array here for HALF_FLOAT.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -739,7 +739,16 @@ var glTypeToTypedArrayType = function(gl, type) {
...
+    case gl.FLOAT:
+      return window.Float32Array;
+    case gl.FLOAT_32_UNSIGNED_INT_24_8_REV:
```
Richard-Yunchao: Again, no combined typed array for both float and uint, so I use Float64Array...

sdk/tests/js/webgl-test-utils.js
zhenyao: We use Uint16Array for this I think

Richard-Yunchao: @zhenyao , you are correct. Done.

zhenyao: Sorry I didn't catch this in the first round, but you will want to add HALF_FLOAT_OES (since it requires an extension to access the enum, it's probably better to just use the number with a comment saying it's HALF_FLOAT_OES).

sdk/tests/js/webgl-test-utils.js
```diff
@@ -761,9 +769,15 @@ var getBytesPerComponent = function(gl, type) {
     case gl.UNSIGNED_SHORT_5_6_5:
     case gl.UNSIGNED_SHORT_4_4_4_4:
     case gl.UNSIGNED_SHORT_5_5_5_1:
+    case gl.HALF_FLOAT:
```
zhenyao: Here also, HALF_FLOAT_OES

sdk/tests/js/webgl-test-utils.js
Richard-Yunchao: Done, @zhenyao. The actual values of HALF_FLOAT and HALF_FLOAT_OES are different.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -744,7 +744,8 @@ var glTypeToTypedArrayType = function(gl, type) {
     case gl.UNSIGNED_INT_2_10_10_10_REV:
     case gl.UNSIGNED_INT_24_8:
       return window.Uint32Array;
-    case gl.HALF_FLOAT:
+    case gl.HALF_FLOAT:      // 0x140B
+    case gl.HALF_FLOAT_OES:  // 0x8D61
```
zhenyao: You should use 0x8D61 and put HALF_FLOAT_OES in the comment. HALF_FLOAT_OES is not part of WebGLRenderingContext.

Richard-Yunchao: Got it. It was done, @zhenyao

zhenyao:
lgtm

PR #2166, opened 2016-11-22
Update the WebGL 2 top-of-tree conformance version to 2.0.1 (beta).

kenrussell:
Note that this will enable several new tests, some of which are known
to not yet pass on all browsers and operating systems.


kenrussell:
@jdashg @zhenyao @kainino0x please review.

@qkmiao @Richard-Yunchao FYI.

We'll need to add several more suppressions in Chromium to handle this change, so if there are any concerns, please raise them. Thanks.

kainino0x:
LGTM

jdashg:
Should be fine.

PR #2168, opened 2016-11-22
Fix a typo in es3fFboInvalidateTest.js

Jiawei-Shao:


kenrussell:
Yikes! Good catch @Jiawei-Shao . What is the impact on these tests?

In this pull request, could you please make the same change under conformance-suites/2.0.0/ ? The fix is needed there too. Thanks.

Jiawei-Shao:
The fix makes the background cleared correctly and won't affect the result.

kenrussell:
Thanks again for this fix. Looks good.

PR #2170, opened 2016-11-22
skip the integer cube_map texture related tests

Richard-Yunchao:
Fix a small bug. I suppose that the original code want to skip these tests. 

PTAL. Thanks a lot!


kenrussell:
Yikes! Good catch @Richard-Yunchao .

In this pull request, could you please make the same change under conformance-suites/2.0.0/ ? The same fix is needed there. Thanks.

@kainino0x FYI -- may explain some test failures.

Richard-Yunchao:
@kenrussell , done.

kenrussell:
Thanks @Richard-Yunchao . Looks good.

PR #2171, opened 2016-11-22
Remove checkSourceCanvasImageData for texImage2D tests

qkmiao:
getImageData() will disable hardware acceleration for canvas2d in chrome.
So, we cannot tests GPU-to-GPU path for texImage2D taking a canvas as source.
If we need to check if pixel values in source canvas change after canvas
used as webgl texture, we can add a standalone test for it.

See GetImageDataForcesNoAcceleration flag in blink:
https://cs.chromium.org/chromium/src/third_party/WebKit/Source/platform/graphics/ExpensiveCanvasHeuristicParameters.h?rcl=0&l=68.

### DIFF
--- a/sdk/tests/conformance2/textures/misc/00_test_list.txt
+++ b/sdk/tests/conformance2/textures/misc/00_test_list.txt
@@ -1,4 +1,5 @@
 --min-version 2.0.1 active-3d-texture-bug.html
+canvas-remains-unchanged-after-used-in-webgl-texture.html
 copy-texture-image.html
 copy-texture-image-luma-format.html
 copy-texture-image-webgl-specific.html
--- a/sdk/tests/conformance2/textures/misc/canvas-remains-unchanged-after-used-in-webgl-texture.html
+++ b/sdk/tests/conformance2/textures/misc/canvas-remains-unchanged-after-used-in-webgl-texture.html
@@ -1,0 +1,94 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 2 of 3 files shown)


kenrussell:
Thanks for pointing this out @qkmiao . I'm sure it was not intended that hardware acceleration for these canvases was disabled, but this was definitely added for a reason, as a regression test. If you would like to remove it from this test please add a separate test verifying this. (It doesn't have to cover all texture format and internalformat combinations -- RGBA/UNSIGNED_BYTE is fine.)

CC'ing @junov and @xidachen because I vaguely remember one of them adding this part of this test.

qkmiao:
@xlai-o added this code. 
I added a separate test to verify it now.

xidachen:
@qkmiao @kenrussell I think I know the reason why GPU acceleration is disabled. Chromium has several heuristics to disable GPU acceleration. In this case, we believe that if a page does getImageData once, it is more likely to do that again. Now as you know, getImageData requires GPU readback if GPU acceleration is enabled. Doing readback again and again is super expensive, and we just turn it off if we see getImageData once. I think for you case, the solution would be adding a new heuristic, such as "WebGLConformanceTesting", and we just keep it enabled always. Does that make sense?

xidachen:
@qkmiao @kenrussell Please ignore my previous comment, I didn't realize that there are changes being made to the tests. I think the change looks good. With this change, the test uses "tex-image-and-sub-image-2d-with-canvas.js" will still be able to verify that TexImage2D taking a canvas element works properly with different kinds of dest texture format. And doing checkSourceCanvasImageData in a separate test makes sense to me.

zhenyao:
Per @xidachen comment above, merging this.

sdk/tests/conformance2/textures/misc/00_test_list.txt
```diff
@@ -1,4 +1,5 @@
 --min-version 2.0.1 active-3d-texture-bug.html
+canvas-remains-unchanged-after-used-in-webgl-texture.html
```
kenrussell: Note: since this was added after the 2.0.0 snapshot was taken, for bookkeeping purposes it should say --min-version 2.0.1. Not a big deal.

PR #2172, opened 2016-11-22
Add ASTC conformance test

DavidPeicho:
Adds conformance test for the ASTC texture compression format.
The file `webgl-compressed-texture-astc.html` tests several compression size on LDR encoded and HDR encoded textures.

Tested on Ubuntu 16.04 with Chrome (Version 54.0.2840.100).
The tests ran on the Intel Core i7-6700HQ integrated gpu.


kenrussell:
Thank you for this contribution @DavidPeicho !

On which browser/OS/GPU did you verify this new test? Unfortunately due to the US holidays I probably won't be in front of a machine supporting ASTC to verify it until next Monday the 28th.

Also, how was the test data generated?

Thanks again!

DavidPeicho:
Sorry @kenrussell forgot to write information about the machine running the tests.
The machine is running Ubuntu 16.04, and the tests ran on the Intel Core i7-6700HQ integrated gpu.
Only Chrome (Version 54.0.2840.100) has been used for the tests.

The data has been generated using [astc-encoder](https://github.com/ARM-software/astc-encoder) from ARM.

kenrussell:
Thanks again @DavidPeicho for supplying this test and I'm sorry for taking so long to review it.

On a Google Pixel phone, several of the formats pass. The ones with higher compression ratios fail:

LDR:
COMPRESSED_RGBA_ASTC_8x8_KHR (needs increased tolerance -- like 30)
COMPRESSED_RGBA_ASTC_10x8_KHR
COMPRESSED_RGBA_ASTC_10x10_KHR
COMPRESSED_RGBA_ASTC_12x10_KHR
COMPRESSED_RGBA_ASTC_12x12_KHR

HDR:
COMPRESSED_RGBA_ASTC_10x8_KHR (needs increased tolerance -- like 40)
COMPRESSED_RGBA_ASTC_10x10_KHR (would need hugely increased tolerance -- over 100)
COMPRESSED_RGBA_ASTC_12x10_KHR (would need hugely increased tolerance -- over 100)
COMPRESSED_RGBA_ASTC_12x12_KHR (would need hugely increased tolerance -- over 100)

The problem with the test as written is that it expects the compressed texture to be comparable to the uncompressed data. This is an incorrect assumption; it should be comparing against the output of "astcenc -d". This is one reason the test fails at the higher block sizes.

Could you please do the following:

1) Provide the source image you passed to astcenc, putting it in sdk/tests/resources/ .
2) For each format where you generated the compressed data, decompress that data and put it in webgl-compressed-texture-astc.html alongside the compressed data.
3) Compare the sampling of the ASTC compressed texture to the decompressed data for that format.

Could you also please document in a comment in webgl-compressed-texture-astc.html the exact astcenc command you ran (i.e., which quality level did you use?), and how you identified which part of the output file to extract?

Thanks. (I investigated doing this starting from your compressed data but it wasn't trivial -- would require writing the file headers by hand -- so since you clearly already have the setup to regenerate the data, I'd appreciate your help doing so.)

Thanks again.

DavidPeicho:
The image being really easy to encode, I did not think about cases close to 0.89 bpp, and, obviously, the difference it makes is important, it does not make sense to compare the decoded version with the original non-compressed image, my bad.

It should now be good @kenrussell, if there is anything else, let me know.

kenrussell:
Awesome! Thank you @DavidPeicho for this superb test. It passes 100% on a conformant ASTC implementation (verified on a Pixel phone). Merging now!

PR #2174, opened 2016-11-24
Cover ANGLE bug with large UBO readback.

null77:
ANGLE only stores starting chunks of large UBOs because of Win7
D3D11 limitations. This was causing problems when reading back
from large UBOs in some cases.

See also http://crbug.com/660670

@zhenyao @kenrussell PTAL. Issue was a bit old but was on my action items list to make a test.


kenrussell:
Very nice. Thanks Jamie for improving the tests.

Merging now. Unless you feel strongly, let's leave this new test out of the 2.0.0 snapshot which was taken a few days ago.

null77:
No, leaving it out of the snapshot is the best, thanks!

PR #2175, opened 2016-11-25
Check TexSubImage3D works with pixel buffer

qkmiao:
TexSubImage3D is buggy when update texture data from pixel buffer on
Linux Mesa. Add this test to track the bug.


zhenyao:
LGTM

PR #2176, opened 2016-11-25
Address unspecified behavior in the WEBGL_multiview spec

Oletus:
The native extension spec recommends an error in case the number of
views doesn't match between the framebuffer and the shader. This
should be required behavior in the WebGL version of the extension.

When the number of views specified in the active program is one,
it needs to be specified what happens with gl_ViewID_OVR. The simplest
option is to specify that it always evaluates to zero in this case.
Another alternative would be to forbid statically using gl_ViewID_OVR
in this case, but this could only be validated at link time.

### DIFF
--- a/extensions/proposals/WEBGL_multiview/extension.xml
+++ b/extensions/proposals/WEBGL_multiview/extension.xml
@@ -185,6 +185,12 @@
       <addendum>
         Attempting to enable both <code>GL_OVR_multiview</code> and <code>GL_OVR_multiview2</code> in the same shader results in a compile error.
       </addendum>
+      <addendum>
+        Attempting to draw when the number of views in the draw framebuffer does not match the number of views specified in the active program generates an <code>INVALID_OPERATION</code> error.
+      </addendum>
+      <addendum>
+        When the number of views specified in the active program is one, <code>gl_ViewID_OVR</code> will always evaluate to zero.
+      </addendum>
     </mirrors>
     <features>
       <feature>
@@ -223,6 +229,10 @@ interface WEBGL_multiview {
   <history>
     <revision date="2016/11/11">
       <change>Initial revision.</change>
+    </revision>
+
+    <revision date="2016/11/25">
+      <change>Specified what happens when the number of views doesn't match or if the number of views is one.</change>
     </revision>
   </history>
 </proposal>


zhenyao:
This change seems fine. Let's merge this.

zhenyao:
While reviewing this change, I went through the entire spec proposal (for the first time).  I am a bit confused by the structure of the extension.  It seems to me the first half and the second half are dis-integrated.  For the first half, we add a context creation attrib to allow it to be stereo and use drawBuffer() to decide which one to draw, but not both. For the second half, it's OVR_multiview, which attach a texture array to a framebuffer and render to multiple ones at the same time.

So that's my understanding.  It seems to me that's actually two different extensions.  That's where I am confused.  My apology if you already explained this before (I was focusing on shipping WebGL2 so I wasn't paying attention to anything else). If that's the case, could you please point me to the discussion? Otherwise, can you clarify a bit?

Oletus:
Yep, the first half of the extension document contains proposed changes to the core spec - the core spec needs to be changed so that the default framebuffer can be made into a stereo framebuffer. The core spec changes are intended to be merged to the core WebGL specification and removed from the extension spec later. The second half contains the parts that are proposed to be made into an extension that would be accessed through the normal extension API.

zhenyao:
What I am confused about is the discrepancy between the two.  The core part seems disallow rendering into two views at the same time, whereas the OVR_multiview part seems to do exactly that (rendering into multiple views in one draw call).

shahmeeresmail:
If the spec itself is being changed to allow for drawbuffers to specify left or right, why not go one step further and allow both left and right to be specified, similar to OpenGL but of course with some caveats ( no transform feedback, queries etc)? Each webgl implementation could then have the option of executing 2 stand alone draws or using command list playback or instancing with RT redirect for each eye or some other method for drawing to both views simultaneously.

Oletus:
zhenyao: That's the intent. The core part just adds the two views/buffers and makes it possible to render into either one of them. The extension makes rendering into both more efficient. Maybe different wording could make things clearer? Another alternative would be to leave the drawBuffer entry point out of the core spec, and only allow rendering to the right buffer using multiview, but I think it would be quite odd if the core spec allowed you to create a buffer which is inaccessible within the core spec. Having drawBuffer in core also means that developers have baseline stereo functionality they can rely on.

shahmeeresmail: The core spec itself is changed in a minimal way mostly because changing the default framebuffer doesn't really fit into an extension. I'm not quite sure what kind of multiview rendering you're proposing to include to the core spec on top of that, but I think the best approach is to keep the core spec changes to a minimum. The extension is targeted only at enabling higher performance, so it is reasonable to keep it optional to implement for current WebGL versions.

zhenyao:
I see.  I agree drawBuffer() should be part of core if an attribute is added to allow two views in the core. My perspective is I feel drawBuffer() is disconnected with the OVR_multiview.  Maybe it's cleaner we define WebGL_multiview which differ from the OVR_multiview by simply adding LEFT_AND_RIGHT (something like that) to drawBuffer()? Implementations can use the OVR_multiview behind the scene to implement it.  It feels to me cleaner and more integrated semantic. Just some thoughts.

Oletus:
The current WEBGL_multiview proposal does add LEFT_AND_RIGHT to drawBuffer(), but since there are other use cases for WEBGL_multiview besides stereo, keeping the functionality for rendering to array textures from OVR_multiview makes sense as well. Multiview could be used to render a cube map, for example.

shahmeeresmail:
I agree with @zhenyao that drawBuffer could implement simultaneous LR rendering behind the scenes.

@Oletus, what I mean is that the OpenGL spec for drawBuffers today states that if GL_BACK is set, that left and right buffers are set as well. What I am saying is to make GL_BACK part of the WebGL spec in the same way that it is part of the OpenGL spec. Yes, it would be an extra change to the core webgl spec but it would still be very similar to how other Khronos graphics specs are defined today.  

Essentially, if we're going to add drawbuffers to the spec, we shouldn't change it to a new beast in the process. Keep it as close to the original spec as it is defined for other Khronos apis. The addition of LEFT_AND_RIGHT is potentially confusing to those who may be using drawbuffers on OpenGL or elsewhere.

Oletus:
Right, maybe adding GL_BACK to the core spec in some way would be reasonable. It might have to be slightly different in WebGL compared to OpenGL, and it does require a fair amount of spec work, but the spec might end up cleaner in the end. How I'd do it is:

In the core spec, if drawBuffer(gl.BACK) is set, drawing commands have the same effect as:

drawBuffer(gl.BACK_LEFT);
<drawing-command>
drawBuffer(gl.BACK_RIGHT);
<drawing-command>

with the exception that transform feedback is not allowed and occlusion queries work as in OVR_multiview.

And if you use a multiview shader from the extension, it basically adds gl_ViewID_OVR, so that drawing commands when drawBuffer(gl.BACK) is set would have the same effect as:

drawBuffer(gl.BACK_LEFT);
View(0);
<drawing-command>
drawBuffer(gl.BACK_RIGHT);
View(1);
<drawing-command>

@shahmeeresmail, is this what you're after? @zhenyao, I'm not sure if this is in line with what you meant by your suggestions, but would this make the spec cleaner in your opinion as well?

zhenyao:
Actually I would suggest leaving gl.BACK to the extension instead of core. The reason is that if OVR_multiview isn't available, we won't be able to implement gl.BACK efficiently (we end up emulating it by draw twice). So that sounds like an extension more than a core function to me.  Otherwise it sounds like a good plan.

shahmeeresmail:
@Oletus Yes, that is what I was thinking :)

@zhenyao, even OVR_multiview may draw twice as part of a compatibility mode or fallback for hardware and drivers that do not fully support other more advanced mechanisms for drawing to two targets in a single draw call or if there are bugs in those implementations.

msc-:
> On 2016 Dec 8, at 8:47, Shahmeer Esmail <notifications@github.com> wrote:
> 
> @Oletus <https://github.com/Oletus> Yes, that is what I was thinking :)
> 
> @zhenyao <https://github.com/zhenyao>, even OVR_multiview may draw twice as part of a compatibility mode or fallback for hardware and drivers that do not fully support other more advanced mechanisms for drawing to two targets in a single draw call or if there are bugs in those implementations.
> 

Why not just require that WEBGL_multiview be supported if the browser provides a stereo canvas. The browser can provide the functionality by drawing twice if the underlying hardware doesn’t provide anything smarter. It won’t be any worse than the application drawing twice.

Regards

    -Mark

msc-:
> Why not just require that WEBGL_multiview be supported if the browser provides a stereo canvas?

Any reactions to this idea?

Oletus:
@msc- We didn't have time to discuss the open questions in the last WebGL WG call. Expect more discussion after the holidays.

PR #2177, opened 2016-11-29
Updated bad-arguments-test for new non-nullable entry points.

### LINKED ISSUE #2173: The test harness hides errors in asynchronous tests.
Some tests are asynchronous which makes the call to notifyFinishedToHarness race with the actual completion of the test. While running the test in its own page works fine, when running inside the harness any call to assert() will trigger an exception that stops the test. This combined with the harness thinking the test is done already will ignore the failures inside asynchronous code.

An example of this is with conformance/misc/bad-arguments-test.html that passes with 1/1 in the harness but fails when run directly due to the recent changes to the optional objects in the WebGL IDL.

kenrussell:
Follow-on to #2099.

Also fixes #2173; race condition in this asynchronous test.

### DIFF
--- a/conformance-suites/2.0.0/conformance/misc/bad-arguments-test.html
+++ b/conformance-suites/2.0.0/conformance/misc/bad-arguments-test.html
@@ -45,13 +45,13 @@ description("Tests calling WebGL APIs with wrong argument types");
 
 var testArguments = [
   { value: "foo",
-    throws: true },
+    throwsForNullables: true },
   { value: 0,
-    throws: true },
+    throwsForNullables: true },
   { value: null,
-    throws: false },
+    throwsForNullables: false },
   { value: undefined,
-    throws: false }
+    throwsForNullables: false }
 ];
 
 var argument;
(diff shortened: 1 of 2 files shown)


kenrussell:
@zhenyao @jdashg @kainino0x please review.

@Kangz thanks for catching this. Also feel free to review.

Verified the updated test with Chrome.

kenrussell:
Also verified in Firefox nightly.

conformance-suites/2.0.0/conformance/misc/bad-arguments-test.html
```diff
@@ -99,23 +111,13 @@
...
-      func2("context.getUniform(program, argument)");
-      func2("context.getUniformLocation(argument, 'u_modelViewProjMatrix')");
-
+      func("context.useProgram(argument)");
     }
+    finishTest();
   });
 });
 
 var successfullyParsed = true;
```
zhenyao: This line should be removed also.

sdk/tests/conformance/misc/bad-arguments-test.html
```diff
@@ -99,23 +111,13 @@
...
-      func2("context.getUniform(program, argument)");
-      func2("context.getUniformLocation(argument, 'u_modelViewProjMatrix')");
-
+      func("context.useProgram(argument)");
     }
+    finishTest();
   });
 });
 
 var successfullyParsed = true;
```
zhenyao: Same here

conformance-suites/2.0.0/conformance/misc/bad-arguments-test.html
kenrussell: Fixing.

sdk/tests/conformance/misc/bad-arguments-test.html
kenrussell: Thanks, fixing.

PR #2178, opened 2016-11-29
Clarify deletion/mark for deletion behavior.

jdashg:


specs/latest/1.0/index.html
```diff
@@ -453,13 +453,13 @@ <h2 class="no-toc">Table of contents</h2>
...
-        by these objects are guaranteed to exist as long as the object exists.
-        Furthermore, the DOM object is guaranteed to exist as long as the author has an explicit
-        valid reference to it <b>or</b> as long as it is bound by the underlying graphics library.
-        When none of these conditions exist the user agent can, at any point, delete the object
-        using the equivalent of a delete call (e.g., deleteTexture). If authors wish to control
-        when the underlying resource is released then the delete call can be made explicitly.
+        stored in these objects and are fully managed by them.
+        The DOM object is guaranteed to exist at least as long as the author has an explicit
+        valid reference to it <b>and</b> at least as long as it is bound by the underlying graphics
+        library.
```
kenrussell: This update is incorrect. For example, calling bindTexture keeps the associated WebGLTexture object alive, even if the author releases all of their references to it, because they can call for example getParameter(TEXTURE_BINDING_2D) and get the reference back. The expando-loss conformance tests verify these behaviors.

The conjunction should be "or".

kenrussell:
Note that this is somewhat related to #2002 and follow-on pull requests. Referencing it.

specs/latest/1.0/index.html
jdashg: "and" most closely matches my understanding, and does not seem wrong based on your explanation. Can you rephrase?

kenrussell: Re-reading, I see your point, but think the original text was more clear. Can I suggest:

The DOM object is guaranteed to exist while either the author has an explicit valid reference to it, or it is bound by the underlying graphics library.

sdk/tests/conformance/extensions/oes-vertex-array-object.html
```diff
@@ -624,25 +624,31 @@
         gl.deleteBuffer(colorBuffer);
         gl.deleteBuffer(positionBuffer);
 
-        // The buffers should not be accessible at this point. Deleted objects that are bound
-        // in the current context undergo an automatic unbinding
+        var expectRetained = (ii != 0);
```
kenrussell: Could you adjust the comment above rather than removing it? Maybe point out that the deleteBuffer calls are no-ops after the first iteration, so they are only automatically unbound from vaos[0], or something similar.

sdk/tests/conformance/extensions/oes-vertex-array-object.html
```diff
@@ -624,25 +624,31 @@
...
-        var isPositionBuffer = gl.isBuffer(positionBuffer);
-        var isColorBuffer    = gl.isBuffer(colorBuffer);
+        var expectedError = (expectRetained ? gl.NO_ERROR : gl.INVALID_OPERATION);
+        wtu.glErrorShouldBe(gl, expectedError,
+                            "Draw call should " + (expectRetained ? "not " : "") + "fail.");
 
-        if(isPositionBuffer)  testFailed("Position buffer should no longer exist after last ref removed");
-        if(isColorBuffer)     testFailed("Color buffer should no longer exist after last ref removed");
+        if (!gl.isBuffer(positionBuffer)) {
+            testFailed("Position buffer should count for isBuffer.");
```
kenrussell: Could you rewrite this? Something like:

"The references from the unbound VAOs should keep positionBuffer's buffer object alive"

or similar? Same for below.

kenrussell:
Looks good modulo updating the older conformance suite snapshots. It would be ideal to update them all in one pull request; it's been difficult to track these down after the fact.

jdashg:
@kenrussell Updated.

zhenyao:
Jeff, the 2.0.0 snapshot changes make chrome fail these tests.  Could you please change the 2.0.0 snapshot changes so that it allows both the old and the new behavior?  Otherwise Chrome will not be able to pass 2.0.0 in M56.

jdashg:
@kenrussell I removed the subtests from the 2.0.0 snapshot, since they'd need to accept both 'true' and 'false'.

Good to merge?

sdk/tests/conformance/extensions/oes-vertex-array-object.html
```diff
@@ -624,25 +624,33 @@
...
+                            "Draw call should " + (expectRetained ? "not " : "") + "fail.");
 
-        if(isPositionBuffer)  testFailed("Position buffer should no longer exist after last ref removed");
-        if(isColorBuffer)     testFailed("Color buffer should no longer exist after last ref removed");
+        if (!gl.isBuffer(positionBuffer)) {
+            testFailed("References from unbound VAOs keep Position buffer alive.");
+        }
+        if (!gl.isBuffer(colorBuffer)) {
+            testFailed("References from unbound VAOs keep Color buffer alive");
+        }
```
kenrussell: Without having implemented this in Chrome, I think there may be a problem with having isBuffer return a different answer for a deleted buffer depending on whether it's been detached from all unbound VAOs. The basic problem is that it seems work needs to be done on the WebGLBuffer when GC'ing the VAO, but this is forbidden in Chrome's implementation because of how DOM objects are GC'd. Still, based on my earlier review, and your thinking through this change, let's move forward with it and reconsider if there are problems in other WebGL implementations.

zhenyao: @jdashg I agree with the spec text change, but I disagree with isBuffer() behavior.  From ES spec 3.0, D.1.3 Deleted Object and Object Name Lifetimes, after calling glDeleteBuffers(), the names become invalid immediately. So isBuffer() should really return false whether the underlying object delete happens immediately or delayed.

kenrussell: Thanks @zhenyao for considering this carefully. @jdashg , sorry I haven't yet tried implementing this in Chrome, but I have a feeling that the new semantics being required by this test can't work in that browser. Due to the GC algorithm built into the DOM, Chrome can't increment a reference count on the WebGLBuffers when they're attached to a VAO and decrement it when they're detached -- the reference count can't be managed correctly when the container object is GC'd.

I've filed #2456 to reconsider this part of the test.

Issue #2143, opened 2016-11-11
New bufferData/bufferSubData overloads need some prose defining what they actually do


bzbarsky:
WebGL2 adds a new overload of `bufferData` that looks like this:

    void bufferData(GLenum target, ArrayBufferView srcData, GLenum usage,
                              GLuint srcOffset, optional GLuint length = 0);

but doesn't define the behavior anywhere.  The usual "just do what GL ES 3.0 does" thing doesn't work here, because GL ES 3.0 has a single glBufferData which as far as I can tell just takes a pointer to the data and how many bytes to grab.  That has no equivalent for the above overload....  In particular, something needs to define what "srcOffset" and "length" actually mean here.

Similar problem for `bufferSubData`.

Please make sure to define what "srcOffset" and "length" mean here when srcData is a DataView, not a typed array.

kenrussell:
Fixed in above pull request. A follow-on one will address minor concerns with it.

PR #2179, opened 2016-11-30
Document bufferData and bufferSubData overloads taking srcOffset and …

kenrussell:
…length.

The new spec text documents the behavior when DataView is passed in. In order to unify the text, the documentation of getBufferSubData was revised to have a similar form to the other methods, and to also document the behavior when DataView is passed.

Addresses #2143.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -1156,6 +1156,77 @@ WebGL2RenderingContext implements WebGL2RenderingContextBase;
     <h4>Buffer objects</h4>
 
     <dl class="methods">
+
+      <dt class="idl-code">void bufferData(GLenum target, ArrayBufferView srcData, GLenum usage, GLuint srcOffset, optional GLuint length = 0);
+          <span class="gl-spec">
+          (<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.10.2">OpenGL ES 3.0.4 &sect;2.10.2</a>,
+           <a href="http://www.khronos.org/opengles/sdk/docs/man3/html/glBufferData.xhtml" class="nonnormative">man page</a>)
+          </span>
+      </dt>
+      <dd>
+        Set the size of the currently bound WebGLBuffer object, then copy a sub-region of <code>srcData</code> to the buffer object.
+        <br><br>
+        Let <code>buf</code> be the buffer bound to <code>target</code>.
+        <br><br>
+        If <code>length</code> is 0:
+        <ul>
+        <li> If <code>srcData</code> is a <code>DataView</code>, let <code>copyLength</code> be <code>srcData.byteLength - srcOffset</code>; the typed elements in the text below are bytes.
+        <li> Otherwise, let <code>copyLength</code> be <code>srcData.length - srcOffset</code>.
+        </ul>
+        Otherwise, let <code>copyLength</code> be <code>length</code>.
+        <br><br>
+        Set the size of <code>buf</code> to <code>copyLength * srcData.BYTES_PER_ELEMENT</code>, or <code>copyLength</code> in the case of <code>DataView</code>.
+        <br><br>
+        If <code>copyLength</code> is greater than zero, copy <code>copyLength</code> typed elements
+        (each of size <code>srcData.BYTES_PER_ELEMENT</code>, or 1 in the case
+        of <code>DataView</code>) from <code>srcData</code> into <code>buf</code>,
+        reading <code>srcData</code> starting at element index <code>srcOffset</code>.
+        If <code>copyLength</code> is 0, no data is written to <code>buf</code>, but this does
+        not cause a GL error to be generated.
+        <ul>
+          <li> If no WebGLBuffer is bound to <code>target</code>, generates an <code>INVALID_OPERATION</code> error.
+          <li> If <code>srcOffset + copyLength</code> is greater than <code>srcData.length</code> (or <code>srcData.byteLength</code> for <code>DataView</code>), generates an <code>INVALID_VALUE</code> error.
+        </ul>
+        If any error is generated, <code>buf</code>'s size is unmodified, and no data is written to it.
+      </dd>
+
+      <dt class="idl-code">void bufferSubData(GLenum target, GLintptr dstByteOffset, ArrayBufferView srcData, GLuint srcOffset, optional GLuint length = 0);
+          <span class="gl-spec">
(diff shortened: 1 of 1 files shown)


kenrussell:
@jdashg @bzbarsky @zhenyao @kainino0x please review.

It might be easiest to review by pulling this change into your local repository, and viewing the HTML page, rather than looking at the raw HTML in the review tool.

specs/latest/2.0/index.html
```diff
@@ -1156,6 +1156,77 @@ <h2 class="no-toc">Table of contents</h2>
...
+        <br><br>
+        If <code>copyLength</code> is greater than zero, copy <code>copyLength</code> typed elements
+        (each of size <code>srcData.BYTES_PER_ELEMENT</code>, or 1 in the case
+        of <code>DataView</code>) from <code>srcData</code> into <code>buf</code>,
+        reading <code>srcData</code> starting at element index <code>srcOffset</code>.
+        If <code>copyLength</code> is 0, no data is written to <code>buf</code>, but this does
+        not cause a GL error to be generated.
+        <ul>
+          <li> If no WebGLBuffer is bound to <code>target</code>, generates an <code>INVALID_OPERATION</code> error.
+          <li> If <code>srcOffset + copyLength</code> is greater than <code>srcData.length</code> (or <code>srcData.byteLength</code> for <code>DataView</code>), generates an <code>INVALID_VALUE</code> error.
```
zhenyao: This does not cover the error case where srcOffset > srcData.length and copyLength = srcData.length - srcOffset. In getBufferSubData we do, so we should add it here also.

specs/latest/2.0/index.html
```diff
@@ -1156,6 +1156,77 @@ <h2 class="no-toc">Table of contents</h2>
...
+        of <code>DataView</code>) from <code>srcData</code> into <code>buf</code>,
+        reading <code>srcData</code> starting at element index <code>srcOffset</code>, and
+        writing <code>buf</code> starting at byte offset <code>dstByteOffset</code>.
+        If <code>copyLength</code> is 0, no data is written to <code>buf</code>, but this does
+        not cause a GL error to be generated.
+        <ul>
+          <li> If no WebGLBuffer is bound to <code>target</code>, generates an <code>INVALID_OPERATION</code> error.
+          <li> If <code>dstByteOffset</code> is less than zero, generates an <code>INVALID_VALUE</code> error.
+          <li> If <code>dstByteOffset + copyByteLength</code> is greater than the size of <code>buf</code>, generates an <code>INVALID_VALUE</code> error.
+          <li> If <code>srcOffset + copyLength</code> is greater than <code>srcData.length</code> (or <code>srcData.byteLength</code> for <code>DataView</code>), generates an <code>INVALID_VALUE</code> error.
```
zhenyao: Same here, srcOffset might be greater than srcData.length and copyLength = srcData.length - srcOffset.

specs/latest/2.0/index.html
```diff
@@ -1156,6 +1156,77 @@ <h2 class="no-toc">Table of contents</h2>
...
+        Let <code>buf</code> be the buffer bound to <code>target</code>.
+        <br><br>
+        If <code>length</code> is 0:
+        <ul>
+        <li> If <code>srcData</code> is a <code>DataView</code>, let <code>copyLength</code> be <code>srcData.byteLength - srcOffset</code>; the typed elements in the text below are bytes.
+        <li> Otherwise, let <code>copyLength</code> be <code>srcData.length - srcOffset</code>.
+        </ul>
+        Otherwise, let <code>copyLength</code> be <code>length</code>.
+        <br><br>
+        Set the size of <code>buf</code> to <code>copyLength * srcData.BYTES_PER_ELEMENT</code>, or <code>copyLength</code> in the case of <code>DataView</code>.
```
bzbarsky: This part is kind of confusing, because it says to do one thing... but then oh, maybe another.  Would be better to make it explicit that if srcData is a DataView, then buf gets set to copyLength, else to copyLength * srcData.BYTES_PER_ELEMENT.

specs/latest/2.0/index.html
```diff
@@ -1156,6 +1156,77 @@ <h2 class="no-toc">Table of contents</h2>
...
+        <br><br>
+        If <code>copyLength</code> is greater than zero, copy <code>copyLength</code> typed elements
+        (each of size <code>srcData.BYTES_PER_ELEMENT</code>, or 1 in the case
```
bzbarsky: Again, this is confusing in the DataView case, because it looks like it still looks at its BYTES_PER_ELEMENT...

specs/latest/2.0/index.html
```diff
@@ -1156,6 +1156,77 @@ <h2 class="no-toc">Table of contents</h2>
...
+        If <code>copyLength</code> is 0, no data is written to <code>buf</code>, but this does
+        not cause a GL error to be generated.
+        <ul>
+          <li> If no WebGLBuffer is bound to <code>target</code>, generates an <code>INVALID_OPERATION</code> error.
+          <li> If <code>srcOffset + copyLength</code> is greater than <code>srcData.length</code> (or <code>srcData.byteLength</code> for <code>DataView</code>), generates an <code>INVALID_VALUE</code> error.
+        </ul>
+        If any error is generated, <code>buf</code>'s size is unmodified, and no data is written to it.
+      </dd>
+
+      <dt class="idl-code">void bufferSubData(GLenum target, GLintptr dstByteOffset, ArrayBufferView srcData, GLuint srcOffset, optional GLuint length = 0);
```
bzbarsky: Similar comments apply here.

specs/latest/2.0/index.html
kenrussell: Thanks for your feedback. I'll address this in my next pull request.

specs/latest/2.0/index.html
kenrussell: Sorry, but I don't know how to make this less confusing without duplicating a lot of text, which I'm unwilling to do. If you have a concrete suggestion about how to improve it please put up a pull request. Thanks.

specs/latest/2.0/index.html
kenrussell: Will fix the first of these two per above.

specs/latest/2.0/index.html
kenrussell: Same here, thanks, will address.

specs/latest/2.0/index.html
bzbarsky: Concrete suggestion: define a variable for the element size, and explicitly assign a value to it conditional on the thing you want to condition on.

I'm not going to have time to do a pull request here; too many other things on my plate right now.

PR #2183, opened 2016-12-05
Depth stencil feedback loop

Richard-Yunchao:
piman pointed out that there should be feedback loop detection for depth/stencil buffer, see https://codereview.chromium.org/2461973002/diff/40001/gpu/command_buffer/service/gles2_cmd_decoder.cc. 
Add test case to expose this feature.
Also see the comments at #2160. 

PTAL. @zhenyao

### DIFF
--- a/sdk/tests/conformance2/rendering/depth-stencil-feedback-loop.html
+++ b/sdk/tests/conformance2/rendering/depth-stencil-feedback-loop.html
@@ -1,0 +1,179 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/rendering/depth-stencil-feedback-loop.html
```diff
@@ -0,0 +1,179 @@
...
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "The test sampls from a image. The same image is used as depth buffer. But depth mask is false.");
+
+    gl.depthMask(gl.TRUE);
+    gl.disable(gl.DEPTH_TEST);
+    wtu.clearAndDrawUnitQuad(gl);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "The test samples from a image. The same image is used as depth buffer. But depth test is not enabled during rendering.");
+
+    // Test rendering and sampling feedback loop for stencil buffer
+    gl.bindTexture(gl.TEXTURE_2D, tex2);
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.DEPTH_STENCIL_ATTACHMENT, gl.TEXTURE_2D, tex2, 0);
```
zhenyao: It's better to only attach to STENCIL_ATTACHMENT, so you don't have to worry about depthMask and DEPTH_TEST at this point (although the test is still correct because you disabled DEPTH_TEST from before.

Richard-Yunchao: I tried to only attach to STENCIL_ATTACHMENT when I wrote this patch, but I can not found any format only for STENCIL_ATTACHMENT, although I am not 100% sure about this. You can refer to the Table 3.2 on page 112 for format/type/sized_internalformat, and Table 3.3 on page 113 for format/type/unsized_internalformat: https://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf.

Richard-Yunchao:
Thanks for merging, @zhenyao .

sdk/tests/conformance2/rendering/depth-stencil-feedback-loop.html
zhenyao: You can use a DEPTH_STENCIL format and only attach to STENCIL_ATTACHMENT.

PR #2184, opened 2016-12-05
Use wtu.startPlayingAndWaitForVideo in image-bitmap-from-video tests.

kenrussell:
Attempt to address flakiness reported in http://crbug.com/671209.

Verified this change in both Chrome and Firefox.


kenrussell:
@zhenyao @jdashg @kainino0x PTAL.

zhenyao:
lgtm

PR #2185, opened 2016-12-06
Add tolerance to pixel comparison in clear-srgb-color-buffer.html

qkmiao:
This should make the test pass on Linux AMD and Intel.


zhenyao:
lgtm if that makes the test pass

PR #2186, opened 2016-12-06
Print failure message to log for debugging

qkmiao:
With the log, it's efficient to debug chrome build bots failures.


sdk/tests/conformance2/textures/misc/tex-srgb-mipmap.html
```diff
@@ -176,7 +176,7 @@
                 Math.abs(cmp[i * 4 + 1] - ref[i * 4 + 1]) > tolerance ||
                 Math.abs(cmp[i * 4 + 2] - ref[i * 4 + 2]) > tolerance ||
                 Math.abs(cmp[i * 4 + 3] - ref[i * 4 + 3]) > tolerance) {
-                debug("Pixel " + i + ": expected (" +
+                testFailed("Pixel " + i + ": expected (" +
```
zhenyao: Will this pollute the log? because we could end up failing every pixel comparison and that's a lot of text in the log.

PR #2188, opened 2016-12-07
COLOR_ATTACHMENT1000 can be either INVALID_ENUM or _OPERATION.

jdashg:
GLES 3.0.5 p219-220, plus Table 4.6 on p206.

### DIFF
--- a/sdk/tests/conformance2/renderbuffers/invalidate-framebuffer.html
+++ b/sdk/tests/conformance2/renderbuffers/invalidate-framebuffer.html
@@ -109,10 +109,10 @@ if (!gl) {
   var maxColorAttachments = gl.getParameter(gl.MAX_COLOR_ATTACHMENTS);
   gl.invalidateSubFramebuffer(gl.READ_FRAMEBUFFER, [gl.COLOR_ATTACHMENT0 + maxColorAttachments], 5, 5, 10, 10);
   wtu.glErrorShouldBe(gl, [gl.INVALID_OPERATION, gl.INVALID_ENUM],
-      "calling invalidateSubFramebuffer to invalidate a COLOR_ATTACHMENT that exceeds MAX_COLOR_ATTACHMENT should generate INVALID_OPERATION.");
+      "calling invalidateSubFramebuffer to invalidate a COLOR_ATTACHMENT that exceeds MAX_COLOR_ATTACHMENT should generate INVALID_ENUM or INVALID_OPERATION.");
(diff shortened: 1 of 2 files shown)


kenrussell:
Looks good

Could you include the update to conformance-suites/2.0.0 in this patch too please?

qkmiao:
Will it be INVALID_ENUM when hitting "An INVALID_ENUM error is generated if attachments contains a value other than COLOR_ATTACHMENTm for m in the range [0, 31]."?

Do you know why range [0, 31] is specified here? We already have MAX_COLOR_ATTACHMENTS enum.

sdk/tests/conformance2/renderbuffers/invalidate-framebuffer.html
```diff
@@ -108,10 +108,10 @@
 
   var maxColorAttachments = gl.getParameter(gl.MAX_COLOR_ATTACHMENTS);
   gl.invalidateSubFramebuffer(gl.READ_FRAMEBUFFER, [gl.COLOR_ATTACHMENT0 + maxColorAttachments], 5, 5, 10, 10);
-  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+  wtu.glErrorShouldBe(gl, [gl.INVALID_OPERATION, gl.INVALID_ENUM],
       "calling invalidateSubFramebuffer to invalidate a COLOR_ATTACHMENT that exceeds MAX_COLOR_ATTACHMENT should generate INVALID_OPERATION.");
   gl.invalidateFramebuffer(gl.READ_FRAMEBUFFER, [gl.COLOR_ATTACHMENT0 + maxColorAttachments]);
-  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION,
+  wtu.glErrorShouldBe(gl, [gl.INVALID_OPERATION, gl.INVALID_ENUM],
       "calling invalidateFramebuffer to invalidate a COLOR_ATTACHMENT that exceeds MAX_COLOR_ATTACHMENT should generate INVALID_OPERATION.");
```
qkmiao: Same here.

jdashg:
I think the range is specified to potentially ease error checking in implementations. Since it's not worth fretting over which error code it should be when in WebGL we generate error text explainations, we should just take both errors here.

jdashg:
@kenrussell

zhenyao:
LGTM

kenrussell:
LGTM after the fact. Thanks for updating the snapshot.

PR #2190, opened 2016-12-07
Add a tricky recursive macro expansion test

Oletus:
The test is crashing in the latest Firefox and the stable branch of
Chrome. It's passing in Edge and in Chrome Canary, which is using the
latest version of ANGLE shader translator.


zhenyao:
lgtm

PR #2191, opened 2016-12-07
Add tests for WebGL 2 float parsing corner cases

Oletus:
The tests are only for WebGL 2.0 and not 1.0, since ESSL 3.00.6 added
more explicit rules for parsing floats.


zhenyao:
The test cases look good and thank you for adding them.
Do you have data how various drivers, especially desktop GL drivers handle them? What does desktop GL say about such situations?

zhenyao:
Thanks.

PR #2192, opened 2016-12-07
Update a broken link to the canvas element

grorg:
Fix this link for the current 1.0 spec and all existing stable versions (no change in functionality).


zhenyao:
Thanks for the fix.

PR #2194, opened 2016-12-08
CompressedTexSubImage can generate INVALID_OP instead of INVALID_ENUM…

jdashg:
… if CompressedTexImage failed to specify the texture image.


kenrussell:
LGTM

PR #2195, opened 2016-12-08
Error for querying FB-dependent limit should error as if asking to re…

jdashg:
…ad. (that is, INVALID_FRAMEBUFFER_OPERATION)

@kenrussell @zhenyao @RafaelCintron @grorg 

This behavior is specified in the spec as INVALID_OPERATION instead of GLES2's "returns undefined values".
However this behavior differs in its error code from trying to read from or draw to an incomplete framebuffer.

Firefox's implementation returns INVALID_FRAMEBUFFER_OPERATION, because it shares validation code with the other entrypoints. Instead of special-casing this, I propose we change the error code for 1.0.4/2.0.1+.

### DIFF
--- a/conformance-suites/2.0.0/conformance/extensions/oes-texture-half-float.html
+++ b/conformance-suites/2.0.0/conformance/extensions/oes-texture-half-float.html
@@ -415,10 +415,10 @@ function runFramebufferTest() {
         shouldBe("gl.checkFramebufferStatus(gl.FRAMEBUFFER)", "gl.FRAMEBUFFER_INCOMPLETE_ATTACHMENT");
 
         shouldBeNull("gl.getParameter(gl.IMPLEMENTATION_COLOR_READ_FORMAT)");
-        wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "IMPLEMENTATION_COLOR_READ_FORMAT should fail for incomplete framebuffers.");
+        wtu.glErrorShouldBe(gl, [gl.INVALID_OPERATION, gl.INVALID_FRAMEBUFFER_OPERATION], "IMPLEMENTATION_COLOR_READ_FORMAT should fail for incomplete framebuffers.");
 
         shouldBeNull("gl.getParameter(gl.IMPLEMENTATION_COLOR_READ_TYPE)");
-        wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "IMPLEMENTATION_COLOR_READ_TYPE should fail for incomplete framebuffers.");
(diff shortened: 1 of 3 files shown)


kenrussell:
I'm not sure I agree with this change. INVALID_OPERATION is raised in general when the OpenGL state is in a configuration where the current call doesn't make sense (or arguments to the current call don't make sense), but would make sense if the context state were different.

INVALID_FRAMEBUFFER_OPERATION seems to me like more of a specialized error when attempting to touch the actual framebuffer (via draw, clear, or readback).

I don't feel strongly about this. If we're going to make the change then let's just change the top-of-tree tests to only allow the single intended error.

msc-:
> On 2016 Dec 8, at 18:01, Ken Russell <notifications@github.com> wrote:
> 
> I'm not sure I agree with this change. INVALID_OPERATION is raised in general when the OpenGL state is in a configuration where the current call doesn't make sense (or arguments to the current call don't make sense), but would make sense if the context state were different.
> 

I agree with Ken.  It is not a framebuffer operation that is invalid. It is a query. That its target is framebuffer information should not change the error that is returned.

Regards

    -Mark

jdashg:
I'm going to push back on this. I think INVALID_FB_OP is more natural, and definitely reflects GLES3.0.5 p18, which says that the INVALID_FRAMEBUFFER_OPERATION error's description is "Framebuffer object is not complete".

Aside, I'm getting pretty frustrated moving code around to try to issue the correct error code, even though we give full-text console.log warnings detailing what's wrong.

kenrussell:
I hadn't seen the taxonomy in the OpenGL ES 3.0.5 spec p. 18. Good point Jeff.

Agree, let's make this change. The top-of-tree oes-texture-half-float.html test should require INVALID_FRAMEBUFFER_OPERATION instead of allowing INVALID_OPERATION.

Merging now.

kenrussell:
Note: filed http://crbug.com/672711 about changing Chrome's behavior and making the conformance test more strict.

zhenyao:
Sorry for the late reply, but the spec is clear in this situation: (ES 3.0.5, page 193, GetInteger generates an INVALID_OPERATION error in these cases if the object bound to READ_FRAMEBUFFER_BINDING is not framebuffer complete, or if READ_BUFFER is NONE< or if the GL is using a framebuffer object and the read buffer selects an attachment that has no image attached).

So this overwrites any general rules in p.18.

zhenyao:
@kenrussell Sorry, but I'll have to revert this.

kenrussell:
No worries @zhenyao . Thanks for pointing out the spec text covering this.

(It's in section 4.3.2 "Reading Pixels".)

PR #2196, opened 2016-12-08
Took WebGL 2.0.0 specification snapshot.

kenrussell:
- Copied specs/latest/2.0 to specs/2.0.0.
 - Adjusted all relative URLs to the "resources" directory.
 - Set up symlink from specs/2.0 to specs/2.0.0.

specs/2.0.0 is still considered an editor's draft until it's ratified
by Khronos.


kenrussell:
PTAL @jdashg @zhenyao @kainino0x

zhenyao:
LGTM
There is a minor file permission issue caught by travis, but it doesn't seem matter. Feel free to merge yourself.

kenrussell:
Thanks @zhenyao and sorry for the delay re-visiting this. It looks strongly to me like pylint's checks are bogus and don't understand symlinks, so I'm adding a whitelist entry for them.

PR #2198, opened 2016-12-08
Add tests checking for non-constant infinity

Oletus:
This targets an ANGLE issue where infinity literals are not written
to the shader output as infinity, but rather as max finite float.


zhenyao:
LGTM

PR #2199, opened 2016-12-08
Apply #2161 to 2.0.0 conformance snapshot

kainino0x:
#2161: GREEN, BLUE, and NUM_SAMPLE_COUNTS enums were removed from WebGL2 core.

These changes didn't make it into the snapshot but are needed.


kenrussell:
Thanks for cleaning this up. LGTM after the fact.

Issue #2193, opened 2016-12-07
Disable WebGL 2.0.0 conformance tests affected by driver bugs


kainino0x:
There are a bunch of failing conformance tests which we agreed to disable (in [the 2.0.0 snapshot](https://github.com/KhronosGroup/WebGL/tree/master/conformance-suites/2.0.0)) to allow browsers to pass conformance on buggy drivers.

kainino0x:
Between #2201, #2311, and #2326, this should be pretty much done.

PR #2201, opened 2016-12-10
Disable WebGL 2.0.0 snapshot tests failing due to Mac AMD/Intel drivers

kainino0x:
Issue #2193


zhenyao:
Thanks for putting this together. for the sub-rectangle one, I feel we should either disable the tests, or workaround in browsers, not to hack the test for passing.

zhenyao:
lgtm

PR #2202, opened 2016-12-10
Add test verifying sizing of non-power-of-two video uploads.

kenrussell:
This is effectively a regression test for http://crbug.com/672895, in which
the size of a WebGL texture which was uploaded from a video was incorrect.
Artifacts were seen when this bug was present.

It was difficult to write a test for this bug. Attempts were made to
produce lower-resolution videos, but they did not provoke the bug. Also,
there are discrepancies in the reported videoWidth and videoHeight between
browsers, though Chrome and Firefox both report a height 1080 for this
video (at least, once the Chrome bug is fixed).


kenrussell:

This was tested in Chrome and Firefox on Windows. Firefox passes the test already; Chrome will pass it once http://crbug.com/672895 is fixed.

zhenyao:
LGTM, but can you fix the Travis issues before merging yourself?

kenrussell:
Thanks @zhenyao and sorry for the delay revisiting this. Fixed. Merging now.

PR #2204, opened 2016-12-13
Move WEBGL_compressed_texture_astc to community approved status.

kenrussell:


kenrussell:
Discussed on public_webgl list. There's a conformance test for it now.

@jdashg @zhenyao @grorg @RafaelCintron

PR #2207, opened 2016-12-15
Sub-texture uploads with too-large level can generate INVALID_OP as w…

jdashg:
…ell as INVALID_VALUE.

Given that the level is too large, texture specification must have failed, thus there is no valid tex-image to select, yielding INVALID_OPERATION.


zhenyao:
lgtm

PR #2208, opened 2016-12-16
Let WebGL2RenderingContext implement WebGLRenderingContextBase

foolip:
To have one mixin implement another is unusual, and does not match how
this is implement in Chromium:
https://cs.chromium.org/chromium/src/third_party/WebKit/Source/modules/webgl/WebGL2RenderingContext.idl

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -968,11 +968,11 @@ interface <dfn id="WebGL2RenderingContextBase">WebGL2RenderingContextBase</dfn>
   [WebGLHandlesContextLoss] GLboolean isVertexArray(WebGLVertexArrayObject? vertexArray);
   void bindVertexArray(WebGLVertexArrayObject? array);
 };
-WebGL2RenderingContextBase implements WebGLRenderingContextBase;
 
 interface <dfn id="WebGL2RenderingContext">WebGL2RenderingContext</dfn>
 {
 };
+WebGL2RenderingContext implements WebGLRenderingContextBase;
 WebGL2RenderingContext implements WebGL2RenderingContextBase;
 
 </pre>


zhenyao:
I think the original idl implements is correct.  According to the Web IDL spec, 
"An implements statement is a definition (matching ImplementsStatement) used to declare that all objects implementing an interface A (identified by the first identifier) must additionally implement interface B (identified by the second identifier), including all other interfaces that B inherits from."

I don't know why Chromium did it differently, so I'll take a look there.

foolip:
Note that the suggested change amounts to the same thing.

I couldn't quite figure out if the existing way is merely unusual or also problematic in some way. Implementation-wise, I suppose it's more work to support this than not. @heycam @bzbarsky, what do you think would be idiomatic here? What would best fit with a future mixins model?

kenrussell:
At the C++ level, anyway, the intent was for WebGL2RenderingContextBase to inherit from WebGLRenderingContextBase -- that's how Chromium's implementation works. The intent was for WebGL2RenderingContext to *not* derive from WebGLRenderingContext, though.

I don't remember why the IDL files in Chromium differ from the spec's. Maybe there was a limitation in the IDL parser at the time they were written.

bzbarsky:
The idea with mixins in terms of the future mixins model is they are just that: mixins.  They are purely a specification device to cut down on the copy/paste.  In terms of implementation, they are the same as if you copy/pasted everything in the mixin directly into the interfaces it's mixed into, unless you mean implementation in the actual IDL parser or something.

So on the "implementation level" there is no WebGL2RenderingContextBase or WebGLRenderingContextBase or anything like that.  Those are purely specification artifacts.

I have no idea what WebGLRenderingContextBase is, actually, since it's not defined anywhere...  That makes it pretty hard to say what this spec is really trying to do.  But if the intention is for WebGLRenderingContext to be redefined as an empty interface, with all its guts moved to WebGLRenderingContextBase, so WebGL2RenderingContext can pull them in, then that would make sene.  I guess the intent of doing it that way is so when WebGL3RenderingContext is defined it only has to implement one thing that itself implements WebGL2RenderingContextBase and hence transitively WebGLRenderingContextBase?

kenrussell:
The goal here was to have WebGL2RenderingContext inherit the behavior from WebGLRenderingContext, but not have an "is-a" relationship between the contexts, so that developers would get the idea that WebGL contexts are not necessarily backward compatible.

For this reason, all the behavior of both WebGL 1.0 and 2.0 was factored into the "base" contexts, and the inheritance relationship (defined by Web IDL) was set up in this way.

kenrussell:
P.S. WebGLRenderingContextBase is defined in the WebGL 1.0 specification.

bzbarsky:
> For this reason, all the behavior of both WebGL 1.0 and 2.0 was factored into the "base" contexts

If you don't mean to imply that anything that implements the 2.0 behavior (e.g. a hypothetical 3.0 context) will also implement the 1.0 behavior, then they should be mixed in separately, I would think.  If the implication is that something that implements the 2.0 behavior will always implement the 1.0 behavior, then the current setup in the spec makes the most sense.

> P.S. WebGLRenderingContextBase is defined in the WebGL 1.0 specification.

Ah, starting with 1.0.3.  The problem is that https://www.khronos.org/registry/webgl/specs/latest/2.0/ links to 1.0.2 as the "previous version", and that does not have WebGLRenderingContextBase.  And discovering from the 2.0 draft that there is a 1.0.3 (or, indeed the current version, which is <https://www.khronos.org/registry/webgl/specs/latest/1.0/>) is a bit of a pain.

But the fact that I even have to guess at what spec it's defined in is ridiculous.  Why doesn't webgl2 just link to the definition?

bzbarsky:
Ideally the linking to the parent interface definition would be automatically handled by whatever spec-HTML-generation framework you're using, in fact... @tabatkins, will bikeshed will do this?

tabatkins:
Yes, in IDL like `interface Foo : Bar { ... };`, the `Bar` is auto-linked (assuming it's possible to do so; it throws a linking error otherwise).

foolip:
It is worth noticed that in Gecko, [`WebGL2RenderingContext`](https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/WebGL2RenderingContext.webidl) actually does inherit directly from `WebGLRenderingContext`. It looks like @jdashg has been poking at this file most recently.

@zhenyao, it'd be great if you could figure out why Blink's IDL is different. If it can trivially be made to match the spec, then great. But if this is literally the only case on the platform of `A implements B` + `B implements C`, then I wonder if it's actually worth teaching everyone's bindings generators about transitive `implements`.

Tangentially, versioned things on the web is a bit scary. Should we expect to see a WebGL 3, 4 and 5, each of which adds a complete copy of the previous API surface? At a certain point that has to show up in the size of the generated bindings.

bzbarsky:
> It is worth noting that in Gecko, WebGL2RenderingContext actually does inherit directly from WebGLRenderingContext

Huh.  That's really odd.  I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1324543

> if this is literally the only case on the platform of A implements B + B implements C

Looking at Gecko's IDL, we have this in a few cases but they don't seem to be web-exposed.  That said, this seems like a totally reasonable thing to want to do with mixins.

kenrussell:
Anything implementing the WebGL 2.0 behavior will also implement the WebGL 1.0 behavior.

The current thinking is that WebGL 3.0 will break backward compatibility, and not be a superset of WebGL 2.0. We wanted to prepare developers for this possibility, which is why WebGL2RenderingContext doesn't derive from WebGLRenderingContext, thereby providing an implicit guarantee that it's fully backward compatible. (Even WebGL 2.0 is not 100% backward compatible to WebGL 1.0 -- in some corner cases, that was not possible.)

We plan to move the WebGL spec to Bikeshed. It's near the top of the list now that WebGL 2.0 has shipped. That should clear up a lot of the problems around the current lack of inter-spec linkage.

saschanaz:
Seems that this is already fixed in master?

kenrussell:
Yes, I guess this was fixed during another mixin-related refactoring proposed by a contributor.

Closing this out.

PR #2209, opened 2016-12-16
Increase the error tolerance in certain readPixels test cases

zhenyao:
They are exposed by https://codereview.chromium.org/2577293002/

Verified on Mac/Linux that the tolerance increase is necessary.


zhenyao:
@kainino0x

kainino0x:
LGTM

PR #2210, opened 2016-12-17
null coerces to 0 for the PBO entrypoint for compressed tex uploads, …

jdashg:
…yielding INVALID_OP.

### DIFF
--- a/conformance-suites/2.0.0/conformance/extensions/webgl-compressed-texture-etc.html
+++ b/conformance-suites/2.0.0/conformance/extensions/webgl-compressed-texture-etc.html
@@ -61,7 +61,8 @@ var COMPRESSED_RGBA8_ETC2_EAC                 = 0x9278;
 var COMPRESSED_SRGB8_ALPHA8_ETC2_EAC          = 0x9279;
 
 var wtu = WebGLTestUtils;
-var gl = wtu.create3DContext(undefined, undefined);
+var contextVersion = wtu.getDefault3DContextVersion();
+var gl = wtu.create3DContext();
 var WEBGL_compressed_texture_etc;
 
 var formats = null;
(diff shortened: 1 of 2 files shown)


kenrussell:
Thanks for this patch. It seems reasonable but I think we need to figure out why Chrome's currently passing these tests on Windows. I'm not in front of a Windows machine right now but @zhenyao or @kainino0x , can we work together tomorrow (Monday) to get to the bottom of this?

conformance-suites/2.0.0/conformance/extensions/webgl-compressed-texture-etc.html
```diff
@@ -138,10 +139,18 @@
     shouldBe("formats.length", isPositive ? "10" : "0");
 
     debug("");
-    shouldThrow("gl.compressedTexImage2D(gl.TEXTURE_2D, 0, COMPRESSED_R11_EAC, 4, 4, 0, null)");
-    shouldThrow("gl.compressedTexSubImage2D(gl.TEXTURE_2D, 0, 0, 0, 0, 0, COMPRESSED_R11_EAC, null)");
-    shouldThrow("gl.compressedTexImage3D(gl.TEXTURE_2D_ARRAY, 0, COMPRESSED_R11_EAC, 4, 4, 4, 0, null)");
-    shouldThrow("gl.compressedTexSubImage3D(gl.TEXTURE_2D_ARRAY, 0, 0, 0, 0, 0, 0, 0, COMPRESSED_R11_EAC, null)");
+    if (contextVersion >= 2) {
+      // `null` coerces into `0` for the PBO entrypoint, yielding INVALID_OP due to no PBO bound.
```
zhenyao: We have absolutely no tests to cover the compressedTex{Sub}Image from PBOs in the conformance test suite until this case.  It turns out Chrome didn't add the idl function signatures for compressedTex{Sub}Image with PBOs, that's why we are throwing in WebGL1 and WebGL2.  At this point, the ToT change is fine, but the 2.0.0 case needs to allow both behaviors.  Let's merge this and I'll upload a pull request for that.

zhenyao: Yunchao,

If you update your chromium source, you will see the new function signatures being added to the IDL, compressed* with offset instead of srcData.  So null will be 0 and will match these newly added signatures.

kainino0x: Yunchao, apologies if I sent you down a wrong path with my list of test failures. I ran those on 2.0.1 and I forgot that this had been a very recent change. I probably should have expected this failure.

Richard-Yunchao: @zhenyao , If we want to list the definition of compressedTexImage2D:
 1) void compressedTexImage2D(GLenum target, GLint level, GLenum internalformat, GLsizei width,
                            GLsizei height, GLint border, GLsizei imageSize, GLintptr offset);
  2) void compressedTexImage2D(GLenum target, GLint level, GLenum internalformat, GLsizei width,
                            GLsizei height, GLint border, ArrayBufferView srcData,
                            optional GLuint srcOffset = 0, optional GLuint srcLengthOverride = 0);

If we only have 7 parameters and call compressedTexImage2D, the 7th parameters should be always "ArrayBufferView", which is required, not optional. So if it can not be "null".  However, in texImage2D, the same parameter is  "ArrayBufferView?", it can be null in this case. 

I suppose that both you and Jeff want to use the first definition of compressedTexImage2D, so we need 8 parameters. I have updated the fix at https://github.com/KhronosGroup/WebGL/pull/2253, could you take another look? 

@kainino0x , no problem. :)

zhenyao: Thanks.  This is last minute spec change (from 7 args to 8), and both Chrome/Firefox won't support these new signatures in our initial WebGL2 launch.

Richard-Yunchao: @zhenyao , Ah... I see...

PR #2212, opened 2016-12-19
Add some tests when fbo miss image

Richard-Yunchao:
@zhenyao , PTAL. Thanks a lot! 

A small change to expose a bug in Chromium. See the fix at https://codereview.chromium.org/2588653002/


zhenyao:
lgtm

PR #2213, opened 2016-12-19
Add test to cover ANGLE sticking depth textures.

null77:
ANGLE would not release a particular resource cache after recreating
an internal reference to the depth texture, causing rendering to
get stuck with one version of the texture. The test covers that bug.

See http://anglebug.com/1664 for more info.

PTAL @kenrussell @zhenyao

### DIFF
--- a/sdk/tests/conformance2/textures/misc/angle-stuck-depth-textures.html
+++ b/sdk/tests/conformance2/textures/misc/angle-stuck-depth-textures.html
@@ -1,0 +1,219 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/textures/misc/angle-stuck-depth-textures.html
```diff
@@ -0,0 +1,219 @@
...
+</head>
+<body>
+<canvas id="example" width="128" height="128"></canvas>
+<div id="description"></div>
+<div id="console"></div>
+
+<script>
+"use strict";
+
+description("This test covers an ANGLE bug where an AMD workaround would cause depth textures to stick. See http://angle.bug.com/1664.");
```
kenrussell: angle.bug.com -> anglebug.com

null77:
Thanks Ken, fixed the description and updated the line endings. Seems Travis is happy.

kenrussell:
Fantastic. Thanks Jamie for this great test. LGTM

PR #2214, opened 2016-12-19
Remove R11_G11_B10 entries from skip list

zhenyao:
Since the MacOSX driver bug is fixed in 10.12.2 release


zhenyao:
See https://github.com/KhronosGroup/WebGL/issues/1832

zhenyao:
Merging myself.  Should we also update the 2.0.0 snapshot or just leave that alone?

kenrussell:
Thanks @zhenyao for cleaning these up. Let's leave the 2.0.0 snapshot alone.

PR #2215, opened 2016-12-19
Unskip a few DEQP cases due to the OSX bug being fixed in 10.12.2

zhenyao:


zhenyao:
These are fixed in 10.12.2.

Merging myself.  Should be also unskip these cases in the 2.0.0 snapshot? @kenrussell @jdashg

kenrussell:
Thanks @zhenyao . Let's leave the 2.0.0 snapshot as is.

PR #2217, opened 2016-12-20
MAX_CLIENT_WAIT_TIMEOUT_WEBGL is in nanoseconds, not milliseconds.

jdashg:


kenrussell:
Thanks for the fix. Looks good.

PR #2219, opened 2016-12-22
Generic vertex attrib values are vertex shader state.

jdashg:


zhenyao:
I disagree.  If generic attrib values are per vertex shader, then VertexAttrib{I}* function should generate INVALID_OPERATION of no program is bound, but we don't have that error specified.

jdashg:
GLES 2.0.25 p150:
https://www.khronos.org/registry/gles/specs/2.0/es_full_spec_2.0.25.pdf

GL2.0 p291:
https://www.opengl.org/registry/doc/glspec20.20041022.pdf

I agree, they should generate INVALID_OP. I have patches to add this to Firefox.

jdashg:
Oh, yeah, I guess they might be saying "this is state for vertex shaders", but not actually "this is state attached to vertex shaders?". That does seem more likely.

jdashg:
Yeah, that must be it. If only I could PR the GL specs!

PR #2223, opened 2016-12-28
Fix wrong size of buffer when filling texture

qkmiao:
Buffer is already a TypedArray type. We don't need to take care of bytes
per component when calculating its size.

### DIFF
--- a/sdk/tests/js/webgl-test-utils.js
+++ b/sdk/tests/js/webgl-test-utils.js
@@ -843,12 +843,21 @@ var fillTexture = function(gl, tex, width, height, color, opt_level, opt_format,
   var numComponents = color.length;
   var bytesPerComponent = getBytesPerComponent(gl, opt_type);
   var rowSize = numComponents * width * bytesPerComponent;
-  var paddedRowSize = Math.floor((rowSize + pack - 1) / pack) * pack;
-  var size = rowSize + (height - 1) * paddedRowSize;
-  size = Math.floor((size + bytesPerComponent - 1) / bytesPerComponent) * bytesPerComponent;
+  // See equation 3.10 in ES 2.0 spec and equation 3.13 in ES 3.0 spec for paddedRowLength calculation.
+  // k is paddedRowLength.
+  // n is numComponents.
+  // l is width.
+  // a is pack.
+  // s is bytesPerComponent.
+  var paddedRowLength;
+  if (bytesPerComponent >= pack)
+    paddedRowLength = numComponents * width;
+  else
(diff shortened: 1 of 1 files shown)


sdk/tests/js/webgl-test-utils.js
```diff
@@ -841,11 +841,9 @@ var fillTexture = function(gl, tex, width, height, color, opt_level, opt_format,
   opt_internalFormat = opt_internalFormat || opt_format;
   var pack = gl.getParameter(gl.UNPACK_ALIGNMENT);
   var numComponents = color.length;
-  var bytesPerComponent = getBytesPerComponent(gl, opt_type);
-  var rowSize = numComponents * width * bytesPerComponent;
+  var rowSize = numComponents * width;
   var paddedRowSize = Math.floor((rowSize + pack - 1) / pack) * pack;
```
zhenyao: I agree we have a bug here, but this paddedRowSize is wrong if the rowSize isn't in bytes. Some extra attention is needed to get the size right.

qkmiao: Thanks, you are right. I change the behavior to align with ES spec. Can you take another look?

sdk/tests/js/webgl-test-utils.js
```diff
@@ -843,12 +843,21 @@ var fillTexture = function(gl, tex, width, height, color, opt_level, opt_format,
...
-  var size = rowSize + (height - 1) * paddedRowSize;
-  size = Math.floor((size + bytesPerComponent - 1) / bytesPerComponent) * bytesPerComponent;
+  // See equation 3.10 in ES 2.0 spec and equation 3.13 in ES 3.0 spec for paddedRowLength calculation.
+  // k is paddedRowLength.
+  // n is numComponents.
+  // l is width.
+  // a is pack.
+  // s is bytesPerComponent.
+  var paddedRowLength;
+  if (bytesPerComponent >= pack)
```
zhenyao: Is this correct?  For example, the bytesPerComponent is 3 and pack is 2, we still need to add an extra byte in the end of a row is width is an odd number.

qkmiao: bytesPerComponent can only be 1, 2, and 4. See getBytesPerComponent = function(gl, type). 
This implementation follows the spec. So I think it should be conformant with any allowed formats.

zhenyao: You are right.  I confused bytesPerComponent with componentsPerPixel.

PR #2224, opened 2016-12-31
Fix bitmaps

jdashg:
After these changes, Firefox passes these tests.

Firefox does not yet implement support for the options dict in the createImageBitmap function.
The tests seem to want this to be optional, but the test fails to complete instead of failing softly.

### DIFF
--- a/conformance-suites/2.0.0/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image-bitmap.js
+++ b/conformance-suites/2.0.0/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image-bitmap.js
@@ -56,9 +56,15 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
                                       0, 255, 0, 0]),
                                       2, 2);
 
-        createImageBitmap(imageData, {imageOrientation: "none", premultiplyAlpha: "none"}).then(function(bitmap) {
-            // bitmap is in unpremultiplied form
-            runImageBitmapTest(bitmap, 0, internalFormat, pixelFormat, pixelType, gl, tiu, wtu, false);
+        createImageBitmap(imageData, {imageOrientation: "none", premultiplyAlpha: "none"})
+        .catch( () => {
+            testPassed("createImageBitmap with options may be rejected if it is not supported. Retrying without options.");
+            return createImageBitmap(imageData);
+        }).then( bitmap => {
+            return runImageBitmapTest(bitmap, 0, internalFormat, pixelFormat, pixelType, gl, tiu, wtu, false);
+        }, () => {
+            testFailed("createImageBitmap(imageData) should succeed.");
+        }).then(() => {
             finishTest();
         });
     }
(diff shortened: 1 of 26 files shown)


conformance-suites/2.0.0/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image-bitmap.js
```diff
@@ -56,9 +56,15 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
                                       0, 255, 0, 0]),
                                       2, 2);
 
-        createImageBitmap(imageData, {imageOrientation: "none", premultiplyAlpha: "none"}).then(function(bitmap) {
-            // bitmap is in unpremultiplied form
-            runImageBitmapTest(bitmap, 0, internalFormat, pixelFormat, pixelType, gl, tiu, wtu, false);
+        createImageBitmap(imageData, {imageOrientation: "none", premultiplyAlpha: "none"})
+        .catch( () => {
+            testPassed("createImageBitmap with options may be rejected if it is not supported. Retrying without options.");
```
zhenyao: Will this accidentally allow other failures from image bitmap creation with options?

jdashg: I think so. We might be able to catch it by catching the DOM exception specifically?

conformance-suites/2.0.0/js/tests/tex-image-and-sub-image-with-image-bitmap-utils.js
```diff
@@ -333,76 +333,83 @@ function runOneIterationImageBitmapTestSubSource(useTexSubImage, bindingTarget,
...
-            runOneIterationImageBitmapTestSubSource(cases[i].sub, bindingTarget, program, cases[i].bitmap,
-                cases[i].flipY, cases[i].premultiply, optionsVal, internalFormat, pixelFormat, pixelType, gl, tiu, wtu);
-        }
+        (bindingTarget == gl.TEXTURE_CUBE_MAP || bindingTarget == gl.TEXTURE_2D_ARRAY))
+    {
+        // Skip testing source sub region on TEXTURE_CUBE_MAP and TEXTURE_2D_ARRAY on WebGL2 to save
+        // running time.
+        return;
     }
+
```
Richard-Yunchao: This code snippet above is a little different from the original one. 

In the original code, it will never execute runOneIterationImageBitmapTestSubSource() under WebGL 1 context. 

However, after this change is applied, it will execute runOneIterationImageBitmapTestSubSource() in WebGL 1 context when bindTarget is gl.TEXTURE_2D or gl.TEXTURE_3D. This make Chromium failed to find texImage2D API in WebGLRenderingContext and then context lost finally. (I am not familiar with sub resource related code, I suppose that sub resource is not supported in webgl 1?) . 

See the small change at https://github.com/KhronosGroup/WebGL/pull/2251/.

PR #2226, opened 2016-12-31
Changes to conformance2/textures/misc/copy-texture-image.html

jdashg:
* Sized formats can still drop components. (e.g. RGBA->RGB)
* Both RGBA8 and RGBA8_SNORM count as fixed-point.
* Depth formats may generate INVALID_OP in addition to INVALID_ENUM.

### DIFF
--- a/conformance-suites/2.0.0/conformance2/textures/misc/copy-texture-image.html
+++ b/conformance-suites/2.0.0/conformance2/textures/misc/copy-texture-image.html
@@ -118,35 +118,21 @@ var testInternalformat = function () {
         "RGBA8UI",
         "RGBA16UI",
     ];
-    var badByteFormats = [
+    const snormFormats = [
         "R8_SNORM",
         "RG8_SNORM",
         "RGB8_SNORM",
         "RGBA8_SNORM",
     ];
-    var badIntFormats = [
-        "RGB8I",
-        "RGB16I",
-        "RGB32I",
-    ];
-    var badUnsingedIntFormats = [
-        "RGB8UI",
-        "RGB16UI",
-        "RGB32UI",
-    ];
     var badFloatFormats = [
         "R16F",
-        "R32F",
         "RG16F",
-        "RG32F",
         "R11F_G11F_B10F",
         "RGB9_E5",
         "RGB16F",
-        "RGB32F",
         "RGBA16F",
-        "RGBA32F",
     ];
-    var badDepthStencilFormats = [
+    const depthAndOrStencilFormats = [
         "DEPTH_COMPONENT16",
         "DEPTH_COMPONENT24",
         "DEPTH_COMPONENT32F",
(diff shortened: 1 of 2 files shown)


conformance-suites/2.0.0/conformance2/textures/misc/copy-texture-image.html
```diff
@@ -80,8 +80,12 @@
         "RG8",
         "RGB8",
         "RGBA8",
+        "R8_SNORM",
```
zhenyao: Originally this test is written against the list of valid internalformats from https://www.khronos.org/opengles/sdk/docs/man3/html/glCopyTexImage2D.xhtml

I understand the man page isn't the spec, but I am curious why the man page so confidently offers a list of the valid internalformats?

zhenyao: @kenrussell @msc- @shannonwoods Anyone knows?

jdashg: Interesting!

shannonwoods: Looks like it was purposeful and explicit: https://cvs.khronos.org/bugzilla/show_bug.cgi?id=10001

zhenyao: Thanks Shannon for digging this up.

kenrussell: Yes, thanks @shannonwoods for finding this!

To summarize for people who don't have access to the bug: SNORM formats aren't supported in CopyTexImage2D.

zhenyao: Also the added RGB integer formats which are not listed in https://www.khronos.org/opengles/sdk/docs/man3/html/glCopyTexImage2D.xhtml

Assume that list is deliberate, I suggest the testing follow it for now.

I agree with the changes related with float formats that become valid after enabling the extension.

shannonwoods:
Dropping channels is permitted for either sized or unsized formats, but for sized formats, the bit depth for each channel that is retained must match the source buffer's effective internal format (see table 3.17 in the ES 3.0.5 spec).

conformance-suites/2.0.0/conformance2/textures/misc/copy-texture-image.html
shannonwoods: I'm less sure about the omission of RGB  integer formats being deliberate-- they're not mentioned explicitly on the bug in the way that SNORM formats are. RGB integer formats aren't color renderable, but RGBA integer formats are, and it should be legal, I think, to copy from a source buffer created from, e.g., a RGBA8I texture, to a texel array using internalformat RGB8I, unless I'm forgetting something. CopyTexImage internalformat validation is a particularly hairy topic that the working group has revisited a number of times, so it is quite possible that the formats were missed.

zhenyao:
Then can we not testing RGB integer formats for now? i.e., not verifying it should work, and not verifying it should fail.  At least in the 2.0.0 suite.

jdashg:
@kenrussell Updated, ptal

The 2.0.0 changes just remove the conflicting testing now, so browsers should continue to pass the 2.0.0 with no changes.

kenrussell:
Thanks for revising this @jdashg . Your changes look good. I'm a little nervous that some OpenGL drivers may not handle some RGB formats (like RGB32I) well, but let's move forward and file bugs with vendors as they're uncovered.

PR #2227, opened 2016-12-31
Fix deqp/functional/gles3/lifetime.html

jdashg:
* WebGL objects always linger when attached to containers.
* Don't detach from a VAO before querying which buffer to delete.

### DIFF
--- a/conformance-suites/2.0.0/deqp/functional/gles3/es3fLifetimeTests.js
+++ b/conformance-suites/2.0.0/deqp/functional/gles3/es3fLifetimeTests.js
@@ -100,12 +100,11 @@ es3fLifetimeTests.ScaleProgram.prototype.draw = function(vao, scale, tf, dst) {
 es3fLifetimeTests.ScaleProgram.prototype.setPos = function(buffer, vao) {
     gl.bindBuffer(gl.ARRAY_BUFFER, buffer);
     gl.bindVertexArray(vao);
-    if (buffer) {
-        gl.vertexAttribPointer(this.m_posLoc, NUM_COMPONENTS, gl.FLOAT, false, 0, 0);
-    } else {
+    if (!buffer) {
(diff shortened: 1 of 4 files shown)


kenrussell:
Sorry for the delay reviewing this.

This change exposes a bug in Chrome around the deletion/mark for deletion behavior in #2178. The changes cause the following tests to fail in Chrome:

lifetime.attach.deleted_name.texture_framebuffer
lifetime.attach.deleted_name.renderbuffer_framebuffer
lifetime.attach.deleted_name.buffer_vertex_array
lifetime.attach.deleted_name.buffer_transform_feedback

It also causes a GL error in lifetime.attach.deleted_output.texture_framebuffer , but currently that isn't caught until a later test, lifetime.delete_active.transform_feedback.

We'll be happy to update Chrome's deletion semantics on top-of-tree, but could you please modify the 2.0.0 snapshot so that it passes both with Firefox nightly as well as Chrome Canary?

jdashg:
We're trying to pick a set of conformance tests that we're comfortable with for 2.0.0, whereas 2.0.1/tip has a more complete set of tests. This way we can make our way forward incrementally.

I honestly don't know if I can modify that test. It was excruciating just to get to this point. I don't think I've ever seen more obtuse or extensive boilerplate.

I'd be fine leaving this test as-is for 2.0.0 and having it among the Firefox51 test result errata.

kenrussell:
Jeff, in that case, could you add the tests that are failing in Firefox to conformance-suites/2.0.0/deqp/framework/common/tcuSkipList.js , and let's undo the other code changes there? Please add a good _setReason description for the failures, pointing to this pull request.

Thanks.

jdashg:
Following #2456, I think these tests are correct as-is, but I'll double-check.

jdashg:
Abandoning in favor of #2731.

PR #2228, opened 2016-12-31
Add clear test that vertexAttribPointer with null buffer and zero off…

jdashg:
…set succeeds.

This is de-facto tested in deqp's lifetime.html, but I'm really surprised we don't seem to test it elsewhere.

### DIFF
--- a/conformance-suites/2.0.0/conformance/attribs/gl-vertexattribpointer.html
+++ b/conformance-suites/2.0.0/conformance/attribs/gl-vertexattribpointer.html
@@ -73,10 +73,10 @@ if (!gl) {
   gl.vertexAttribPointer(0, 1, gl.FLOAT, false, 0, 0);
   wtu.glErrorShouldBe(gl, gl.NO_ERROR);
 
-  gl.bindBuffer(gl.ARRAY_BUFFER, null);
-  gl.vertexAttribPointer(0, 1, gl.FLOAT, false, 0, 0);
-  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
-      "vertexAttribPointer should succeed if no buffer is bound and `offset` is zero.");
+  //gl.bindBuffer(gl.ARRAY_BUFFER, null);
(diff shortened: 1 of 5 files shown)


jdashg:
Oh, the way we clarified this for WebGL 1 was to forbid any VertexAttribPointer call when null is bound.
In WebGL 2, it is specified as I test in GLES 3.0.5 p25 for non-zero VAOs, presumably leaving the zero-VAO behavior as 'anything goes' as in GLES2.

conformance-suites/2.0.0/conformance/attribs/gl-vertexattribpointer.html
```diff
@@ -59,14 +59,28 @@
...
+      "vertexAttribPointer should fail if no buffer is bound and `offset` is non-zero.");
+
+  gl.bindBuffer(gl.ARRAY_BUFFER, vertexObject);
+  gl.vertexAttribPointer(0, 1, gl.FLOAT, false, 0, 0);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+
+  gl.bindBuffer(gl.ARRAY_BUFFER, null);
+  gl.vertexAttribPointer(0, 1, gl.FLOAT, false, 0, 0);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+      "vertexAttribPointer should succeed if no buffer is bound and `offset` is zero.");
```
kenrussell: Could you please also allow INVALID_OPERATION here, in the 2.0.0? Chrome currently rejects this call, and we don't want to invalidate our 2.0.0 conformance results at this point. Thanks.

jdashg:
@kenrussell updated

kenrussell:
Thanks Jeff. Confirmed in Chrome Canary that the 2.0.0 snapshot continues to pass, but that the new top-of-tree tests fail. Filed http://crbug.com/678850 about fixing them. Merging now.

PR #2229, opened 2017-01-03
Add srgb extension format checking for generateMipmap.

JiangYizhou:
This patch checks if SRGB_EXT format works for generateMipmap.
Also move some common functions to webgl-test-utils.js.
@zhenyao, @kenrussell, @qkmiao and @Richard-Yunchao, PTAL.


zhenyao:
lgtm

PR #2231, opened 2017-01-04
Clarify an undefined behavior related with MRT in WebGL2.

zhenyao:
I think we already have tests in this, and the language was meant for this but
ended up narrower than intended.


zhenyao:
Thanks @null77 for catching this
@kenrussell @jdashg PTAL

kenrussell:
Thanks for the clarification. Looks good.

Issue #2233, opened 2017-01-04
compressedTex{Sub}Image{2|3}D from PBO should have an arg |imageSize| as ES signature


zhenyao:
Also, we need to add tests to cover these cases

zhenyao:
@kenrussell @kainino0x @jdashg

zhenyao:
So the current situation is:

Chrome doesn't expose the PBO APIs at all
Firefox exposes these APIs (I assume) without |imageSize|

The only tests in 2.0.0 suites are the ones from https://github.com/KhronosGroup/WebGL/pull/2232.  When Firefox/Chrome update to APIs with |imageSize|, they should still be passing.

PR #2234, opened 2017-01-05
Add imageSize to compressedTex{Sub}Image{2|3}D.

zhenyao:
This is to address https://github.com/KhronosGroup/WebGL/issues/2233, the spec part.

@kenrussell @kainino0x @jdashg PTAL

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -791,19 +791,19 @@ interface <dfn id="WebGL2RenderingContextBase">WebGL2RenderingContextBase</dfn>
                          GLint x, GLint y, GLsizei width, GLsizei height);
 
   void compressedTexImage2D(GLenum target, GLint level, GLenum internalformat, GLsizei width,
-                            GLsizei height, GLint border, GLintptr offset);
+                            GLsizei height, GLint border, GLsizei imageSize, GLintptr offset);
   void compressedTexImage2D(GLenum target, GLint level, GLenum internalformat, GLsizei width,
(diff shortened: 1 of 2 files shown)


kenrussell:
Thanks for putting this together Mo.

Should we update the detailed documentation for these entry points to discuss the imageSize argument?

@jdashg please review this -- it's an oversight in the specification for these APIs. Arguably the imageSize argument is redundant -- for all of the existing compressed texture formats, it can be computed from the internal format, width and height -- but we specify it for the ArrayBufferView variants of these methods, so to avoid deviation from the ES3 spec, it should be added here.

zhenyao:
imageSize related documentation is exactly the same as the man pages, usually we avoid repeating it if there isn't any WebGL specific rules.

jdashg:
I think it makes the most sense to also insert imageSize args before the ArrayBufferView args, and remove the optional length arg.
This would most match the GLES spec, and be most consistent between entrypoints.

Also I don't think I can take this change in Firefox 51, but we can errata it for 51 and fix in 52.

LGTM otherwise.

zhenyao:
Same here with Chrome. We have to wait for the next release to update this.

Let me merge this to move forward.

As for the imageSize with ArrayBufferView, in theory I agree it's a good idea, but since both Firefox and Chrome will launch WebGL2 with a different signature, I have two concerns: 1) we might break contents if they are out already 2) make conformance tests pass for both versions will be challenging.

We can discuss further in the working group meeting today.

PR #2235, opened 2017-01-05
Expand copyTexSubImage partial upload test.

jdashg:


sdk/tests/conformance/textures/misc/copy-tex-sub-image-2d-partial-texture.html
```diff
@@ -43,36 +43,141 @@
...
+                  new Uint8Array(view.buffer));
+}
+
+////
+
 function runTest() {
-    var wtu = WebGLTestUtils;
-    var canvas = document.getElementById("canvas");
-    canvas.width = 256;
-    canvas.height = 256;
```
kenrussell: This test (added in #1992 ) specifically allocated a large canvas which would be GPU-accelerated in Chrome, and acted as a regression test for a bug. The new test is great, but it has significantly different behavior. Rather than changing this test, could you please add a new one? Feel free to move this one into something like copy-tex-sub-image-2d-partial-texture-from-large-canvas.html if desired.

jdashg:
@kenrussell Made a new file for the test, and renamed the old test to be more clear.

sdk/tests/conformance/textures/misc/copytexsubimage2d-subrects.html
```diff
@@ -0,0 +1,191 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
```
kenrussell: Should be 2017 going forward.

PR #2236, opened 2017-01-05
Add fbo complete test for fbo attaching base level of a texture

qkmiao:
This failed on NVIDIA Linux with ANGLE and Intel Mac OSX. Chrome bug is filed: https://bugs.chromium.org/p/chromium/issues/detail?id=678526.

### DIFF
--- a/sdk/tests/conformance2/renderbuffers/framebuffer-test.html
+++ b/sdk/tests/conformance2/renderbuffers/framebuffer-test.html
@@ -292,12 +292,35 @@ function testFramebufferRenderbuffer() {
   wtu.glErrorShouldBe(gl, gl.NO_ERROR, "bind draw framebuffer to default (null) framebuffer.");
 }
 
+function testFramebufferTextureWithNonZeroBaseLevel() {
+  var width = 32;
+  var height = 16;
+  var level = 1;
+  var texture = gl.createTexture();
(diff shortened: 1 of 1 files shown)


kainino0x:
LGTM

sdk/tests/conformance2/renderbuffers/00_test_list.txt
```diff
@@ -1,3 +1,4 @@
+--min-version 2.0.1 framebuffer-completeness.html
```
zhenyao: Can we change the name to be more specific, like framebuffer-texture-with-non-zero-base-level.html, something like that?

qkmiao: We can add framebuffer completeness related tests in this file and use different function names like runBaseLevelTest to differentiate them.  Does this make sense? If not, I can change the file name.

zhenyao: Then framebuffer-test.html sounds like a perfect place for this test case.

qkmiao: Move the test to framebuffer-test.html

zhenyao:
Thank you. LGTM

PR #2237, opened 2017-01-05
Add conformance2/programs/active-built-in-attribs.html

jdashg:


sdk/tests/conformance2/programs/00_test_list.txt
```diff
@@ -1 +1,2 @@
+active-built-in-attribs.html
```
kenrussell: At this point should say --min-version 2.0.1, just for bookkeeping reasons.

sdk/tests/conformance2/programs/active-built-in-attribs.html
```diff
@@ -0,0 +1,107 @@
+﻿<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
```
kenrussell: 2017, now.

PR #2239, opened 2017-01-05
Fix nullable behaviors in 1.0.3

zhenyao:
This is following https://github.com/KhronosGroup/WebGL/pull/2124


zhenyao:
@kenrussell @kainino0x PTAL

zhenyao:
I can do 1.0.2 and 1.0.1 after this

kenrussell:
Great. Thanks Mo for taking care of this. LGTM. Merging now.

PR #2240, opened 2017-01-06
Use vp9+webm video tagged with bt601 instead of undefined colorspace.

jdashg:
Firefox chooses bt709 for vp9 when colorspace isn't tagged on a file.
Tagging the colorspace as bt601 causes textures/video/* to pass in Firefox.


kenrussell:
Thanks, nice! Tested on Chrome on macOS and Windows; still working there.

Out of curiosity, what tool did you use to do the transcoding? Could you put the command line you used here?

Merging now.

jdashg:
My colleague Thomas Daede from the Daala team handled it. I will ask.
We should consider getting some alternate colorspace files and adding tests for those.

jdashg:
We used vpxenc to just re-encode the video explicitly tagging as bt601.

qkmiao:
This pr caused some video related fail. I fixed them at https://github.com/KhronosGroup/WebGL/pull/2247.

msc-:
> We used vpxenc to just re-encode the video explicitly tagging as bt601.

Does this actually do a color space conversion from bt709 or does it simply label the video as bt601?

PR #2243, opened 2017-01-06
Add OES_EGL_image_external extension

xingri:
* Add VideoFrameInfo interface and function to return it.

### DIFF
--- a/extensions/proposals/OES_EGL_image_external/extension.xml
+++ b/extensions/proposals/OES_EGL_image_external/extension.xml
@@ -58,10 +58,8 @@ interface OES_EGL_image_external {
     const GLenum TEXTURE_BINDING_EXTERNAL_OES     = 0x8D67;
     const GLenum REQUIRED_TEXTURE_IMAGE_UNITS_OES = 0x8D68;
 
-    [RaisesException] void EGLImageTargetTexture2DOES(
+    [RaisesException] WebGLVideoFrameInfo EGLImageTargetTexture2DOES(
         GLenum target, HTMLVideoElement video);
-    [RaisesException] WebGLVideoFrameInfo getVideoFrameInfo(
-        HTMLVideoElement video); 
 };
   </idl>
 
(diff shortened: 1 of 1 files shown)


extensions/proposals/OES_EGL_image_external/extension.xml
```diff
@@ -44,6 +45,13 @@
 
   <idl xml:space="preserve">
 [NoInterfaceObject]
+interface WebGLVideoFrameInfo {
+  readonly attribute double currentTime;
+  readonly attribute unsigned long textureWidth;
+  readonly attribute unsigned long textureHeight;
+};
```
RafaelCintron: Instead of adding this structure, consider adding these attributes to the video element itself.  currentTime, in particular, seems like it would have applications outside of WebGL.

xingri: @RafaelCintron Thanks for the review. It is added to notify application what currently WebGL is rendering of which frame of Video. Application can detect the delaying time and adjust audio synchronization etc.

extensions/proposals/OES_EGL_image_external/extension.xml
```diff
@@ -44,13 +45,20 @@
...
+
+[NoInterfaceObject]
 interface OES_EGL_image_external {
     const GLenum TEXTURE_EXTERNAL_OES             = 0x8D65;
     const GLenum SAMPLER_EXTERNAL_OES             = 0x8D66;
     const GLenum TEXTURE_BINDING_EXTERNAL_OES     = 0x8D67;
     const GLenum REQUIRED_TEXTURE_IMAGE_UNITS_OES = 0x8D68;
 
-    [RaisesException] void EGLImageTargetTexture2DOES(
+    [RaisesException] WebGLVideoFrameInfo EGLImageTargetTexture2DOES(
```
kenrussell: I've studied Chrome's media implementation a bit and it looks like this form of the API can work. The last video frame returned by the decoder can be held on to indefinitely. If you're OK with calling this explicitly once per frame, then this is a pretty simple solution to the problem of knowing exactly which frame from the video was latched in to the TEXTURE_EXTERNAL_OES texture.

PR #2245, opened 2017-01-06
Fix nullable behaviors in 1.0.1

zhenyao:
This is following https://github.com/KhronosGroup/WebGL/pull/2124
and is identical to https://github.com/KhronosGroup/WebGL/pull/2239 and https://github.com/KhronosGroup/WebGL/pull/2244


zhenyao:
@kenrussell @kainino0x FYI
Merging myself

PR #2247, opened 2017-01-07
Fix vp9 webm video resource name

qkmiao:
Updated file name according https://github.com/KhronosGroup/WebGL/pull/2240. 
PTAL @kenrussell @zhenyao @jdashg


kainino0x:
LGTM

PR #2249, opened 2017-01-10
Fix vp9 webm resource name in 2.0.0 snapshot.

kenrussell:
This is a port of #2247 to the 2.0.0 snapshot, which was broken with #2240 .


kenrussell:
Merging myself. Tested locally.

PR #2250, opened 2017-01-10
Tests rendering with integer texture non-zero base level

qkmiao:


qkmiao:
This test fails in chrome on NVIDIA Windows. I filed chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=679639.

kenrussell:
Is this a Chrome bug or does it fail in Firefox in the same configuration? CC'ing @jdashg .

qkmiao:
It's not a chrome bug. It's a NVIDIA Windows driver bug. Firefox also fails on this test.

zhenyao:
Looks good.

sdk/tests/conformance2/rendering/draw-with-integer-texture-base-level.html
```diff
@@ -0,0 +1,85 @@
...
+var canvas;
+var wtu = WebGLTestUtils;
+canvas = document.getElementById("example");
+var gl = wtu.create3DContext(canvas, undefined, 2);
+var tiu = TexImageUtils;
+
+function testDrawIntegerTextureBaseLevel()
+{
+    description("This test verifies the functionality of rendering with integer texture non-zero base level.");
+
```
zhenyao: If you have a chromium bug, can you add a comment to point to the bug link? Or provide some details like on which platform it fails, the fashion of the failure.

qkmiao: Added failed platforms in comments.

zhenyao:
Thanks.

PR #2251, opened 2017-01-10
fix a bug in image_bitmap related tests

Richard-Yunchao:
PTAL. Thanks!
@zhenyao , @kenrussell , @jdashg . Please see my comment at https://github.com/KhronosGroup/WebGL/pull/2224/files


zhenyao:
LGTM

PR #2252, opened 2017-01-10
Fix in npot-video-sizing.html

Richard-Yunchao:
Just a typo, but make the test fail. 

PTAL. @kenrussell , @zhenyao .


kenrussell:
Thanks @Richard-Yunchao . For the record, this was a new test added in #2202 .

PR #2253, opened 2017-01-10
Fix bug in webgl-compressed-texture-etc.html

Richard-Yunchao:
@kenrussell , @zhenyao , @jdashg , PTAL. 

FYI, @qkmiao 

Please see my comment at https://github.com/KhronosGroup/WebGL/pull/2210/files.


zhenyao:
See my reply in https://github.com/KhronosGroup/WebGL/pull/2210/

I don't think this pull request is correct.

Richard-Yunchao:
@zhenyao , @kenrussell , @jdashg , Could you take another look? Thanks a lot.

zhenyao:
lgtm

PR #2254, opened 2017-01-11
Fix bugs in attrib-type-match.html

Richard-Yunchao:
Sorry, I made a mistake for this test. PTAL. @zhenyao  and @kenrussell . 

vertexAttribIPointer have no parameter of normalized... 

I am surprised that the test can pass in Chromium. Per my understanding of IDL file and JS binding grammar, if the function have 5 parameters, and we fill it with 6 and call the function, then it can not be found in WebGLRenderingContext( or WebGL2RenderingContext) during function overloading.


kenrussell:
Thanks @Richard-Yunchao for the fix. I think these bugs were present since the test was added.

ECMAScript allows functions to be called with "too many" arguments. The remaining arguments are ignored. I think they're accessible via the "arguments" array though (the use of which is strongly discouraged).

"false" coerces to 0, which is why these tests were passing.

FYI, this refers to earlier pull requests #1863 and #1941.

Looks good. Merging now.

Richard-Yunchao:
Got it. Thanks for your explanation, @kenrussell .

PR #2255, opened 2017-01-12
Rejected EXT_texture_storage per discussion on public_webgl list.

kenrussell:


kenrussell:
As discussed. Merging myself. @jdashg @zhenyao @grorg @RafaelCintron

PR #2256, opened 2017-01-12
Fill in missing details in WEBGL_multiview spec

Oletus:
This adds specs for getFramebufferAttachmentParameter(), and extension
macros in shaders. The added tokens, added functions and errors are
documented with the correct syntax.


Oletus:
The specs for errors that are specific to the WebGL version of the extension are now duplicated between the sections for addendums and errors. What's the convention here, should they be removed from the addendums section?

kenrussell:
I don't think there's a strong convention yet regarding in which section these errors should be documented. Let's merge this now and refine it later. Good work moving this forward.

PR #2257, opened 2017-01-12
Add a l-value nested dynamic indexing test

Oletus:
This is a repro case for the following ANGLE bug:
https://bugs.chromium.org/p/angleproject/issues/detail?id=1672


kenrussell:
Excellent reduced test case Olli. LGTM.

PR #2261, opened 2017-01-15
Fix texImage3D and clearBuffer spec bugs found by Kathleen Mattson.

kenrussell:
Clarified use of clearBufferfi by incorporating some text from the ES
3.0 manual page.


kenrussell:
@grorg @jdashg @zhenyao @RafaelCintron @kainino0x FYI.

I'm taking the liberty of merging this myself so that Kathleen and Khronos have a fixed spec to refer to while creating the WebGL 2.0 reference card.

PR #2159, opened 2016-11-17
Added conformance tests for OffscreenCanvas

xidachen:
The spec for OffscreenCanvas is here:
https://html.spec.whatwg.org/multipage/scripting.html#the-offscreencanvas-interface

In this pull request, we add two simple tests only, which is to ensure that webgl and webgl2 context creation is correct.

Let's put this pull request on hold for few days until this change lands because currently creating webgl2 context is not supported in chromium yet and the test will fail:
https://codereview.chromium.org/2509193002/

@kenrussell @zhenyao Please review.

PR #2264, opened 2017-01-16
Skip OffscreenCanvas test if it's not supported

qkmiao:
finishTest() doesn't break the execution of left part of a test. We
should explicitly stop doing following test after finishTest().

This fixes some issues introduced in
https://github.com/KhronosGroup/WebGL/pull/2159.

### DIFF
--- a/sdk/tests/conformance/offscreencanvas/context-creation-worker.html
+++ b/sdk/tests/conformance/offscreencanvas/context-creation-worker.html
@@ -39,21 +39,22 @@
 <script>
 description("This test ensures that the WebGL context can be created on an OffscreenCanvas.");
 
-if (!window.OffscreenCanvas)
+if (!window.OffscreenCanvas) {
+  testSkipped("OffscreenCanvas is not supported.");
   finishTest();
-
-var worker = new Worker('context-creation-worker.js');
-worker.postMessage("Start worker");
-worker.onmessage = function(e) {
-  if (e.data == "Test passed") {
-    testPassed("All tests have passed");
-  } else {
-    testFailed("Some tests failed");
+} else {
+  var worker = new Worker('context-creation-worker.js');
+  worker.postMessage("Start worker");
+  worker.onmessage = function(e) {
+    if (e.data == "Test passed") {
+      testPassed("All tests have passed");
+    } else {
+      testFailed("Some tests failed");
+    }
+    finishTest();
   }
-  finishTest();
 }
-var successfullyParsed = true;
-</script>
 
+</script>
 </body>
 </html>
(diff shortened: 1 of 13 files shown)


qkmiao:
@kenrussell @zhenyao @xidachen PTAL

kainino0x:
LGTM but I'm not really qualified to comment on the addition of testSkipped.

sdk/tests/conformance/offscreencanvas/context-creation-worker.html
```diff
@@ -39,21 +39,22 @@
 <script>
 description("This test ensures that the WebGL context can be created on an OffscreenCanvas.");
 
-if (!window.OffscreenCanvas)
+if (!window.OffscreenCanvas) {
+  testSkipped("OffscreenCanvas is not supported.");
```
kenrussell: Here and throughout, please don't use testSkipped(). It prevents a 100% pass rate of the conformance suite when it's run via the webgl-conformance-tests.html harness.

Instead, please either just use debug() to report the fact that the test was skipped, or use testPassed().

qkmiao: Thanks for pointing this. Done.

sdk/tests/conformance/offscreencanvas/context-lost-restored.html
```diff
@@ -39,38 +39,46 @@
...
             // we didn't call prevent default so we should not be able to restore the context
             if (!compareGLError(gl.INVALID_OPERATION, "WEBGL_lose_context.restoreContext()")) {
                 testFailed("Some test failed");
                 finishTest();
+                return;
             }
             testLosingAndRestoringContext().then(function() {
                 testPassed("Test passed");
                 finishTest();
+                return;
```
kenrussell: This "return" looks unnecessary since it's at the end of a function.

sdk/tests/conformance/offscreencanvas/context-lost-restored.html
```diff
@@ -39,38 +39,46 @@
...
             }, function() {
                 testFailed("Some test failed");
                 finishTest();
+                return;
```
kenrussell: Same here.

sdk/tests/conformance/offscreencanvas/context-lost-restored.html
```diff
@@ -39,38 +39,46 @@
...
             });
        }, 0);
     });
     canvas.addEventListener("webglcontextrestored", function() {
         testFailed("Some test failed");
         finishTest();
+        return;
```
kenrussell: Same here.

PR #2265, opened 2017-01-16
Report INVALID_ENUM for ETC2/EAC formats without WEBGL_compressed_texture_etc

qkmiao:


sdk/tests/conformance/extensions/webgl-compressed-texture-etc.html
```diff
@@ -140,11 +140,12 @@
 
     debug("");
     if (contextVersion >= 2) {
+      var expectedError = isPositive ? gl.INVALID_OPERATION: [gl.INVALID_ENUM, gl.INVALID_OPERATION];
```
kainino0x: It seems a little odd to allow two either INVALID_ENUM or INVALID_OPERATION. But the difference may not matter enough to specify and fix.

kainino0x:
@qkmiao thanks for this fix. I added one comment for discussion.

kenrussell:
It looks like these calls can generate one or the other error, which is why both have to be allowed for WebGL 2.0.

Thanks @qkmiao for this fix. Merging now.

sdk/tests/conformance/extensions/webgl-compressed-texture-etc.html
qkmiao: We don't know which error will be raised early by implementations, either INVALID_ENUM or INVALID_OPERATION is possible in this case.

PR #2266, opened 2017-01-18
Fixed nullability of getUniform's arguments in WebGL 2.0.

kenrussell:
The arguments were made non-nullable in WebGL 1.0. Updating this
detail documentation was overlooked.


kenrussell:
@jdashg @grorg @zhenyao @kainino0x @RafaelCintron FYI. This was a small oversight so I'm merging it myself.

PR #2268, opened 2017-01-19
Disable multisample.fbo_max_samples.* in 2.0.0

zhenyao:
Due to NVidia Quadro 600 driver bug.


zhenyao:
@kenrussell @kainino0x @jdashg @Oletus FYI. Merging myself.

kenrussell:
LGTM after the fact. These are just a few small suppressions.

PR #2270, opened 2017-01-22
Split gl-uniform-arrays.html to avoid timeout

Jiawei-Shao:
There are a lot of calls of gl.getUniform() in gl-uniform-arrays.html,
which is very time consuming on a debug version of Chrome and will
make the whole test spend quite a long time to finish.

See http://crbug.com/603906 for more info.

This patch intends to fix this issue by splitting this test into smaller ones.


sdk/tests/conformance/uniforms/gl-uniform-arrays.html
```diff
@@ -305,206 +288,9 @@
...
-  for (var ii = 0; ii < test.maxUniformVectors; ++ii) {
-    gl.uniform4fv(locations[ii], [ii + 4, ii + 2, ii + 3, ii + 1]);
-  }
-
-  var kEpsilon = 0.01;
-  var isSame = function(v1, v2) {
-    return Math.abs(v1 - v2) < kEpsilon;
-  };
-
-  for (var ii = 0; ii < test.maxUniformVectors; ++ii) {
```
Jiawei-Shao: It takes about 10 seconds to finish this loop on a debug version of Chrome. 
As the loop is totally executed for 18 times, it is the most time consuming part in this test.

Jiawei-Shao:
@qkmiao PTAL. Thanks!

sdk/tests/conformance/uniforms/gl-uniform-arrays.html
```diff
@@ -305,206 +288,9 @@
...
-    var location = locations[ii];
-    if (location) {
-      var value = gl.getUniform(program, locations[ii]);
```
Jiawei-Shao: It takes about 10 seconds to finish this loop on a debug version of Chrome.
As the loop is totally executed for 18 times, it is the most time consuming part in this test.

qkmiao:
Looks good to me. This should fix some timeout for Debug bot.

Jiawei-Shao:
Now I am investigating why gl.getUniform is so slow on Debug Chromium.

zhenyao:
Thanks.  Once we fix the slowness, we might want to merge back tests into one.

PR #2272, opened 2017-01-24
Fix integer precision issue for cube map texture sampling on Android

qkmiao:
Highp integers have exactly 32 bits of precision, while mediump and lowp
integers have implementation-defined numbers of bits, see section 4.1.3
Integers in ESSL 3.00 spec. Default precision of integer in fragment
shader is mediump, see section 4.5.4 Default Precision Qualifiers in
ESSL 3.00 spec. So the condition expression in cube map texture sampling
shaders will be undefined. This caused cube map related texture
conformance tests to fail on Andorid where mediump may be implemented in
16 bits.


qkmiao:
With this pr, all conformance2/textures/* passed on Nexus 6.  This fixed chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=569836.

kenrussell:
*Thank you* Qiankun for tracking down this test bug.

Would you please merge it back to the conformance-suites/2.0.0 snapshot, too? This has been a longstanding bug preventing conformance on Android.

Thank you again!

qkmiao:
Backported to 2.0.0 snapshot.

zhenyao:
Thank you for getting the bottom of this!

LGTM

PR #2275, opened 2017-01-26
Fix timeout when video type is unsupported

qkmiao:


zhenyao:
LGTM
Can you also do the same with 2.0.0 snapshot?

PR #2276, opened 2017-01-27
Fix timeout in 3D image-bitmap-from-video tests and backport 2D fixup to 2.0.0 snapshot

qkmiao:
Fixed 3D image-bitmap-from-video tests and backported following two prs:
1. Fix timeout when video type is unsupported (#2275).
2. Use wtu.startPlayingAndWaitForVideo in image-bitmap-from-video tests.
(#2184).


zhenyao:
lgtm
Thank you

Issue #2273, opened 2017-01-24
Test suite depends on random images on the web


bzbarsky:
Specifically, the test suite requires loading of http://www.opengl.org/img/opengl_logo.jpg (which is not good at all, but at least you could make the argument that the Khronos group controls that image) and of http://mashable.com/wp-content/uploads/2008/08/thunderbird-logo-64x64.png (which is just completely unacceptable, because it could disappear at any time, breaking the test suite).

zhenyao:
Can you point out which tests use these images?

bzbarsky:
readPixelsBadArgs.html, origin-clean-conformance.html, texImage2DHTML.html, texSubImage2DHTML.html

But really, you can probably use grep just as well as I can...

zhenyao:
They are used for cross origin testing and there is an option to use the local url.

bzbarsky:
I understand what they are used for.  I'm saying that using files that you don't control for that purpose is completely nuts.

bzbarsky:
For practical purposes in terms of running this stuff in automation, by the way, it would be a lot more useful if the harness allowed providing a list of possible hostnames it might run at as opposed to hardcoding localhost/127.0.0.1.  But we can solve this in other ways in automation too.  This issue is really about the hosted version of the tests at https://www.khronos.org/registry/webgl/sdk/tests/webgl-conformance-tests.html for example.

kenrussell:
Fixed per above. These resources all now point to the Khronos-controlled domain get.webgl.org.

bzbarsky:
Thank you!

PR #2277, opened 2017-01-27
Put WebGL conformance resources that must be cross-domain on get.webg…

kenrussell:
…l.org.

Thanks to webmaster James Riordon for the idea.

Partially addresses #2273. Once resources are verified, tests will be updated.


kenrussell:
I'm merging this myself to test the availability of these resources. Updates to the tests will follow.

PR #2278, opened 2017-01-27
Updated links to cross-origin resources to point to get.webgl.org.

kenrussell:
Will test these live on the khronos.org site and then backport the
changes to the earlier conformance snapshots.

Partially addresses #2273.


kenrussell:
Merging myself. Tested locally and also modified webgl-test-utils.js locally to force loading of the resources from get.webgl.org. Need to test on the live server before backporting.

PR #2279, opened 2017-01-27
Relax image comparison threshold a bit to allow rounding up/down beha…

zhenyao:
…viors.


zhenyao:
Simple fix. Merging myself.

@kenrussell @kainino0x @jdashg PTAL.

kainino0x:
LGTM

kenrussell:
LGTM after the fact.

PR #2280, opened 2017-01-28
Update conformance snapshots to pull cross-origin resources from get.…

kenrussell:
…webgl.org.

Addresses #2273.


kenrussell:
Double-checked these by hand. Merging myself.

PR #2281, opened 2017-01-29
Add tolerance for pixels checking in image_bitmap_from_video tests

qkmiao:
This makes the new added video type tests pass.

I saw these tests failed on chromium bots in roll: https://codereview.chromium.org/2664653002/. But I cannot reproduce them locally. Failed message is something like: "at (8, 8) expected: 255,0,0 was 253,1,0".


zhenyao:
lgtm

zhenyao:
does this affect older snapshot?

kenrussell:
Thanks for tracking this down @qkmiao .

Issue #2282, opened 2017-02-03
Improve WebGL's power handling


kenrussell:
Use of the WebGL API currently implies use of the high-power discrete GPU on dual-GPU systems. A "preferLowPowerToHighPerformance" context creation attribute was specified some time ago, but it hasn't been widely implemented, nor used by web pages.

After discussion in the working group, the following decisions were made:

1) replace preferLowPowerToHighPerformance with an enumerant:
powerPreference = "low_power", "high_performance", "default"

2) getContextCreationAttributes should return the user's powerPreference request *unless* the system had to override it for some reason.

3) Advertise to WebGL developers that the WEBGL_lose_context is a way to forcibly terminate their context.

4) Advertise to WebGL developers that their context may be lost at any time -- for example, if a tab using WebGL has been in the background for a long time.

The working group notes that it's very difficult to handle this automatically. On macOS, it appears to be possible to dynamically switch between the integrated and discrete GPU for running WebGL content without losing the context ( https://codereview.chromium.org/2627323007/ ). The same is not possible on Windows; it is not possible to seamlessly migrate multisampled renderbuffers between GPUs, for example.

Edge attempted to dynamically switch between the integrated and discrete GPUs based on how much GPU load the WebGL application induced. (Multisampled renderbuffers were resolved to single-sample textures before migration, and other corner cases were handled as well.) The result was flaky; developers could not reliably use either GPU, so the mechanism was removed from Edge. Other browser implementers should learn from this experience.

grorg:
Oops. I didn't notice this when I started my work :(

RafaelCintron:
@kenrussell , the section in the issue about Edge is not quite correct.  

In previous versions of Microsoft's WebGL implementation, there was a feature that monitored WebGL performance of the system using disjoint timer queries.  To avoid runaway WebGL pages from adversely interfering with the responsiveness of the desktop, the feature kicked the website into software mode if it encountered too many long frames.  The rationale is that CPUs are generally much better able to preemptively schedule workloads than GPUs.  

The feature was ultimately removed from the product because CPU workloads were easy to become entangled with GPU measurements using disjoint timer queries. This led to pages getting pushed into software mode unintentionally, which caused developer frustration.  We decided to instead rely on the system being able to let the user close runaway browser tabs when they are not satisfied with performance.   

[...]

PR #2283, opened 2017-02-03
Rename preferLowPowerToHighPerformance to powerPreference.

kenrussell:
Document that WEBGL_lose_context extension should destroy graphics
resources.

### DIFF
--- a/extensions/WEBGL_lose_context/extension.xml
+++ b/extensions/WEBGL_lose_context/extension.xml
@@ -37,13 +37,19 @@ interface WEBGL_lose_context {
   </idl>
   <newfun>
     <function name="loseContext" type="void">
-      When this function is called and the context is not lost, simulate
+      <p>When this function is called and the context is not lost, simulate
       losing the context so as to trigger the steps described in the WebGL
       spec for handling context lost. The context will remain in the lost
       state according to the WebGL specification until
       <code>restoreContext()</code> is called.  If the context is already
       lost when this function is called, generate an
-      <code>INVALID_OPERATION</code> error.
+      <code>INVALID_OPERATION</code> error.</p>
+
+      <p>Implementations should destroy the underlying graphics context and
+      all graphics resources when this method is called. This is the
+      recommended mechanism for applications to programmatically halt their
+      use of the WebGL API.</p>
+
     </function>
     <function name="restoreContext" type="void">
       When this function is called while the context is lost, and the conditions
@@ -99,6 +105,10 @@ interface WEBGL_lose_context {
     </revision>
     <revision date="2014/07/15">
       <change>Added NoInterfaceObject extended attribute.</change>
+    </revision>
+    <revision date="2017/02/03">
+      <change>Per discussion in working group, document that this extension
+      should destroy the underlying context and all graphics resources.</change>
     </revision>
   </history>
 </ratified>
--- a/extensions/WEBGL_lose_context/extension.xml
+++ b/extensions/WEBGL_lose_context/extension.xml
@@ -37,19 +37,13 @@ interface WEBGL_lose_context {
   </idl>
   <newfun>
     <function name="loseContext" type="void">
-      <p>When this function is called and the context is not lost, simulate
+      When this function is called and the context is not lost, simulate
       losing the context so as to trigger the steps described in the WebGL
       spec for handling context lost. The context will remain in the lost
       state according to the WebGL specification until
       <code>restoreContext()</code> is called.  If the context is already
       lost when this function is called, generate an
-      <code>INVALID_OPERATION</code> error.</p>
-
-      <p>Implementations should destroy the underlying graphics context and
-      all graphics resources when this method is called. This is the
-      recommended mechanism for applications to programmatically halt their
-      use of the WebGL API.</p>
-
+      <code>INVALID_OPERATION</code> error.
     </function>
     <function name="restoreContext" type="void">
       When this function is called while the context is lost, and the conditions
@@ -105,10 +99,6 @@ interface WEBGL_lose_context {
     </revision>
     <revision date="2014/07/15">
       <change>Added NoInterfaceObject extended attribute.</change>
-    </revision>
-    <revision date="2017/02/03">
-      <change>Per discussion in working group, document that this extension
-      should destroy the underlying context and all graphics resources.</change>
     </revision>
   </history>
 </ratified>
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -807,7 +807,6 @@ enum WebGLPowerPreference { "default", "low-power", "high-performance" };
                 only a hint and a WebGL implementation may choose to ignore it.
               </p>
               <p>
-
                 WebGL implementations use context lost and restored events to regulate power and
                 memory consumption, <em>regardless</em> of the value of this attribute. <em>At
                 WebGLRenderingContext creation time, this attribute will be ignored if no event
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -807,12 +807,12 @@ enum WebGLPowerPreference { "default", "low-power", "high-performance" };
                 only a hint and a WebGL implementation may choose to ignore it.
               </p>
               <p>
+
                 WebGL implementations use context lost and restored events to regulate power and
-                memory consumption, <em>even if</em> this attribute is not specified. <em>The
-                application must have registered an event listener which would handle the
-                <code>"webglcontextlost"</code> and <code>"webglcontextrestored"</code> events if
-                they were dispatched to the canvas element before fetching a WebGLRenderingContext
-                specifying this attribute, or "default" will be used.</em>
+                memory consumption, <em>regardless</em> of the value of this attribute. <em>At
+                WebGLRenderingContext creation time, this attribute will be ignored if no event
+                listeners exist which would handle both <code>"webglcontextlost"</code>
+                and <code>"webglcontextrestored"</code> events dispatched to the canvas element.</em>
               </p>
               <p>
                 The allowed values are:
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -829,7 +829,7 @@ enum WebGLPowerPreference { "default", "low-power", "high-performance" };
                     <em>high-performance</em>
                   </p>
                   <p>
-                    Indicates a request for a GPU configuration that values high rendering
+                    Indicates a request for a GPU configuration that prioritizes high rendering
                     performance over power consumption.  Any content that chooses this value should
                     be aware that it is more likely that the user agent will force a lost context at
                     any time. Implementations may decide to initially respect this request and,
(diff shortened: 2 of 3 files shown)


kenrussell:
Partially addresses #2282 .

@grorg @jdashg @RafaelCintron @zhenyao @kainino0x in particular please review.

Thanks.

specs/latest/1.0/index.html
```diff
@@ -723,14 +723,19 @@ <h2 class="no-toc">Table of contents</h2>
         The <code>WebGLContextAttributes</code> dictionary contains drawing surface attributes and
         is passed as the second parameter to <code>getContext</code>.
     </p>
-    <pre class="idl">dictionary <dfn id="WebGLContextAttributes">WebGLContextAttributes</dfn> {
+    <!-- The capitalization rules here match those of the crossOrigin
+    attribute, which was added to HTMLMediaElement on behalf of
+    WebGL. -->
+    <pre class="idl">enum PowerPreference { "default", "low-power", "high-performance" };
```
jdashg: I do sort of like "balanced" instead of "default", but "default" is fine.

extensions/WEBGL_lose_context/extension.xml
```diff
@@ -37,13 +37,19 @@ interface WEBGL_lose_context {
...
+      <p>When this function is called and the context is not lost, simulate
       losing the context so as to trigger the steps described in the WebGL
       spec for handling context lost. The context will remain in the lost
       state according to the WebGL specification until
       <code>restoreContext()</code> is called.  If the context is already
       lost when this function is called, generate an
-      <code>INVALID_OPERATION</code> error.
+      <code>INVALID_OPERATION</code> error.</p>
+
+      <p>Implementations should destroy the underlying graphics context and
```
jdashg: Thanks for adding this text, though it'd be nice to have it in a separate PR.

specs/latest/1.0/index.html
kainino0x: I'm not sure if "balanced" describes what we'll be doing. On an ACPI system, "balanced" usually refers to automatic frequency scaling. If this mostly controls which GPU is used, "default" makes more sense.

kainino0x:
Hm, any chance we can agree on meanings for all 3 values ("default", "low-power", "high-performance")? Specifically, the meaning of "default" is unclear to me.

specs/latest/1.0/index.html
grorg: This needs to have a WebGL prefix.

`enum WebGLPowerPreference { ...`

specs/latest/1.0/index.html
```diff
@@ -723,14 +723,19 @@ <h2 class="no-toc">Table of contents</h2>
...
+
+dictionary <dfn id="WebGLContextAttributes">WebGLContextAttributes</dfn> {
     GLboolean alpha = true;
     GLboolean depth = true;
     GLboolean stencil = false;
     GLboolean antialias = true;
     GLboolean premultipliedAlpha = true;
     GLboolean preserveDrawingBuffer = false;
-    GLboolean preferLowPowerToHighPerformance = false;
+    PowerPreference powerPreference = "default";
```
grorg: Should be WebGLPowerPreference.

grorg:
The commit wording suggests that this change is just about the context lost extension, but the title is about the renaming of the attribute. The commit does a bit of both.

grorg:
I'm not sure if this should be called `performancePreference` or `powerPreference`. I don't have strong opinions.

In my PR #2295 , I used the following wording.

performancePreference

 Provides a hint to the user agent indicating what configuration of GPU is suitable
for this WebGL context. This may influence which GPU is used in a system with
multiple GPUs. For example, a dual-GPU system might have one GPU that consumes less
power at the expense of rendering performance. Note that this property is only a hint and
a WebGL implementation may choose to ignore it. The allowed values are:

*default*

Let the user agent decide which GPU configuration is most suitable. This is
the default value.

*high-performance*

Indicates a request for a GPU configuration that values high
rendering performance over power consumption.
Any content that chooses this value should
be aware that it is much more likely that the user agent will
force a lost context at any time. Developers should always provide
event handlers for the context lost and
context restored events if they specify this value. Implementations may
decide to initially respect this request and, after some time, choose to
lose the context and restore a new context ignoring the request. Developers
are encouraged to only specify this value if they believe it is absolutely
necessary, since it may degrade the overall system performance and consume more
of the user's battery.

*low-power*

Indicates a request for a GPU configutation that values power consumption
over rendering performance. Generally this means the content is unlikely to be
constrained by drawing. For example, it may render only one frame a
second, it does not issuing draw calls with a lot of geometry or complex
shaders, or it is a very small HTML canvas element. Developers are encouraged
to specify this value if their content allows, since it may consume less
of the user's battery.

kainino0x:
I'll move my comments from #2295 to this thread.

> _high-performance_

Overall I really like this wording, but I'm concerned that developers may read this as "you can either get high performance with low reliability or vice versa" which sounds scary and isn't accurate.

Specifically, I think we should make it clear that if a user is actively using "high" content, it shouldn't spontaneously lose context to save power (dunno how that's implemented). Here is also a good place to make it clear that recovering from a lost context isn't that hard - since the JS state is preserved.

> _low-power_

How about this slightly tweaked wording?

Indicates a request for a GPU configuration that prioritizes power saving over rendering performance. Generally, content should use this if it is unlikely to be constrained by drawing performance; for example, if it renders only one frame a second, draws only relatively simple geometry with simple shaders, or uses a very small HTML canvas element. Developers (...)

specs/latest/1.0/index.html
grorg: Also, isn't this a pre class idl inside an existing pre class idl? No need to wrap it.

I suggest that the enum go one section above though, in the types section.

RafaelCintron:
I am fine with wording that provides examples of instances where the browser may decide to lose-context on your canvas.  However, I think we should set clear expectations that context lost can happen no matter what, even when you ask for "low" and get "low".  On memory constrained machines like low end phones (which only have one GPU) dumping canvas contexts that use lots of memory is a good way to make room for other foreground applications or browser tabs.

grorg:
I've landed the rename in WebKit, so I'm no longer proposing a name change :)
https://trac.webkit.org/changeset/212316

Now I'm starting on implementing the modes a bit better. @kenrussell, any thoughts on the proposed wording?

kenrussell:
@grorg Working on incorporating the above feedback. Will push a new revision for review shortly.

extensions/WEBGL_lose_context/extension.xml
kenrussell: Undid this change; will submit it in a separate pull request.

kenrussell:
@grorg @jdashg @RafaelCintron revised based on all of your feedback. Dean, in particular, thanks for your starting text; I edited it somewhat but used large portions.

Please re-review. Thanks.

specs/latest/1.0/index.html
```diff
@@ -790,14 +797,60 @@ <h2 class="no-toc">Table of contents</h2>
...
+                  </p>
+                  <p>
+                    Let the user agent decide which GPU configuration is most suitable. This is the
+                    default value.
+                  </p>
+                  <p>
+                    <em>high-performance</em>
+                  </p>
+                  <p>
+                    Indicates a request for a GPU configuration that values high rendering
```
kainino0x: I would change "values" to "prioritizes" here as well.

kenrussell: Thanks for the suggestion. Done.

specs/latest/1.0/index.html
```diff
@@ -790,14 +797,60 @@ <h2 class="no-toc">Table of contents</h2>
...
+                less power at the expense of rendering performance. Note that this property is
+                only a hint and a WebGL implementation may choose to ignore it.
+              </p>
+              <p>
+                WebGL implementations use context lost and restored events to regulate power and
+                memory consumption, <em>even if</em> this attribute is not specified. <em>The
+                application must have registered
+                <code>"webglcontextlost"</code> and <code>"webglcontextrestored"</code> event
+                listeners on the canvas element before fetching a WebGLRenderingContext specifying
+                this attribute, or "default" will be used.</em>
```
grorg: I like this rule. I thought we only talked about it informally, but I'm glad to see it in the specification.

specs/latest/1.0/index.html
```diff
@@ -790,14 +797,60 @@ <h2 class="no-toc">Table of contents</h2>
...
+                  </p>
+                  <p>
+                    Let the user agent decide which GPU configuration is most suitable. This is the
+                    default value.
+                  </p>
+                  <p>
+                    <em>high-performance</em>
+                  </p>
+                  <p>
+                    Indicates a request for a GPU configuration that prioritizes high rendering
```
grorg: Maybe just "rendering performance" - removing "high"?

specs/latest/1.0/index.html
```diff
@@ -790,14 +797,60 @@ <h2 class="no-toc">Table of contents</h2>
...
+                  </p>
+                  <p>
+                    <em>low-power</em>
+                  </p>
+                  <p>
+                    Indicates a request for a GPU configuration that prioritizes power saving over
+                    rendering performance. Generally, content should use this if it is unlikely to
+                    be constrained by drawing performance; for example, if it renders only one frame
+                    per second, draws only relatively simple geometry with simple shaders, or uses a
+                    small HTML canvas element. Developers are encouraged to specify this value if
```
grorg: "are encouraged to use this value", rather than "specify"?

specs/latest/1.0/index.html
grorg: Actually, now that I think about this, the developer is free to add an event listener on any parent of the canvas element, up to the Document (and maybe even the window object).

So maybe the wording should be that there is a registered listener that would handle the lost/restore events if they were dispatched to the canvas element.

kenrussell:
Thanks @grorg for your review and further suggestions. Incorporated them. Please re-review.

kenrussell:
Thanks @grorg . Any other comments @jdashg @kainino0x @RafaelCintron ?

specs/latest/1.0/index.html
```diff
@@ -790,14 +797,60 @@ <h2 class="no-toc">Table of contents</h2>
...
+                less power at the expense of rendering performance. Note that this property is
+                only a hint and a WebGL implementation may choose to ignore it.
+              </p>
+              <p>
+                WebGL implementations use context lost and restored events to regulate power and
+                memory consumption, <em>even if</em> this attribute is not specified. <em>The
+                application must have registered an event listener which would handle the
+                <code>"webglcontextlost"</code> and <code>"webglcontextrestored"</code> events if
+                they were dispatched to the canvas element before fetching a WebGLRenderingContext
+                specifying this attribute, or "default" will be used.</em>
```
kainino0x: I find this hard to parse, how about:

WebGL implementations use context loss/restoration events to regulate power and memory consumption, <em>regardless</em> of the value of this attribute. At WebGLRenderingContext creation time, this attribute will be ignored if no event listeners exist which would handle both <code>"webglcontextlost"</code> and <code>"webglcontextrestored"</code> events dispatched to the canvas element.

kenrussell: Thanks for the suggestion. Incorporated your text.

kenrussell:
Further thoughts?

RafaelCintron:
LGTM.

kenrussell:
Thanks everyone. Merging now.

grorg:
Awesome. Thanks @kenrussell . WebKit's implementation will land over the next few days.

PR #2053, opened 2016-09-21
Remove ETC2/EAC compressed texture formats from WebGL 2.0 core

kainino0x:
This fixes #2030.

- [x] #2047
- [x] #2083

PR #2285, opened 2017-02-04
Update more extension tests to account for removal of ETC2/EAC

kainino0x:
This is fallout from #2053, in which I forgot to update the ATC tests.

This wasn't noticed because it runs only on limited configurations (as ATC is an AMD extension). I found it while debugging test failures on Google Pixel (Qualcomm Adreno 530), which supports it.

This corrects a failure on this device on both Chrome and Firefox.


zhenyao:
lgtm

PR #2289, opened 2017-02-08
Split attaching texture level 1 to framebuffer to a different test

Oletus:
This test is new in 2.0.1 and is failing on at least Chrome OpenGL
Windows using an NVIDIA GPU. It's easier to keep track of the failure
if the test is in a separate test file, rather than in a catch-all
framebuffer test file with lots of unrelated tests.

Also fix wrong order of shouldBe parameters in the test to get a
better error message.


zhenyao:
Yes, this is better.

PR #2290, opened 2017-02-09
Make compressed texture max size test reliable

Oletus:
ArrayBuffer allocations larger than 1 GB are not reliable on 32-bit
browsers. To make the compressed texture size conformance test
reliable, the max allocation size it uses needs to be limited down to
1 GB from the theoretical limit it was previously using.

To restore as much of the lost coverage as possible, cover max texture
size with 2D textures that are large only in one dimension and small
in the other dimension.

A separate stress test that uses ArrayBuffers up to the theoretical 4
GB limit is also added. This is valuable to have since it can reveal
bugs in 64-bit browsers that the lighter conformance test does not
reveal anymore. It also uses larger cube map textures in its positive
tests, raising the size limit from 2K x 2K to 8K x 8K.

### DIFF
--- a/conformance-suites/1.0.3/conformance/extensions/webgl-compressed-texture-size-limit.html
+++ b/conformance-suites/1.0.3/conformance/extensions/webgl-compressed-texture-size-limit.html
@@ -125,8 +125,9 @@ function getSharedArrayBufferSize() {
         sharedArrayBufferSize = bufferSizeNeeded;
       }
       bufferSizeNeeded = test.func(t.maxSize + test.sizeStep, t.maxSize + test.sizeStep);
-      // ArrayBuffers can be at most 4GB (minus 1 byte)
-      if (bufferSizeNeeded > sharedArrayBufferSize && bufferSizeNeeded <= 4294967295) {
+      // ArrayBuffers can be at most 4GB (minus 1 byte), but any allocations larger than 1 GB are unreliable in practice. So limit allocations to 1 GB.
+      // Textures that are wide in just one dimension can still be used to test max TEXTURE_2D size limit even if we don't allocate space for huge square textures.
+      if (bufferSizeNeeded > sharedArrayBufferSize && bufferSizeNeeded <= Math.pow(2, 30)) {
         sharedArrayBufferSize = bufferSizeNeeded;
       }
     }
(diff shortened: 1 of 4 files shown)


Oletus:
This affects conformance coverage of compressed cube map max size checks on a subset of platforms, and the changes need to be done in older versions of the conformance suite as well. Can I get comments from a few different people before merging?

kenrussell:
Thanks @Oletus for putting this together. The changes seem like a reasonable compromise given the realities of browsers on lower-end systems.

I looked a bit into trying to detect whether running on a 32-bit or 64-bit browser. It looks complicated and unreliable.

Any other comments @zhenyao @jdashg @kainino0x @RafaelCintron @grorg (in particular) before merging?

kainino0x:
This adds test expectations to 2.0.0 (in the "else"), but if it passes on Chrome+FF then I assume that's fine?

conformance-suites/1.0.3/conformance/extensions/webgl-compressed-texture-size-limit.html
```diff
@@ -217,11 +218,25 @@
       var dataSize = test.func(t.maxSize + test.sizeStep, t.maxSize + test.sizeStep);
       // this check assumes that each element is 1 byte
       if (dataSize > sharedArrayBuffer.byteLength) {
-        testPassed("Unable to test texture larger than maximum size due to ArrayBuffer size limitations -- this is legal");
+        if (t.target == gl.TEXTURE_CUBE_MAP) {
+          testPassed("Unable to test texture larger than maximum size due to ArrayBuffer size limitations -- this is legal");
+        } else {
```
kainino0x: (here)

Oletus: You are right that this technically increases coverage, but the intent has been to maintain the existing coverage as closely as possible. Since max size square textures can't be tested as much after this change, it's been replaced with textures that are large only in one dimension.

If people would rather just remove coverage from the older conformance tests, I'd be okay with that, but if this isn't a big issue to anyone, I could merge the existing version. Any more comments on this?

RafaelCintron:
Confirmed that WebGL 1.0 test pass on Edge.  The change LGTM.

zhenyao:
lgtm

kenrussell:
@Oletus the way you've phrased it is fine in my opinion.

Since all parties seem to be in agreement, let's make this change. Thanks for thinking about and fixing this.

PR #2292, opened 2017-02-09
Update more conformance tests to OffscreenCanvas

xlai-o:
@kenrussell 
The two new tests that are adding here are adapted from existing Chromium layout tests:
1. offscreencanvas-resize.html
https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/fast/canvas/OffscreenCanvas-resize.html
The resizing behavior is specified in WHATWG spec https://html.spec.whatwg.org/multipage/scripting.html#the-offscreencanvas-interface,

2. context-attribute-preserve-drawing-buffer.html
https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/fast/canvas/OffscreenCanvas-webgl-preserveDrawingBuffer.html

### DIFF
--- a/sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
+++ b/sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
@@ -1,0 +1,76 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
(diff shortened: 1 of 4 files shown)


sdk/tests/conformance/offscreencanvas/offscreencanvas-resize.html
```diff
@@ -0,0 +1,109 @@
...
+      asyncStepsCompleted++;
+      if (asyncStepsCompleted == 2) {
+        finishTest();
+      }
+    });
+  }, 0);
+}
+
+if (!window.OffscreenCanvas) {
+    testPassed("No OffscreenCanvas support");
```
qkmiao: finishTest() here.

qkmiao: Keep same indent in the file

sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
```diff
@@ -0,0 +1,76 @@
...
+
+    // Test if OffscreenCanvas.webgl loses context if presereDrawingbuffer is false.
+    var pixelsNoPreserve = getPixelsFromOffscreenWebgl(false);
+    shouldBe(pixelsNoPreserve, [0,0,0,0]);
+}
+
+var successfullyParsed = true;
+
+</script>
+<script src="../../js/js-test-pre.js"></script>
```
qkmiao: js-test-pre.js -> js-test-post.js

kainino0x:
I think you might not need finishTest if you have js-test-post. (You can
double check this though)

sdk/tests/conformance/offscreencanvas/offscreencanvas-resize.html
xlai-o: Thanks. I've updated a new commit to address all your comments.

xlai-o:
>

xlai-o:
@kainino0x Not really; the test that I've used finishTest() are async test so in that particular case I can't just include js-test-post as the other non-async test.

kainino0x:
Oh, yes, you're right. Everything looks correct. Sorry, didn't look at the
actual code from my phone when I replied.

On Fri, Feb 10, 2017, 9:55 AM xlai <notifications@github.com> wrote:

@kainino0x <https://github.com/kainino0x> Not really; the test that I've
used finishTest() are async test so in that particular case I can't just
include js-test-post as the other non-async test.

—
You are receiving this because you were mentioned.


Reply to this email directly, view it on GitHub
<https://github.com/KhronosGroup/WebGL/pull/2292#issuecomment-279016459>,
or mute the thread

kenrussell:
Thanks Olivia for these new tests. LGTM too. Thanks @qkmiao and @kainino0x for your reviews as well.

PR #2293, opened 2017-02-09
Add a test for a chrome crash related with compressedTex* and PBO.

zhenyao:


sdk/tests/conformance2/textures/misc/compressed-tex-from-pbo-crash.html
```diff
@@ -0,0 +1,63 @@
...
+
+var gl = wtu.create3DContext("example", undefined, 2);
+
+var tex = gl.createTexture();
+gl.bindTexture(gl.TEXTURE_2D, tex);
+// The following statement caused a crash in Chrome. See crbug.com/690335.
+// Because no PBO is bound, the last parameter is interprected as a pointer
+// to the client side data.
+gl.compressedTexImage2D(gl.TEXTURE_2D, 0, 0, 4, 4, 0, 8, 3);
+
```
toji: I think we should be checking for some sort of GL error here, given that this is an invalid call. Otherwise LGTM.

kainino0x:
It could be invalid operation, value, or enum, I think.

sdk/tests/conformance2/textures/misc/compressed-tex-from-pbo-crash.html
zhenyao: Done.

zhenyao:
Addressed toji's comment. Merging myself.

PR #2294, opened 2017-02-10
Make some GLSL tests use the GLSL test framework

Oletus:
Use the GLSL test runner instead of having custom procedural code in
the tests. This makes the test code cleaner and the tests easier to
debug, since the GLSL runner prints out the shader source.


zhenyao:
Nice.  Thanks Olli.

PR #2295, opened 2017-02-13
Add performancePreference to WebGLContextAttributes

grorg:
- Draft wording for `performancePreference`, with values default, low and high.
- Remove `preferLowPowerToHighPerformance`

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -712,6 +712,10 @@ typedef unsigned short GLushort;
 typedef unsigned long  GLuint;
 typedef unrestricted float GLfloat;
 typedef unrestricted float GLclampf;
+
+enum WebGLPerformancePreference {
+    "default", "high", "low"
+};
 </pre>
 
 <!-- ======================================================================================================= -->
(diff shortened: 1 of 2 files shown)


grorg:
Ignore this. I hadn't noticed #2283

specs/latest/1.0/index.html
```diff
@@ -790,14 +794,40 @@ <h2 class="no-toc">Table of contents</h2>
...
+                        necessary, since it may degrade the overall system performance and consume more
+                        of the user's battery.</dd>
+                    <dt>low</dt>
+                    <dd>Indicates a request for a GPU configutation that values power consumption
+                        over rendering performance. Generally this means the content is unlikely to be
+                        constrained by drawing. For example, it may render only one frame a
+                        second, it does not issuing draw calls with a lot of geometry or complex
+                        shaders, or it is a very small HTML canvas element. Developers are encouraged
+                        to specify this value if their content allows, since it may consume less
+                        of the user's battery.
```
kainino0x: How about this slightly tweaked wording?

Indicates a request for a GPU configuration that prioritizes power saving over rendering performance. Generally, content should use this if it is unlikely to be constrained by drawing performance; for example, if it renders only one frame a second, draws only relatively simple geometry with simple shaders, or uses a very small HTML canvas element. Developers (...)

specs/latest/1.0/index.html
```diff
@@ -790,14 +794,40 @@ <h2 class="no-toc">Table of contents</h2>
...
+                        Any content that chooses this value should
+                        be aware that it is much more likely that the user agent will
+                        force a lost context at any time. Developers should always provide
+                        event handlers for the <a href="#WEBGLCONTEXTEVENT">context lost and
+                        context restored events</a> if they specify this value. Implementations may
+                        decide to initially respect this request and, after some time, choose to
+                        lose the context and restore a new context ignoring the request. Developers
+                        are encouraged to only specify this value if they believe it is absolutely
+                        necessary, since it may degrade the overall system performance and consume more
+                        of the user's battery.</dd>
```
kainino0x: Overall I really like this wording, but I'm concerned that developers may read this as "you can either get high performance with low reliability or vice versa" which sounds scary and isn't accurate.

Specifically, I think we should make it clear that if a user is actively using "high" content, it shouldn't spontaneously lose context to save power (dunno how that's implemented). Here is also a good place to make it clear that recovering from a lost context isn't that hard - since the JS state is preserved.

kainino0x:
Oh, didn't realize this wasn't orthogonal to #2283. I think we need to incorporate some of the wording here into #2283.

grorg:
I'm wondering if #2283 is complete. It seems the commit is doing two things at once, is missing descriptions of the values, and the commit message doesn't align with the title.

kenrussell:
#2283 has been revised to address these concerns. Please re-review there. Thanks.

PR #2296, opened 2017-02-14
Improve the description of WebGLContextEvent

grorg:
A slight tweak to the wording that indicates what
would cause such an event to fire. The motivation
behind this is that it was possible to read the
specification and conclude that a lost + restored
must be fired if a GPU switch occurs.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -3135,14 +3135,14 @@ WebGLRenderingContext implements WebGLRenderingContextBase;
     <h3><a name="WEBGLCONTEXTEVENT">WebGLContextEvent</a></h3>
 
     <p>
-        WebGL generates a <code>WebGLContextEvent</code> event in response to a status change to the
-        WebGL rendering context associated with the HTMLCanvasElement which has a listener for this
-        event. Events are sent using the DOM Event
-        System <a href="#refsDOM3EVENTS">[DOM3EVENTS]</a>. Event types can include the loss or
-        restoration of state, or the inability to create a context.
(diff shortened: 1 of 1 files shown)


specs/latest/1.0/index.html
```diff
@@ -3135,14 +3135,14 @@ <h2 class="no-toc">Table of contents</h2>
...
-        event. Events are sent using the DOM Event
-        System <a href="#refsDOM3EVENTS">[DOM3EVENTS]</a>. Event types can include the loss or
-        restoration of state, or the inability to create a context.
-        <a href="http://www.w3.org/TR/domcore/#eventinit">EventInit</a> is defined in the DOM4
-        specification <a href="#refsDOM4">[DOM4]</a>.
+        WebGL generates a <code>WebGLContextEvent</code> event in response to important changes in status of a
+        WebGL rendering context. Events are sent using the DOM Event System <a href="#refsDOM3EVENTS">[DOM3EVENTS]</a>,
+        and are dispatched to the HTMLCanvasElement associated with the WebGL rendering context.
+        The types of status changes that can trigger a <code>WebGLContextEvent</code> event are <a href="#CONTEXT_LOST">
+        the loss of state</a>, <a href="#CONTEXT_RESTORED">the restoration of state</a>, and <a href="#CONTEXT_CREATION_ERROR">
```
kenrussell: What about changing "the loss/restoration of state" to "the loss/restoration of the context"?

RafaelCintron:
LGTM

specs/latest/1.0/index.html
kenrussell: @grorg to move this along let me merge this and make that small change as a follow-up.

PR #2298, opened 2017-02-14
Fix negative case of webgl-draw-buffers-feedback-loop.html

kainino0x:
Fix negative case of conformance/extensions/webgl-draw-buffers-feedback-loop.html

This fixes a test failure when WEBGL_draw_buffers is unavailable.

@Richard-Yunchao @zhenyao PTAL


zhenyao:
LGTM
Thanks for fixing this.

PR #2301, opened 2017-02-16
Change "loss/restoration of state" to "loss/restoration of the context".

kenrussell:
Follow-up to #2296.


kenrussell:
@grorg @kainino0x @RafaelCintron @jdashg please review.

kainino0x:
LGTM

kenrussell:
I think this is non-controversial so merging it now to get it off my plate. Please revert if there's some concern.

PR #2302, opened 2017-02-17
Simplify a test by consilidating two test paths.

zhenyao:
Pure refactoring, no logic changing


zhenyao:
@toji Can you review this?
@kenrussell @kainino0x FYI

toji:
LGTM

kenrussell:
Very nice Mo. LGTM after the fact.

PR #2303, opened 2017-02-17
Test we don't set more uniform values than it should be with sub sour…

zhenyao:
…ce uniform*().

This is for crbug.com/693266

### DIFF
--- a/sdk/tests/conformance2/uniforms/gl-uniform-arrays-sub-source.html
+++ b/sdk/tests/conformance2/uniforms/gl-uniform-arrays-sub-source.html
@@ -63,6 +63,10 @@ var gl = wtu.create3DContext("example", undefined, 2);
 var vSrc = wtu.getScript("vshader");
 var fTemplate = wtu.getScript("fshader");
 
+// |color| is defined in fragment shader as an array of 3 elements.
+var uniformArraySize = 3;
+var initialUniformValue = 99;
+
 var typeInfos = [
   { type: 'float',
     setter: 'uniform1fv',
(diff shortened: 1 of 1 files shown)


zhenyao:
@kainino0x @kenrussell PTAL

This test passes in Firefox and Chrome with https://codereview.chromium.org/2703593003/
This test fails in Chrome without https://codereview.chromium.org/2703593003/

sdk/tests/conformance2/uniforms/gl-uniform-arrays-sub-source.html
```diff
@@ -258,11 +258,19 @@
 
 ];
 
-function checkUniformValues(program, typeInfo) {
+function checkUniformValues(program, typeInfo, initialValues) {
```
kenrussell: Could you rename "initialValues" to "useInitialValues"? Right now it sounds like an array rather than a boolean.

sdk/tests/conformance2/uniforms/gl-uniform-arrays-sub-source.html
```diff
@@ -258,11 +258,19 @@
...
   var numSrcValues = typeInfo.srcValues.length / typeInfo.numElementsPerType;
-  for (var ii = 0; ii < numSrcValues; ++ii) {
-    var expectedValues = typeInfo.srcValues.slice(
-        typeInfo.numElementsPerType * ii, typeInfo.numElementsPerType * (ii + 1));
+  for (var ii = 0; ii < 3; ++ii) {
```
kenrussell: What does the "3" mean here? Is it that the tests above (used to) test setting 3 elements of the associated type (vec3, vec4, etc.)?

sdk/tests/conformance2/uniforms/gl-uniform-arrays-sub-source.html
```diff
@@ -331,6 +340,16 @@
     }
   }
 
+  // Initialize uniforms to value 99 and verify it.
+  var initValues = new Array(typeInfo.numElementsPerType * 3);
+  for (var ii = 0; ii < typeInfo.numElementsPerType * 3; ++ii) {
```
kenrussell: Could you please define a variable with some documentation and use it in place of these instances of "3" and above?

sdk/tests/conformance2/uniforms/gl-uniform-arrays-sub-source.html
zhenyao: Done

sdk/tests/conformance2/uniforms/gl-uniform-arrays-sub-source.html
zhenyao: It's the fragment shader's uniform |color| array size. I used a var with comment.

sdk/tests/conformance2/uniforms/gl-uniform-arrays-sub-source.html
zhenyao: Done

kenrussell:
Merging now.

PR #2304, opened 2017-02-18
fix deqp testing 8-sample FBOs on hardware which doesn't support it

kainino0x:
This fixes these tests on Chrome and Firefox on the Google Pixel.


kenrussell:
Great diagnosis and fix @kainino0x .

For a little more background: the comment about the GL_SAMPLES query not working is from the original dEQP tests. We assume it is referring to older, buggy drivers. If that query isn't working as advertised then we do want the test to fail. Expecting the call to glRenderbufferStorageMultisample to generate a GL_INVALID_OPERATION error, and attempting to recover from that, was poor behavior. (A bug was also introduced in that logic during the port to JavaScript.)

Merging now.

PR #2305, opened 2017-02-22
Remove ATC tests from 2.0.0 tests

kainino0x:
It is unclear whether this is a driver bug on Android/Qualcomm,
or if the test itself was wrong (didn't allow all possible errors).

I think the NO_ERRORs are a driver bug, and the others might not be.

This fixes this failure on both Chrome and Firefox (in exactly the same way) on Google Pixel.

### DIFF
--- a/conformance-suites/2.0.0/conformance/extensions/00_test_list.txt
+++ b/conformance-suites/2.0.0/conformance/extensions/00_test_list.txt
@@ -25,7 +25,7 @@
 --min-version 1.0.2 --max-version 1.9.9 oes-element-index-uint.html
 webgl-debug-renderer-info.html
 webgl-debug-shaders.html
---min-version 1.0.3 webgl-compressed-texture-atc.html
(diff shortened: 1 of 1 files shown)


zhenyao:
I am not sure I agree with this CL.  I'd rather add this test to skip list than making the test incorrect (therefore pass on buggy drivers)

kenrussell:
Per offline discussion let's comment out the sections of the test affected by the apparent driver bug in the 2.0.0 snapshot (as was done for other vendors' driver bugs), and file an issue to revisit this in the top-of-tree conformance tests. Thanks.

kainino0x:
Per further investigation and yet another offline discussion I'm disabling this test wholesale in 2.0.0, and pushing to have it removed entirely. #2310

Patch updated.

kainino0x:
(PTAL)

kenrussell:
LGTM

PR #2306, opened 2017-02-23
fix error in conformance2/textures/misc/tex-storage-2d.html

kainino0x:


kenrussell:
Good catch Kai. Presumably this was failing on most non-desktop machines? LGTM.

kainino0x:
It must have been. But it was passing on the Pixel C, which has S3TC.

PR #2307, opened 2017-02-23
Don't skip deqp tests in WebGL 2.0.1 if the corresponding tests are n…

Richard-Yunchao:
…ot driver/hw bugs in c++ deqp.

These tests can be skipped in WebGL 2.0.0. But I think we should recover the tests and try to fix them in WebGL 2.0.1.

PTAL. Thanks a lot! @kenrussell @zhenyao and all owners.

### DIFF
--- a/sdk/tests/deqp/framework/common/tcuSkipList.js
+++ b/sdk/tests/deqp/framework/common/tcuSkipList.js
@@ -118,13 +118,6 @@ goog.scope(function() {
         _skip("blit.rect.nearest_consistency_out_of_bounds_min_reverse_src_x");
         _skip("blit.rect.nearest_consistency_out_of_bounds_min_reverse_src_y");
 
-        _setReason("Tricky blit rects can result in imperfect copies on Mac Intel driver.");
-        // crbug.com/658724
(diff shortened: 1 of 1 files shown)


kenrussell:
Yunchao, agree that we should push for these tests to be re-enabled. Could you please write a Radar template for these bugs? According to crbug.com/658724, one has not yet been filed. We must file them with Apple if we expect them to ever be fixed.

Richard-Yunchao:
@kenrussell , actually, I have wrote a Radar template (https://docs.google.com/document/d/1820bNVEsqW0sJ3nX77T29UQUq66UCborO1u0RNJunbY/edit), although the current tests can pass because the failed cases are skipped.

Richard-Yunchao:
@jdashg , Done! I don't know why I need to sign CLA. I thought I signed it long before.

kenrussell:
It seems fine to me to un-skip these tests at this point; it looks like the underlying driver bugs were fixed a couple of years ago. Hopefully they haven't regressed.

kenrussell:
I'm also assuming that the CLA assistant tool is buggy.

PR #2308, opened 2017-02-23
Add a GLSL3 test for a HLSL compiler freeze.

Kangz:
The HLSL compiler doesn't handle gradient in discontinuous control flow
nicely, and ANGLE was missing handling for new GLSL ES 300 gradient
builtins in its workaround. This is a regression test.

See http://anglebug.com/1915


kenrussell:
Excellent. Thanks Corentin for adding this important regression test.

PR #2311, opened 2017-02-23
2.0.0: disable tests failing on Pixel in Chrome+FF

kainino0x:
I'll be updating all of the crbug links to point at specific bugs.


kainino0x:
Done. Ready for review.

kenrussell:
Very nice work and good bookkeeping job @kainino0x .

For the record, other tests were disabled in the WebGL 2.0.0 conformance suite snapshot because of graphics driver bugs on other platforms (macOS and Linux, for example), so this isn't a special case for Android. Disabling these tests to enable shipment of WebGL 2.0.0 on Android was agreed upon by the working group.

Let's follow up and file bugs against the vendors so these bugs get fixed.

Merging now.

PR #2312, opened 2017-02-24
readPixels RGB565 needs slightly extra tolerance on Google Pixel

kainino0x:


kenrussell:
Looks fine. I can't imagine this is controversial.

PR #2313, opened 2017-02-24
fix "runTests" to "runRenderTests" in compare-structs-containing-arrays

kainino0x:
These tests are clearly intended to be render tests (i.e. have their output be checked to be green). This passes at least on some machines (e.g. Linux NVIDIA).


kenrussell:
Very good Kai. Let's strengthen this test.

@Oletus FYI we found that these tests don't pass on all OpenGL ES 3.0 implementations. We should double-check the tests to make sure they're correct.

PR #2314, opened 2017-02-25
Updated WebGL logo to Khronos' current version.

kenrussell:
Size has changed to 240x100. Updated sizes in all affected HTML files
and templates.

Some files contained incorrect HTML for the logos; intent was that
WebGL and Khronos logos be side-by-side. Fixed this throughout.


kenrussell:
All: I've verified these changes locally. They were made upon request of the Khronos webmaster. Merging myself now.

PR #2315, opened 2017-02-25
Fixed preexisting whitespace errors highlighted in #2314.

kenrussell:


kenrussell:
Trivial change. Confirmed locally. Merging myself.

PR #2316, opened 2017-02-25
2.0.0: disable multisample.fbo_{4,max}_samples.constancy_*

kainino0x:
These tests are failing in Chrome on Google Pixel (7.1.1).

We've investigated this, and we're pretty confident that it's a driver bug, even though these tests pass in ES3 dEQP and in Firefox - it seems to be due to some unknown interaction with Chrome's architecture.

http://crbug.com/695742


kenrussell:
To the other working group members: this is as agreed, to get 2.0.0 passing on Android. We will continue to investigate this, and probably file a bug with Qualcomm.

PR #2318, opened 2017-02-27
negativetextureapi: reduce allocations to prevent OOM

kainino0x:
This was required to get the test to pass on Android (Pixel).


kenrussell:
This is excellent Kai. There are no semantic changes to the test, just (significantly) improving its memory consumption. Thank you for tracking this down and fixing it.

PR #2319, opened 2017-02-27
Add "download text summary" button

kainino0x:
Works around 100k clipboard limit on Android, which makes capturing conformance results very difficult.
Also much more convenient than copy-pasting.


kenrussell:
Awesome. Thanks Kai.

PR #2320, opened 2017-03-01
Added more logging to diagnose http://crbug.com/690122 .

kenrussell:


kenrussell:
Tested locally. Merging myself.

PR #2321, opened 2017-03-02
Specify current tested behavior on drawElement offset < 0

### LINKED ISSUE #2287: What error for glDrawElements with a negative offset?
In the WebGL CTS there is a test checking that INVALID_VALUE is generated when the offset parameter to DrawElements is negative. The check was [originally added](https://github.com/KhronosGroup/WebGL/commit/7980f486de20ed486371c345c43e6eefbbdc3b36#diff-a43003bac86566881758d8c82e4fbb46R57) in a very old commit and is now in [common code](https://github.com/KhronosGroup/WebGL/blob/5f654b7708979d45a8144c6280f1c30db8f52014/sdk/tests/js/tests/out-of-bounds-test.js#L30) used in conformance/rendering/draw-elements-out-of-bounds.html.

Neither of the WebGL or OpenGL ES specifications include an error case for a negative offset. The WebGL specification section 6.5 has an error case for OOB element buffer reads that gives INVALID_OPERATION if offset < 0 and count > 0 but that's it.

Mo pointed out that in multiple places in the OpenGL ES specification there are `offset` intptr arguments that produce INVALID_VALUE if their are negative (such as BindBufferRange and CopyBufferSubData).

Should we specify the behavior tested in the CTS or change the CTS to follow the rules outlined in Section 6.5?

Kangz:
Fixes #2287

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -3043,10 +3043,11 @@ WebGLRenderingContext implements WebGLRenderingContextBase;
         <dd>
             Draw using the currently bound element array buffer.  The given offset is in bytes, and
             must be a valid multiple of the size of the given type or an <code>INVALID_OPERATION</code>
-            error will be generated; see <a href="#BUFFER_OFFSET_AND_STRIDE">Buffer Offset and Stride
-            Requirements</a>. If <code>count</code> is greater than zero, then a non-null
-            <code>WebGLBuffer</code> must be bound to the <code>ELEMENT_ARRAY_BUFFER</code> binding
-            point or an <code>INVALID_OPERATION</code> error will be generated.<br><br>
+            error will be generated; in addition the offset must be non-negative or an
+            <code>INVALID_VALUE</code> error will be generated; see <a href="#BUFFER_OFFSET_AND_STRIDE">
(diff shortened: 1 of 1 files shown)


specs/latest/1.0/index.html
```diff
@@ -3043,10 +3043,11 @@ <h2 class="no-toc">Table of contents</h2>
         <dd>
             Draw using the currently bound element array buffer.  The given offset is in bytes, and
             must be a valid multiple of the size of the given type or an <code>INVALID_OPERATION</code>
-            error will be generated; see <a href="#BUFFER_OFFSET_AND_STRIDE">Buffer Offset and Stride
-            Requirements</a>. If <code>count</code> is greater than zero, then a non-null
-            <code>WebGLBuffer</code> must be bound to the <code>ELEMENT_ARRAY_BUFFER</code> binding
-            point or an <code>INVALID_OPERATION</code> error will be generated.<br><br>
+            error will be generate; in addition the offset must be non-negative or an
```
zhenyao: generate -> generated

zhenyao:
LGTM with one type.

Kangz:
Addressed comment, thanks for the review!

PR #2323, opened 2017-03-06
fix typo at renderbuffers-feedback-loop.html

ChenHanTsai:
This test, it uses glBindAttribLocation, but the string name is wrong, so it can't bind the correct vertex attribute, and in the shader it also doesn't use layout syntax, so the index will be undefined, the index is based on OpenGL automatic assigned . The Chrome with AMD can work is because just luck that the position and uv is the same as enableattribute index, but Firefox with AMD is position and uv reverse order.


zhenyao:
Looks good.  Thanks for the fix.

PR #2324, opened 2017-03-06
Fix typos in renderbuffers/feedback-loop.html in snap shots.

zhenyao:


zhenyao:
This is a followup to  https://github.com/KhronosGroup/WebGL/pull/2323

Merging myself.

kenrussell:
LGTM after the fact. Thank you for backporting this fix.

Issue #2262, opened 2017-01-15
TexImageSource is not defined in the WebGL2 spec


greggman:
I'm guessing TexImageSource means HTMLImageElement, HTMLCanvasElement, HTMLVideoElement, ImageBitmap, ImageData?  But it's not defined in the spec or am I missing it?

kenrussell:
The typedef's in the WebGL 1.0 spec. The WebGL 2.0 spec is currently defined as a difference spec relative to WebGL 1.0.

Khronos has pointed out that this is confusing and that there should be at least an option of generating a full spec. We don't have the toolchain in place to do this yet, but we plan to move the spec to Bikeshed, at which point we intend to offer both.

Closing this.

greggman:
Ok, turns out I was looking at 1.0.2 spec, not 1.0.3 spec.

At the same time the 1.0.3 spec doesn't really spell it out. It also does not inckude `ImageBimtap` which is mentioned in the 2.0 spec

kenrussell:
This is covered in the latest spec at https://www.khronos.org/registry/webgl/specs/latest/1.0/ . The snapshots are really only useful as archival material. Please look at the top-of-tree specs for the behavior that browsers should be tracking.

greggman:
I figure out why I was looking at the old spec. It's because that's what's linked from the WebGL 2.0 spec

https://www.khronos.org/registry/webgl/specs/latest/2.0/

Links to 

https://www.khronos.org/registry/webgl/specs/1.0.2/

not to 

https://www.khronos.org/registry/webgl/specs/latest/

PR #2325, opened 2017-03-07
Fix latest/2.0 spec link to 1.0.

zhenyao:
Use latest/1.0 instead of 1.0.2


zhenyao:
@kenrussell PTAL

zhenyao:
This is to fix https://github.com/KhronosGroup/WebGL/issues/2262

kenrussell:
Great. Thanks Mo for picking this up. LGTM

PR #2326, opened 2017-03-07
2.0.0: disable tests needed for Android/NVIDIA (SHIELD)

kainino0x:


kenrussell:
This is as agreed upon in the working group. Thanks Kai for assembling these. Looks good; merging now.

PR #2327, opened 2017-03-07
Fix texture filtering param for test case integer-cubemap-specificati…

JiangYizhou:
…on-order-bug

Internal format RG8I cannot do texture filtering according to es 3.1,
table 8.13, modify LINEAR to NEAREST.


JiangYizhou:
@zhenyao @kenrussell PTAL.

zhenyao:
lgtm

PR #2328, opened 2017-03-08
Add a test for dynamically indexing a swizzled l-value vector

Oletus:
This exposes a bug that was recently fixed in ANGLE as well as bugs in
various OpenGL drivers.


zhenyao:
lgtm

PR #2329, opened 2017-03-08
Add a test for a uniform struct with a non-square matrix

Oletus:
This exposes a bug in ANGLE's D3D backend.


zhenyao:
LGTM

What is the nature of the bug?  It fails to compile/link? Or something else?

Oletus:
The rendering results are incorrect. I suspect it has something to do with how uniform registers are assigned for the non-square matrix, folks in the ANGLE team will take a look at the bug.

PR #2330, opened 2017-03-08
Use GLSL test framework for tests that need to set uniforms

Oletus:
This unifies test output and reduces duplication of trivial code
without losing coverage.


zhenyao:
Great refactoring!

LGTM

PR #2331, opened 2017-03-08
Further reduce unnecessary JS allocations

kainino0x:
This doesn't change test semantics.

This coalesces a few more allocations in negativetextureapi.html (see
also #2318) and many allocations in uninitialized-test-2.html.


zhenyao:
Great!
LGTM

PR #2332, opened 2017-03-08
negativetextureapi.html: remove tests which cause huge allocations

kainino0x:
Even after reducing the memory usage a few times, these allocations are too huge (>256MB on Android/NVIDIA). We decided to just remove them; (max+1,max+1) doesn't add significant coverage over (max+1,1) and (1,max+1), and these tests are just too unreliable on low-memory or 32-bit devices.

### DIFF
--- a/conformance-suites/2.0.0/deqp/functional/gles3/es3fNegativeTextureApiTests.js
+++ b/conformance-suites/2.0.0/deqp/functional/gles3/es3fNegativeTextureApiTests.js
@@ -367,33 +367,29 @@ goog.scope(function() {
             bufferedLogToConsole('gl.INVALID_VALUE is generated if width or height is greater than gl.MAX_TEXTURE_SIZE.');
 
             var maxSideSize = Math.max(maxCubemapSize, maxTextureSize);
-            var scratchBuffer = new ArrayBuffer(es3fNegativeTextureApiTests.etc2EacDataSize(maxSideSize, maxSideSize));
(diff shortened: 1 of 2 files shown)


kainino0x:
@jdashg @zhenyao PTAL.

kenrussell:
LGTM.

For more background: the native tests in this area rely on the fact that these CompressedTexImage2D calls are supposed to return early because the width and height are outside the legal range, and don't touch the incoming data (for which the tests illegally pass NULL). In JavaScript it isn't possible to pass NULL for the "data" argument.

I firmly believe that these tests don't provide useful added coverage over the other ones which already test (max_width + 1) and (max_height + 1). There are other examples elsewhere in the conformance suite where tests performing huge allocations were removed because they were unreliable on lower-memory devices. (In other words, this pull request isn't simply working around a browser bug.)

LGTM, but let's wait to hear back from Mozilla (as the other current WebGL 2.0 implementer) before merging this.

kainino0x:
Oh, here are the native tests: [compressedteximage2d_max_width_height](https://android.googlesource.com/platform/external/deqp/+/refs/heads/master/modules/gles3/functional/es3fNegativeTextureApiTests.cpp#333)

zhenyao:
LGTM
The original C++ test passes data=null, which is not spec compliant.

jdashg:
Cool. I proposed something similar to the non-compressed tests, so this gets no pushback from me!

PR #2333, opened 2017-03-09
fence_sync.client_wait_sync_finish: Allow TIMEOUT_EXPIRED

kainino0x:
In WebGL, a finish() or flush() may not have any effect. This means
TIMEOUT_EXPIRED is still possible in this test case, even though it
doesn't call clientWaitSync until a frame after the fenceSync() and
finish() operations.

This was seen on Pixel (Android/Qualcomm).


zhenyao:
To be clear, Chrome implements finish() as a flush(), so it doesn't really guarantee the command to finish, so the sync object may not be signaled.  This is a difference between C++ version and WebGL 2.0 version.

Thanks for tracking this down. LGTM.

PR #2334, opened 2017-03-10
Remove proposed core spec changes from WEBGL_multiview extension

Oletus:
Rendering to a stereo default framebuffer should be replaced with
something else defined in another extension spec - either a new type
of framebuffer abstraction that also knows the HMD it is targeting, or
by rendering to the default framebuffer side-by-side.

### DIFF
--- a/extensions/proposals/WEBGL_multiview/extension.xml
+++ b/extensions/proposals/WEBGL_multiview/extension.xml
@@ -14,100 +14,6 @@
     <api version="2.0"/>
   </depends>
   <overview>
-
-    <h1>Proposed changes to the core WebGL specification</h1>
-
-    <p>These changes are written here in the extension document in order to facilitate discussion of the proposal. They are intended to be merged to the core WebGL specification and removed from here at a later date.</p>
-
-    <h2>The following is added to the WebGLRenderingContextBase interface:</h2>
-
-    <pre>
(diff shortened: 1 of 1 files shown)


RafaelCintron:
Previous to this change, the 'stereo' context attribute made the default framebuffer be multi-view.  When the web developer passes the Canvas to the VRLayer, it's clear to WebVR which framebuffer is to be shown in the HMD. 

With this new change, how are we going to hookup the multiview framebuffer created using the extension to what WebVR draws to the HMD?

Oletus:
A stereo framebuffer that can be directly sent to WebVR will need to be handled by a separate multiview extension that would also be accessible from WebGL 1.0. I'm going to draft proposals for that separate extension soon. There are a few alternative ways to do that, like a new abstraction or simply accelerating side-by-side stereo rendering.

After these changes, this extension's role would simply be to enable rendering to texture arrays, just like the native OVR extension.

Oletus:
Could this be merged, or should I put up proposals for the other multiview extensions that would complement this one in the same pull request?

zhenyao:
@RafaelCintron Do you have more concerns?

RafaelCintron:
I am OK with merging this change and dealing with the other multiview extensions as a separate pull request.

PR #2336, opened 2017-03-14
Update WebGL 2.0 text on get.webgl.org's "get a WebGL implementation"…

kenrussell:
… page.


kenrussell:
Textual updates. Merging myself.

PR #2337, opened 2017-03-14
Removed unneeded meta tag.

kenrussell:


kenrussell:
Trivial. Merging myself.

Issue #2300, opened 2017-02-15
Add 'webglcontextchanged' event


grorg:
When an implementation swaps GPUs the various limits and values exposed to the developer may change. They should be made aware of this, in case they've cached the values.

I suggest a new WebGLContextEvent with name 'webglcontextchanged'. It's basically indicating to the developer that something about the context is different. They don't have to do anything in response, but this gives them the opportunity.

If people agree I'll file a PR. My suggested wording would be along the lines of "fired when any detectable attribute for the WebGL context is changed by the user agent".

And maybe an example that shows max texture size or something changing.

kenrussell:
This seems redundant. The webglcontextlost and webglcontextrestored events should already provide enough information to the developer that they need to re-cache any GPU related information. We could expand the documentation around these events, in particular webglcontextrestored, but I don't think we should add a new event type.

grorg:
We originally considered just using contextrestored for this, sending such an event without a corresponding contextlostevent. But we were concerned that the majority of content in the wild was written to treat contextrestored as the time to reset everything. That's not what we want.

A GPU switch might happen quite often, and to a lot of content, no matter whether it requested low or high power. Forcing a real lost context at this time would be bad - you'd get churn on basically every running WebGL instance. 

Also, we don't think a GPU switch will cause many issues. Hopefully "it just works". What we're trying to provide is some way for content that really understands it is pushing the limits to just get a notification. It might be that it doesn't need to react to it. Most content won't.

grorg:
During the WebGL conference call, @kenrussell asked for a list of situations where a GPU switch might not "work" (on macOS).

I'll try to gather more exact data, but at the moment we don't have anything official documenting it.

At the moment we think that WebGL 1 will never cause an error or crash that warrants a context lost. There might be cases where incorrect rendering is possible, such as if the point size was set to something bigger than GL_POINT_SIZE_MAX before the switch. Maybe that warrants a lost context, but it seems to be overkill - even if they restore the context they can't exceed that point size.

WebGL 2 might be more problematic. We haven't done a full review yet. Clearly there are a lot more limits to check, but in most cases the driver will just clamp to its max.

kenrussell:
It's unappealing to rush in changes to this area of the spec. They'll stick around forever.

[...]

PR #2338, opened 2017-03-16
Add a 'webglcontextchanged' event

grorg:
Addresses the first part of #2300.

The 'reason' field and enum is coming in a follow-up pull request.


grorg:
I suggest waiting until I've included the 'reason' attribute and enum before merging. I'll do that today.

kenrussell:
Thanks Dean. Please do add the enum, and let's call it a "reasons" field (sequence of that enum).

PR #2339, opened 2017-03-17
Prompt user before reloading webgl-conformance-tests.html

kainino0x:
Accidental reload makes it easy to lose results in the middle of a conformance run. In Chrome, it's particularly easy if you don't disable "pull-to-refresh" in about:flags. This causes the browser to prompt when reloading.


kenrussell:
LGTM. Thanks for improving the harness.

Issue #2340, opened 2017-03-20
Diff between WebGL and GLES 2.0: no shader binary, SHADER_COMPILER should be always true


Richard-Yunchao:
I read the WebGL spec, but I didn't find such a section among all differences between WebGL and OpenGL ES 2.0: https://www.khronos.org/registry/webgl/specs/latest/1.0/#6. 

In WebGL 2 spec, there is a section <no program binary>: https://www.khronos.org/registry/webgl/specs/latest/2.0/#5.4. 

So should we add a section to WebGL spec to explicitly say that there is no shader binary in WebGL? @kenrussell and @zhenyao

zhenyao:
Right, we should mention it in the difference section.

Richard-Yunchao:
This issue has been resolved. Closing...

PR #2341, opened 2017-03-21
WebGL 1.0 spec update: no shader binaries in WebGL

Richard-Yunchao:
PTAL, @zhenyao and @kenrussell .

See the discussion at #2340.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -3481,6 +3481,14 @@ bound in order for texture-related operations and queries to succeed.
 
 </p>
 
+    <h3>No Shader Binaries</h3>
+
+<p>
+
+Accessing binary representations of compiled shaders is not supported in the WebGL API. This includes OpenGL ES 2.0 ShaderBinary entry point. In addition, querying shader binary formats and shader compiler with getParameter is not supported in the WebGL API.
+
+</p>
+
     <h3><a name="BUFFER_OFFSET_AND_STRIDE">Buffer Offset and Stride Requirements</a></h3>
 
 <p>
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -3485,7 +3485,14 @@ bound in order for texture-related operations and queries to succeed.
 
 <p>
 
-Accessing binary representations of compiled shaders is not supported in the WebGL API. This includes OpenGL ES 2.0 ShaderBinary entry point. In addition, querying shader binary formats and shader compiler with getParameter is not supported in the WebGL API.
+Accessing binary representations of compiled shaders is not supported in the WebGL API. This
+includes the OpenGL ES 2.0 ShaderBinary entry point. In addition, querying shader binary formats
+and the availability of a shader compiler via getParameter is not supported in the WebGL API.
+
+</p>
+<p>
+
+All WebGL implementations must implicitly support an on-line shader compiler.
 
 </p>
 
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -3486,8 +3486,9 @@ bound in order for texture-related operations and queries to succeed.
 <p>
 
 Accessing binary representations of compiled shaders is not supported in the WebGL API. This
-includes the OpenGL ES 2.0 ShaderBinary entry point. In addition, querying shader binary formats
-and the availability of a shader compiler via getParameter is not supported in the WebGL API.
+includes the OpenGL ES 2.0 <code>ShaderBinary</code> entry point. In addition, querying shader
+binary formats and the availability of a shader compiler via <code>getParameter</code> is not
+supported in the WebGL API.
 
 </p>
 <p>


dmikis:
Anchor's missing. Also it there a need to specify this since there's simply no `shaderBinary` call in WebGL API?

dmikis:
This bit is very unclear:

>  Shader compiler should be always true

I assume it refer to `GL_SHADER_COMPILER` enum, but, as with `shaderBinary` entry point, there's no such enum in WebGL.

Richard-Yunchao:
@dmikis , thanks for your review. I think it is better to explicitly say this. It means that we deliberately don't want to expose shaderBinary to web developers, instead of we FORGET to expose the API. And that is also the reason why we have a standalone section to enumerate all differences between WebGL 1 and GLES 2.0.

Per the statements "Shader compiler should be always true", maybe we can remove this sentence, maybe we can have a better words to express this meaning. Any suggestions about this? @zhenyao and @kenrussell ?

specs/latest/1.0/index.html
```diff
@@ -3481,6 +3481,14 @@ <h2 class="no-toc">Table of contents</h2>
 
 </p>
 
+    <h3>No Shader Binaries</h3>
+
+<p>
+
+Accessing binary representations of compiled shaders is not supported in the WebGL API. This includes OpenGL ES 2.0 ShaderBinary entry point. Shader compiler should be always true. In addition, querying shader binary formats and shader compiler with getparameter is not supported in the WebGL API.
```
zhenyao: Shader compiler is a separate issue from shader binaries.  Let's address it in a separate point (or a separate CL).

Add <code /> around ShaderBinary and getParameter (capital P)

dmikis:
@Richard-Yunchao IMHO, it's obvious that it's not forgotten. "Differences Between WebGL and OpenGL ES 2.0" section is more about explicitly stating differences in behaviour between GLES entry points and corresponding WebGL ones. And it's clearly needed there since we refer to GLES man pages as description of entry points. Why we need to explicitly state that we omit some entry points isn't obvious for me, to be honest.

specs/latest/1.0/index.html
Richard-Yunchao: Thanks for your review, @zhenyao . I have updated the pr according to your comments. Please take another look. Thanks a lot!

Richard-Yunchao:
@dmikis , You know, shader binary is not only an API, but a standalone feature. we also remove a couple of GLenums like NUM_SHADER_BINARY_FORMAT and SHADER_BINARY_FORMAT from getParameter. 

If someone think the differences between WebGL and GLES is obvious, so we don't clearly state it in spec. IMHO, that is not a good way to draft a spec. I think we should clearly state ALL differences between WebGL and GLES. We should not let users to GUESS. Someone think some differences are obvious, the others may don't think so. As a result, why not clearly state it in the spec? It is not a big burden to clearly point it out in the spec. In addition, we already have had a subsection <No program binaries> in WebGL 2 spec, so <No shader binaries> is a counterpart in WebGL 1 spec.

specs/latest/1.0/index.html
```diff
@@ -3481,6 +3481,14 @@ <h2 class="no-toc">Table of contents</h2>
 
 </p>
 
+    <h3>No Shader Binaries</h3>
+
+<p>
+
+Accessing binary representations of compiled shaders is not supported in the WebGL API. This includes OpenGL ES 2.0 ShaderBinary entry point. In addition, querying shader binary formats and shader compiler with getParameter is not supported in the WebGL API.
```
kenrussell: Can we rephrase this a little? Also, could you please wrap it to 100 characters?

"This includes OpenGL ES" -> "This includes the OpenGL ES"
"and shader compiler with" -> "and the availability of a shader compiler via"

Suggest adding:

"All WebGL implementations must implicitly support an on-line shader compiler."

Thanks.

Richard-Yunchao: Thanks for your review, @kenrussell . I have updated this pull request accordingly, please take another look.

specs/latest/1.0/index.html
```diff
@@ -3481,6 +3481,21 @@ <h2 class="no-toc">Table of contents</h2>
 
 </p>
 
+    <h3>No Shader Binaries</h3>
+
+<p>
+
+Accessing binary representations of compiled shaders is not supported in the WebGL API. This
+includes the OpenGL ES 2.0 ShaderBinary entry point. In addition, querying shader binary formats
+and the availability of a shader compiler via getParameter is not supported in the WebGL API.
```
zhenyao: Again, please wrap <code> </code> around ShaderBinary and getParameter.

zhenyao: wrap "code" tag around ...

Richard-Yunchao: Sorry. @zhenyao, the pull request has been updated accordingly.  Please take another look.

kenrussell:
Looks good. Thanks @Richard-Yunchao for your contribution.

PR #2345, opened 2017-03-23
Add minimized conformance test for transform feedbacks' buffer state.

kenrussell:
Regression test for http://crbug.com/696345 .

Thanks to @greggman for the test.

### DIFF
--- a/sdk/tests/conformance2/transform_feedback/transform_feedback.html
+++ b/sdk/tests/conformance2/transform_feedback/transform_feedback.html
@@ -67,6 +67,7 @@ var gl = wtu.create3DContext(canvas, null, 2);
 var WEBGL_get_buffer_sub_data_async = gl.getExtension("WEBGL_get_buffer_sub_data_async");
 var tf = null;
 var tf1 = null;
+var buf = null;
 var program = null;
 var activeInfo = null;
 var query = null;
@@ -78,6 +79,7 @@ if (!gl) {
     testPassed("WebGL context exists");
 
     runBindingTest();
+    runTFBufferBindingTest();
     runObjectTest();
     runGetBufferSubDataTest();
     runOneOutFeedbackTest();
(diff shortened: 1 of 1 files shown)


kenrussell:

Tested in Chrome with the fix for http://crbug.com/696345, and in Firefox Nightly; Firefox already passes this new test.

sdk/tests/conformance2/transform_feedback/transform_feedback.html
```diff
@@ -114,6 +116,38 @@
...
+    // gl.TRANSFORM_FEEDBACK_BUFFER is part of Transform Feedback objects'
+    // state. See OpenGL ES 3.0.5 Section 6.24.
+    //
+    // Since the TRANSFORM_FEEDBACK was just unbound, there should be nothing
+    // bound.
+    shouldBeNull('gl.getIndexedParameter(gl.TRANSFORM_FEEDBACK_BUFFER_BINDING, 0)');
+
+    // Binding the buffer to the ARRAY_BUFFER binding point should succeed.
+    gl.bindBuffer(gl.ARRAY_BUFFER, buf);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "binding buffer to ARRAY_BUFFER");
```
zhenyao: This is against WebGL 2 spec: https://www.khronos.org/registry/webgl/specs/latest/2.0/#5.1

greggman: That's not my understanding of the spec.

The spec says to me there are 3 types of buffer. 

1. Undefined
2. Element Array
3. Other Data

Binding to `TRANSFORM_FEEDBACK` means the buffer is now type "Other Data". A Buffer that is type "Other Data" and be bound to any bind point except `ELEMENT_ARRAY_BUFFER`

zhenyao: There is another section that says you can't bind a buffer TRANSFORM_FEEDBACK buffer binding and another binding point at the same time.

greggman: They aren't bound at the same time. That's the bug this test is checking for. TRANSFORM_FEEDBACK bindings are part of the TransformFeedback object, not global state. So, unbinding the Transformfeedback object (a few lines above) means the buffer is no longer bound to TRANSFORM_FEEDBACK.

That was the bug being fixed in Chrome. Chrome wrongly had TRANSFORM_FEEDBACK as global state. It's not, it's state of the current TransformFeedback object.

zhenyao: But if we relax this, then BindTransformFeedback() can generate INVALID_OPERATION because now you suddenly have a buffer bound to both transform feedback buffer binding and another binding point.  Isn't this super weird?

That only left the option of doing validation at draw time, but that also include readPixels from buffer, all Texture uploading path from buffer, draw, buffer data uploading, etc ...

greggman: The issue above has nothing to do with relaxing anything. It has to do with actually working period. The test above works as OpenGL is suppose to work

1. create transformfeedback
2. bind transformfeedback
3. bind buffer to TRANSFORM_FEEDBACK
    
    this bind the buffer to a particular index on the TransformFeedback object the same way vertexAttribPointer binds a buffer to an attribute

4.  Unbind the transform feedback

There are now NO buffers bound. There's above attached to a currently unused Transformfeedback. The same way buffers can be attached to unused VertexArrayObjects and the same way textures can be attached to framebuffers. 

Binding to ARRAY_BUFFER shoud not error at this point period.

zhenyao: I understand that.  What I am saying is 

5 Bind buffer to ARRAY_BUFFER (let's say we allow this)

6 bind transformfeedback

6 can't be allowed because it will create a dual binding buffer. To me that's a less natural restriction.

greggman: Thinking about it, I don't know what checks are happening now but I assume the point is you don't want the same buffer being both an input and an output. (just like with textures and framebuffer attachments).

In that case you can only check this at draw time. Input buffers are attached to vertexarrayobjects, output buffers are attached to transformfeedbackobjects. You only know if there is will be a conflict at draw time. The user selects one set of buffer inputs with`gl.bindVertexArray`, And one set of buffer outputs with `gl.bindTransformFeedback`.

There is arguably no reason to ever error on calling `gl.bindBuffer(gl.ARRAY_BUFFER)`. ARRAY_BUFFER is never an input, the inputs come from the attributes which have their own buffers attached.

greggman: See https://github.com/KhronosGroup/WebGL/issues/2346

zhenyao: "Buffers should not be bound or in use for both transform feedback and other
purposes in the GL. Specifically, if a buffer object is simultaneously bound to a
transform feedback buffer binding point and elsewhere in the GL, any writes to
or reads from the buffer generate undefined values. Examples of such bindings
include ReadPixels to a pixel buffer object binding point and client access to a
buffer mapped with MapBufferRange."

So yeah, it's way beyond just draw, and ARRAY_BUFFER is definitely problematic.

kenrussell:
Let's not merge this yet. I've commented on #2346 regarding a larger spec and conformance test change that should be done first.

jdashg:
I believe this is obsolete. #2602 added tests to verify the spec behavior we ended up with.

Issue #2347, opened 2017-03-24
WebGL2 implementations not supporting the default transform feedback


greggman:
I might be missing something but ... according to the OpenGL ES 3.0 spec, if there is no transform feedback bound then the default transform feedback object is bound and unless I'm missing it should act exactly the same as any other transform feedback

I thought WebGL2 would probably disallow the default transformfeedback the same way it disallows the default textures.

But, the WebGL2 spec says it's supported

[Section 3.7.15](https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.7.15)

> `void bindTransformFeedback (GLenum target, WebGLTransformFeedback? transformFeedback)` (OpenGL ES 3.0.4 §2.14.1, man page)
> Bind the given WebGLTransformFeedback object. **If transformFeedback is `null`, the default transform feedback object provided by the context is bound**. 

So I tried it but I got errors in both firefox and chrome.

https://jsfiddle.net/greggman/rc2ja9u7/

It's possible I'm doing something wrong. If you just change line

From:

    const tf = null; // Use the default transform feedback  ... gl.createTransformFeedback(); 
 
To:

    const tf = gl.createTransformFeedback(); 

Then it works but AFAICT it should work as well with the default transformfeedback object.

If the test is correct either the implementations need to change or the spec needs to change. If it's the spec to disallow the default transformfeedback object it would seem like it should also require more errors (calling gl.bindBufferBase/gl.bindBufferRange) when no transformfeedback is bound etc. Otherwise the implementations need to change

zhenyao:
Chrome has a fake "default" transformfeedback in GPU command buffer, so in theory you never see the real default one and the fake one just acts the same as any other transformfeedback.

However, we have a bug in WebGL because WebGL needs to create its own "fake" default one. @kenrussell has a CL that might just fix this.

If you can minimize your code into a test case, that would be really appreciated.

greggman:
This brings up the question what happens if I call `gl.getParameter(gl.TRANSFORM_FEEDBACK_BINDING)` when the default transform feedback is bound do I get a `WebGLTransformFeedback`? I assume no, I get `null` but that should probably be specced and tested.

kenrussell:
[...]

PR #2348, opened 2017-03-25
Test for using the default transform feedback object

greggman:
For issue #2347 

Note: 2 lines are marked as FIXME and should be removed when/if
impls and the spec are updated to make `gl.bindBuffer` less
strict.

### DIFF
--- a/sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
+++ b/sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
@@ -1,0 +1,142 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL Transform Feedback Conformance Test - Default Transform Feedback</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<canvas id="canvas" style="width: 50px; height: 50px;"> </canvas>
+<div id="console"></div>
+<script id="vshader" type="x-shader/x-vertex">#version 300 es
+in float in_value;
+out float out_value;
+
+void main() {
+   out_value = in_value * 2.;
+}
+</script>
+<script id="fshader" type="x-shader/x-fragment">#version 300 es
+precision mediump float;
+out vec4 dummy;
+void main() {
+  dummy = vec4(1);
+}
+</script>
+<script>
+"use strict";
+description("This test verifies using the default transform feedback object");
+
+debug("");
+
+var wtu = WebGLTestUtils;
+var canvas = document.getElementById("canvas");
+var gl = wtu.create3DContext(canvas, null, 2);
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
```diff
@@ -0,0 +1,142 @@
...
+}
+
+function runDefaultTransformFeedbackTest() {
+  const prog = wtu.setupTransformFeedbackProgram(gl, ["vshader", "fshader"],
+      ["out_value"], gl.SEPARATE_ATTRIBS,
+      ["in_value"]);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "linking transform feedback shader should not set an error");
+
+  const inLoc = 0;
+  const outLoc = 0;
```
zhenyao: outLoc not used at all

sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
```diff
@@ -0,0 +1,142 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
```
zhenyao: 2017

sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
```diff
@@ -0,0 +1,142 @@
...
+  gl.bindVertexArray(srcVAO);
+
+  gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, tf);
+  gl.beginTransformFeedback(gl.TRIANGLES);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no errors");
+  gl.drawArrays(gl.TRIANGLES, 0, 3);
+  gl.endTransformFeedback();
+
+  // FIXME: this should not be needed once the spec and implementationsa are fixed
+  gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, null);
```
zhenyao: Since we are testing default tf, this should have no effect regardless of spec fix

sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
```diff
@@ -0,0 +1,142 @@
...
+  gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, tf);
+  gl.useProgram(prog);
+  gl.bindBufferBase(gl.TRANSFORM_FEEDBACK_BUFFER, 0, dstBuffer);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no errors");
+
+  runFeedback(gl, prog, srcVAO1, tf);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no errors");
+
+  const expected = [2, 4, 6];
+  gl.bindBuffer(gl.ARRAY_BUFFER, dstBuffer);
```
zhenyao: This is incorrect.  Even with the potential rule relax, this will still be illegal because now dstBuffer is bound to the default TF and ARRAY_BUFFER at the same time, and the wtu.checkFloatBuffer will result in an INVALID_OP.

sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
```diff
@@ -0,0 +1,142 @@
...
+}
+
+function createVAO(gl, buf, inLoc) {
+  const vao = gl.createVertexArray();
+  gl.bindVertexArray(vao);
+  gl.bindBuffer(gl.ARRAY_BUFFER, buf);
+  gl.enableVertexAttribArray(inLoc);
+  gl.vertexAttribPointer(inLoc, 1, gl.FLOAT, false, 0, 0);
+  // FIXME: this should not be needed once the spec and implementationsa are fixed
+  gl.bindBuffer(gl.ARRAY_BUFFER, null);  // this is not needed
```
zhenyao: This is unnecessary even for now - srcBuffer1 is never bound to TF, so there is no conflict at all, no need to unbind it. Therefore, this FIXME is misleading.

sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
greggman: Either I'm not getting it or you're not getting. There is NO conflict between having an array in ARRAY_BUFFER and a TF. That's not a conflict in OpenGL ES 3.0 and it can NOT be a conflict in WebGL2. That part of the spec needs to change. That's the whole point of all these things. ARRAY_BUFFER is NEVER an input to anything. It is not used during rendering. It is only an internal bind point used as a buffer reference in various function calls. It is not used during a draw call.

greggman: As I mentioned in the other PR, no this is not the case. ARRAY_BUFFER is never an input during rendering. It's just an internal bind point use in various function calls like `bufferData` and `vertexAttribPointer`. There should never be a conflict no matter what is attached to ARRAY_BUFFER.

sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
greggman: fixed

sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
greggman: fixed

sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
greggman: fixed

greggman:
Note: This version actually works in both firefox and chrome. The actual issues are the binding conflicts which only show up if you delete the `FIXME` lines

sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
zhenyao: It really depends on how much we relax the rules. As the spec specifically pointed out that MapBufferRange() can be a potential conflict situation, where getBufferSubData() may internally use. However, if we relax the rule the finest grid, then there is no conflict.

jdashg:
Superseded by #2802.

Issue #2346, opened 2017-03-24
WebGL2 Buffer restrictions overly restrictive?


greggman:
[The WebGL2 spec, section 5.1 says](https://www.khronos.org/registry/webgl/specs/latest/2.0/#5.1)

> In addition, a buffer can not be simultaneously bound to a transform feedback buffer binding point and another buffer binding point in the WebGL 2 API. Attempting to violate this rule generates an INVALID_OPERATION error, and the state of the binding point will remain untouched.

I agree with the need for the restriction but the actual enforcement seems not  really in the spirit of OpenGL. It seems like this would be checked at draw time not bind time.

Just like it's okay to do this

1. Upload RGB to level 1 of a texture
2. Upload RGBA to level 0 of a texture 
3. Call gl.generateMips

Level 1 will be overwritten. No error was issued on step 2 that the mips didn't match format.

Another example.

1. make framebuffer
2. make texture
3. attach texture
4. call texImage2d to set format of texture

Some implementations would consider issuing an error at step 3 because the texture is not yet renderable but that's not the OpenGL way. The checks don't happen until draw time.

Maybe a better example

1. make and bind framebuffer
2. make and bind texture
3. attach texture to framebuffer
4. unbind texture

If I was to render at step 3 I'd get an error because the texture is both an input and and output. I don't get an error though because it's not checked until draw time. If restriction similar to the buffer restriction was added step 4 would have to come before step 3.

It seems like these buffer binding restrictions should work the same texture binding restrictions and other similar things, checked for at draw time. In fact I'd suspect checking and bind time will break lots of ported content that expects to be able to have things temporarily bound to multiple bind points and work like it does for textures.

Unless I'm mis-understanding how transformfeedback state works it seems like the spec should say along the lines of 

> In addition, when calling any drawing function a buffer can not be simultaneously bound to a transform feedback buffer binding point and attached to an attribute in the currently bound vertexarrayobject or a global attribute if no vertexarrayobject is bound. Attempting to violate this rule generates an INVALID_OPERATION error.

greggman:
Here's a test showing Chrome is broken in this area

https://jsfiddle.net/greggman/vp1p7973/

Firefox passes the test

It's related to the way Chrome is (or was) handling transform feedback.

You can see how in the test, checking at `gl.bindBuffer` time is NOT a useful test. The test has to happen at draw time. The test shows Firefox catches the error and Chrome does not.

[...]

PR #2349, opened 2017-03-25
Test for having the same read and write buffer using a transform feedback

greggman:
Firefox passes, Chrome fails

note: there are 2 lines marked as "FIXME" that I believe should be removed
assuming the spec changes so that `gl.bindBuffer` is less strict


greggman:
This is related to #2346

sdk/tests/conformance2/transform_feedback/transform_feedback_same_buffers_as_input_and_output.html
```diff
@@ -0,0 +1,153 @@
+<!--
+
+/*
+** Copyright (c) 2015 The Khronos Group Inc.
```
zhenyao: 2017

sdk/tests/conformance2/transform_feedback/transform_feedback_same_buffers_as_input_and_output.html
```diff
@@ -0,0 +1,153 @@
...
+}
+
+function runSameInputOutputTest() {
+  const prog = wtu.setupTransformFeedbackProgram(gl, ["vshader", "fshader"],
+      ["out_value"], gl.SEPARATE_ATTRIBS,
+      ["in_value"]);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "linking transform feedback shader should not set an error");
+
+  const inLoc = 0;
+  const outLoc = 0;
```
zhenyao: outLoc is not used.

sdk/tests/conformance2/transform_feedback/transform_feedback_same_buffers_as_input_and_output.html
```diff
@@ -0,0 +1,153 @@
...
+  gl.useProgram(prog);
+  gl.bindBufferBase(gl.TRANSFORM_FEEDBACK_BUFFER, 0, dstBuffer);
+  // this binds the default (id = 0) TRANSFORM_FEEBACK buffer
+  gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, null);
+
+  runFeedback(gl, prog, srcVAO1, tf);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no errors");
+
+  const expected = [2, 4, 6];
+  gl.bindBuffer(gl.ARRAY_BUFFER, dstBuffer);
```
zhenyao: I am assuming this is where Chrome fails? According to the current spec restrictions, Chrome generates an INVALID_OPERATION here.

sdk/tests/conformance2/transform_feedback/transform_feedback_same_buffers_as_input_and_output.html
```diff
@@ -0,0 +1,153 @@
...
+  wtu.checkFloatBuffer(gl, gl.ARRAY_BUFFER, expected);
+
+  // this should fail because the transform feedback's buffer #0 and the
+  // srcVAO2's buffer #0 are the same buffer
+  runFeedback(gl, prog, srcVAO2, tf);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "reading and writing to same buffer");
+
+  gl.bindBuffer(gl.ARRAY_BUFFER, dstBuffer);
+  wtu.checkFloatBuffer(gl, gl.ARRAY_BUFFER, expected, "should be the same as before as nothing has executed");
```
zhenyao: Even if it's executed, the result is still the same expected because it's the same program? So this is a useless and misleading check.

sdk/tests/conformance2/transform_feedback/transform_feedback_same_buffers_as_input_and_output.html
```diff
@@ -0,0 +1,153 @@
...
+
+}
+
+function runFeedback(gl, prog, srcVAO, tf, dstBufferInfo) {
```
zhenyao: dstBufferInfo isn't used.

kenrussell:
I'd like to hold off merging this until Chrome's behavior has been updated per #2346 .

sdk/tests/conformance2/transform_feedback/transform_feedback_same_buffers_as_input_and_output.html
greggman: If incorrectly executed (as it currently is in chrome) the result is 4, 8, 12 because it's manipulating the result from the first `* 2.0` operation.

That result is probably GPU dependent since the whole problem is results of having the same input and output buffer is undefined. In any case it shouldn't execute and the buffer should still have 2, 4, 6 in it. It doesn't in Chrome currently on my MBP. Seems like the test is valid

sdk/tests/conformance2/transform_feedback/transform_feedback_same_buffers_as_input_and_output.html
greggman: fixed

sdk/tests/conformance2/transform_feedback/transform_feedback_same_buffers_as_input_and_output.html
greggman: fixed

sdk/tests/conformance2/transform_feedback/transform_feedback_same_buffers_as_input_and_output.html
greggman: this is one place where chrome fails if the "FIXME" lines are removed. Otherwise Chrome fails because it allows calling `drawArrays` even though the same buffers are bound to a VAO and and TF.

sdk/tests/conformance2/transform_feedback/transform_feedback_same_buffers_as_input_and_output.html
greggman: fixed

zhenyao:
lgtm
but let's hold off until spec is updated as @kenrussell suggested

PR #2354, opened 2017-03-29
Add case for default fbo content would not changed between two readPxels calls.

xhcao:
It seems that every readPixels call will trigger CMAA operation in
Chromium if CMAA is enabled. If there are not drawing operations
between two readPixels calls, fbo content should not be changed.

### DIFF
--- a/sdk/tests/conformance/reading/00_test_list.txt
+++ b/sdk/tests/conformance/reading/00_test_list.txt
@@ -1,3 +1,3 @@
+--min-version 1.0.4 fbo-remains-unchanged-after-read-pixels.html
 read-pixels-pack-alignment.html
 read-pixels-test.html
-
--- a/sdk/tests/conformance/reading/fbo-remains-unchanged-after-read-pixels.html
+++ b/sdk/tests/conformance/reading/fbo-remains-unchanged-after-read-pixels.html
@@ -1,0 +1,125 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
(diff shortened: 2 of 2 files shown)


xhcao:
This case is testing for https://codereview.chromium.org/2597933002/

sdk/tests/conformance/reading/00_test_list.txt
```diff
@@ -1,3 +1,3 @@
 read-pixels-pack-alignment.html
 read-pixels-test.html
-
+fbo-remains-unchanged-after-read-pixels.html
```
zhenyao: Please add 1.0.4 condition to the new test

xhcao: @zhenyao , why need to add 1.0.4 for the new test? Is the reason that it is CMAA special case.

zhenyao: All new tests should have the min version condition so that browsers passing older version of conformance won't regress; it's also for bookkeeping purpose so people know this is a newly added tests after 1.0.3 snapshot.

xhcao: I had added the min version for this test, please help to review it again, thank you.

sdk/tests/conformance/reading/00_test_list.txt
```diff
@@ -1,3 +1,3 @@
 read-pixels-pack-alignment.html
 read-pixels-test.html
-
+--min-version 1.0.4 fbo-remains-unchanged-after-read-pixels.html
```
zhenyao: Sorry I didn't catch this earlier, but please follow alphabetical order.

xhcao: Sorry, late reply for vacation. I had ordered these files according alphabetical order.

zhenyao:
lgtm
thanks

ds-hwang:
lgtm after mass. Thanks!

PR #2355, opened 2017-03-30
Add a proposal for a multiview spec for side-by-side rendering

Oletus:
The new proposal makes it possible to expose multiview drawing to
WebGL 1.0, and to draw in a multiview fashion straight into the
default framebuffer. It works with the current WebVR 1.1 API, which
requires frames to be submitted as split into left and right halves.

This proposal has WEBGL_multiview_side_by_side written as a difference
spec from WEBGL_multiview. This is easier to maintain than having
duplication between these specs. A new native spec document which both
would refer to might be a better longer-term solution, though.

The extension is mutually exclusive with regular WEBGL_multiview. This
simplifies the spec and implementation compared to the situation where
both methods of multiview drawing might be activated at the same time.

### DIFF
--- a/extensions/proposals/WEBGL_multiview_side_by_side/extension.xml
+++ b/extensions/proposals/WEBGL_multiview_side_by_side/extension.xml
@@ -1,0 +1,165 @@
+<?xml version="1.0"?>
+
+<proposal href="proposals/WEBGL_multiview_side_by_side/">
+  <name>WEBGL_multiview_side_by_side</name>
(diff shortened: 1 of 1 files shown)


Oletus:
Abandoning this, we'll go with a cleaner approach with WebVR API changes rather than expose side-by-side rendering explicitly.

kenrussell:
OK. Thanks Olli for continuing to give this thought.

PR #2358, opened 2017-04-01
add 'varying arrays should not be reversed' to GLSL bug conformance tests

Treeki:
I've created a GLSL bug conformance test that checks an ugly bug I ran into in Chrome for Android on my Adreno 300-series devices (Nexus 7 2013, Moto G3).

Firefox successfully passes the test on the same devices. Unfortunately I don't currently have access to any Android devices with other GPUs so I cannot check to see if it passes in Chrome on those.

### DIFF
--- a/sdk/tests/conformance/glsl/bugs/varying-arrays-should-not-be-reversed.html
+++ b/sdk/tests/conformance/glsl/bugs/varying-arrays-should-not-be-reversed.html
@@ -1,0 +1,101 @@
+<!--
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/glsl/bugs/varying-arrays-should-not-be-reversed.html
```diff
@@ -0,0 +1,101 @@
...
+    [0, 255, 0],
+    [0, 0, 255],
+    [0, 128, 255],
+    [255, 128, 0]
+  ];
+
+  for (var i = 0; i < triples.length; i++) {
+    var triple = triples[i];
+    var x = i * 64;
+    gl.viewport(x, 0, 64, 256);
```
zhenyao: Does this work? I think it goes way beyond the WebGL's drawing buffer size.

kenrussell:
@zhenyao FYI I am testing this test now on a Google Pixel.

sdk/tests/conformance/glsl/bugs/varying-arrays-should-not-be-reversed.html
kenrussell: This does work -- all it's doing is rendering into vertical slices of the destination canvas. The maximum x coordinate fed in is 5 * 64 = 320, and the canvas is 512 pixels wide per above.

zhenyao:
Thanks.  I thought by default canvas is 256x256.

kenrussell:
@zhenyao the test explicitly sets it to 512x256. I think the default size for canvases is actually something like 300x150.

@Treeki for the record, your test passes on this very old version of Android on the Nexus 7. Working on updating it to a more recent version.

Android version 4.4.4
Kernel version 3.4.0-g03485a6
android-build@vpbs1.mtv.corp.google.com #1
Tue Mar 18 15:02:27 PDT 2014

Build number
KTU84P

kenrussell:
Confirmed broken on Nexus 7 in:
Android 5.1
Kernel version 3.4.0-g7597781
Build number LMY47O

Continuing to upgrade to top-of-tree Android on this device.

kenrussell:
Still broken in top-of-tree Android on this device.
Will try to file a bug with Qualcomm, but no guarantees it will ever be fixed since I'm not sure this device is going to receive further Android (and consequently graphics driver) updates.

kainino0x:
I've continued looking into this bug and moved the discussion here:
https://crbug.com/709317

PR #2359, opened 2017-04-01
Add new test for corner case videos, and the first one.

kenrussell:
The video being added covers Chromium bug http://crbug.com/701060 .
Verified that all cases fail in current Chromium, and pass with
https://codereview.chromium.org/2791813003 applied.

Tests already all pass in Firefox.

The expectation is that more videos will be added to this test as more
scenarios with uncommon video metadata are uncovered.


kenrussell:
@zhenyao @kainino0x @jdashg @RafaelCintron @grorg please review.

The test will be skipped wherever the video format isn't supported.

Note that I had difficulty running any of the WebGL conformance tests using videos in current Safari on macOS 10.12.3. We'll have to look into why separately.

sdk/tests/conformance/textures/misc/00_test_list.txt
```diff
@@ -25,6 +25,7 @@ texture-active-bind.html
 --min-version 1.0.2 texture-attachment-formats.html
 --min-version 1.0.2 texture-clear.html
 texture-complete.html
+--min-version 1.0.4 texture-corner-case-videos.html
```
zhenyao: nit: out of order

sdk/tests/conformance/textures/misc/texture-corner-case-videos.html
```diff
@@ -0,0 +1,61 @@
...
+<div id="description"></div>
+<div id="console"></div>
+<script>
+"use strict";
+function testPrologue(gl) {
+    return true;
+}
+
+generateTest(
+    'Verify uploading corner-case videos to 2D textures',
```
zhenyao: nit: can you add what corner-case it is? It's not apparent by looking through this test.

sdk/tests/conformance/textures/misc/00_test_list.txt
kenrussell: Oops, sorry. Fixed.

sdk/tests/conformance/textures/misc/texture-corner-case-videos.html
kenrussell: Yes, sorry about that. Have added a comment to each test (only the single test, for now) and displayed it in the test's output.

PR #2360, opened 2017-04-04
Draw into source of copyTexSubImage2D shouldn't crash

kainino0x:
Regression test for https://crbug.com/707445 (crash in Apple Intel driver, OS regression in 10.12.4).


kenrussell:
Excellent minimized test case Kai. Thanks for adding it.

PR #2362, opened 2017-04-07
Added test from Andre Weissflog (@floooh) for draw buffers hang on ma…

kenrussell:
…cOS.

This was originally reported under http://crbug.com/696187 and was
already fixed by the time it was reported to Apple via Radar.

Confirmed that this test runs properly on drivers that don't contain
the bug, and doesn't execute properly (causes the browser to lock up)
on drivers that do. There will be no false positives.

### DIFF
--- a/sdk/tests/conformance2/rendering/draw-buffers-driver-hang.html
+++ b/sdk/tests/conformance2/rendering/draw-buffers-driver-hang.html
@@ -1,0 +1,208 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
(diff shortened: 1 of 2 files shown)


kenrussell:
@floooh sorry I didn't send you this to submit yourself as I'd originally planned. Have not been able until now to find the time to test on the affected hardware and OS version, and fit this into the test harness.

@zhenyao @kainino0x @jdashg in particular please review. This test catches the original driver bug, though it doesn't seem feasible to do so without hanging the browser. On drivers unaffected by the bug, this runs fine, and quickly.

floooh:
@kenrussel no problem! I'm glad you took care of it, thanks! :)

sdk/tests/conformance2/rendering/draw-buffers-driver-hang.html
```diff
@@ -0,0 +1,208 @@
...
+  gl.bindFramebuffer(gl.FRAMEBUFFER, res_fb0);
+  gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, tex0, 0);
+  gl.checkFramebufferStatus(gl.FRAMEBUFFER);
+  let res_fb1 = gl.createFramebuffer();
+  gl.bindFramebuffer(gl.FRAMEBUFFER, res_fb1);
+  gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT1, gl.TEXTURE_2D, tex1, 0);
+  gl.checkFramebufferStatus(gl.FRAMEBUFFER);
+  let res_fb2 = gl.createFramebuffer();
+  gl.bindFramebuffer(gl.FRAMEBUFFER, res_fb2);
+  gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT2, gl.TEXTURE_2D, tex2, 0);
```
zhenyao: why the previous two fb calls checkFramebufferStatus and the last one doesn't?

floooh: probably a copy-paste bug, I created the JS test case from an iOS GL trace log, and left out the iOS specific calls, and with this most likely the 3rd checkFramebufferStatus(), it won't change the test behaviour though.

floooh: It was unclear what exactly caused the hang, just that the sequence of GL calls within the BEGIN/END comments triggered the hang (it was a bit tricky to test since each run made the machine unusable for around a minute) ;)

floooh:
I think @kenrussell needs to press the merge button, I'm not seeing it :)

zhenyao:
Let's merge for now.

kenrussell:
Thanks @zhenyao for merging. It's a good question whether the test case could be reduced further, but the machine which repro'd the hang has since been upgraded so we can't take the risk of modifying the test at this point.

PR #2363, opened 2017-04-11
Increase tolerance to 18 to handle colorspace conversion differences.

kenrussell:


kenrussell:
There's a need for this to handle platform differences -- see https://codereview.chromium.org/2791813003/ -- so merging myself now.

PR #2364, opened 2017-04-11
Add a test for sub source clearBuffer*().

zhenyao:
Surprisingly this is not tested, so here you are!


zhenyao:
@kenrussell @kainino0x @jdashg

kenrussell:
Excellent work Mo. LGTM.

PR #2365, opened 2017-04-12
Remove "normative references" sub-section.

kenrussell:
The use of the term "normative references" here has implications that
were unintended and undesired for the WebGL specifications. Since the
distinction is not critical for this specification, remove it.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -4092,7 +4092,6 @@ extensions.
 
     <h2>References</h2>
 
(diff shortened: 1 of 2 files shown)


kenrussell:
@jdashg @grorg Spec editors: this small change is needed in order to address concerns raised by one of the working group participants. Since I think this is non-controversial, and since it's urgent, I'm taking the liberty of merging it myself now. Please raise concerns or revert if this was a problem.

PR #2366, opened 2017-04-13
Spec & test that local and global variables are initialized

kainino0x:
This test is currently failed by Chrome and Firefox (it needs to be implemented in the ANGLE shader translator, http://anglebug.com/1966

According to @kenrussell,
> the WebGL WG resolved to make this spec and conformance test change a while ago.


zhenyao:
The spec change and test cases look good.  I suggest adding two more test cases

1)

bool flag;
if (flag) {
  gl_fragColor = vec4(1.0, 0.0, 0.0, 1.0);
} else {
  gl_fragColor = vec4(0.0, 0.0, 0.0, 0.0);
}

2) loop variables
for (int i; ...)

kenrussell:
Thanks @kainino0x for adding this spec text and the test. Now we just have to fix the various WebGL implementations to actually enforce this. :)

PR #2367, opened 2017-04-13
Test more types of uninitialized local/global variable

kainino0x:
(Addition to #2366.) Also corrects the expected result (should be [0,0,0,0], not [0,0,0,255]).


kenrussell:
Thanks for adding these new tests @kainino0x and thanks @zhenyao for reviewing them.

I'd hoped we would be farther along in actually implementing this uninitialized variable initialization in ANGLE's shader translator before merging these new tests, but agree that having failing tests is a great motivator to fixing them. :)

PR #2368, opened 2017-04-15
Add regression test for BlitFramebuffer to no-alpha contexts.

kenrussell:


kenrussell:
Note: this isn't ready to be merged yet.

I thought that this would reproduce the behavior of Epic's Zen Garden in http://crbug.com/699566, but am now not sure.

It passes with a prototype fix to Chromium, but fails the "should be legal to blit" check in Firefox (both 52 and Nightly), when trying to blit and resolve a multisampled RBGA8 renderbuffer to the (implicitly RGB) WebGL default back buffer.

It also fails in Chromium when a real RGB texture is used for the back buffer on macOS when specifying:

--disable-features=WebGLImageChromium

even when Chromium's rigid checks about source and destination formats are removed. It looks like there is a real restriction in the OpenGL driver that's not documented in the spec (even in the OpenGL 4.1 spec, which is used on macOS).

I thought that it was legal to drop channels during calls to BlitFramebuffer, like it is during calls to CopyTex(Sub)Image2D. The ES 3.0.5 spec says about BlitFramebuffer:

"If the color formats of the read and draw buffers do not match, and mask includes COLOR_BUFFER_BIT, pixel groups are converted to match the destination format."

In current Chromium, the RGB back buffer is emulated with an RGBA one, and that's why the Zen Garden *used* to not render correctly, but it does now. I wonder whether Firefox's checks were made more strict recently, and/or whether the Zen Garden demo was updated specifically because of this issue. I'm not sure why it is rendering correctly in Chromium right now unless it's doing user agent detection for selecting its back buffer format.

@jdashg @juj @shannonwoods @zhenyao I'd appreciate any help you can offer finding the text that defines BlitFramebuffer's behavior in this case. Also, any hints on changes that were made recently to the Zen Garden demo would be great, since the WebGL 2.0 spec probably needs to be tightened up in this area, and tests added.

zhenyao:
In ES 3.0.4 spec, page 199, Section 4.3.3 Copying Pixels:

"If SAMPLE_BUFFERS for the read framebuffer is one, no copy is performed
and an INVALID_OPERATION error is generated if the formats of the read and
draw framebuffers are not identical or if the source and destination rectangles are
not defined with the same (X0, Y 0) and (X1, Y 1) bounds."

kenrussell:
Thanks @zhenyao for pointing out that spec text. I completely missed it.

This pull request is wrong; don't merge it. However, I'd like to leave it open for the moment since I'm moving the test between multiple machines and this is an easy way to grab the diffs.

kenrussell:
This test is incorrect so I'm closing this pull request. A correct and more comprehensive conformance test has been posted in #2377 .

PR #2369, opened 2017-04-17
Copy into integer cube map texture should work

qkmiao:
This is a regression test for crbug.com/712117.

### DIFF
--- a/sdk/tests/conformance2/textures/misc/00_test_list.txt
+++ b/sdk/tests/conformance2/textures/misc/00_test_list.txt
@@ -2,7 +2,7 @@
 --min-version 2.0.1 angle-stuck-depth-textures.html
 canvas-remains-unchanged-after-used-in-webgl-texture.html
 --min-version 2.0.1 compressed-tex-from-pbo-crash.html
-copy-texture-cube-map-AMD-bug.html
+--min-version 2.0.1 copy-texture-cube-map-AMD-bug.html
 copy-texture-image.html
 copy-texture-image-luma-format.html
 copy-texture-image-webgl-specific.html
--- a/sdk/tests/conformance2/textures/misc/00_test_list.txt
+++ b/sdk/tests/conformance2/textures/misc/00_test_list.txt
@@ -2,6 +2,7 @@
 --min-version 2.0.1 angle-stuck-depth-textures.html
 canvas-remains-unchanged-after-used-in-webgl-texture.html
 --min-version 2.0.1 compressed-tex-from-pbo-crash.html
+copy-texture-cube-map-AMD-bug.html
 copy-texture-image.html
 copy-texture-image-luma-format.html
 copy-texture-image-webgl-specific.html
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/textures/misc/00_test_list.txt
```diff
@@ -2,6 +2,7 @@
 --min-version 2.0.1 angle-stuck-depth-textures.html
 canvas-remains-unchanged-after-used-in-webgl-texture.html
 --min-version 2.0.1 compressed-tex-from-pbo-crash.html
+copy-texture-cube-map-AMD-bug.html
```
zhenyao: Can you add --min-version here?

sdk/tests/conformance2/textures/misc/copy-texture-cube-map-AMD-bug.html
```diff
@@ -0,0 +1,92 @@
...
+<script>
+"use strict";
+var wtu = WebGLTestUtils;
+var gl;
+
+function runTest() {
+  var width = 16;
+  var height = 16;
+  var level = 0;
+  var internalformat = gl.RGBA8UI;
```
zhenyao: Can we expand the test to cover also RGBA8I, RGBA8, RGBA32F?

qkmiao: I think RGBA32F isn't supported by CopyTexImage2D in chrome now. Should we support floating formats  if https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float/ is enabled? I didn't see explicit specification about this in the extension spec.

sdk/tests/conformance2/textures/misc/00_test_list.txt
qkmiao: I added min-version 2.0.1 for this test.

sdk/tests/conformance2/textures/misc/copy-texture-cube-map-AMD-bug.html
qkmiao: I added other formats. We can easily add floating formats in the test if we decide to.

qkmiao:
Thanks @zhenyao . I updated this pr to fix your comments. PTAL.

msc-:
> Should we support floating formats if https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float/ is enabled? I didn't see explicit specification about this in the extension spec.

That's because it is mostly covered in the core spec. The extension spec. modifies the CopyTexImage INVALID_OPERATION error on p.139 so it is only raised when trying to copy floating point data to a format that is not floating- or fixed-point.

ReadPixels is modified to support RGBA and FLOAT. (Table 3.15 should have been modified to include RGBA and FLOAT but I the modification to ReadPixels makes this combination valid.

Table 3.17 is modified to include RG{B,BA}32F.

The core spec. says

> If internalformat is sized, the internal format of the new texel array is internalformat, and this is also the new texel array’s effective internal format. If the component sizes of internalformat do not exactly match the corresponding component sizes of the source buffer’s effective internal format, described below, an INVALID_OPERATION error is generated.

So the conclusion is that copying RGBA32F to RGBA32F is supported.

qkmiao:
@msc- I think you are right. Thanks. I will fix this issue in chrome and add conformance test to check this.

zhenyao:
LGTM
Please add float test case with extension check in a followup CL

Issue #2370, opened 2017-04-17
Clarify intended usage of STENCIL_INDEX constant


grovesNL:
Originally I had [posted this question on StackOverflow](http://stackoverflow.com/questions/43445727/where-is-the-stencil-index-constant-used-in-webgl) and @greggman's answer mentioned I should probably cross-post it here.

> In the [WebGL 1 specification][1], I noticed there is a constant `const GLenum STENCIL_INDEX = 0x1901;` and was wondering how it is used. I have been attempting to use it with either textures or renderbuffers although none of the combinations I've attempted have worked. For example, I've attempted using `STENCIL_INDEX` instead of `STENCIL_INDEX8` for the renderbuffer in the example below, or calling `texImage2D` with `STENCIL_INDEX8` as internal format and `STENCIL_INDEX` as format.
> 
> In either WebGL 1 or 2 (with any extensions enabled if required), I am wondering where the constant `STENCIL_INDEX` could be used in either a texture, renderbuffer, or some attachment configuration. Even if this is hardware specific I'm still interested in knowing how it is used.
> 
> For example, I am curious how the code below could be modified to use `STENCIL_INDEX` (*not* `STENCIL_INDEX8` -- note that this constant already being used in a renderbuffer):
> 
> 
> 
> 
> 
>     const pixel = new Uint8Array([0, 0, 0, 0]);
> 
>     for (const version of ['webgl', 'webgl2']) {
>       const canvas = document.createElement('canvas'),
>         gl = canvas.getContext(version, {stencil: true}),
>         framebuffer = gl.createFramebuffer(),
>         stencilbuffer = gl.createRenderbuffer(),
>         texture = gl.createTexture();
> 
>       canvas.width = 1;
>       canvas.height = 1;
> 
>       gl.bindFramebuffer(gl.FRAMEBUFFER, framebuffer);
> 
>       gl.bindRenderbuffer(gl.RENDERBUFFER, stencilbuffer);
>       gl.renderbufferStorage(gl.RENDERBUFFER, gl.STENCIL_INDEX8, 1, 1);
>       gl.framebufferRenderbuffer(gl.FRAMEBUFFER, gl.STENCIL_ATTACHMENT, gl.RENDERBUFFER, stencilbuffer);
> 
>       gl.bindTexture(gl.TEXTURE_2D, texture);
[...]

PR #2371, opened 2017-04-17
Remove STENCIL_INDEX from WebGL 1.0 spec

zhenyao:
This enum is not used at all in WebGL 1.0 and 2.0.

This is pointed out in https://github.com/KhronosGroup/WebGL/issues/2370

### DIFF
--- a/conformance-suites/1.0.1/conformance/context/constants.html
+++ b/conformance-suites/1.0.1/conformance/context/constants.html
@@ -391,7 +391,6 @@ RGBA4                          : 0x8056,
 RGB5_A1                        : 0x8057,
 RGB565                         : 0x8D62,
 DEPTH_COMPONENT16              : 0x81A5,
-STENCIL_INDEX                  : 0x1901,
 STENCIL_INDEX8                 : 0x8D48,
 DEPTH_STENCIL                  : 0x84F9,
 
--- a/conformance-suites/1.0.2/conformance/context/constants.html
+++ b/conformance-suites/1.0.2/conformance/context/constants.html
@@ -414,7 +414,6 @@ RGBA4                          : 0x8056,
 RGB5_A1                        : 0x8057,
 RGB565                         : 0x8D62,
 DEPTH_COMPONENT16              : 0x81A5,
-STENCIL_INDEX                  : 0x1901,
 STENCIL_INDEX8                 : 0x8D48,
 DEPTH_STENCIL                  : 0x84F9,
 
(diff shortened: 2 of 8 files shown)


zhenyao:
@kenrussell Please review
@kainino0x @jdashg @grorg @RafaelCintron FYI

kenrussell:
I'm afraid this needs larger changes. The conformance tests need to be updated. The top-of-tree tests should require that this enum not be present, but the older snapshots need to be updated to allow the property to be either present or absent. See ignoredProperties in for example conformance-suites/1.0.3/conformance/context/constants-and-properties.html .

zhenyao:
Updated tests also. @kenrussell Please take another look.

zhenyao:
Tested with Chrome with/without STENCIL_INDEX removed from WebGLRenderingContext

specs/1.0.0/index.html
```diff
@@ -1214,7 +1214,6 @@ <h2 class="no-toc">Table of contents</h2>
     const GLenum RGB5_A1                        = 0x8057;
     const GLenum RGB565                         = 0x8D62;
     const GLenum DEPTH_COMPONENT16              = 0x81A5;
-    const GLenum STENCIL_INDEX                  = 0x1901;
```
kenrussell: Let's not modify the old specs, just the old conformance tests.

specs/1.0.1/index.html
```diff
@@ -1434,7 +1434,6 @@ <h2 class="no-toc">Table of contents</h2>
     const GLenum RGB5_A1                        = 0x8057;
     const GLenum RGB565                         = 0x8D62;
     const GLenum DEPTH_COMPONENT16              = 0x81A5;
-    const GLenum STENCIL_INDEX                  = 0x1901;
```
kenrussell: Same here.

specs/1.0.2/index.html
```diff
@@ -1448,7 +1448,6 @@ <h2 class="no-toc">Table of contents</h2>
     const GLenum RGB5_A1                        = 0x8057;
     const GLenum RGB565                         = 0x8D62;
     const GLenum DEPTH_COMPONENT16              = 0x81A5;
-    const GLenum STENCIL_INDEX                  = 0x1901;
```
kenrussell: Same here.

specs/1.0.3/index.html
```diff
@@ -1509,7 +1509,6 @@ <h2 class="no-toc">Table of contents</h2>
     const GLenum RGB5_A1                        = 0x8057;
     const GLenum RGB565                         = 0x8D62;
     const GLenum DEPTH_COMPONENT16              = 0x81A5;
-    const GLenum STENCIL_INDEX                  = 0x1901;
```
kenrussell: Same here.

specs/1.0.0/index.html
zhenyao: Done.  Here and other spec snapshots.

grovesNL:
Thanks @zhenyao and @kenrussell

PR #2373, opened 2017-04-19
Add a regression test for crbug.com/676719

zhenyao:


zhenyao:
This test is modified from the test case provided by reporter for crbug.com/676719
@kenrussell @kainino0x Please review

kainino0x:
As long as we can rely on the video rendering the same way twice (which I think is necessary), and I think we can, LGTM

sdk/tests/conformance/textures/misc/tex-video-using-tex-unit-non-zero.html
```diff
@@ -0,0 +1,175 @@
...
...IjeDLC8VARZe6K/kyAhQ4ahp8lRpy95iUXvqnHM+JvMl5iUF2fyHOCgq8rS5fgfrD0b+uq+nbVwAAJtMAAABHfq1Na1DAOo0YYEFpJdrEwUsXFlPvJh77UfjoJVvapBvh/JsOuz6RawoR2u9xIs1MQtz441KWOR4FxE6DUmqdhwtABuhbDIwuWlnUjzILqWcfaMP28wbT+Zu2TAsgAoHdKYAtbVFyro3bgASgoADdlUGBEEck8Z7XaOo/EAAA1hprNARKQunZugP/kVjlnOxVG38NMmQVV0RUVnFdAKwZQA9naG6wR+ci75gyT6HQe6ObBccm1VzKMkLOkoYu+PCAAAAAIUAAAAA0AAADV/jFDAvLmm384wUAA1MLiBcoL5AC2h4diIuEdM18TI/2P8DK+oBzNQr57D9qKS0BxE3oAcKcH1TFsGY6AAAAAAAAACghFBnJ9dvl5GoafC7bhk1AQeQz3jwAA+Olf3ew2JE8l4FM80HcAAB/7PT0z1MjflBwcgF15L7srY/mL2cZB+Vcc9vXXGAXHAtXf7BAvrTUyfEDBwh3HI+8dgD58r/qUmVVWOF7IJKMPi/hQA6ZZ+jyYTA7kwIvY+qUvpVNUsjGNeCEBr3bQanZh2mkCAVqolsAAPpfpzQX8eQyAMgA8n36orO1Mfkjp8ZIzpW+nrNJ3S22uX1r5uPvo5jX/z3pxVGL4wRHH29p+1+7TcLnjwioYnxANpM8H8zG6lbJQDgm5kDbBp8M1XyWUpPDjLx+GoBPtw0WDEKr92kxS1diwUilTAn7wavvbVhDyYSfF6C5ZDMMvEIKWnFAUIM7XsBe8b04fc89d+kAagdwCd1IhxgCV2DKlOLO+IHqAMkhR5DhXileE7SrPxEVUyOECNU4ModYiBlGAS/9R+2RG60n4SYMqEr8oPMP61rVnfUir/UFgfgtDEvOrniAAmHjcM3+mAAAAikXAAT9d10oLjHck5F1OPGgb5CHYidLeaFUOerzkcTgOZiGzGPxuvgFYjDbO/hU+WLUgIfTNADXr7Wo00nYCa/Cn1KUcea+pjhjBTq5ZB4oTEcBHpV8Xqm9vJ8AAAHoENY/yywOSAAAASz31TZ1YUAAAACvIAAADxkAGplRAAZyNmES+sALzD/0baSIAABxTu2sBAAAAAAAAEbuPs4EAt4r3gQHxggFu8IED"></video>
```
kenrussell: Does this video contain more than a single frame? The "loop autoplay" is concerning because it means the video could conceptually advance in time between the calls to texImage2D, below.

It would be better if the video were paused at the beginning of runTest(), or if it weren't played at all, but instead just displayed the first frame.

sdk/tests/conformance/textures/misc/tex-video-using-tex-unit-non-zero.html
```diff
@@ -0,0 +1,175 @@
...
+gl.uniform1i(loc1, 1);
+
+wtu.setupUnitQuad(gl, 0, 1);
+
+// Load textures
+var imageElement = document.getElementById("src-image");
+var videoElement = document.getElementById("src-video");
+
+// Need to wait for video to be ready before continuing
+videoElement.addEventListener('canplay', runTest);
```
kenrussell: I think you should listen for the 'canplaythrough' event instead. See these video-related pixel tests in Chromium, which have been reliable:
https://cs.chromium.org/chromium/src/content/test/data/gpu/pixel_video_mp4.html

sdk/tests/conformance/textures/misc/tex-video-using-tex-unit-non-zero.html
```diff
@@ -0,0 +1,175 @@
...
+  gl.bindTexture(gl.TEXTURE_2D, texture1);
+  gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, imageElement);
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
+  gl.bindTexture(gl.TEXTURE_2D, null);
+
+  //gl.clearColor(1.0, 0.0, 1.0, 1.0);
+  //gl.clear(gl.COLOR_BUFFER_BIT);
```
kenrussell: Could you remove this commented-out code?

sdk/tests/conformance/textures/misc/tex-video-using-tex-unit-non-zero.html
```diff
@@ -0,0 +1,175 @@
...
+// Load textures
+var imageElement = document.getElementById("src-image");
+var videoElement = document.getElementById("src-video");
+
+// Need to wait for video to be ready before continuing
+videoElement.addEventListener('canplay', runTest);
+
+function runTest() {
+  // Prevent further calls to oncanplay
+  videoElement.removeEventListener('canplay', runTest);
```
kenrussell: If the video actually needs to play at all, this would be a good place to pause it.

sdk/tests/conformance/textures/misc/tex-video-using-tex-unit-non-zero.html
zhenyao: Done

sdk/tests/conformance/textures/misc/tex-video-using-tex-unit-non-zero.html
zhenyao: Removed loop autoplay, and changed to event canplaythrough instead.

zhenyao:
Revised. @kenrussell Please take another look.

sdk/tests/conformance/textures/misc/tex-video-using-tex-unit-non-zero.html
```diff
@@ -0,0 +1,169 @@
...
+gl.uniform1i(loc1, 1);
+
+wtu.setupUnitQuad(gl, 0, 1);
+
+// Load textures
+var imageElement = document.getElementById("src-image");
+var videoElement = document.getElementById("src-video");
+
+// Need to wait for video to be ready before continuing
+videoElement.addEventListener('canplaythrough', runTest);
```
kenrussell: Sorry, one other comment. It's possible that some browsers will not have support for the webm format, and that's legal. It's important to check whether the video is playable on the browser using videoElement.canPlayType(). See:

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-video.js

Look for the call to canPlayType(), and also the types listed at the top of the file.

It'll be necessary to figure out what the format is of the webm stream in this file, and call canPlayType() with the appropriate format string.

zhenyao:
Updated the tests with four different formats.  On my mac, three formats are supported, and they all fail without the fix in https://codereview.chromium.org/2825223002/, and pass with the fix.

@kenrussell Please take another look.

sdk/tests/conformance/textures/misc/tex-video-using-tex-unit-non-zero.html
```diff
@@ -0,0 +1,218 @@
...
+      }
+      if(!video.canPlayType(info.type).replace(/no/, '')) {
+        debug(info.type + " unsupported");
+        runNextVideo();
+        return;
+      };
+      document.body.appendChild(video);
+      video.type = info.type;
+      video.src = info.src;
+      wtu.startPlayingAndWaitForVideo(video, runTest);
```
kenrussell: I still wonder whether because the video will be playing during the test, whether this test will be flaky.

sdk/tests/conformance/textures/misc/tex-video-using-tex-unit-non-zero.html
```diff
@@ -0,0 +1,218 @@
...
+    }
+    function runTest() {
+      var texture0 = gl.createTexture();
```
kenrussell: Should we pause the current video here, so that we're guaranteed that the two texImage2D calls taking the video element receive the same data?

zhenyao: Done. @kenrussell Please review again.

kenrussell:
Very nice. Good work on the test. LGTM

PR #2375, opened 2017-04-19
Fix #2274: Add GL_RGB10_A2 to texImageSource variants of tex(Sub)Image*

### LINKED ISSUE #2273: Test suite depends on random images on the web

### LINKED ISSUE #2274: Spec: GL_RGB10_A2 format missing  in texImageSource variants of tex(Sub)Image2D
An entry:
- Internal format: GL_RGB10_A2
- format: GL_RGBA
- type: GL_UNSIGNED_INT_2_10_10_10_REV

is missing from the table defining supported internal_format&format&type combinations to use with with TexImageSource source  [1].

As it is normalized fixed-point format [2] with higher precision and the note [1] is not applicable to it:
> When the data source is a DOM element (HTMLImageElement, HTMLCanvasElement, or HTMLVideoElement), or is an ImageBitmap or ImageData object, commonly each channel's representation is an unsigned integer type of at least 8 bits. Converting such representation to signed integers or unsigned integers with more bits is not clearly defined. For example, when converting RGBA8 to RGBA16UI, it is unclear whether or not the intention is to scale up values to the full range of a 16-bit unsigned integer. Therefore, only converting to unsigned integer of at most 8 bits, half float, or float is allowed.

[1] 
https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.7.6
table under  [throws] void texImage2D(GLenum target, GLint level, GLint internalformat, GLsizei width, GLsizei height, GLint border, GLenum format, GLenum type, TexImageSource source) 

[2]
https://www.khronos.org/opengl/wiki/Image_Format#Special_color_formats

astojilj:
Chromium implementation is in review request: [https://crrev.com/2821363003/](https://crrev.com/2821363003/).
Verified by running tests locally on OSX.

### DIFF
--- a/sdk/tests/conformance2/textures/canvas/00_test_list.txt
+++ b/sdk/tests/conformance2/textures/canvas/00_test_list.txt
@@ -25,6 +25,7 @@ tex-2d-rgba8-rgba-unsigned_byte.html
 tex-2d-srgb8_alpha8-rgba-unsigned_byte.html
 tex-2d-rgb5_a1-rgba-unsigned_byte.html
 tex-2d-rgb5_a1-rgba-unsigned_short_5_5_5_1.html
+tex-2d-rgb10_a2-rgba-unsigned_int_2_10_10_10_rev.html
 tex-2d-rgba4-rgba-unsigned_byte.html
 tex-2d-rgba4-rgba-unsigned_short_4_4_4_4.html
 tex-2d-rgba16f-rgba-half_float.html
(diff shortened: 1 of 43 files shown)


astojilj:
@kenrussell , @zhenyao, PTAL. 
FYI @Richard-Yunchao.

astojilj:
@kenrussell, a note related to [our discussion](https://github.com/KhronosGroup/WebGL/issues/2274#issuecomment-282441796) about  [Chromium implementation](https://crrev.com/2821363003/): both GPU [copy](https://codereview.chromium.org/2821363003/patch/60001/70001) and [CPU](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/platform/graphics/gpu/WebGLImageConversion.cpp?type=cs&l=638) side conversion were already implemented - mostly, just had to enable it. Patching the https://crrev.com/2821363003/ to Chromium on OSX makes the test pass.

zhenyao:
You should also check in the change in tex_image_test_generator.py

astojilj:
Thanks @zhenyao. Fixed.

zhenyao:
LGTM

kenrussell:
I'm sorry but I'm reverting this because https://codereview.chromium.org/2821363003/ hasn't been sent out for review yet. (You need to press 'm' to officially publish it and select reviewers.) Landing this now would block WebGL conformance rolls into Chromium until many new suppressions were added. Please put this back up for review when https://codereview.chromium.org/2821363003/ is closer to landing.

PR #2377, opened 2017-04-19
Specify, and add tests for, resolving multisampled renderbuffers to t…

kenrussell:
…he back buffer.

This behavior was implicitly defined earlier and is now clearly
specified and tested. This acts as a regression test for
http://crbug.com/699566 , a failure to correcly render Epic's Zen
Garden demo on macOS. The new test already passes in Firefox.

### DIFF
--- a/sdk/tests/conformance2/rendering/blitframebuffer-resolve-to-back-buffer.html
+++ b/sdk/tests/conformance2/rendering/blitframebuffer-resolve-to-back-buffer.html
@@ -65,7 +65,7 @@ function runTest(testParams) {
 
     // Find the supported samples for a multisampled renderbuffer of the appropriate internal format.
     let samples = gl.getInternalformatParameter(gl.RENDERBUFFER, gl[testParams.internalformat], gl.SAMPLES);
-    if (!samples) {
+    if (!samples || !samples.length) {
         testFailed("At least one multisampled format is required to be supported");
         return;
     }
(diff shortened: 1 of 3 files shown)


kenrussell:
@zhenyao @jdashg @kainino0x in particular please review.

This fails on Chrome Canary, but passes with the forthcoming fix for http://crbug.com/699566 .

The test already passes on Firefox.

sdk/tests/conformance2/rendering/blitframebuffer-resolve-to-back-buffer.html
```diff
@@ -0,0 +1,237 @@
...
+
+    var canvas = document.createElement('canvas');
+    canvas.width = sz;
+    canvas.height = sz;
+    document.getElementById('canvasHeader').appendChild(canvas);
+    var gl = wtu.create3DContext(canvas, testParams.attribs, 2);
+
+    // Find the supported samples for a multisampled renderbuffer of the appropriate internal format.
+    let samples = gl.getInternalformatParameter(gl.RENDERBUFFER, gl[testParams.internalformat], gl.SAMPLES);
+    if (!samples) {
```
zhenyao: Just curious: are we returning an empty set or None when failing? If empty set, does !empty_set work?

kenrussell: Excellent question. You're right, this check was wrong. It needed to be:
  if (!samples || !samples.length)

Fixed in the current version.

zhenyao:
LGTM

PR #2378, opened 2017-04-20
Add tex-base-level-bug.html to detect driver bug of texImage2D.

JiangYizhou:
This is a regression test for Chromium Issue 705865.
First, Set texture base level > 0, then set texture level 0
result in pixel comparison error of its base level.

The test case fails on mac and windows.

### DIFF
--- a/sdk/tests/conformance2/textures/misc/00_test_list.txt
+++ b/sdk/tests/conformance2/textures/misc/00_test_list.txt
@@ -13,6 +13,7 @@ mipmap-fbo.html
 --min-version 2.0.1 npot-video-sizing.html
 --min-version 2.0.1 tex-3d-mipmap-levels-intel-bug.html
 tex-3d-size-limit.html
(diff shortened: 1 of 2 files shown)


JiangYizhou:
I add a conformance test to catch this regression.
A related chromium patch is here(https://codereview.chromium.org/2772193003/).
@zhenyao @kenrussell @qkmiao @Richard-Yunchao PTAL.

qkmiao:
This looks good to me. I discussed this pr with yizhou already before she sent it. Please help to review @zhenyao @kenrussell

kenrussell:
Thanks @JiangYizhou and @qkmiao for this test. It's very small and concise.

I'm surprised to see it fail on Windows, because ANGLE's OpenGL ES validation is generally very close to the spec, but as far as I can tell, the test is correct.

Merging now. @zhenyao @kainino0x @null77 @vonture FYI.

PR #2379, opened 2017-04-21
Verify that EXT_texture_norm16 enums are not supported.

astojilj:
See https://crrev.com/2767063002/#msg48

### DIFF
--- a/sdk/tests/conformance2/textures/misc/tex-image-with-bad-args.html
+++ b/sdk/tests/conformance2/textures/misc/tex-image-with-bad-args.html
@@ -39,7 +39,7 @@
 <script>
 "use strict";
 description('Tests texImage2D with invalid internalformat/format/type combinations');
-debug("This is a regression test for <a href='https://bugs.chromium.org/p/chromium/issues/detail?id=656889'>Chromium Issue 656889</a>");
+debug("This includes a regression test for <a href='https://bugs.chromium.org/p/chromium/issues/detail?id=656889'>Chromium Issue 656889</a>");
 
 var wtu = WebGLTestUtils;
 var gl = wtu.create3DContext("c", null, 2);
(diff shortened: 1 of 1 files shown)


astojilj:
@kenrussell, PTAL. Thanks.

sdk/tests/conformance2/textures/misc/tex-image-with-bad-args.html
```diff
@@ -65,6 +65,8 @@
     { internalformat: gl.RGBA, format: gl.RGBA, type: gl.HALF_FLOAT, errors: [gl.INVALID_OPERATION] },
     { internalformat: gl.LUMINANCE, format: gl.LUMINANCE, type: gl.FLOAT, errors: [gl.INVALID_OPERATION] },
     { internalformat: gl.LUMINANCE_ALPHA, format: gl.LUMINANCE_ALPHA, type: gl.HALF_FLOAT, errors: [gl.INVALID_OPERATION] },
+    { internalformat: 0x822A /*GL_R16_EXT*/ , format: gl.RED, type: gl.UNSIGNED_SHORT, errors: [gl.INVALID_VALUE] },
+    { internalformat: gl.RED, format: gl.RED, type: gl.UNSIGNED_SHORT, errors: [gl.INVALID_VALUE] },
```
kenrussell: Are these guaranteed to generate INVALID_VALUE, or if the validation occurs in a different order, could they generate INVALID_OPERATION? I'm having difficulty finding the place in the ES 3.0.5 spec defining this.

astojilj: I've found this:
"GL_INVALID_VALUE is generated if internalFormat is not one of the accepted resolution and format symbolic constants."
https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glTexImage2D.xhtml
INVALID_VALUE looks consistent with other out-of-range handling of individual parameters (and not only related to this method).

kenrussell: Thanks, but I also see this, from section 3.8.3 "Texture Image Specification" from page 127 of the ES 3.0.5 specification:

"Specifying a combination of values for format, type, and internalformat that is not listed as a valid combination in tables 3.2 or 3.3 generates the error INVALID_OPERATION."

The second call here with internalformat=GL_RED, format=GL_RED, and type GL_UNSIGNED_SHORT, can be considered to violate multiple rules -- e.g. if GL_RED should be considered GL_R8, the type would still be invalid (UNSIGNED_SHORT instead of UNSIGNED_BYTE).

Please change the second of these to also accept INVALID_OPERATION and then I'll merge it. Thanks.

msc-:
> Please change the second of these to also accept INVALID_OPERATION and then I'll merge it. Thanks.

I think that case should accept only INVALID_OPERATION. gl.RED and gl.UNSIGNED_SHORT are valid formats and type enum values. It is the combination that is incorrect, therefore INVALID_OPERATION.

kenrussell:
Mark, I don't think it's clear. GL_RED isn't a legal internalformat for ES 3.0 under any circumstances because it's unsized, and only a limited number of unsized internal formats are supported for backward compatibility. That's probably the reason Chrome at least is producing INVALID_VALUE. However different browsers might do their validation differently. Up to this point, WebGL's negative tests which could plausibly generate multiple OpenGL errors have allowed all of them.

msc-:
> GL_RED isn't a legal internalformat for ES 3.0 under any circumstances
Yes, that's true. It's probably best to allow both errors.

astojilj:
Done - both errors allowed. PTAL.
Checked only with Firefox and Chrome and they generate INVALID_VALUE.

kenrussell:
Very nice. Thank you. LGTM

PR #2381, opened 2017-04-23
Fetch WebGL 2.0 context by default in object-deletion-behaviour-2 test.

kenrussell:
Makes it easier to run it from its URL rather than through the harness.


kenrussell:
Trivial change. Merging myself.

PR #2383, opened 2017-04-25
Verify that samplers' state doesn't leak across contexts.

kenrussell:
Regression test for http://crbug.com/713127 .


kenrussell:
@kainino0x in particular please review.

@zhenyao @jdashg and others: FYI.

kenrussell:
Merging in order to be able to start running this test on our automated test systems.

PR #2387, opened 2017-05-01
Support SharedArrayBuffers

domenic:
This annotates all ArrayBuffer or ArrayBufferView types with [AllowShared], per the proto-specification at https://tc39.github.io/ecmascript_sharedmem/dom_shmem.html#webgl, extended to also modify the Web GL 2.0 entrypoints. This allows the backing data to be a SharedArrayBuffer, instead of only non-shared ArrayBuffers, per https://github.com/heycam/webidl/pull/353.

This also makes some related editorial changes, removing the redundant BufferDataSource typedef in favor of the defined-in-Web IDL BufferSource type, and consolidating two overloads of bufferData (one accepting ArrayBuffer? and the other accepting ArrayBufferView) into a single overload accepting BufferSource?.

---

This is my first time contributing to the Web GL spec, so help appreciated with any policies, procedures, or style issues.

I also imagine we'll want to update the conformance test suite as well, but that's outside my capabilities. @binji or @flagxor may be able to help with that. I've left "allow edits from maintainers" checked so that others can push such test additions directly to this PR, if they like.

This should not be merged until https://github.com/heycam/webidl/pull/353 is.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -1238,8 +1238,8 @@ typedef (ImageBitmap or
          HTMLCanvasElement or
          HTMLVideoElement) TexImageSource;
 
-typedef (Float32Array or sequence&lt;GLfloat&gt;) Float32List;
-typedef (Int32Array or sequence&lt;GLint&gt;) Int32List;
+typedef ([AllowShared] Float32Array or sequence&lt;GLfloat&gt;) Float32List;
+typedef ([AllowShared] Int32Array or sequence&lt;GLint&gt;) Int32List;
 
 [NoInterfaceObject]
 interface <dfn id="WebGLRenderingContextBase">WebGLRenderingContextBase</dfn>
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -1232,7 +1232,6 @@ for (var i = 0; i < numVertices; i++) {
     See <a href="#CONTEXT_LOST">the context lost event</a> for further details.
     </p>
     <pre class="idl">
-typedef (ArrayBuffer or ArrayBufferView) BufferDataSource;
 typedef (ImageBitmap or
          ImageData or
          HTMLImageElement or
@@ -1690,9 +1689,8 @@ interface <dfn id="WebGLRenderingContextBase">WebGLRenderingContextBase</dfn>
                            GLenum srcAlpha, GLenum dstAlpha);
 
     void bufferData(GLenum target, GLsizeiptr size, GLenum usage);
-    void bufferData(GLenum target, ArrayBuffer? data, GLenum usage);
-    void bufferData(GLenum target, ArrayBufferView data, GLenum usage);
-    void bufferSubData(GLenum target, GLintptr offset, BufferDataSource data);
+    void bufferData(GLenum target, [AllowShared] BufferSource? data, GLenum usage);
+    void bufferSubData(GLenum target, GLintptr offset, [AllowShared] BufferSource data);
 
     [WebGLHandlesContextLoss] GLenum checkFramebufferStatus(GLenum target);
     void clear(GLbitfield mask);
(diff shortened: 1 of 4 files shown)


binji:
When I [enabled this in Chrome](https://chromium.googlesource.com/chromium/src/+/ed24fe201ef7afca59b9562443cdbce3b1ce8142%5E%21/#F43), I also added the `[AllowShared]` attribute to `Float32Array` (e.g. `uniformMatrix2fv`, among many others) , `Int32Array` (`vertexAttribI4iv`), and `Uint32Array` (`vertexAttribI4uiv`) parameters.

domenic:
Ah, great to know. That wasn't suggested in https://tc39.github.io/ecmascript_sharedmem/dom_shmem.html#webgl, but it does seem like the principal is to be maximal rather than minimal. I'll upload another commit to add those.

/cc @lars-t-hansen

domenic:
What about the functions like getUniform and getVertexAttrib that return typed arrays? Are they backed by SABs, or still normal ABs?

binji:
Just scanning the code in Chrome, these seem to generally return new non-shared typed arrays w/ new backing stores. The one exception seems to be calling `getVertexAttrib` with `ARRAY_BUFFER_BINDING`, which seems to return an existing non-shared ArrayBuffer.

But I'm no expert here, so I may have missed something.

lars-t-hansen:
@domenic, there was no grand principle at work, I think.  @juj provided for shared memory on APIs that were essential to make WebGL work with shared memory from compiled asm.js code (now webassembly code).

Note that adding `[AllowShared]` is likely to lead to situations where memory might be accessed racily by the implemenation.  It is important to consider the security consequences of that, esp time-of-check-to-time-of-use consequences.  IMO the principle should be to be closer to minimal, not closer to maximal.

domenic:
Some offline discussion between WebAssembly/SharedArrayBuffer implementers concluded that the more maximal scope in this PR is good to go, but we should carefully review any further additions for the reasons @lars-t-hansen states.

domenic:
The prerequisite Web IDL pull request has been merged. So this is ready for editor review.

specs/latest/1.0/index.html
```diff
@@ -1690,9 +1689,8 @@ <h2 class="no-toc">Table of contents</h2>
                            GLenum srcAlpha, GLenum dstAlpha);
 
     void bufferData(GLenum target, GLsizeiptr size, GLenum usage);
-    void bufferData(GLenum target, ArrayBuffer? data, GLenum usage);
-    void bufferData(GLenum target, ArrayBufferView data, GLenum usage);
-    void bufferSubData(GLenum target, GLintptr offset, BufferDataSource data);
+    void bufferData(GLenum target, [AllowShared] BufferSource? data, GLenum usage);
```
kenrussell: I assume that passing null for this union type to bufferData won't cause a failure to resolve the overload -- that it will still generate an INVALID_OPERATION OpenGL error per the spec.

specs/latest/2.0/index.html
```diff
@@ -699,21 +699,20 @@ <h2 class="no-toc">Table of contents</h2>
   /* Buffer objects */
   // WebGL1:
   void bufferData(GLenum target, GLsizeiptr size, GLenum usage);
-  void bufferData(GLenum target, ArrayBuffer? srcData, GLenum usage);
-  void bufferData(GLenum target, ArrayBufferView srcData, GLenum usage);
-  void bufferSubData(GLenum target, GLintptr dstByteOffset, BufferDataSource srcData);
+  void bufferData(GLenum target, [AllowShared] BufferSource? srcData, GLenum usage);
```
kenrussell: Same comment here.

specs/latest/1.0/index.html
domenic: Indeed, `null` is a valid value of type `[AllowShared] BufferSource? data`, so it'll still end up hitting the prose which causes the error.

kenrussell:
Since the spec editors haven't commented, let's merge this now. Thanks @domenic for adding this support.

PR #2388, opened 2017-05-01
Regression test for cached-load of programs w/ uniform blocks

kainino0x:
Regression test for http://crbug.com/716018


kainino0x:
This is a very targeted test. Would it benefit from covering more types of shader variables?

kenrussell:
Very nice test. While I do think we should test more types of shader variables, let's leave this as a very targeted test and add more as they're found to be necessary.

PR #2389, opened 2017-05-02
transform_feedback: test multiple compiles of the same shaders

kainino0x:
This extra test recompiles the transform feedback shader program while
preventing getShaderParameter from forcing a bypass of Chromium's shader
cache.

Passes in Firefox, fails in Chrome. Passes in Chrome with a patch.

### DIFF
--- a/sdk/tests/conformance2/transform_feedback/transform_feedback.html
+++ b/sdk/tests/conformance2/transform_feedback/transform_feedback.html
@@ -449,6 +449,7 @@ function runVaryingsTest() {
     verifyTransformFeedbackVarying(program, 2, false);
 
     // Test recompiling/relinking the program
+    // Regression test for http://crbug.com/716018
     var skipCompileStatus = true;
     program = wtu.setupTransformFeedbackProgram(gl, ["vshader", "fshader"],
         ["out_add", "out_mul"], gl.SEPARATE_ATTRIBS,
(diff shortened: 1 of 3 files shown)


kainino0x:
I realized there's something going wrong here, please don't merge yet.

sdk/tests/conformance2/transform_feedback/transform_feedback.html
```diff
@@ -439,8 +439,12 @@
     verifyTransformFeedbackVarying(program, 1, true, "out_mul");
     verifyTransformFeedbackVarying(program, 2, false);
 
-    // Now relink.
-    gl.linkProgram(program);
+    // Now recompile/relink the program
+    var skipCompileStatus = true;
+    program = wtu.setupTransformFeedbackProgram(gl, ["vshader", "fshader"],
+        ["out_add", "out_mul"], gl.SEPARATE_ATTRIBS,
+        ["in_data"], skipCompileStatus);
```
kenrussell: It looks to me like "skipCompileStatus" is matching the "opt_locations" argument of setupTransformFeedbackProgram instead of "opt_skipCompileStatus".

kainino0x: Thanks, fixed

kainino0x:
Okay, problems fixed; PTAL

kainino0x:
Added a tiny fixup to uniform-buffers-second-compile.html as well.

kenrussell:
LGTM. Good work tracking this down.

PR #2390, opened 2017-05-03
Add the concept of an opaque multiview framebuffer to WEBGL_multiview

Oletus:
An opaque multiview framebuffer acts as a multiview framebuffer, but
has immutable attachments defined by another web API. Under the hood,
the views of an opaque multiview framebuffer could be implemented in
whichever way is most optimal for a given situation. For example, they
could be texture layers, or side-by-side viewports on a 2D texture.

This will make it possible to create a zero-copy display pipeline that
makes use of multiview rendering. Another web API such as WebVR 2.0
could expose an opaque multiview framebuffer for rendering, and
specify rules when it gets displayed or invalidated. Since the
framebuffer is owned entirely by the other web API and only interacts
with WebGL in the role of a render target, the interactions between
WebGL and the other API remain fairly simple.

The opaque multiview framebuffer would be possible to support in
WebGL 1.0 as well, since it does not require API user facing support
for texture arrays. This revision doesn't yet fully address WebGL 1.0
compatibility of the extension, but is intended as one step towards
that.

This spec also leaves open the possibility to expose the underlying
memory of an opaque multiview framebuffer in some other way as well.
It could be complemented by a separate API for rendering to only one
of the views at a time, for example.


Oletus:
@RafaelCintron @shahmeeresmail @toji thoughts on this proposal? I've tried to incorporate all feedback I've gotten and requirements of WebVR. Sorry it's taken a while for me to get around to this again! I'll follow this up with a proposal to make the extension WebGL 1.0 compatible if there's agreement on this general direction, and go for another round of feedback on the public mailing list as well.

kenrussell:
Any comments? @toji and I discussed this general direction in person a couple of days ago and it sounds reasonable. If no feedback, let's merge this to make discussion easier.

RafaelCintron:
Thank you for putting this together, @Oletus 

The WebVR community group has been discussing approaches to doing layers.  The [current proposal](https://github.com/w3c/webvr/pull/218) is to continue to use side-by-side textures.  The WebVR API would provide the developer with a framebuffer for each layer.  The web developer would render to each layer using normal WebGL or a (new) side-by-side multiview extension.  Another difference between WebVR 2.0 and v1.1 is the potential of having N-views instead of just 2.

@shahmeersmail is no longer working at Intel, and therefore no longer working on multiview. However, he has made his [multiview prototype](https://github.com/shahmeeresmail/angle/tree/MultiviewPrototype) available on GitHub.  The prototype is flexible enough to do side-by-side or array textures.  It will be up to us to finish it out.  

I am fine with keeping the current, texture array based multiview extension for developers who wish to use texture arrays for other (non-WebVR) purposes. But it would be good to have an accompanying side-by-side version as well.

Oletus:
I'd rather avoid exposing side-by-side multiview rendering explicitly. With an opaque multiview framebuffer primitive, it can still be done behind the scenes if needed. WebVR could expose an opaque multiview framebuffer layer as an alternative to a side-by-side 2D framebuffer layer. Again, behind the scenes these could be the same thing on some platforms, but the opaque multiview framebuffer would make it possible to do optimizations on platforms that natively support some other form of multiview rendering, or need the framebuffer in a different format for submission to the HMD. I think that this is fairly important for forward compatibility.

N-views is not a problem, though some multiview optimizations may only be possible for up to a certain number of views.

RafaelCintron:
In WebVR 2.0, the framebuffers provided by the implementation behave like the default framebuffer in that we do not allow the developer to switch out the textures using any of the framebufferXXXX WebGL calls. We do not want the developer to "lose" the system provided texture and replace it with their own. The reason we provide the developer a fully constructed framebuffer instead of something more granular is because WebGL 1.0 does not allow developers to create and mix MSAA and non-MSAA textures with each other. 

The current multiview proposal does let you switch out textures from the framebuffer:  
> void framebufferTextureMultiviewWEBGL(GLenum target, GLenum attachment, GLuint texture, GLint level, GLint baseViewIndex, GLsizei numViews)

Later on, the example code specifically talks about texture arrays:
```JS
    var gl = document.createElement('canvas').getContext('webgl2');
    var ext = gl.getExtension('WEBGL_multiview');
    var tex = gl.createTexture();
    gl.bindTexture(gl.TEXTURE_2D_ARRAY, tex);
    gl.texStorage3D(gl.TEXTURE_2D_ARRAY, 1, gl.RGBA8, 512, 512, 2);
    ext.framebufferTextureMultiviewWEBGL(gl.TEXTURE_2D_ARRAY, gl.COLOR_ATTACHMENT0, tex, 0, 0, 2);
    gl.drawElements(...);  // draw will be broadcasted to the layers of the texture.
```
In order to create a opaque multiview extension, it seems like we'll need to adjust the multiview extension to deal with more generic "view" objects instead of textures or 2D array textures. Is this something you had planned in a future pull request? What am I missing?

Oletus:
@RafaelCintron the intent of the opaque multiview framebuffer is exactly what you're saying: the attachments are set up by the browser, and can't be changed by using the WebGL API. I'm not sure if you looked at the right diff. The spec in this pull request says:

"Opaque multiview framebuffers are <code>WebGLFramebuffer</code> objects that act as if they have multi-view attachments, but their attachments are not exposed as textures or renderbuffers and can not be changed."

"Calling <code>framebufferRenderbuffer</code>, <code>framebufferTexture2D</code>, <code>framebufferTextureLayer</code>, <code>framebufferTextureMultiviewWEBGL</code>, or any other call that could change framebuffer attachments with an opaque multiview framebuffer bound to <code>target</code> generates an <code>INVALID_OPERATION</code> error."

I think the spec text should be pretty clear like that, but I can add an example showing how you interact with opaque multiview framebuffers to illustrate it better.

Oletus:
I added a code example to the pull request now.

RafaelCintron:
Thank you for the clarification, @Oletus.

I agree we should not allow the developer to change the attachments.  However, I think we should also provide an API that allows the developer to filter rendering to only one of the views.  In this mode, we can allow the developer to use readPixels and copyTexImage2D to more easily do mirroring and/or debug their content. 

What do you think?

Oletus:
@RafaelCintron I agree that access to a single view, complete with read access, is a desirable feature. However, I think that it is cleanest if accessing single views of an opaque framebuffer is handled outside the multiview spec. For example an opaque non-multiview framebuffer concept could be introduced to the core WebGL spec. Then WEBGL_multiview would be just about multiview rendering. Do you think that this is reasonable? I'd be open to other alternatives for addressing this need as well.

RafaelCintron:
We could introduce the concept of an opaque switchable "multi-view" framebuffer outside of the multiview extension.  However, I think we would end up needing to solve the same problems that the multview extension already does for us: virtualizing viewports, scissor rects, getting gl_FragCoord to work correctly, etc.  

Do you see obstacles having the multiview extension support doing one view at a time?

Oletus:
I think that getting gl_FragCoord and viewports to work correctly will be the same issue in the implementation regardless of if the high-level spec exposes single views as a core spec object or as an object in the extension. It also seems that some kind of a core spec "opaque framebuffer" will probably be needed either way to expose the WebVR framebuffer in case the multiview extension is not supported. At least I've gotten the idea that multiview should be orthogonal from core WebVR support, neither depending on the other. This "opaque framebuffer" concept could be reused for accessing single views of a opaque multiview framebuffer.

But if the important thing for you would be getting the view-switching API for opaque multiview framebuffers inside the WEBGL_multiview extension rather than outside of it, maybe that would make sense. However, I think that's getting more complex with its own pros and cons and would probably be more productive to discuss in an entirely separate pull request once this one is merged.

kenrussell:
@RafaelCintron how does it sound to you to merge this pull request now and discuss extending it further in a separate pull request?

RafaelCintron:
For WebVR, the current plan is for the multiview extension to be optional.  If you decide not to use it, you will be given a framebuffer where the N-views are laid out side-by-side with each other.  You, as the developer, are responsible for managing this when you render. 

So as things stand now, multiview is either completely on or completely off for a given VRLayer.  The framebuffer is either "opaque" and forces you to draw to every view at once OR the views are side-by-side and you pay the performance penalty of not having multi-view. There's no middle ground.  I think giving developers the option of rendering to a single view of an opaque framebuffer would be a good thing to have.  

I agree we can discuss single view rendering as a separate full request.  If we don't get to it by the time the spec is finalized, developers can workaround the limitation by creating degenerate triangles based on the value of the gl_ViewID_OVR builtin.   

One additional limitation I think we should consider is that getFramebufferAttachmentParameter not work at all for opaque framebuffers.  If the opaque framebuffer is antialias, then retrieving any of the attachments and using it in a non-antialias framebuffer will not work properly, at least in WebGL 1.0.

Oletus:
The pull request already contains proposed restrictions for getFramebufferAttachmentParameter, like disallowing querying the attachment object name. I think it makes sense that you can still query the number of views, the pixel format related parameters or the framebuffer type (returning FRAMEBUFFER_DEFAULT) from an opaque multiview framebuffer using getFramebufferAttachmentParameter. Or do you think restrictions on some of these are necessary?

I sent a new version of the patch that clarified the getFramebufferAttachmentParameter spec a bit by including the restriction on FRAMEBUFFER_ATTACHMENT_OBJECT_NAME in the addendums in addition to the errors section.

RafaelCintron:
The FRAMEBUFFER_ATTACHMENT_OBJECT_NAME is the main one I think needs to be restricted.  Allowing the others is OK with me.

Issue #2391, opened 2017-05-05
No test case covering sampler operands


jchen10:
ESSL 1.00 section 4.1.7, and ESSL 3.00 section.
Except for parameters to texture lookup functions, array indexing, structure field selection, and parentheses, samplers are not allowed to be operands in expressions.

Better to have some test cases to cover the rule. An ANGLE bug(https://bugs.chromium.org/p/angleproject/issues/detail?id=2028)  can then be exposed.

PR #2392, opened 2017-05-05
Add sampler operands validation tests (#2391)

jchen10:
According to ESSL spec, sampler operands can only be used in array
indexing, structure field selection, and parentheses expresions.
This validates this rule.

### DIFF
--- a/sdk/tests/conformance/glsl/misc/sampler-operand.html
+++ b/sdk/tests/conformance/glsl/misc/sampler-operand.html
@@ -1,0 +1,72 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL GLSL Conformance Tests - sampler operands</title>
+<link rel="stylesheet" href="../../../resources/js-test-style.css"/>
+<link rel="stylesheet" href="../../../resources/glsl-feature-tests.css"/>
+<script src="../../../js/js-test-pre.js"></script>
+<script src="../../../js/webgl-test-utils.js"></script>
+<script src="../../../js/glsl-conformance-test.js"></script>
+</head>
+<body>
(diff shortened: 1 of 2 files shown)


jchen10:
@gyagp @Richard-Yunchao @qjia7 Please have a look!

zhenyao:
The test looks fine, but you will need to add it to the sdk/tests/conformance/glsl/misc/00_test_list.txt with --min-version 1.0.4 qualifier

jchen10:
@zhenyao Added to the list, thanks for the review!

qjia7:
LGTM. Thanks.

sdk/tests/conformance/glsl/misc/sampler-operand.html
```diff
@@ -0,0 +1,114 @@
...
+    { fShaderId: 'samplerInParentheses',
+      fShaderSuccess: true,
+      linkSuccess: true,
+      passMsg: 'Samplers in parentheses should succeed'
+    },
+
+    // ANGLE bug.
+    { fShaderId: 'samplerAdding',
+      fShaderSuccess: false,
+      linkSuccess: false,
```
Richard-Yunchao: You check the linkStatus here. But I suppose that the compile status should fail (before linkage), according to the ANGLE bug at https://bugs.chromium.org/p/angleproject/issues/detail?id=2028.

Richard-Yunchao: My fault, the fShaderSuccess is the compile status.
LGTM then.

jchen10:
@Oletus @zhenyao Please have a look, Thanks!

sdk/tests/conformance/glsl/misc/sampler-operand.html
```diff
@@ -0,0 +1,114 @@
...
+<div id="description"></div>
+<div id="console"></div>
+
+<script>
+// ESSL(1.00, 3.00, 3.10) section 4.1.7
+// Except for array indexing, structure field selection, and
+// parentheses, samplers are not allowed to be operands in expressions.
+</script>
+
+<script id="samplerArrayIndexing" type="x-shader/x-fragment">
```
Oletus: This is already covered by other tests like sdk/tests/conformance/glsl/bugs/sampler-array-using-loop-index.html

Should leave this out to avoid adding duplicate coverage.

sdk/tests/conformance/glsl/misc/sampler-operand.html
```diff
@@ -0,0 +1,114 @@
...
+uniform sampler2D s2D[2];
+void main() {
+    gl_FragColor = texture2D(s2D[1], vec2(0.5, 0.5));
+}
+</script>
+
+<script id="samplerMemberSelection" type="x-shader/x-fragment">
```
Oletus: Also already covered by sdk/tests/conformance/glsl/bugs/sampler-struct-function-arg.html

sdk/tests/conformance/glsl/misc/sampler-operand.html
```diff
@@ -0,0 +1,114 @@
...
+uniform struct S {
+    sampler2D a;
+} s;
+void main() {
+    gl_FragColor = texture2D(s.a, vec2(0.5, 0.5));
+}
+</script>
+
+<script id="samplerInParentheses" type="x-shader/x-fragment">
```
Oletus: Not sure how necessary this test is either. I don't think that parentheses are systematically covered in general, and it seems odd to cover it especially for samplers here. I'd maybe leave this out too, though don't feel too strongly about it.

zhenyao:
@jchen10 Can you address @Oletus comments and remove the duplicated test cases? Then I am happy to merge. Thank  you for your contribution.

jchen10:
@zhenyao @Oletus The duplicated cases removed. Thanks for the review.

sdk/tests/conformance/glsl/misc/sampler-operand.html
```diff
@@ -0,0 +1,72 @@
...
+// parentheses, samplers are not allowed to be operands in expressions.
+</script>
+
+<script id="samplerAdding" type="x-shader/x-fragment">
+// This covers an ANGLE bug.
+// See https://bugs.chromium.org/p/angleproject/issues/detail?id=2028.
+uniform sampler2D s1;
+uniform sampler2D s2;
+void main() {
+    gl_FragColor = texture2D(s1 + s2, vec2(0.5, 0.5));
```
Oletus: I don't think that the texture2D call should be here. To keep the test minimal, the statement should be just s1 + s2. An implementation might generate an error on using the result of the addition even if it doesn't generate an error on the addition itself. This could happen especially if the implementation prunes statements that have no effect.

jchen10:
@Oletus Make sense to keep test minimal. Fixed, thanks!

Oletus:
Great, looks good!

zhenyao:
LGTM
Thanks @Oletus for your careful review

PR #2394, opened 2017-05-05
Fix a test coverage issue in webgl-depth-texture.html

zhenyao:
According to the extension spec, INVALID_OPERATION is generated if
1) level > 0 or
2) target != TEXTURE_2D

When we test with CUBEMAP targets, we used level == 1, so target != TEXTURE_2D
isn't really tested (and therefore, Chrome behaved incorrectly).

ALso, we do test level > 0 with target == TEXTURE_2D and make sure
INVALID_OPERATION is generated in this case.  So it's clear to me when we test
with CUBEMAP targets with level == 1, the intention is to test with level == 0.


zhenyao:
@greggman That's why Chrome has a bug in it.
@kenrussell @kainino0x FYI. It's simple enough, so I merge myself to move forward.

kenrussell:
LGTM after the fact. Thanks @zhenyao for catching and fixing this.

PR #2396, opened 2017-05-08
Replace get.webgl.org Safari troubleshooting URL to fix broken link

reidab:
The page I chose as a replacement isn't perfect, but it's the best I could find on either apple.com or webkit.org that gets the message across that Safari hides its WebGL settings under "Security".

The previous URL (http://www.webkit.org/blog/919/webgl-draft-specification-now-available/) now leads to a 404.


kainino0x:
LGTM, but I would of course like to have a more helpful troubleshooting page. @grorg do you know of one?

kenrussell:
Since @grorg hasn't had a chance to reply, merging this now.

PR #2398, opened 2017-05-11
Expand testing uninitialized variables

Oletus:
Added tests with two declarators in the same declaration. These cases
need to be taken into account when transforming the shader source to
initialize unininitialized variables.

Also change shader precision to mediump. Supporting highp float in
fragment shaders is not mandatory in WebGL 1.0.


zhenyao:
LGTM
Thanks @Oletus for picking up this.

PR #2399, opened 2017-05-13
Backport null-vbo vertexAttribPointer(offset!=0) test, but not offset…

jdashg:
…==0 test.

Backporting the offset==0 test would cause previously conformant implementations to fail.


kainino0x:
LGTM. Could you clarify in the commit message (and PR description) that the test case you're editing was wrong and was fixed in #2228? (unless I'm reading the history wrong)

kenrussell:
We can adjust the commit message during "squash and merge" -- let me do that to save time.

PR #2400, opened 2017-05-14
Fix #2274: Add GL_RGB10_A2 to texImageSource variants of tex(Sub)Image*

### LINKED ISSUE #2274: Spec: GL_RGB10_A2 format missing  in texImageSource variants of tex(Sub)Image2D

astojilj:
Chromium implementation landed: https://crrev.com/2821363003/.
Verified by running tests locally on OSX.

@kenrussell , @zhenyao, PTAL.
FYI @Richard-Yunchao.

Sorry for the delay on this. Thanks.

### DIFF
--- a/sdk/tests/conformance2/textures/canvas/00_test_list.txt
+++ b/sdk/tests/conformance2/textures/canvas/00_test_list.txt
@@ -25,6 +25,7 @@ tex-2d-rgba8-rgba-unsigned_byte.html
 tex-2d-srgb8_alpha8-rgba-unsigned_byte.html
 tex-2d-rgb5_a1-rgba-unsigned_byte.html
 tex-2d-rgb5_a1-rgba-unsigned_short_5_5_5_1.html
+tex-2d-rgb10_a2-rgba-unsigned_int_2_10_10_10_rev.html
 tex-2d-rgba4-rgba-unsigned_byte.html
 tex-2d-rgba4-rgba-unsigned_short_4_4_4_4.html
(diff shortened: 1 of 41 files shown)


zhenyao:
LGTM

kenrussell:
Great! Thanks @astojilj for pushing this through and @zhenyao for reviewing.

PR #2401, opened 2017-05-16
Test fbo attached to a non-level-0 texture image completeness behavior.

zhenyao:
The issue is raised in crbug.com/722684

### DIFF
--- a/sdk/tests/conformance2/rendering/framebuffer-texture-level1.html
+++ b/sdk/tests/conformance2/rendering/framebuffer-texture-level1.html
@@ -44,6 +44,9 @@ var wtu = WebGLTestUtils;
 var gl;
 
 function testFramebufferTextureWithNonZeroBaseLevel(level) {
+  if (level < 1) {
+    throw "This test is incorrect if level < 1";
+  }
   var width = 32;
   var height = 16;
   var texture = gl.createTexture();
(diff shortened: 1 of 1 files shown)


zhenyao:
@kenrussell @kainino0x Please review (we might already have been testing this since Chrome is passing, but this is simple change, so let's just add it.)

sdk/tests/conformance2/rendering/framebuffer-texture-level1.html
```diff
@@ -43,13 +43,11 @@
 var wtu = WebGLTestUtils;
 var gl;
 
-function testFramebufferTextureWithNonZeroBaseLevel() {
+function testFramebufferTextureWithNonZeroBaseLevel(level) {
```
kenrussell: Do you want to add an assertion that level > 1? Because otherwise someone else may come in here and find that the test fails because the texture is actually complete if called with level=0.

zhenyao: Done.  Please take another look @kenrussell

kenrussell:
LGTM

grovesNL:
Thanks! FYI I can confirm https://www.khronos.org/registry/webgl/sdk/tests/conformance2/rendering/framebuffer-texture-level1.html?webglVersion=2 incorrectly reports `36053` on Windows/ANGLE with Chromium 60.0.3103, and crashes Firefox.

zhenyao:
Thanks for reporting the issues. That's how we can increase the test coverage and improve implementation qualities. Really appreciate it.

PR #2403, opened 2017-05-17
Allow all isEnabled queries in getParameter

### LINKED ISSUE #2384: Clarify intended behavior for `isEnabled` and `getParameter` for certain capabilities
As I wrote [on StackOverflow](http://stackoverflow.com/questions/43613793/difference-between-webglrenderingcontext-isenabledgetparameter-for-capabili/43622669) (some of the question snipped for brevity), the code below shows an inconsistency between browsers responding to `isEnabled` and `getParameter` for `SAMPLE_ALPHA_TO_COVERAGE`, `SAMPLE_COVERAGE`, and `STENCIL_TEST`.

> 
> 
> 
>     var canvas = document.createElement('canvas')
>     ,   gl = canvas.getContext('webgl') || canvas.getContext('experimental-webgl')
>     ,   capabilities = [ 'SAMPLE_ALPHA_TO_COVERAGE', 'SAMPLE_COVERAGE', 'STENCIL_TEST' ];
>     for (var i = 0; i < capabilities.length; i++) {
>         var capability = gl[capabilities[i]];
>         gl.enable(capability);
>         var isEnabledResult = gl.isEnabled(capability)
>         ,   getParameterResult = gl.getParameter(capability);
>         if (isEnabledResult !== getParameterResult) {
>             console.log(capabilities[i], isEnabledResult, getParameterResult);
>         }
>     }
> 
> 
> - With WebGL1 on Chrome/Canary I receive the following (`STENCIL_TEST` will be removed here if the stencil buffer is enabled while creating the context):
> 
>         SAMPLE_ALPHA_TO_COVERAGE true null
>         SAMPLE_COVERAGE true null
>         STENCIL_TEST true false
> 
> - With WebGL1 on Firefox/Nightly, IE11, and Edge (regardless of stencil buffer being enabled):
> 
>         SAMPLE_ALPHA_TO_COVERAGE true null 
>         SAMPLE_COVERAGE true null
> 
> - With WebGL2 on Chrome/Canary (`STENCIL_TEST` will be removed here if stencil buffer is enabled):
> 
>         STENCIL_TEST true false
> 
> - With WebGL2 on Firefox/Nightly, I receive nothing.

@jdashg's [answer to my question](http://stackoverflow.com/a/43622669/3190758) suggested that there may be some issues with the behavior of `getParameter` for the above cases and that `isEnabled` and `getParameter` should behave the same in these cases.

There are a couple issues that seem to require clarification in the specification:

- Should both `gl.isEnabled(gl.STENCIL_TEST)` and `gl.getParameter(gl.STENCIL_TEST)` return `true` or `false` when the context is created without the stencil buffer (i.e. without `{stencil: true}`)?
- Should `gl.getParameter(gl.SAMPLE_ALPHA_TO_COVERAGE)` and `gl.getParameter(gl.SAMPLE_COVERAGE)` match behavior between WebGL1 and WebGL2 contexts (i.e. backport WebGL2 `getParameter` behavior for these constants)?

kainino0x:
* Spec update
* Add tests verifying getParameter works like isEnabled
* Test enable queries (via isEnabled and getParameter) for multiple sets of context creation attributes

Fixes #2384

### DIFF
--- a/sdk/tests/conformance/state/gl-enable-enum-test.html
+++ b/sdk/tests/conformance/state/gl-enable-enum-test.html
@@ -1,7 +1,7 @@
 <!--
 
 /*
-** Copyright (c) 2012 The Khronos Group Inc.
+** Copyright (c) 2017 The Khronos Group Inc.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
 ** copy of this software and/or associated documentation files (the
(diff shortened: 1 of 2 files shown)


kainino0x:
SAMPLE_ALPHA_TO_COVERAGE and SAMPLE_COVERAGE fail in both Chrome and Firefox.

STENCIL_TEST also fails in Chrome.

Note the alternate context creation attributes do not trigger any different failures on either Chrome or Firefox, but the coverage is good to have anyway.

sdk/tests/conformance/state/gl-enable-enum-test.html
```diff
@@ -47,110 +47,137 @@
...
-    'SAMPLE_ALPHA_TO_COVERAGE',
-    'SAMPLE_COVERAGE',
-    'SCISSOR_TEST',
-    'STENCIL_TEST'
-  ];
+function runPositiveTestsWithParameters(params) {
+  debug("");
+  debug("Running positive tests with parameters: " + JSON.stringify(params));
+
+  gl = wtu.create3DContext("canvas", params);
```
zhenyao: Are we able to create multiple WebGL rendering contexts with different params from the same canvas element? If not, you should pass in null instead of "canvas" here to create a new canvas element each time.

kainino0x: Excellent point, I forgot about that. I will check and fix if necessary.

kainino0x: Fixed. Now the STENCIL_TEST test only fails on some context types.

zhenyao:
LGTM

PR #2406, opened 2017-05-26
Expose Intel 6000/6100 clear to 0/1 bug

qjia7:
This test is ported from crbug.com/710443's sample.mm provided
    by bsalomon.
    
    If we bind a texture to a frame buffer's color attachment, clear
    it using the composition of 0 and 1. The render color is not
    correct when sampling that texture to another frame buffer.
    It seems that the clear command didn't work in some situations
    and the result is random.
    
    This bug can be reproduced in below situations:
    case1 : Set TexParameteriv(GL_TEXTURE_2D, GL_TEXTURE_SWIZZLE_RGBA*)
    when bind sampler.
    case2 : Set TexParameteriv(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER*)
    when bind sampler and don't use it before bind sampler.
    
    This patch will use case2 to reproduce Mac bug since WebGL don't
    support swizzle.
    
    Change the alpha component of clear color can also workaround this bug.

### DIFF
--- a/sdk/tests/conformance/rendering/00_test_list.txt
+++ b/sdk/tests/conformance/rendering/00_test_list.txt
@@ -26,3 +26,4 @@ point-size.html
 --min-version 1.0.2 simple.html
 triangle.html
 line-loop-tri-fan.html
+--min-version 1.0.4 framebuffer-texture-clear.html
--- a/sdk/tests/conformance/rendering/framebuffer-texture-clear.html
+++ b/sdk/tests/conformance/rendering/framebuffer-texture-clear.html
@@ -1,0 +1,118 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL framebuffer clearColor with pure 0/1</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<canvas id="canvas" width="2" height="2"> </canvas>
+<script>
+"use strict";
+// This test verifies a Mac Intel HD 6000/6100 driver bug. See crbug.com/710443.
+var wtu = WebGLTestUtils;
+var gl = wtu.create3DContext("canvas");
+
+function InitializeRGBAData(width, height)
+{
+  var size = 4 * width * height;
+  var data = new Uint8Array(size);
+  for (var i = 0; i < size; i++) {
+    data[i] = 128;
+  }
+  return data;
+}
+
+function testFramebufferTextureClearWithPureZeroOrOne(clearColor, expectedColor) {
+  var width = 32;
+  var height = 32;
+  var texture0 = gl.createTexture();
+  gl.bindTexture(gl.TEXTURE_2D, texture0);
+  // It seems that if we add texParameteri here, all cases can pass no matter
+  // you use texParameteri or not in Line 85.
+  // gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
+  var texData = InitializeRGBAData(width, height);
+  gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, width, height, 0, gl.RGBA, gl.UNSIGNED_BYTE, texData);
+
+  var fbo0 = gl.createFramebuffer();
+  gl.bindFramebuffer(gl.FRAMEBUFFER, fbo0);
+  gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture0, 0);
+  gl.viewport(0, 0, width, height);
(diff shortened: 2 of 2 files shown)


sdk/tests/conformance2/rendering/framebuffer-texture-clear.html
```diff
@@ -0,0 +1,190 @@
...
+  gl.bindBuffer(gl.ARRAY_BUFFER, vertexObject);
+  gl.bufferData(gl.ARRAY_BUFFER, new Float32Array([
+      1, 1,
+      0, 1,
+      0, 0,
+      1, 1,
+      0, 0,
+      1, 0]), gl.STATIC_DRAW);
+  gl.enableVertexAttribArray(1);
+  gl.vertexAttribPointer(1, 2, gl.FLOAT, false, 0, 0);
```
qjia7: From line 81 to line 138, if I use 'tiu.setupTexturedQuad(gl, "RGBA8")' to realize it, I can't reproduce the error on Intel 6000 Mac. But using above code, the error can be seen.

qjia7:
@zhenyao @kenrussell Please take a look.

sdk/tests/conformance2/rendering/framebuffer-texture-clear.html
```diff
@@ -0,0 +1,190 @@
...
+      + "}\n";
+
+  var program = wtu.loadProgram(gl, vSrc, fSrc);
+  if (!program) {
+    testFailed("Failed to create the program");
+    return;
+  }
+
+  gl.bindAttribLocation(program, 0, "inPosition");
+  gl.bindAttribLocation(program, 1, "inLocalCoord");
```
zhenyao: These calls have no effects until you relink the program.  So it only works because of pure luck that the binding auto-generated is the same as you wanted. Instead, you should do getAttribLocation() here.

sdk/tests/conformance2/rendering/framebuffer-texture-clear.html
zhenyao: I think this test should really be a WebGL 1 test if you just switch internal format from RGBA8 to RGBA, and then you can use wtu.setupTexturedQuad instead of tiu.setupTexturedQuad.  This will simplify the test code a lot (and remember to move the test from conformance2/ to conformance/)

sdk/tests/conformance2/rendering/framebuffer-texture-clear.html
```diff
@@ -0,0 +1,190 @@
...
+  var buffer = new Uint8Array(4);
+  gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, buffer);
+  if (buffer[0] == expectedColor[0] &&
+      buffer[1] == expectedColor[1] &&
+      buffer[2] == expectedColor[2] &&
+      buffer[3] == expectedColor[3]) {
+    testPassed("clearColor sets the render texture with the correct data");
+  } else {
+    testFailed("clearColor fails to work. Expected: " + expectedColor + ", got: " + buffer);
+  }
```
zhenyao: use wtu.checkCanvasRect instead.

sdk/tests/conformance2/rendering/framebuffer-texture-clear.html
qjia7: You are right. That's the problem why I can't get consistent result between current code with directly using setupTexturedQuad. But a new problem blocks me now that I can't reproduce crbug.com/710443's issue using current code.

In fact, I can reproduce that bug by writing ANGLE's end to end test. See https://chromium-review.googlesource.com/c/513435/. I just ported it to a WebGL conformance test. The code can be one-one mapping. However, I can't reproduce this bug using current webgl test. This confuses me. @zhenyao  can you give me some ideas why the result is different even though the current code looks like same with the angle native test?

zhenyao:
I don't know on top of my head, but Chrome's command buffer adds a lot of code (driver bug workarounds, etc).  You can try with --disable-gpu-driver-bug-workarounds and see if it helps to reproduce the bug.

sdk/tests/conformance2/rendering/framebuffer-texture-clear.html
```diff
@@ -0,0 +1,190 @@
...
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+<script src="../../js/tests/tex-image-and-sub-image-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<canvas id="canvas" width="2" height="2"> </canvas>
+<script>
+"use strict";
```
qjia7: Done

sdk/tests/conformance2/rendering/framebuffer-texture-clear.html
qjia7: Done

sdk/tests/conformance2/rendering/framebuffer-texture-clear.html
```diff
@@ -0,0 +1,190 @@
...
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
+  gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA8, width, height, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+  gl.bindTexture(gl.TEXTURE_2D, null);
+  var fbo1 = gl.createFramebuffer();
+  gl.bindFramebuffer(gl.FRAMEBUFFER, fbo1);
+  gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture1, 0);
+
+  var vSrc =
```
qjia7: Use wtu.setupTexturedQuad

sdk/tests/conformance2/rendering/framebuffer-texture-clear.html
qjia7: @zhenyao I have updated this patch. Now the bug can be reproduced on Mac. I traced the api call stack in ANGLE and found that the 'TexParameter' APIs are cached and only be really executed when apply samplers before drawing. So I changed the api sequence to keep same with angle. Now the bug can be reproduced. It seems that the time of doing TexParameter will affect this bug. 

sample.mm can reproduce this bug even don't follow angle's logic. That's because it will call glTexParameteriv(GL_TEXTURE_2D, GL_TEXTURE_SWIZZLE_RGBA, swiz). And this can also affect that bug. However, webgl don't support swizzle. So we couldn't reproduce that bug in old logic.

qjia7:
@zhenyao Please take another look.

Paste the reason again in case someone wants to know:
I traced the api call stack in ANGLE and found that the 'TexParameter' APIs are cached and only be really executed when apply samplers before drawing. So I changed the api sequence to keep same with angle. Now the bug can be reproduced. It seems that the time of doing TexParameter will affect this bug.

sample.mm can reproduce this bug even don't follow angle's logic. That's because it will call glTexParameteriv(GL_TEXTURE_2D, GL_TEXTURE_SWIZZLE_RGBA, swiz). And this can also affect that bug. However, webgl don't support swizzle. So we couldn't reproduce that bug in old logic.

zhenyao:
Looks good to me.  Due to your analysis, by calling glTexParamateri() early will fix this bug.  Can we come up with a driver bug workaround based on this?

qjia7:
>Due to your analysis, by calling glTexParamateri() early will fix this bug. Can we come up with a driver bug workaround based on this?

calling glTexParamateri() early is only a fix for this WebGL test (see my commit description). But if some third party can call TexParameteriv(GL_TEXTURE_2D, GL_TEXTURE_SWIZZLE_RGBA*)
when bind sampler. This bug can still be exposed even we call TexParameteriv(GL_TEXTURE_2D, GL_TEXTURE_SWIZZLE_RGBA*) early. Such as the skia meets this bug in crbug.com/710443. So far, the best workaround seems still to change the alpha value like the skia did(https://skia-review.googlesource.com/c/17327/) and port that fixing to command buffer.

PR #2407, opened 2017-05-26
Add test for resizing multisampled renderbuffers

kainino0x:
This is a regression test for Chrome on Adreno 4xx/5xx:
http://crbug.com/696126

### DIFF
--- a/sdk/tests/conformance2/renderbuffers/multisampled-renderbuffer-initialization.html
+++ b/sdk/tests/conformance2/renderbuffers/multisampled-renderbuffer-initialization.html
@@ -42,6 +42,10 @@
 var wtu = WebGLTestUtils;
 description('Verify multisampled renderbuffers are initialized to 0 before being read in WebGL');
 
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/renderbuffers/multisampled-renderbuffer-initialization.html
```diff
@@ -52,8 +52,10 @@
 
     runTest(gl, gl.canvas.width, gl.canvas.height, 0);
```
zhenyao: Can you define some const variable (or whatever in js) to be 0, 1, 2, like ALLOCATE_ATTACH = 0, ATTACH_ALLOCATE = 1, and ALLOCATE_ATTACH_ALLOCATE_ANOTHER_SIZE = 2? This will make code much easier to read.

kainino0x: Done

zhenyao:
Thanks. LGTM.

PR #2410, opened 2017-05-30
Add `clear` call to test for resizing multisampled renderbuffers

kainino0x:
Followup to #2407.

This clear is needed to expose this Qualcomm bug via Firefox as well.


zhenyao:
LGTM

Thanks for getting to the bottom of this and making sure this test demonstrate the driver bug in both Chrome and Firefox

kenrussell:
Excellent work @kainino0x demonstrating this longstanding driver bug in all WebGL implementations on Android. LGTM after the fact.

PR #2412, opened 2017-06-01
Fix dEQP multisample test tolerance for sample counts >= 64

Oletus:
Previously the test wasn't able to handle sample counts >= 64. This
was because it's counting unique colors, and it was detecting colors
with <= 3 difference in one channel as the same. But if you have 65
different gray levels between 0 and 255, at least two of them will be
set apart by only 3.

This made the test fail on Quadro drivers that supported 64 samples.

Now the test automatically lowers the tolerance if it is required by
having a high sample count.

After this fix the test at least has a chance of handling high sample
counts correctly, though it still needs to be verified if this fix is
sufficient on its own.

### DIFF
--- a/sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
+++ b/sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
@@ -617,13 +617,17 @@ goog.scope(function() {
 
         /** @type {number} */ var requiredNumDistinctColors = this.m_numSamples + 1;
 
+        // If the number of samples is high (64 or more), we need to lower the threshold for detecting unique colors, otherwise two expected unique colors would be treated as the same color.
+        var threshold = Math.min(3, Math.floor(255 / this.m_numSamples) - 1);
+        var thresholdRGBA = tcuRGBA.newRGBAComponents(threshold, threshold, threshold, threshold);
+
         for (var y = 0; y < renderedImg.getHeight() && this.m_detectedColors.length < requiredNumDistinctColors; y++)
         for (var x = 0; x < renderedImg.getWidth() && this.m_detectedColors.length < requiredNumDistinctColors; x++) {
             /** @type {tcuRGBA.RGBA} */ var color = new tcuRGBA.RGBA(renderedImg.getPixel(x, y));
 
             /** @type {number} */ var i;
             for (i = 0; i < this.m_detectedColors.length; i++) {
-                if (tcuRGBA.compareThreshold(color, this.m_detectedColors[i], tcuRGBA.newRGBAComponents(3, 3, 3, 3)))
+                if (tcuRGBA.compareThreshold(color, this.m_detectedColors[i], thresholdRGBA))
                     break;
             }
 


zhenyao:
Olli, thanks for the fix. I am away from my Linux today, so I'll verify if it fixes the failures on my Linux tomorrow.

sdk/tests/deqp/functional/gles3/es3fMultisampleTests.js
```diff
@@ -617,13 +617,17 @@ goog.scope(function() {
 
         /** @type {number} */ var requiredNumDistinctColors = this.m_numSamples + 1;
 
+        // If the number of samples is high (64 or more), we need to lower the threshold for detecting unique colors, otherwise two expected unique colors would be treated as the same color.
+        var threshold = Math.min(3, Math.floor(255 / this.m_numSamples) - 1);
```
zhenyao: Sorry, after merge, just realized this can end up negative.  @Oletus Can you address it in a follow up CL?

kenrussell:
Thanks @Oletus for this fix and @zhenyao for reviewing.

Is this a change that should be merged into the upstream dEQP tests?

kenrussell:
Note: filed internal bug 62259905 against the dEQP regarding considering upstreaming this fix.

Oletus:
Yep, I think this should be in upstream dEQP as well. Mo is also right that this fails for sample counts > 255. I don't know of any existing hardware supporting such sample counts, but I guess that could be addressed. The best thing to do to address that would be to limit the testing to sample count 128 I think.

kenrussell:
Note that @phaulos provided some feedback on internal bug 62259905. There he pointed out that these internal formats aren't actually conformant. The hardware in question supports 16x MSAA, and 2x supersampling in the X and Y dimensions is used to simulate 64x MSAA.  https://www.khronos.org/registry/OpenGL/extensions/NV/NV_internalformat_sample_query.txt was intended to allow this divergence from the spec. There was disagreement about whether it should be allowed to affect the behavior of multisampled renderbuffers' formats in this way, and the OpenGL ES working group decided not to allow this, while the OpenGL working group did.

Not sure whether this patch is likely to be accepted upstream.

WebGL 2.0 implementations should filter out non-conformant renderbuffer sample count values. This is what NVIDIA's OpenGL ES driver for desktops does.

zhenyao:
Interesting. How can we tell coding wise if a #sample is conformant or not?

kenrussell:
If the https://www.khronos.org/registry/OpenGL/extensions/NV/NV_internalformat_sample_query.txt extension is present, then we should use it to query all of the numbers of supported samples the driver returns, and filter them so we only return the conformant numbers.

This isn't that important for ES 3.0, but @phaulos pointed out that it would matter for ES 3.1, where the samples are exposed directly to the shader, and the supersampling approach would break down.

It does seem that Olli's patch here is correct though (though could use the robustness improvements you pointed out, Mo). It appears there are some ES 3.2 conformance tests which break when non-conformant multisampling is used, so that's detected there.

Oletus:
Thanks for the additional analysis, I wasn't aware of the supersampling path used on Quadro. It might be possible to write a test to detect non-conformant multisampling on GLES 3.0 / WebGL 2.0 as well, if that's desired. Are you going to work on using NV_internalformat_sample_query in Chromium / ANGLE or should I?

kenrussell:
Could you? I'd really appreciate it. It's not on my radar right now. It's not the highest priority thing, but at least filing an ANGLE bug regarding using this extension to filter out non-conformant sample counts would be helpful. Thanks.

Oletus:
I'm writing a patch for ANGLE to filter out non-conformant formats. I'll use the Chromium bug to track it, should be done with it pretty soon.

PR #2413, opened 2017-06-01
Enable reading from 1-view opaque multiview framebuffers

Oletus:
Opaque multiview framebuffers can now be attached to READ_FRAMEBUFFER
so that reading from opaque multiview framebuffers that have just a
single view will work.


kenrussell:
Sounds good. I didn't go through the spec thoroughly to see whether this generalizes only the 1-view case, but let's merge this for easier broader review.

PR #2415, opened 2017-06-02
Add more cases to multisampled-renderbuffer-initialization

kainino0x:
These test changing renderbuffers from non-multisampled to multisampled
and vice versa. This caught an extra bug in my workaround in Chrome.
The test function is refactored a little bit to be more flexible.

I spent quite a bit of time trying to add a test for depth renderbuffers, but I wasn't able to get it working. blitFramebuffer on the depth bit from a depth-only FBO to a depth+color FBO doesn't seem to do anything, though I don't know why.

### DIFF
--- a/sdk/tests/conformance2/renderbuffers/multisampled-renderbuffer-initialization.html
+++ b/sdk/tests/conformance2/renderbuffers/multisampled-renderbuffer-initialization.html
@@ -42,10 +42,6 @@
 var wtu = WebGLTestUtils;
 description('Verify multisampled renderbuffers are initialized to 0 before being read in WebGL');
 
-const ALLOCATE_ATTACH = 0;
-const ATTACH_ALLOCATE = 1;
-const ALLOCATE_ATTACH_CLEAR_REALLOCATE = 2;
-
 var gl = wtu.create3DContext("testbed", null, 2);
 
 if (!gl) {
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/renderbuffers/multisampled-renderbuffer-initialization.html
```diff
@@ -54,22 +50,32 @@
...
-    runTest(gl, gl.canvas.width, gl.canvas.height, ALLOCATE_ATTACH);
-    runTest(gl, gl.canvas.width, gl.canvas.height, ATTACH_ALLOCATE);
-    runTest(gl, gl.canvas.width, gl.canvas.height, ALLOCATE_ATTACH_CLEAR_REALLOCATE);
-    runTest(gl, gl.canvas.width, gl.canvas.height, ALLOCATE_ATTACH);
-    runTest(gl, gl.canvas.width, gl.canvas.height, ATTACH_ALLOCATE);
-    runTest(gl, gl.canvas.width, gl.canvas.height, ALLOCATE_ATTACH_CLEAR_REALLOCATE);
+    let c = gl.canvas;
+    var maxSamples = gl.getInternalformatParameter(
+        gl.RENDERBUFFER, gl.RGBA8, gl.SAMPLES)[0];
+    for (let i = 0; i < 2; ++i) {
```
kenrussell: Is it intentional that this loops twice? Or was that done for debugging earlier? If intentional, please add a comment why.

kainino0x: Originally it did 2 tests, then it did the same 2 tests again. (It resizes the canvas between each test.) I don't know why, so I just put a loop around the whole thing.

kenrussell: Ah, I see. OK.

PR #2416, opened 2017-06-05
remove moz-webgl and webkit-3d

greggman:
these never existed in shipping webgl AFAIK and even if they
once did that was 6-7 years ago


zhenyao:
Looks fine. Thanks for cleaning this up.

PR #2417, opened 2017-06-08
Expose a clear to non 0/1 bug on Intel D3D drivers

Jiawei-Shao:
This patch exposes an Intel D3D driver bug on clear.

The bug can be reproduced by following steps:
1. Create a WebGL context with antialias
2. Create a texture and call copyTexImage2D on it
3. Clear the default framebuffer by a color that isn't pure 0/1
The final render color isn't consistent with the clear color on
Intel D3D drivers.

### DIFF
--- a/sdk/tests/conformance/rendering/clear-after-copyTexImage2D.html
+++ b/sdk/tests/conformance/rendering/clear-after-copyTexImage2D.html
@@ -1,0 +1,85 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/rendering/clear-after-copyTexImage2D.html
```diff
@@ -0,0 +1,85 @@
...
+  gl.clearColor(clearColor[0], clearColor[1], clearColor[2], clearColor[3]);
+  gl.clear(gl.COLOR_BUFFER_BIT);
+
+  wtu.checkCanvasRect(gl, 0, 0, gl.drawingBufferWidth, gl.drawingBufferHeight, expectedColor);
+
+  gl.bindTexture(gl.TEXTURE_2D, null);
+  gl.deleteTexture(texture);
+}
+
+description("Test that if clear the default frambuffer without pure 0/1 after copyTexImage2D, the final render color should be consistent with the clear color.");
```
zhenyao: Why this only happens on default framebuffer?  In Chrome, the default framebuffer is not really the OpenGL's default framebuffer, but a regular framebuffer created in the command buffer, no different from other user created framebuffers.  Emphasizing default framebuffer is kinda misleading.

zhenyao: without pure 0/1 -> with a color other than [0,0,0,1]

Jiawei-Shao: Thanks to point out this. I have corrected this misleading description in the new patch.

The bug is observed with colors that has non 0/1 RGBA components. For example, the clear is correct not only with [0,0,0,1] but also with the color [1, 0, 0, 1], [0, 1, 0, 1] and [0, 0, 1, 1].

Jiawei-Shao: This bug is only observed when we create a WebGL context with "antialias" == "true". 

In OpenGL ES 3.0 spec it is not allowed to call copyTexImage2D when READ_FRAMEBUFFER_BINDING is non zero and the value of SAMPLE_BUFFERS for the read framebuffer is one. 
(OpenGL ES 3.0.4 SPEC chapter 3.8.5  Page. 145)

So I think this bug cannot be observed on non-zero framebuffers since it isn't allowed to call copyTexImage2D when current read framebuffer has multisampled attachments.

Jiawei-Shao: The bug is observed with colors that has non 0/1 RGB components. For example, the clear is correct not only with [0,0,0,1] but also with the color [1, 0, 0, 1], [0, 1, 0, 1] and [0, 0, 1, 1].

zhenyao: OK for the 0/1 issue. I misunderstood.

As for the default framebuffer, in WebGL default framebuffer, we actually blit from a multisampled DRAW_FRAMEBUFFER to a non-multisampled READ_FRAMEBUFFER internally, which can be implemented in WebGL2, but not in WebGL1.

Jiawei-Shao: Got it. Thanks for your comments!

PR #2418, opened 2017-06-08
Add tests for uninitialized multisampled depth renderbuffers

kainino0x:
Tests for the Qualcomm driver bug in
http://crbug.com/696126
as well as a Mac/Intel bug that I haven't reported yet.

### DIFF
--- a/sdk/tests/conformance2/renderbuffers/multisampled-depth-renderbuffer-initialization.html
+++ b/sdk/tests/conformance2/renderbuffers/multisampled-depth-renderbuffer-initialization.html
@@ -1,7 +1,7 @@
 <!--
 
 /*
-** Copyright (c) 2015 The Khronos Group Inc.
+** Copyright (c) 2017 The Khronos Group Inc.
 **
 ** Permission is hereby granted, free of charge, to any person obtaining a
 ** copy of this software and/or associated documentation files (the
@@ -40,7 +40,7 @@
 <script>
 "use strict";
 var wtu = WebGLTestUtils;
-description('Verify multisampled renderbuffers are initialized to 0 before being read in WebGL');
+description('Verify multisampled depth renderbuffers are initialized to 0 before being read in WebGL');
 
 var gl = wtu.create3DContext("testbed", null, 2);
 
@@ -49,6 +49,7 @@ if (!gl) {
 } else {
     // Set the clear color to green. It should never show up.
     gl.clearColor(0, 1, 0, 1);
+    gl.clearDepth(0);
 
     let c = gl.canvas;
     var maxSamples = gl.getInternalformatParameter(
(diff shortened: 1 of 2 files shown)


kainino0x:
@zhenyao PTAL. This hits the bug I mentioned on macOS that I haven't reported yet.

kainino0x:
P.S. I would like to add stencil tests here as well, but I haven't written any yet.

sdk/tests/conformance2/renderbuffers/multisampled-depth-renderbuffer-initialization.html
```diff
@@ -0,0 +1,172 @@
...
+
+    function allocStorage(width, height, samples) {
+        gl.renderbufferStorageMultisample(
+            gl.RENDERBUFFER, samples, gl.DEPTH_COMPONENT16, width, height);
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR,
+            "should be no error after renderbufferStorageMultisample(DEPTH_COMPONENT16).");
+    }
+
+    shouldBe("gl.checkFramebufferStatus(gl.FRAMEBUFFER)",
+             "gl.FRAMEBUFFER_COMPLETE");
```
zhenyao: Let's return early if this is not FRAMEBUFFER_COMPLETE rather than failing the test.  On desktop GL, FRAMEBUFFER_COMPLETE isn't guaranteed, especially on a depth image only situation.

sdk/tests/conformance2/renderbuffers/multisampled-depth-renderbuffer-initialization.html
```diff
@@ -0,0 +1,172 @@
...
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, 'should be no errors');
+
+    // Blit from multisampled buffer to non-multisampled buffer.
+    gl.bindFramebuffer(gl.READ_FRAMEBUFFER, fbo_m);
+    gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, fbo);
+    gl.clearDepth(0);
```
zhenyao: This will leak to the next run of runTest().  So you need to call gl.clearDepth(1) earlier in this function.

kainino0x: It actually should be set to 0 always. I moved it to the top of the test.

sdk/tests/conformance2/renderbuffers/multisampled-depth-renderbuffer-initialization.html
kainino0x: Done.

zhenyao:
LGTM

PR #2420, opened 2017-06-13
Change some WEBGL_multiview errors to be generated at draw

Oletus:
Make it possible to enable EXT_disjoint_timer_query(2) at the same
time as WEBGL_multiview. This is important particularly for emscripten
compatibility. Emscripten apps enable all available extensions at
initialization. So if a timer query is active, multiview draws and
clears now generate an INVALID_OPERATION error.

Transform feedback related errors are also changed to happen at draw
time rather than at state changes. This is a more elegant API, and
even though making it perform as well is slightly more difficult, a
sufficiently performant implementation should be possible.


Oletus:
@jdashg, could you in particular take a look? I'm addressing some of the feedback you had with regards to the extension proposal.

One thing that's still unresolved is whether we want opaque multiview framebuffers to be WebGLFramebuffer or have a new WebGLOpaqueFramebuffer type. I do prefer the current approach for the following reasons:

-It would seem awkward if there was a different bindOpaqueFramebufferWEBGL entry point for them - it seems more natural that opaque multiview framebuffers are bound the same way as regular FBOs and the default framebuffer.
-Having a WebGLOpaqueFramebuffer would make it tricky to handle getParameter(FRAMEBUFFER_BINDING) which is currently specified in the core spec to return WebGLFramebuffer. I don't know if there's a precedent for a WebGL extension changing the type of object returned by some core spec entry point?
-To me having an immutable WebGLFramebuffer doesn't seem that hacky or weird - it's quite a common pattern in GL that specific properties of an object like its format prevent it being used in certain ways through the API.

One thing that should probably be cleaned up wrt. opaque multiview framebuffers is the return value of FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE. Now it's specified to return FRAMEBUFFER_DEFAULT, a new enum would probably be cleaner.

Thoughts?

RafaelCintron:
I am in favor of not having a new WebGLOpaqueFramebuffer.

kenrussell:
These changes look good. Are there any objections to merging them? I agree that it would be painful to add a new WebGLOpaqueFramebuffer type, and would rather simply have instances of WebGLFramebuffer which obey certain restrictions like not being able to query or modify their attachments.

Oletus:
I updated the pull request to cover a bit more ground, I noticed one necessary condition for framebuffer completeness that was previously omitted.

kenrussell:
Still looks good to me. For the record, since Squash and Merge is now available in Github, it's probably best to add more commits to the pull request, rather than squashing them, so that people can see the history during the review. I just updated https://www.khronos.org/webgl/wiki/Using_Github_To_Contribute indicating this is the preferred workflow now.

kenrussell:
Still looking for review from @jdashg .

Oletus:
Ping @jdashg can you do the review? This has been sitting here for a while now.

extensions/proposals/WEBGL_multiview/extension.xml
```diff
@@ -124,6 +106,15 @@
       <addendum>
         In OpenGL ES shading language version 1.00 <code>gl_ViewID_OVR</code> has the type <code>int</code> as opposed to <code>uint</code>.
       </addendum>
+      <addendum>
+        When a timer query is active and the number of views in the current draw framebuffer is greater than 1, attempting to draw generates an <code>INVALID_OPERATION</code> error.
+      </addendum>
+      <addendum>
+        When a timer query is active and the number of views in the current draw framebuffer is greater than 1, <code>calling clear</code> generates an <code>INVALID_OPERATION</code> error.
```
jdashg: Merge this with the above draw error.

extensions/proposals/WEBGL_multiview/extension.xml
```diff
@@ -240,6 +210,15 @@ interface WEBGL_multiview {
...
       The error <code>INVALID_OPERATION</code> is generated by attempting to draw if the active program declares a number of views and the number of views in the draw framebuffer does not match the number of views declared in the active program.
     </error>
+    <error>
+      The error <code>INVALID_OPERATION</code> is generated by attempting to draw if the number of views in the current draw framebuffer is greater than 1 and transform feedback is active.
+    </error>
+    <error>
+      The error <code>INVALID_OPERATION</code> is generated by attempting to draw if the number of views in the current draw framebuffer is greater than 1 and a timer query is active.
+    </error>
+    <error>
+      The error <code>INVALID_OPERATION</code> is generated by calling <code>clear</code> if the number of views in the current draw framebuffer is greater than 1 and a timer query is active.
```
jdashg: Merge the similar cases here too.

jdashg:
Apologies for holding this back so long.

Oletus:
It's alright, I was hoping to get this in a bit sooner and maybe we can do better in the future, but the spec hasn't been holding the implementation back too much yet.

PR #2421, opened 2017-06-14
Remove extra includes of glsl-conformance-test.js

Oletus:
It was being included in several WebGL 2 conformance tests that were
not using it.


zhenyao:
Thanks for cleaning this up

PR #2422, opened 2017-06-15
Fix bogus vertex attribute name in offset test

Oletus:
The attribute locations are specified in the shader text in
gl-vertexattribipointer-offsets.html. Clean up the non-existent
attribute name "vPosition".

Add a new targeted test to cover bindAttribLocation behavior when the
attribute name doesn't exist in the shader. This didn't seem to be
intentionally covered in existing tests.


kenrussell:
Nice catch Olli and good new test. LGTM.

Issue #2419, opened 2017-06-09
WebGL2 drawArraysInstanced etc seem to be based on ANGLE but differences not documented


greggman:
Here's a small sample that uses `drawArrays` and `drawArraysInstanced` in WebGL2.

https://jsfiddle.net/greggman/pxqvk4pg/

For `drawArraysInstanced` I get the error

    chrome: GL ERROR :GL_INVALID_OPERATION : glDrawArraysIntancedANGLE: attempt to draw with all attributes having non-zero divisors
    firefox: WebGL warning: drawArraysInstanced: at least one vertex attribute divisor should be 0

AFAICT no such limit exists in OpenGL ES 3.0 and the limit is not documented as a difference in the spec in the WebGL2 docs

The docs probably need to mention that limitation

computersarecool:
Is there a reason this has to be an error?  Here is a reason why it'd be great if this didn't print an error / if there was not this limitation:  

Frequently I start drawing a shape while waiting for its shader to load asynchronously. While loading I use a temp shader for which no attributes have divisors.  With the current settings my program prints errors to the console until the real shader has loaded.

I don't see why this has to be an error in WebGL2 but not in OpenGL.

zhenyao:
@greggman is correct the limitation came from ANGLE.  In [ANGLE_instanced_arrays](https://www.khronos.org/registry/OpenGL/extensions/ANGLE/ANGLE_instanced_arrays.txt), we have a limitation due to D3D: """
    INVALID_OPERATION is generated by DrawArraysInstancedANGLE or
    DrawElementsInstancedANGLE if there is not at least one enabled
    vertex attribute array that has a <divisor> of zero and is bound to an
    active generic attribute value in the program used for the draw command.
"""
@vonture @null77 Can we lift this limit in D3D11? If not, we probably need to add the constraints to WebGL2 spec and add a conformance test (the current test is the WebGL1 extension test).

@jdashg: what do you think?

computersarecool:
Additionally, I can think of situations where I want to instance using only `gl_InstanceID` (to apply an offset or something) in a shader with no `in` attributes at all.

kenrussell:
Not sure whether this restriction is still necessary. @null77 @vonture in ANGLE's current D3D11 backend, do we still need to have at least one vertex attribute with a divisor of 0 for efficiency?

[...]

PR #2424, opened 2017-06-22
Remove WebGL2 constraint that requires draw*Instanced() calls to fail

zhenyao:
if non of vertex attribs have zero a divisor.

The original limit was introduced in WebGL1 extension due to D3D9 limit.
Such limit does not exist in D3D11, therefore should not be applied to WebGL2.


zhenyao:
This is addressing https://github.com/KhronosGroup/WebGL/issues/2419

zhenyao:
@kenrussell Please take a look
@jdashg @Kangz @vonture FYI

Kangz:
LGTM

For reference here's the respective ANGLE patch: https://chromium-review.googlesource.com/c/544545/

kenrussell:
Great. Thanks for making this change. LGTM. Merging now.

PR #2425, opened 2017-06-22
Add test for uniform/uniform block naming conflicts.

y-novikov:


zhenyao:
1) you need to add the test to the 00_test_list.txt in the same folder
2) I thought the intention is to test naming conflict between uniform block names and input names and output names across shaders (because they all belong to global namespace and uniform block names belong to shared global namespace)

y-novikov:
1. I thought I did that. Though, I missed it on my first "git push", so I did "git commit --amend", "git push --delete" and "git push" again, maybe this caused some problem? I do see this test in sdk/tests/conformance2/glsl3/00_test_list.txt when I look at the commit. You don't see it? Maybe I should create a new branch, then. How do I edit github commits, anyway?
2. The spec says in 4.2.5:
"Vertex shader outputs are not considered to be shared globals since they
must pass through the rasterization stage before they are used as input by the fragment shader."
There is only 1 mention of "global name space" in the spec, in 4.3.5, where uniform variables are discussed. I concluded that only uniforms exist in "global name space".
In other places, spec says "global scope", which I think is limited to a single shader.

Also, WebGL 1 doesn't test for uniform/varying conflicts, just uniform/attribute conflicts.

I think, even if we ignore the spec, and consider in and out variables as shared, there is no need to check for them during link time, because they will have the same name in vertex and fragment, so they should fail at compile time.

zhenyao:
Now I see the 00_test_list.txt.  Don't know why it wasn't there the first time I checked.

As for the input/output, here is what my understanding is (feel free to disagree).

I think in WebGL1/ES2, we are under testing and under validating in Chrome. Shared global scope means a name defined in any shader is registered in all shaders.  So a uniform name will conflict with global names in any shaders.  In that sense, any global names, including input/output names, also local variables defined in the global scope, even though these are not in the "shared" global namespace, i.e., they will not conflict with names in other shader's "local" global names. However, they will conflict with names in "shared" global namespace.

zhenyao:
I'd like to hear what others think about this, specifically @null77 @kenrussell @Oletus

y-novikov:
I see your point. But I think a local global scope name will conflict with shared global scope name only if shared global scope name is redeclared locally, at which point the shader will fail compilation, so no reason to validate that at link time.

zhenyao:
I am not sure that's the case.  All text about attributes in ES2 is also saying it's in global namespace. It's also not in shared global namespace. Yet we detect the conflicts and reject linking. I don't see how other global variables are different.

zhenyao:
I am happy to be convinced otherwise though.

y-novikov:
The only difference I found between uniforms and attributes on one side and varyings on the other is in GLSL 1.0.17 section 4.3.
varying form linkage between vertex and fragment shader, while attributes and uniforms also form linkage with OpenGL ES. I think this maybe is the meaning of the "shared global namespace" - it is the namespace through which they are accessed from OpenGL ES, not from GLSL ES.
Thus, I think I should remove the check for conflicts with instanced uniform block name and with uniform block members. Probably only uniforms and uniform block names can conflict when accessed from OpenGL ES. Though, an error is returned now for conflict between uniform and uniform block member, probably by the driver.
Can we ask people who wrote the spec?

y-novikov:
I found something interesting in GLSL ES 3.20.0 section 4.3.5:
"Uniforms in shaders all share a single global name space when linked into a program or separable
program. Hence, the types, precisions and any location specifiers of all declared uniform variables with
the same name must match across shaders that are linked into a single program. However it is not
required to repeat the location specifier in all the linked shaders. While this single uniform name space is
cross stage, a uniform variable name's scope is per stage: If a uniform variable name is declared in one
stage (e.g., a vertex shader) but not in another (e.g., a fragment shader), then that name is still available in
the other stage for a different use."

This seems to support the idea that uniform from one shader will not conflict with "in" or "out" of another shader.

y-novikov:
Oh, and in section 4.3.9 there is an explanation of why the driver reported an error for uniform and uniform block member conflict:
"Within a shader interface, all declarations of the same global name must be for the same object and must
match in type and in whether they declare a variable or member of a block with no instance name. The
API also needs this name to uniquely identify an object in the shader interface. It is a link-time error if
any particular shader interface contains
• two different blocks, each having no instance name, and each having a member of the same
name
or
• a variable outside a block, and a block with no instance name, where the variable has the same
name as a member in the block."

The definition of shader interface is:
"A shader interface is defined to be one of these:
• All the uniform variables and uniform blocks declared in a program. This spans all compilation
units linked together within one program.
• All shader storage blocks.
• The boundary between adjacent programmable pipeline stages: This spans all the outputs
declared in the first stage and all the inputs declared in the second stage. Note that for the
purposes of this definition, the fragment shader and the preceding shader are considered to have
a shared boundary even though in practice, all values passed to the fragment shader first pass
through the rasterizer and interpolator."

In 3.0 only the first one applies, I think - "All the uniform variables and uniform blocks declared in a program".

So, I'd say only these 3 can conflict:
1. uniform variable names
2. uniform block names
3. members of uniform blocks with no instance name

Agreed?

null77:
This "local global scope" terminology you introduced is a bit confusing. Here's what the desktop GLSL 450 spec says about shared globals:

> Shared globals are global variables declared with the same name in independently compiled units
> (shaders) within the same language (i.e., same stage, e.g., vertex) that are linked together when making a
> single program.

The "uniforms only" stuff is probably out of date copy-pasted text. It should also include uniform blocks. Varyings seem to be intentionally excluded:

> Globals forming the interface between two different shader languages are discussed in other sections.

Also the language for "shader interfaces" matches what Yuly posted in his comment above. But he missed that "programmable pipeline stages" means vertex and fragment shader boundaries - it even explicitly mentions fragment shaders. 

> Within a shader interface, all declarations of the same global name must be for the same object and must
> match in type and in whether they declare a variable or member of a block with no instance name.

So: conflicts between uniforms and varyings? yes. conflicts between uniform blocks and the uniforms and varyings? yes. conflicts between attributes and fragment outputs? no, those aren't part of the same shader interface. That's my reading anyway.

zhenyao:
Thank you both for digging the spec. I think it's quite clear now that the Chrome's behavior (and that of WebGL conformance) is incorrect, and we should not fail link if fragment shader's uniform name conflict with attribute names.

Do we agree with that?

We will need to dig up the bug where we add the test cases.  If it's from reading the spec incorrectly, we can just relax it. If it's from driver conformance issues, I still think we are OK because we prefix variable names differently depending on the type (at least that's what I remembered).

y-novikov:
Comment in shaders-with-name-conflicts.html says:
"Though most drivers allow this to work, some drivers to do no therefore
WebGL implementation must enforce this restriction to provide consistent
behavior."
So, that test should remain as is, I believe.
I also don't think @null77 said that attributes don't conflict with uniforms (which I think do conflict). He said that attributes don't conflict with fragment outputs.

I need to read the spec one more time to respond to @null77 comment. My basic idea is "names accessible from GL conflict at link time, names accessible from GLSL conflict at compile time".

zhenyao:
It's not a big deal we leave it there, but this seems like driver bugs, and Chrome/ANGLE with name hashing probably already worked around it.

zhenyao:
Correction: we didn't add different prefixes for different variable types in shader name hashing. I was planning to do that, and probably did it in my dreams and thought I did in reality. :(

y-novikov:
So, John Kessinich has confirmed that the only possible conflict is between uniform names and uniform block members. See https://gitlab.khronos.org/opengl/GLSL/issues/12 and the spec that says "If a uniform variable name is declared in one stage (e.g., a vertex shader) but not in another (e.g., a fragment shader), then that name is still available in the other stage for a different use.". There are 4 things we need to do:
1. Make sure that ANGLE name hashing works around driver bugs which don't allow attributes and uniforms to have the same name.
2. Change shaders-with-name-conflicts test to check the opposite of what it checks now.
3. Implement verification of uniform / uniform block member conflicts.
4. Add a test for it.
Going to close this pull request.

PR #2372, opened 2017-04-18
Add CopyTexImage2D tests for EXT_color_buffer_float

qkmiao:

PR #2426, opened 2017-06-22
Fix ext-color-buffer-float.html to pass valid RGB16F/32F texture formats

Samulus:
RGB16F and RGB32F are both valid internal texture formats and
should not produce an error when this extension is enabled. The
extension should not be limiting the enums that can be passed
to CopyTexImage.

### DIFF
--- a/extensions/EXT_color_buffer_float/extension.xml
+++ b/extensions/EXT_color_buffer_float/extension.xml
@@ -39,10 +39,7 @@
       </feature>
 
       <feature>
-        <p>Renderbuffers with these internal formats can be created. These
-        internal formats are valid for CopyTexImage2D. Note that textures
-        with these internal formats can be created with TexImage2D in core
-        WebGL 2.0.</p>
+        <p>Renderbuffers with these internal formats can be created.</p>
       </feature>
 
       <feature>
--- a/sdk/tests/conformance2/extensions/ext-color-buffer-float.html
+++ b/sdk/tests/conformance2/extensions/ext-color-buffer-float.html
@@ -427,11 +427,7 @@ function runCopyTexImageTest(enabled)
         if (enabled) {
             shouldBe("gl.checkFramebufferStatus(gl.FRAMEBUFFER)", "gl.FRAMEBUFFER_COMPLETE");
             gl.copyTexImage2D(gl.TEXTURE_2D, level, destInternalformat, 0, 0, width, height, 0);
-            if (destInternalformat == gl.RGB16F || destInternalformat == gl.RGB32F) {
-                wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "CopyTexImage2D should fail.");
-            } else {
-                wtu.glErrorShouldBe(gl, gl.NO_ERROR, "CopyTexImage2D should succeed.");
-            }
+            wtu.glErrorShouldBe(gl, gl.NO_ERROR, "CopyTexImage2D should succeed.");
         } else {
             shouldBe("gl.checkFramebufferStatus(gl.FRAMEBUFFER)", "gl.FRAMEBUFFER_INCOMPLETE_ATTACHMENT");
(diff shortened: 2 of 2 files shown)


jdashg:
We ran into this as a failure in our top-of-tree webgl2 conformance testing for Firefox. 2.0.0 does not have this issue.

zhenyao:
In general I agree with this change.  However, I don't remember who wrote the original extension spec and add the CopyTexImage sentence that sounds like only these color-renderable float formats are allowed as CopyTexImage internalformat. So I'd double check with that person in case we miss something from the ES3 spec. @shannonwoods ?

kenrussell:
Qiankun added these tests and this restriction in https://github.com/KhronosGroup/WebGL/pull/2372 .

I also agree in general with this change, and am not sure why the restriction was put in place. There may be some subtlety with CopyTexImage's rules. I'm in the process of building Chromium on Windows to see where these INVALID_OPERATIONs are generated (in Chrome or ANGLE).

extensions/EXT_color_buffer_float/extension.xml
```diff
@@ -39,10 +39,7 @@
       </feature>
 
       <feature>
-        <p>Renderbuffers with these internal formats can be created. These
-        internal formats are valid for CopyTexImage2D. Note that textures
-        with these internal formats can be created with TexImage2D in core
-        WebGL 2.0.</p>
+        <p>Renderbuffers with these internal formats can be created.</p>
       </feature>
```
msc-: This change is unrelated to the CTS change. Do you realize this?

sdk/tests/conformance2/extensions/ext-color-buffer-float.html
```diff
@@ -427,11 +427,7 @@
         if (enabled) {
             shouldBe("gl.checkFramebufferStatus(gl.FRAMEBUFFER)", "gl.FRAMEBUFFER_COMPLETE");
             gl.copyTexImage2D(gl.TEXTURE_2D, level, destInternalformat, 0, 0, width, height, 0);
-            if (destInternalformat == gl.RGB16F || destInternalformat == gl.RGB32F) {
-                wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "CopyTexImage2D should fail.");
-            } else {
-                wtu.glErrorShouldBe(gl, gl.NO_ERROR, "CopyTexImage2D should succeed.");
-            }
+            wtu.glErrorShouldBe(gl, gl.NO_ERROR, "CopyTexImage2D should succeed.");
         } else {
```
msc-: This change is incorrect. RGB{16,32}F formats are not color-renderable. Neither the GL_EXT_color_buffer_float extension nor the WebGL wrapper make them so. This test needs to remain as is.

msc-:
> On Jun 23, 2017, at 13:09, zhenyao <notifications@github.com> wrote:
> 
> Mark, I fail to find a spect text that support your claim. To my
> understanding, color-renderable-ness only affect source texture in the FBO,
> not the target texture format in CopyTexImage.
> 
> If I miss something, can you point it to us?

You didn’t miss anything. That is why I deleted my comment, but apparently not fast enough.

The change to the conformance test is correct. The core ES 3.0 spec. says that components may be dropped during CopyTexImage and there is a table that states that a base format of RGB is an acceptable destination for a source base format of RGBA. So copying an RGBA32F framebuffer attachment to an RGB32F texture is definitely supported.

Regards

    -Mark

zhenyao:
Thanks.

Mo

jdashg:
Ohh, I think the origin of this restriction may have been regarding whether dropping components from sized formats is valid for CopyTex*Image. (PR #2226)
Regardless, as that PR clarifies, dropping components is allowed for sized formats.

kenrussell:
Dropping of components is certainly allowed, even for sized internal formats.

I've examined the history of this change and it appears that it was an oversight. Will fix in Chrome in http://crbug.com/736499 . Thanks for finding this and fixing the test.

PR #2427, opened 2017-06-22
Attempt to fix Travis CI; remove apparently unneeded dependency.

kenrussell:


kenrussell:
Good, it passed. Merging myself.

PR #2428, opened 2017-06-23
Don't require INVALID_OPERATION on index validation failure.

jdashg:
I also nuke the redundant conformance/more/functions/drawElementsBadArgs in favor of our other tests around the tree, and moving some tests to conformance/rendering/gl-drawelements. In particular, the the more/functions/* tests don't check the error value, only whether or not an error occurred, making them less useful than tests in other (sub-)frameworks.


zhenyao:
This is really nice. Thank you.

jdashg:
cc @RafaelCintron

zhenyao:
We probably also want to merge this back to 1.0.2 since Edge and Safari haven't passed that version yet.

PR #2430, opened 2017-06-23
Another test case being affected by instanced draws where we no longer

zhenyao:
requires at least one vertex attribs to have zero divisor.

This is addressing https://github.com/KhronosGroup/WebGL/issues/2419


zhenyao:
@Kangz @kenrussell PTAL (I missed this one yesterday)

kenrussell:
LGTM.

PR #2431, opened 2017-06-23
Add a test for a Chrome bug where querying uniform block info fails a…

zhenyao:
…fter detaching shaders from a program

See crbug.com/735784

### DIFF
--- a/sdk/tests/conformance2/programs/query-uniform-blocks-after-shader-detach.html
+++ b/sdk/tests/conformance2/programs/query-uniform-blocks-after-shader-detach.html
@@ -1,0 +1,119 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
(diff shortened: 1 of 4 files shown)


zhenyao:
@kenrussell Please review

sdk/tests/conformance2/programs/query-uniform-blocks-after-shader-detach.html
```diff
@@ -0,0 +1,119 @@
...
+<head>
+<meta charset="utf-8">
+<title>WebGL Uniform Buffers Conformance Tests</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+<script id="vshader" type="x-shader/x-vertex">#version 300 es
+
+in vec4 position;
+in vec4 color;
```
kenrussell: Test case could be simplified by removing "color" and "v_frontColor" and setting "resultColor" to a constant value.

zhenyao:
addressed @kenrussell  comments

Issue #2423, opened 2017-06-20
Calling framebufferRenderbuffer without bindRenderbuffer


RafaelCintron:
While writing conformance tests today, I discovered an interoperability bug between browsers with framebufferRenderBuffer.

Consider the following piece of code:
```javascript
let renderbuffer = gl.createRenderbuffer();
gl.framebufferRenderbuffer(gl.FRAMEBUFFER, attachment.enum, gl.RENDERBUFFER, renderbuffer);
wtu.glErrorShouldBe(gl, gl.NO_ERROR);
assertMsg(renderbuffer === gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME),"renderbuffer attachment suceeded");
```
On Chrome, the call to framebufferRenderbuffer fails with INVALID_OPERATION but the attachment succeeds, yielding a "pass" for the subsequent assertMsg.
On Firefox and Edge, both the call to framebufferRenderbuffer and subsequent assertMsg succeed.

Angle, on the other hand, forbids calling framebufferRenderbuffer without first calling bindRenderbuffer.  
```C++
// [OpenGL ES 2.0.25] Section 4.4.3 page 112
// [OpenGL ES 3.0.2] Section 4.4.2 page 201
// 'renderbuffer' must be either zero or the name of an existing renderbuffer object of
// type 'renderbuffertarget', otherwise an INVALID_OPERATION error is generated.
if (renderbuffer != 0)
{
    if (!context->getRenderbuffer(renderbuffer))
    {
        context->handleError(Error(GL_INVALID_OPERATION));
        return false;
    }
}
```
Such a restriction is certainly necessary for `bindTexture` before `framebufferTexture2D` because, otherwise, you would have no way to know whether to accept cubemap or non-cubemap texture targets.  So it makes sense to have the same restriction for `bindRenderbuffer` before `framebufferRenderbuffer`.

As I see it, our choices are to either have framebufferRenderbuffer flag no error and for the attachment to succeed OR have frambufferRenderbuffer flag an error and have the attachment fail. Either way, Chrome's current behavior will have to change.

The downside to the latter approach is potential compatibility issues for existing content that, today, freely calls `createRenderBuffer` followed by `framebufferRenderbuffer` without first calling `bindRenderbuffer`.  In spite of this, I'm still inclined to favor the latter behavior.  

[...]

PR #2432, opened 2017-06-26
Test calling framebufferRenderbuffer before bindRenderbuffer

RafaelCintron:
Calling framebufferRenderbuffer on a created renderbuffer before calling bindRenderbuffer on the render buffer must flag an invalid operation and the attachment should fail.

From the spec:
[OpenGL ES 2.0.25] Section 4.4.3 page 112
[OpenGL ES 3.0.2] Section 4.4.2 page 201
'renderbuffer' must be either zero or the name of an existing renderbuffer object of
type 'renderbuffertarget', otherwise an INVALID_OPERATION error is generated.

### DIFF
--- a/sdk/tests/conformance/renderbuffers/framebuffer-object-attachment.html
+++ b/sdk/tests/conformance/renderbuffers/framebuffer-object-attachment.html
@@ -426,6 +426,7 @@ if (checkValidColorDepthCombination()) {
     testFramebufferIncompleteMissingAttachment();
     testUsingIncompleteFramebuffer();
     testReadingFromMissingAttachment();
+    testBindRenderbufferBeforeFramebufferAttach();
 }
 
 function checkFramebuffer(expected) {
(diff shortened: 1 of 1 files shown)


kainino0x:
It's not really clear to me from the spec that the renderbuffer has to both exist and be bound (since framebufferRenderbuffer takes a renderbuffer name as well as gl.RENDERBUFFER as arguments). However, I was recently working on renderbuffer stuff and it was definitely true in practice. LGTM.

Would it be worth clarifying this behavior in the WebGL spec, or am I missing some language from the ES specs that implies this?

kenrussell:
Per @RafaelCintron 's spec references above and discussion on #2423 , the specs already define the behavior here.

@RafaelCintron thanks for the new tests. LGTM. Since @zhenyao was working on this simultaneously, let's let him review and merge.

zhenyao:
LGTM

kainino0x:
Ah, thanks, I understand the requirement now.
https://github.com/KhronosGroup/WebGL/issues/2423#issuecomment-310246053

RafaelCintron:
Apologies for the duplicated effort.  I didn't know @zhenyao was already working on the test.

zhenyao:
No apology necessary @RafaelCintron . Now I remembered that in the last meeting you said you would do it.

sdk/tests/conformance/renderbuffers/framebuffer-object-attachment.html
```diff
@@ -655,6 +656,40 @@
...
+
+    attachmentTypes.forEach(function(attachmentType) {
+        let strAttachmentType = wtu.glEnumToString(gl, attachmentType);
+        debug("");
+        debug("Testing " + strAttachmentType);
+        let rbo = gl.createRenderbuffer();
+        gl.framebufferRenderbuffer(gl.FRAMEBUFFER, attachmentType, gl.RENDERBUFFER, rbo);
+        wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "bindRenderbuffer must be called before attachment to " + strAttachmentType);
+        shouldBe("gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl." + strAttachmentType + ", gl.FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE)", "gl.NONE");
+        shouldBeNull("gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl." + strAttachmentType + ", gl.FRAMEBUFFER_ATTACHMENT_OBJECT_NAME)");
```
zhenyao: I just realized a spec difference between ES2/ES3.

In ES2, if FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE is NONE, i.e., no image is attached, querying FRAMEBUFFER_ATTACHMENT_OBJECT_NAME generates INVALID_ENUM.

In ES3, such query is allowed and 0 is returned.

@RafaelCintron Can you address this? This test is for ES2 only, so we should probably just remove this.

zhenyao: I'll add a spec update and clear this up. @RafaelCintron

PR #2433, opened 2017-06-26
Add a test for framebufferRenderbuffer with never bound renderbuffer.

zhenyao:
This addresses the issue raised in https://github.com/KhronosGroup/WebGL/issues/2423


zhenyao:
@kenrussell Please take a look
@kainino0x @RafaelCintron @jdashg FYI

zhenyao:
Never mind.  I saw https://github.com/KhronosGroup/WebGL/pull/2432 already covers this.

Thanks @RafaelCintron

PR #2435, opened 2017-06-27
Removing non-black transparent pixels from createImageBitmap conformance tests

zakerinasab:
This change updates WebGL conformance tests for createImageBitmap from 
ImageBitmap and ImageData to avoid non-black transparent pixels. Skia is
removing support for unpremul operations, which means to do scaling,
color correction, etc. unpremul images must be first converted to
premul. When pixel values are non-black transparent, like 0,255,0,0,
the unpremul->premul->unpremul round-trip results in transparent black,
breaking the previous conformance tests.

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image-bitmap.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-image-bitmap.js
@@ -51,9 +51,9 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
 
         var imageData = new ImageData(new Uint8ClampedArray(
                                       [255, 0, 0, 255,
-                                      255, 0, 0, 0,
+                                      255, 0, 0, 128,
                                       0, 255, 0, 255,
-                                      0, 255, 0, 0]),
+                                      0, 255, 0, 128]),
                                       2, 2);
 
         createImageBitmap(imageData, {imageOrientation: "none", premultiplyAlpha: "none"})
@@ -61,7 +61,7 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
             testPassed("createImageBitmap with options may be rejected if it is not supported. Retrying without options.");
             return createImageBitmap(imageData);
         }).then( bitmap => {
-            return runImageBitmapTest(bitmap, 0, internalFormat, pixelFormat, pixelType, gl, tiu, wtu, false);
+            return runImageBitmapTest(bitmap, 0.5, internalFormat, pixelFormat, pixelType, gl, tiu, wtu, false);
         }, () => {
             testFailed("createImageBitmap(imageData) should succeed.");
         }).then(() => {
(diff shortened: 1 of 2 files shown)


kenrussell:
There is a major overarching rule governing the WebGL conformance tests: tests are not removed or modified in order to work around bugs in browsers. I've personally been guilty (deliberately or accidentally) of violating this rule in the past, so am particularly sensitive to it.

Now, it may be the case that these tests have been relying on behavior that's undefined in the ImageBitmap specification. For example, the 2D canvas context's backing store is allowed per spec to be stored in premultiplied form, so some operations on it are lossy.

Is there similar wording governing ImageBitmap? If not, then some has to be added in order for this change to be accepted.

CC'ing @jdashg explicitly since Firefox is the only other browser that's currently implementing ImageBitmap (even with partial support).

zakerinasab:
As far as I see, there is nothing about the backing store of the
ImageBitmap in the standard. It only says that if options.premultiply is
"none", the output that is not premultiplied by alpha must be left
untouched.

In the part that talks about ImageData, however, it says that due to the
lossy nature of converting to and from premultiplied alpha color
values, the data read by getImageData might be different from what has been
set using putImageData. As I see we have the same issue in these
conformance tests, as we create an ImageBitmap from an ImageData, draw it
to a canvas, and read back the pixels using gl.readPixels or
gl.getImageData. I'm not sure if we need to add similar wording for
ImageBitmap or not, as this is only happening when using ImageData as the
source (other sources are premul, so there should be no change in a
write-read round-trip).

I'll be happy to add a couple of lines to the standard if it is needed to
move forward.

kenrussell:
Hang on a second. Not all of the other ImageBitmap sources are premultiplied. It absolutely must continue to be the case that if you create an ImageBitmap from a Blob representing a PNG with a separate alpha channel, and set premultiplyAlpha:none in the ImageBitmap creation options, that the alpha channel will never be multiplied into the color channels.

This is supposed to be verified by https://github.com/KhronosGroup/WebGL/tree/master/sdk/tests/conformance/textures/image_bitmap_from_blob .

I certainly hope that this test is suitably verifying this. I can't stress enough how important this particular code path is. Can you please help confirm that Skia's removal of its unmultiplied code paths isn't going to affect this?

It would be OK to add some text to https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#dom-imagebitmapoptions-premultiplyalpha saying that if the input is an ImageData, its alpha channel may have already been premultiplied, so premultiplyAlpha:none may not be lossless. (The same would be true for Canvas elements, btw.) But it absolutely must be honored for Blobs that decode to images that contain an alpha channel.

jdashg:
Non-alpha-premult data (from at least PNG) must be retained.

kenrussell:
For the record, it looks like the Chromium CL which informed this pull request is https://chromium-review.googlesource.com/c/542959/ . Based on @junov 's comments there, it's not clear that making this change is correct.

Please comment again on this pull request if you're requesting a re-review.

jdashg:
I'll tentatively close this.

PR #2437, opened 2017-06-28
Add a test exposing non-conformant multisampling

Oletus:
The test checks that the fragment shader is evaluated only once per
framebuffer pixel when drawing into a multisampled framebuffer. This
is specified in GLES 3.0.5 section 3.6.3.

### DIFF
--- a/sdk/tests/conformance2/rendering/multisampling-fragment-evaluation.html
+++ b/sdk/tests/conformance2/rendering/multisampling-fragment-evaluation.html
@@ -1,0 +1,164 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL multisampling fragment shader evaluation</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+
+<script id="vshader" type="x-shader/x-vertex">#version 300 es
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/rendering/multisampling-fragment-evaluation.html
```diff
@@ -0,0 +1,163 @@
...
+    let possibleColors = [
+        [255, 0, 0, 255],
+        [0, 255, 0, 255],
+        [0, 0, 255, 255]
+    ];
+    let anyColorMatched = false;
+    for (let i = 0; i < possibleColors.length; ++i) {
+        let colorMatched = true;
+        for (let j = 0; j < 4; ++j) {
+            if (readBuffer[j] !== possibleColors[i][j]) {
```
kenrussell: Should a tolerance be added here?

sdk/tests/conformance2/rendering/multisampling-fragment-evaluation.html
```diff
@@ -0,0 +1,163 @@
...
+        return;
+    }
+
+    // Create a program that will choose between one of different possible colors in the fragment shader.
+    // It should be evaluated only once per framebuffer pixel, so only one of the colors will end up in the framebuffer.
+    // However, if the multisampling mode is incorrectly implemented by supersampling, the samples may have different
+    // colors.
+    let program = wtu.setupProgram(gl, ["vshader", "fshader"], ["aPosition"]);
+
+    // Render one triangle using the program. The triangle covers the whole viewport.
```
kenrussell: Any chance that the upper left corner of the framebuffer, or the right or bottom edge, will end up participating in a multisample resolve and therefore have a different color than one of the three allowed ones?

sdk/tests/conformance2/rendering/multisampling-fragment-evaluation.html
```diff
@@ -0,0 +1,163 @@
...
+    gl.readPixels(0, 0, size, size, gl.RGBA, gl.UNSIGNED_BYTE, readBuffer);
+
+    // Check that the canvas is one of the colors that the fragment shader may generate, and not a blend of them.
+    let possibleColors = [
+        [255, 0, 0, 255],
+        [0, 255, 0, 255],
+        [0, 0, 255, 255]
+    ];
+    let anyColorMatched = false;
+    for (let i = 0; i < possibleColors.length; ++i) {
```
kenrussell: These loops only work correctly if size=1 (which it is). Should loops be added to iterate the X and Y coordinates too?

sdk/tests/conformance2/rendering/multisampling-fragment-evaluation.html
Oletus: Good point. Sample locations aren't actually very tightly specified - the triangle should be made much bigger to compensate.

sdk/tests/conformance2/rendering/multisampling-fragment-evaluation.html
Oletus: I hardcoded the size to be 1 pixel instead.

sdk/tests/conformance2/rendering/multisampling-fragment-evaluation.html
Oletus: It's reasonable to have a bit of tolerance, the final pixel value is the result of multisample resolve after all.

Oletus:
Thanks for the very good feedback, updated the patch according to the comments!

sdk/tests/conformance2/rendering/multisampling-fragment-evaluation.html
```diff
@@ -0,0 +1,164 @@
...
+
+    // Render one triangle using the program. The triangle needs to extend far outside the viewport on all sides, so
+    // that we can safely assume all samples fall inside the triangle. GLES 3.0.5:
+    // "The sample points associated with a pixel may be located inside or outside of the unit square that is considered to bound the pixel."
+    // Here we assume that sample points are less than 9999 pixels away from the pixel they are associated with.
+    let buffer = gl.createBuffer();
+    gl.bindBuffer(gl.ARRAY_BUFFER, buffer);
+    gl.bufferData(gl.ARRAY_BUFFER, new Float32Array([ 10000,  30000,
+                                                     -30000, -10000,
+                                                      10000, -10000]), gl.STATIC_DRAW);
```
kenrussell: OK, that's quite a huge triangle. :) I suspect it would be OK to use a smaller one.

Oletus: In practice a couple of pixels is probably enough, but since this is testing an area of the spec that is fairly open to interpretation, I decided to make it conservative.

PR #2438, opened 2017-06-28
Clarify what is allowed in global initializers

Oletus:
It should be made clear that referring to built-in globals like
gl_FragCoord is not allowed, but on the other hand referring to
built-in constants like gl_MaxVertexUniformVectors is.

This matches the existing behavior in ANGLE.


jdashg:
LGTM

PR #2439, opened 2017-06-29
Add test where local variable shadows outer function definition.

kenrussell:
Regression test for:
https://bugs.chromium.org/p/angleproject/issues/detail?id=2095


kenrussell:
Could someone please review this? @null77 @kainino0x @Oletus @jdashg @zhenyao 

I don't have a fix for it yet -- will be in ANGLE's HLSL backend.

kenrussell:
Merging myself.

PR #2440, opened 2017-07-01
Test drawing with textures from non-ready video uploads.

jdashg:


kenrussell:
I'm sorry, but this change is problematic. It's been the case for a very long time that Chrome (at least) has deliberately signaled a GL error when attempting to upload an incomplete HTMLImageElement or HTMLVideoElement. The rationale is that it's the only way to tell the developer that they did something wrong; otherwise they'll just get a black texture with no way to query its size, since those APIs don't exist in OpenGL ES.

There is precedence for this in the 2D canvas context spec, where it raises an exception if there's an attempt to draw an incomplete image or non-ready video element:
https://html.spec.whatwg.org/multipage/canvas.html#check-the-usability-of-the-image-argument
https://html.spec.whatwg.org/multipage/canvas.html#drawing-images:check-the-usability-of-the-image-argument

Apologies, but I'm reverting this commit because otherwise Chrome will lose test coverage when rolling forward the WebGL conformance suite. Let's either discuss the rationale further for making this change, or better, make this an error on all browsers. (Right now Chrome produces INVALID_VALUE, but we could change it to INVALID_OPERATION since the video element might be a valid value in different circumstances.)

PR #2441, opened 2017-07-03
Add a test that doesn't write gl_FragColor or gl_FragData

Oletus:
In this case, the fragment color should be left untouched according to
the WebGL spec.

This exposes a bug in ANGLE that writes the fragment color in case it
is read by the shader.

### DIFF
--- a/sdk/tests/conformance/glsl/variables/gl-fragcolor-unwritten.html
+++ b/sdk/tests/conformance/glsl/variables/gl-fragcolor-unwritten.html
@@ -1,0 +1,109 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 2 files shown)


Oletus:
The issue in ANGLE might actually be fairly complex to fix - it might be reasonable to change the spec instead. The alternative spec would be:

If a fragment shader does not statically use either gl_FragColor or gl_FragData, the values of the fragment colors following shader execution are untouched.

Or are shaders like this important to support:

if (!onlyDepth) {
  gl_FragColor = something;
}
gl_FragDepth = somethingElse;

Thoughts?

sdk/tests/conformance/glsl/variables/gl-fragcolor-unwritten.html
```diff
@@ -0,0 +1,109 @@
...
+var wtu = WebGLTestUtils;
+var gl = wtu.create3DContext();
+
+var test = function(fragShader, shaderDescription) {
+    debug("");
+    debug("Clearing canvas to green and drawing with " + shaderDescription);
+    wtu.setupProgram(gl, [wtu.simpleVertexShader, fragShader], undefined, undefined, true);
+    gl.clearColor(0, 1, 0, 1);
+    gl.clear(gl.COLOR_BUFFER_BIT);
+    gl.disable(gl.BLEND);
```
jdashg: Test also with blending enabled, with src-dependent blend funcs.

jdashg:
We should either establish rationale for this spec language, or revert it.

ESSL1 p60:
"If subsequent fixed functionality consumes fragment color and an execution of a
fragment shader does not write a value to `gl_FragColor` then the fragment color consumed is undefined."

The blend functionality likely qualifies for "subsequent fixed functionality".

ES2 is more general:
GLES 2.0.5 p89:
" Any colors, or color components, associated with a fragment that are not written by
the fragment shader are undefined."

@zhenyao Do you remember why this was made 'left untouched' and not 'zero' or (implicitly left as) 'undefined'?

jdashg:
@Oletus It would be great to get a test for ESSL3 also. (frag shader output vars)

Oletus:
We should look into a spec change instead.

kenrussell:
Note: the current spec language was added in #1866 . There's not a lot of context on the pull request and unfortunately no links to associated Chromium or ANGLE bugs that would have shown associated code changes.

zhenyao:
The reason we say untouched instead of initializing to zeros is the concern for perf cost initializing gl_fragData arrays. For example, a user may attach a bunch of color images, but only write to one at a stage.

PR #2442, opened 2017-07-03
Update test suite documentation

Oletus:
Digged out date for when 2.0.1 test suite work was started from git
history.

The command line tip for updating 00_test_list files is outdated since
new tests should be prefixed with --min-version.


kenrussell:
Thanks for updating the docs and sorry for the slow review on my side.

PR #2443, opened 2017-07-03
Add a test case for ambiguous struct parameters HLSL bug

Oletus:
Cover a missing workaround in ANGLE: The shader translation needs to
work around HLSL compiler treating structs with different names but
identical fields as ambiguous.


jdashg:
Ick, ok. :(

kenrussell:
LGTM after the fact. Thanks Olli for the fix and Jeff for reviewing.

PR #2444, opened 2017-07-03
Replace gl.COLOR_ATTACHMENT1 with ext.COLOR_ATTACHMENT1_WEBGL

Samulus:
gl.COLOR_ATTACHMENT[1-9] is undefined on WebGL1. However, when
the WEBGL_draw_buffers extension is enabled on WebGL1 one can use
ext.COLOR_ATTACHMENT[N]_WEBGL in its place. This test checks
for the extension but erroneously uses gl.COLOR_ATTACHMENT1
instead of ext.COLOR_ATTACHMENT1_WEBGL after it finding it.

### DIFF
--- a/sdk/tests/conformance/extensions/webgl-draw-buffers-feedback-loop.html
+++ b/sdk/tests/conformance/extensions/webgl-draw-buffers-feedback-loop.html
@@ -74,6 +74,7 @@ var height = 8;
 var tex0;
 var tex1;
 var fbo;
+var ext;
 var program;
 var positionLoc;
 var texCoordLoc;
@@ -83,7 +84,7 @@ if (!gl) {
 } else {
     testPassed("WebGL context exists");
 
-    var ext = gl.getExtension("WEBGL_draw_buffers");
+    ext = gl.getExtension("WEBGL_draw_buffers");
     if (!ext) {
         testPassed("No WEBGL_draw_buffers support -- this is legal");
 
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/extensions/webgl-draw-buffers-feedback-loop.html
```diff
@@ -129,7 +129,7 @@
 
     gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
     gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, tex0, 0);
-    gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT1, gl.TEXTURE_2D, tex1, 0);
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, ext.COLOR_ATTACHMENT1_WEBGL, gl.TEXTURE_2D, tex1, 0);
```
jdashg: Since you use `ext` down here now, you should move `ext` into the global scope. Move the declaration (but not the definition) up next to `var fbo`.

kenrussell:
Wow, good find. Thanks for fixing this. It looks like the reason this went unnoticed so long is that Chromium has a failure expectation for this test because https://bugs.chromium.org/p/angleproject/issues/detail?id=1619 is still in progress. CC'ing @Richard-Yunchao .

PR #2445, opened 2017-07-04
Add test for uniform blocks containing arrays of structs

Oletus:
This exposes a bug in ANGLE's DirectX/HLSL backend. Works on ANGLE's
GL backend.

### DIFF
--- a/sdk/tests/conformance2/uniforms/uniform-blocks-with-arrays.html
+++ b/sdk/tests/conformance2/uniforms/uniform-blocks-with-arrays.html
@@ -1,0 +1,136 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/uniforms/uniform-blocks-with-arrays.html
```diff
@@ -0,0 +1,136 @@
...
+    var program = wtu.setupProgram(gl, ['vshader', fragShader], ['vPosition'], undefined, true);
+    if (!program) {
+        testFailed("Loading program failed");
+        return;
+    }
+    var blockIndex = gl.getUniformBlockIndex(program, "lightData");
+
+    var uniformBuffer = gl.createBuffer();
+    gl.bindBuffer(gl.UNIFORM_BUFFER, uniformBuffer);
+    gl.bufferData(gl.UNIFORM_BUFFER, new Float32Array(bufferFloatCount), gl.DYNAMIC_READ);
```
kenrussell: Is it worth uploading non-zero data into the uniform block? Maybe set the lighting constant vector to (0, 0.5, 0, 1) and upload (0, 0.5, 0, 0);

PR #2447, opened 2017-07-06
Revert "Test drawing with textures from non-ready video uploads."

kenrussell:
Reverts KhronosGroup/WebGL#2440.

Please see my comment on the other pull request for why this change was problematic.


kenrussell:
Merging myself because this is blocking Chrome from rolling forward the WebGL conformance suite.

PR #2448, opened 2017-07-06
Texture switch performance

tsherif:
A test for the Mac/AMD texture-switching performance problem reported in: https://bugs.chromium.org/p/chromium/issues/detail?id=735483

I've tested on a Mac with an Intel Iris, iPad Air 2, Nexus 10, Pixel Phone, and they all pass as they should. An affected Mac with a Radeon R9 M290 fails as it should.

### DIFF
--- a/sdk/tests/conformance/rendering/texture-switch-performance.html
+++ b/sdk/tests/conformance/rendering/texture-switch-performance.html
@@ -73,13 +73,15 @@ if (!gl) {
       gl.drawArrays(gl.TRIANGLES, 0, 6);
     }
 
-    if (++frameCount === 100) {
-      var elapsed = Date.now() - startTime;
-      if (elapsed < 3333) {  // 30FPS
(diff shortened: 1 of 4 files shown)


sdk/tests/conformance/rendering/texture-switch-performance.html
```diff
@@ -0,0 +1,97 @@
+<!--
+
+/*
+** Copyright (c) 2012 The Khronos Group Inc.
```
kenrussell: Nit: 2017

sdk/tests/conformance2/rendering/texture-switch-performance.html
```diff
@@ -0,0 +1,97 @@
+<!--
+
+/*
+** Copyright (c) 2012 The Khronos Group Inc.
```
kenrussell: 2017 here too.

PR #2450, opened 2017-07-11
Add a test for an array of structs containing samplers as function arg

Oletus:
This exposes a bug in ANGLE's HLSL translator.

ANGLE bug link: http://anglebug.com/2103

### DIFF
--- a/sdk/tests/conformance/glsl/bugs/sampler-struct-function-arg.html
+++ b/sdk/tests/conformance/glsl/bugs/sampler-struct-function-arg.html
@@ -86,16 +86,16 @@ var gl = wtu.create3DContext(canvas);
 
 var textureGreen;
 
-var createTextureGreen = function() {
-    var textureGreen = gl.createTexture();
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/glsl/bugs/sampler-struct-function-arg.html
```diff
@@ -74,24 +83,27 @@
...
 var gl = wtu.create3DContext(canvas);
-if (!gl) {
-    testFailed("context does not exist");
-} else {
-    var textureGreen = gl.createTexture()
+
+var textureGreen;
+
+var createTextureGreen = function() {
+    var textureGreen = gl.createTexture();
```
kainino0x: nit: maybe rename to avoid confusion with the global `textureGreen`?

kenrussell:
Looks good overall. Could you please edit the pull request to point to the ANGLE bug?

Please merge yourself after addressing Kai's and my feedback. Thanks.

PR #2451, opened 2017-07-11
Test uploading HTML data (canvas, image, etc.) to luminance and/or alpha textures.

kenrussell:
Test uploading HTML data (canvas, image, etc.) to luminance and/or alpha textures.

The most extensive new tests are under conformance/textures/canvas and conformance2/textures/canvas, where semi-transparent cases have been added for all of the texture formats.

The canvas tests used to repeat multiple times, to act as regression tests for bugs in early WebGL implementations. The repetition has been kept for WebGL 1.0's RGB and RGBA internal formats, but removed for all other texture formats, to avoid dramatically increasing the tests' run time. Overall, the run time of the test suite will decrease with this pull request.

Chrome, Firefox and Safari pass all of the new WebGL 1.0 tests with WebGL 1.0 contexts. Edge passes them, but generates errors about the UNPACK_COLORSPACE_CONVERSION_WEBGL parameter, which it does not support.

Firefox fails the new WebGL 1.0 luminance, alpha, and luminance_alpha tests when run against a WebGL 2.0 context (tested Linux and macOS). Emulation of these texture formats on the Core Profile is needed, as was done for Chrome. There is a preexisting failure of the RGB10_A2 tests in Firefox.

Chrome passes all of the tests with a forthcoming bug fix, and fails several of the tests without it.

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-canvas.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-canvas.js
@@ -27,8 +27,60 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
     var gl = null;
     var successfullyParsed = false;
     var whiteColor = [255, 255, 255, 255];
-    var redColor = [255, 0, 0];
-    var greenColor = [0, 255, 0];
+    var redColor = [255, 0, 0, 255];
+    var greenColor = [0, 255, 0, 255];
+    var semiTransparentRedColor = [127, 0, 0, 127];
+    var semiTransparentGreenColor = [0, 127, 0, 127];
+    var repeatCount;
+
+    function replicateRedChannel(color)
+    {
+        color[1] = color[0];
+        color[2] = color[0];
+    }
+
+    function zapColorChannels(color)
+    {
+        color[0] = 0;
+        color[1] = 0;
+        color[2] = 0;
+    }
+
+    function setAlphaChannelTo1(color)
+    {
+        color[3] = 255;
+    }
+
+    function replicateAllRedChannels()
+    {
+        replicateRedChannel(redColor);
+        replicateRedChannel(semiTransparentRedColor);
+        replicateRedChannel(greenColor);
+        replicateRedChannel(semiTransparentGreenColor);
+    }
+
+    function setAllAlphaChannelsTo1()
+    {
+        setAlphaChannelTo1(redColor);
+        setAlphaChannelTo1(semiTransparentRedColor);
+        setAlphaChannelTo1(greenColor);
+        setAlphaChannelTo1(semiTransparentGreenColor);
+    }
+
+    function shouldRepeatTestForTextureFormat(internalFormat, pixelFormat, pixelType)
+    {
+        // There were bugs in early WebGL 1.0 implementations when repeatedly uploading canvas
+        // elements into textures. In response, this test was changed into a regression test by
+        // repeating all of the cases multiple times. Unfortunately, this means that adding a new
+        // case above significantly increases the run time of the test suite. The problem is made
+        // even worse by the addition of many more texture formats in WebGL 2.0.
+        //
+        // Doing repeated runs with just a couple of WebGL 1.0's supported texture formats acts as a
+        // sufficient regression test for the old bugs. For this reason the test has been changed to
+        // only repeat for those texture formats.
+        return ((internalFormat == 'RGBA' && pixelFormat == 'RGBA' && pixelType == 'UNSIGNED_BYTE') ||
+                (internalFormat == 'RGB' && pixelFormat == 'RGB' && pixelType == 'UNSIGNED_BYTE'));
+    }
 
     function init()
     {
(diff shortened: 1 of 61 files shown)


kenrussell:
@zhenyao @jdashg in particular please review. @kainino0x @grorg @RafaelCintron @null77 @vonture FYI.

The first commit contains the significant changes, including to the test generator. The second commit contains the new autogenerated test files. I split the commits to hopefully make this easier to review. Please tell me if I should reorganize it.

These tests pass in Chrome with the fix for http://crbug.com/733599 and fail without the fix. I spot-checked some of them with the pass-through command buffer and they pass.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
```diff
@@ -42,11 +58,22 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
             return;
         }
 
+        repeatCount = (shouldRepeatTestForTextureFormat(internalFormat, pixelFormat, pixelType) ? 4 : 1);
```
jdashg: repeatCountForTextureFormat() would be more straightforward, imo.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-canvas.js
```diff
@@ -195,13 +324,41 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
...
+                // formats with lower bit depths.
+                var tolerance = 0;
+                if (semiTransparent) {
+                    tolerance = 3;
+                    if (pixelType == 'UNSIGNED_SHORT_5_6_5' || internalFormat == 'RGB565') {
+                        tolerance = 6;
+                    } else if (pixelType == 'UNSIGNED_SHORT_4_4_4_4' || internalFormat == 'RGBA4') {
+                        tolerance = 9;
+                    } else if (pixelType == 'UNSIGNED_SHORT_5_5_5_1' || internalFormat == 'RGB5_A1') {
+                        // It's ambiguous whether the one bit of alpha should be treated as:
```
jdashg: I think this u8norm->u1norm conversion would fall under the normalized->float and float->normalized rule, applying them sequentially as u8norm->float->u1norm. In this case, this is specified in GLES 3.0.5 p10, and can actually always return either of the nearest values.
I do think we should test that 0/255->0/1 and 255/255->1/1 are preserved. (otherwise this format is useless)

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-canvas.js
```diff
@@ -43,20 +95,68 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
...
+          case gl.SRGB8_ALPHA8:
+            semiTransparentRedColor = wtu.sRGBToLinear(semiTransparentRedColor);
+            semiTransparentGreenColor = wtu.sRGBToLinear(semiTransparentGreenColor);
+            break;
+          case gl.RGBA8UI:
+            // For int and uint textures, TexImageUtils outputs 1.0 for the alpha channel all the
+            // time because of differences in behavior when sampling integer textures with and
+            // without alpha channels. Since changing this behavior may have large impact across the
+            // test suite, leave it as is for now.
+            setAllAlphaChannelsTo1();
```
jdashg: s/outputs 1.0/outputs 255/

Also afaict, sampling int/uint textures with no alpha channel yields an alpha value of 1(u)i.

Can you file an issue for this? It would be good to test this.

zhenyao: +1

zhenyao:
Thanks for adding these test cases. LGTM in general.  Feel free to merge yourself and address jdashg's comments in a followup.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
kenrussell: Thanks, good idea. Will make that change in the next commit.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-canvas.js
kenrussell: Will update the comment in the next commit. Filed #2453 about testing int/uint textures containing an alpha channel.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-canvas.js
kenrussell: Thanks for finding that spec reference. Rewrote the comment to point to it. The opaque case is already tested; the semi-transparent case though can't be tested well with these canvas-to-texture tests, because most browsers use a premultiplied backing store for 2D canvases so nothing except transparent black is representable when alpha=0.

Filed #2454 about expanding another preexisting test to cover these cases.

kenrussell:
Thanks for your reviews @jdashg and @zhenyao . Addressed the review feedback and filed follow-on issues. Merging now.

PR #2452, opened 2017-07-11
Accept INVALID_ENUM or INVALID_OPERATION for glReadPixels invalid parameters.

vonture:
This matches dEQP's readPixels tests and ANGLE's implementation because some format/type combinations may have valid enums but be invalid in combination with the other enum.

### DIFF
--- a/sdk/tests/conformance/reading/read-pixels-test.html
+++ b/sdk/tests/conformance/reading/read-pixels-test.html
@@ -169,7 +169,7 @@ function continueTestPart2() {
     var format = invalidFormat[ff];
     var buf = new Uint8Array(4);
     gl.readPixels(0, 0, 1, 1, format, gl.UNSIGNED_BYTE, buf);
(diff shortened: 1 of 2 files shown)


vonture:
@kenrussell @zhenyao  PTAL

zhenyao:
I did a quick scan and think all the invalid format enums are actually invalid. Which one falls under "valid enum but invalid combination"?

Regardless, I think spec is ambiguous on this so I am OK with relaxing the errors. But can you also update the 2.0.0 / 1.0.3 / 1.0.2 snapshots?

kenrussell:
Relaxing these checks looks OK; agreed that it can be ambiguous whether INVALID_OPERATION or INVALID_ENUM are generated due to the need for both to be correct.

@vonture we can backport this to the earlier conformance snapshots (under conformance-suites/) for you; tell us whether you'd prefer that.

vonture:
I'm going to wait a bit on this and see if it's possible to update dEQP.  The spec is a bit ambiguous but I would prefer to only catch one kind of error here.

PR #2457, opened 2017-07-13
Add a regression test for crbug.com/740603

zhenyao:
@kenrussell PTAL


msc-:
The description is not helpful because crbug.com/740603 says "User is not allowed to view this issue.

kenrussell:
@msc- the bug will become publicly visible soon.

The new test LGTM.

PR #2458, opened 2017-07-17
Add a test for broadcasting gl_FragColor when main() returns

Oletus:
The test verifies that the value written to gl_FragColor is
broadcasted correctly regardless of any return statements in main().

This exposes a bug in ANGLE.

The patch includes refactoring of webgl-draw-buffers.html: Common
functionality with the new test is moved to a new utility file, and
simpleVertexShader from WebGLTestUtils is now used.


kainino0x:
lgtm

zhenyao:
Ah, I see the bug now you write a test.  Thanks for discovering it.

PR #2459, opened 2017-07-18
Add a multiple render targets test for discarding fragment

Oletus:
Add new multiple render target subtests that do a discard in the
fragment shader. The new subtests are added to both WebGL 1.0
EXT_draw_buffers tests and WebGL 2.0 core tests.

The new subtests are not expected to expose existing bugs.

The patch includes refactoring the WebGL 2.0 draw buffers test so that
it shares utility code with the EXT_draw_buffers test.


zhenyao:
Looks good.  Thanks Olli.

kenrussell:
Very nice Olli. LGTM after the fact.

PR #2462, opened 2017-07-19
Revise spec on untouched fragment outputs to match implementations

Oletus:
At least on ANGLE-based WebGL implementations, fragment color is
always written into the framebuffer if gl_FragColor or gl_FragData is
statically used in the shader.

Since this doesn't seem to be causing any problems, the easiest way
to fix this is to change the spec to match the implementation.

### DIFF
--- a/extensions/WEBGL_draw_buffers/extension.xml
+++ b/extensions/WEBGL_draw_buffers/extension.xml
@@ -74,7 +74,7 @@
         If the WEBGL_draw_buffers extension is enabled, but the fragment shader does not contain the <code>#extension GL_EXT_draw_buffers</code> directive to enable it, then writes to <code>gl_FragColor</code> are only written to <code>COLOR_ATTACHMENT0_WEBGL</code>, and not broadcast to all color attachments. In this scenario, other color attachments are guaranteed to remain untouched.
       </addendum>
       <addendum>
-        If a fragment shader writes to neither <code>gl_FragColor</code> nor <code>gl_FragData</code>, the values of
+        If a fragment shader statically uses neither <code>gl_FragColor</code> nor <code>gl_FragData</code>, the values of
         the fragment colors following shader execution are untouched.
 
         If a fragment shader contains the <code>#extension GL_EXT_draw_buffers</code> directive, all
@@ -180,6 +180,9 @@ interface WEBGL_draw_buffers {
     </revision>
     <revision date="2016/07/23">
       <change>Revised behavior of the same image is attached to more than one color attachment point.</change>
+    </revision>
+    <revision date="2017/07/19">
+      <change>Fragment outputs are not guaranteed to be untouched in case they are statically used, even if they are not written.</change>
     </revision>
   </history>
 </ratified>
--- a/extensions/WEBGL_draw_buffers/extension.xml
+++ b/extensions/WEBGL_draw_buffers/extension.xml
@@ -71,12 +71,9 @@
         The value of <code>gl_MaxDrawBuffers</code> must match <code>MAX_DRAW_BUFFERS_WEBGL</code> from the API if the extension is enabled in a WebGL context; otherwise, the value must be 1. Whether or not the extension is enabled with the <code>#extension GL_EXT_draw_buffers</code> directive in a shader does not affect the value of <code>gl_MaxDrawBuffers</code>. The value of <code>gl_MaxDrawBuffers</code> is a constant in the shader, and is guaranteed to be frozen at program link time. It is implementation-dependent whether it is frozen at shader compile time. (A consequence is that if a program is linked, and later the WEBGL_draw_buffers extension is enabled, the value of <code>gl_MaxDrawBuffers</code> seen by that program will still be 1.)
       </addendum>
       <addendum>
-        If the WEBGL_draw_buffers extension is enabled, but the fragment shader does not contain the <code>#extension GL_EXT_draw_buffers</code> directive to enable it, then writes to <code>gl_FragColor</code> are only written to <code>COLOR_ATTACHMENT0_WEBGL</code>, and not broadcast to all color attachments. In this scenario, other color attachments are guaranteed to remain untouched.
+        If the WEBGL_draw_buffers extension is enabled, but the fragment shader does not contain the <code>#extension GL_EXT_draw_buffers</code> directive to enable it, then the value of <code>gl_FragColor</code> is only written to <code>COLOR_ATTACHMENT0_WEBGL</code>, and not broadcast to all color attachments. In this scenario, other color attachments are guaranteed to remain untouched.
       </addendum>
       <addendum>
-        If a fragment shader statically uses neither <code>gl_FragColor</code> nor <code>gl_FragData</code>, the values of
-        the fragment colors following shader execution are untouched.
-
         If a fragment shader contains the <code>#extension GL_EXT_draw_buffers</code> directive, all
         <code>gl_FragData</code> variables (from <code>gl_FragData[0]</code> to <code>gl_FragData[MAX_DRAW_BUFFERS_WEBGL - 1]</code>)
         default to zero if no values are written to them during a shader execution.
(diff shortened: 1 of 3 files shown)


Oletus:
A more comprehensive review of the spec would be useful as previously discussed, but if it's not clear when we'll be able to do that, we could merge this if we want to make the spec and implementation to agree in the meantime.

zhenyao:
Thanks @Oletus for sorting this out. After some checking, I agree the original spec wording doesn't match ANGLE implementation. However, the wording in this still leaves undefined values. I see we have two options:

1) using the wording in this pull request, but add something like "if a fragment output variable is statically used, it is expected to be initialized to 0"

2) simply require all fragment output variables to be initialized to 0 no matter what, and change ANGLE implementation to match that.

Personally I would vote for 1) as it is least amount of coding change.

RafaelCintron:
FWIW, Edge initializes pixel shader output to zero, as well as all local variables.  We have not run into performance or compatibility problems doing so.   

I vote we do 2 and keep things simple.

Oletus:
I think I'd agree with Rafael here, if there's no compatibility impact as he says - it's not that much work to change this in ANGLE, and the code would be simpler in the end.

zhenyao:
No objection from me.

Oletus:
Amended the proposal. No tests yet, but I think I could add them to the pull request later.

extensions/WEBGL_draw_buffers/extension.xml
```diff
@@ -181,5 +178,8 @@ interface WEBGL_draw_buffers {
     <revision date="2016/07/23">
       <change>Revised behavior of the same image is attached to more than one color attachment point.</change>
     </revision>
+    <revision date="2017/07/20">
+      <change>Fragment outputs are always initialized even if they are not written.</change>
```
zhenyao: initialized to zero to be more specific.

msc-:
> Fragment outputs are always initialized even if they are not written

Doesn't this mean the implementation will write to every fragment, in all attached images in the multi-buffer case. You can't initialize something without writing to it. Won't this cause performance problems?

zhenyao:
That was the original reason we don't initialize them, but instead, leave them untouched.  However, according to @RafaelCintron comment above, there isn't any noticeable perf issues.

Oletus:
The color attachments that get written to can be controlled by calling drawBuffers, or by only attaching the framebuffer attachments the app wants to write to. I can't think of a realistic use case where you'd need per-pixel shader-level control for it. So I wouldn't expect performance to be a major problem.

grovesNL:
Shouldn't the specification allow implementations to decide whether to skip unnecessary initialization (especially if there is a potential performance impact)? @zhenyao's option 1 sounds preferable.

kenrussell:
Could we please move this pull request forward? A simpler solution is probably better in this case if there is no compatibility or significant performance impact, so I support just initializing all of the fragment outputs to zero like Edge does per @RafaelCintron 's comment.

specs/latest/1.0/index.html
```diff
@@ -4090,10 +4090,24 @@ <h2 class="no-toc">Table of contents</h2>
...
 <p>
-    If a fragment shader writes to neither <code>gl_FragColor</code> nor <code>gl_FragData</code>,
-    the values of the fragment colors following shader execution are untouched.
+    If a fragment statically uses neither <code>gl_FragColor</code> nor <code>gl_FragData</code>, it
+    behaves as if (0, 0, 0, 0) would be assigned to <code>gl_FragColor</code> in the beginning of
+    the shader.
 </p>
 
+<p>
+    Statically used fragment output variables default to zero if they are not written during a
```
jdashg: Easier to say: "All fragment shader outputs will default to zeros."

specs/latest/2.0/index.html
```diff
@@ -3325,13 +3325,16 @@ <h2 class="no-toc">Table of contents</h2>
...
+        If fragment output variables are not defined in an ESSL3 fragment shader, the values of the
+        fragment colors following shader execution are untouched.
     </p>
     <p>
-        All user-defined output variables default to zero if they are not written during a shader execution.
+        All user-defined fragment output variables default to zero if they are not written during a
+        shader execution.
+    </p>
+    <p>
+        Statically used built-in fragment output variables default to zero if they are not written
```
jdashg: If you s/user-defined/defined/ above, that would cover both these cases.

specs/latest/2.0/index.html
```diff
@@ -3325,13 +3325,16 @@ <h2 class="no-toc">Table of contents</h2>
     </div>
 
     <p>
-        If an ESSL1 fragment shader writes to neither <code>gl_FragColor</code> nor <code>gl_FragData</code>,
-        the values of the fragment colors following shader execution are untouched. If corresponding output
-        variables are not defined in an ESSL3 fragment shader, the values of the fragment colors following
-        shader execution are untouched.
+        If fragment output variables are not defined in an ESSL3 fragment shader, the values of the
+        fragment colors following shader execution are untouched.
```
jdashg: This is ok, but requires draw-time DrawBuffer patching.

GLES3 p185:
"If an OpenGL ES Shading Language 1.00 fragment shader writes to gl_-
FragColor or gl_FragData, DrawBuffers specifies the draw buffer, if any, into
which the single fragment color defined by gl_FragColor or gl_FragData[0]
is written. If an OpenGL ES Shading Language 3.00 fragment shader writes a user-
defined varying out variable, DrawBuffers specifies a set of draw buffers into
which each of the multiple output colors defined by these variables are separately
written. If a fragment shader writes to none of gl_FragColor, gl_FragData,
nor any user-defined output variables, the values of the fragment colors following
shader execution are undefined, and may differ for each fragment color."

I think it's better if we just INVALID_OPERATION if any color draw buffer is non-null+non-NONE.

Oletus: I'm fine with the INVALID_OPERATION if a shader doesn't write all color buffers that are enabled in the framebuffer. But since it is a breaking change so we'd have to watch out for regressions, and I would like to hear other opinions on it too. Maybe that would also be better to handle in a separate commit, with tests changed in sync with the spec.

jdashg: Right, I think we'd probably break some content by generating INVALID_OP whenever /any/ color buffer is bound to an unwritten frag output. However, it should be much less common for content to write to /none/ of the color outputs. I think we'd be safe making that INVALID_OP, particularly since this is WebGL 2 only.

specs/latest/1.0/index.html
jdashg: Yes. They will behave as if they default to zero, but that's different from us writing zeros into both of them. (which would break the mutually-exclusive static-use requirement)
The wording you have right now is technically correct, but requires more knowledge to piece together that this functionally guarantees that whichever output you choose, it will default to zero. Easier to just say they all default to zero.

specs/latest/2.0/index.html
```diff
@@ -3325,13 +3325,16 @@ <h2 class="no-toc">Table of contents</h2>
     </div>
 
     <p>
-        If an ESSL1 fragment shader writes to neither <code>gl_FragColor</code> nor <code>gl_FragData</code>,
-        the values of the fragment colors following shader execution are untouched. If corresponding output
```
msc-: Why have you removed the ESSL1-related lines? You can still use ESSL1 shaders in OpenGL ES 3 & WebGL 2.

jdashg: It is covered by the third paragraph added to this section. ("Statically used built-in fragment output variables default to zero[...]")

jdashg:
We can't just zero undeclared fragment outputs for essl3, because the frag output component type has to match the colorAttachmentN format component type.

I believe we error when we have this sort of component type mismatch, so I think it makes sense to also error if we have a non-null colorAttachmentN format component type for a frag output location not declared in the shader. (And zero all declared frag outputs)

lexaknyazev:
Does this PR change the behavior so that all statically used outputs are always written with zeros, or it just declares what implementations already do?

A possibility to leave some outputs untouched on fragment-shader level may unlock a few interesting techniques, e.g., using instancing to render cubemaps with fewer draw calls.

lexaknyazev:
> viewport/layer selection in vertex or geometry shaders

That's definitely too much to ask from ES 3.0.

I was thinking of a more simple case:
```
out vec4 pos_x_col;
out vec4 pos_y_col;
...
if (something) {
    pos_x_col = vec4(...) // pos_y_col remains untouched
} else {
    pos_y_col = vec4(...) // pox_x_col remains untouched 
}

jdashg:
ES3.0 says unwritten frag outputs have undefined values, hence this
proposed WebGL restriction.

kenrussell:
After extensive discussion and investigation the WebGL working group decided to make this area of undefined behavior in the OpenGL specifications a draw-time error in WebGL per #2780 .

Thanks @Oletus and everyone for your inputs on this. Closing.

PR #2463, opened 2017-07-19
Relax texture switch performance test from #2448 .

kenrussell:
This test assumed it will achieve a steady-state 30 FPS frame rate,
which is optimistic for automated test machines which build and run
debug versions of the browser. Reduce the expectation to 10 FPS.

### DIFF
--- a/sdk/tests/conformance/rendering/texture-switch-performance.html
+++ b/sdk/tests/conformance/rendering/texture-switch-performance.html
@@ -76,7 +76,7 @@ if (!gl) {
     ++frameCount;
 
     if (Date.now() - startTime > 5000) {
-      if (frameCount > 150) {  // 30FPS
+      if (frameCount > 50) {  // 10FPS
         testPassed("Texture switching did not significantly hurt performance");
       } else {
         testFailed("Texture switching significantly hurt performance");
--- a/sdk/tests/conformance/rendering/texture-switch-performance.html
+++ b/sdk/tests/conformance/rendering/texture-switch-performance.html
@@ -66,7 +66,7 @@ if (!gl) {
   var frameCount = 0;
 
   function draw() {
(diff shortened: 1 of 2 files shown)


kenrussell:
@tsherif FYI.

@zhenyao @vonture I'm going to merge this myself to improve the reliability of Chrome's bots, because the debug bots have been intermittently failing this test since it was integrated.

tsherif:
@kenrussell Is that high enough to catch cases where the problem is present (Mac with an AMD GPU)?

kenrussell:
@tsherif I'm pretty sure it will. I'll try to confirm before merging this.

kenrussell:
Unfortunately, this discriminator is too low, and the test starts to pass on the affected hardware :(

tsherif:
Do you guys allow for tests to be relative to a baseline on a given machine? I.e. you could run the render loop in a first pass for 2 seconds or so, without updating the sampler uniform, to get a baseline frame count, and then a second pass would test that the frame count while texture switching isn't less than some percentage of the baseline.

That might have a better chance of covering the range of machines you're dealing with.

kenrussell:
Good idea Tarek. No, I didn't try that. If you'd be willing to make that change that would be nice. In that case the test's going to be quite a bit more complex so I'd suggest refactoring it into sdk/tests/js/tests/texture-switch-performance.js and have both the WebGL 1.0 and 2.0 versions in there, and the .html files import it and trivially run the test with the right arguments.

In absence of that, increasing the workload to 800 draw calls per second allowed increasing the discriminant to 15 FPS, which is still 50% more than the affected hardware can achieve. I hope this will run well on the various machines.

Merging this now.

tsherif:
Sure, I can probably get you something in the next few days.

PR #2464, opened 2017-07-20
Expose TexSubImage2D bad dimensions bug

qjia7:
This change will expose bug http://crbug.com/733172.

### DIFF
--- a/sdk/tests/js/tests/texture-corner-case-videos.js
+++ b/sdk/tests/js/tests/texture-corner-case-videos.js
@@ -187,6 +187,20 @@ function generateTest(desc,
             debug("Checking upper left corner");
             wtu.checkCanvasRect(gl, 4, gl.canvas.height - 8, 2, 2, topColor,
                                 "shouldBe " + topColor, tolerance);
+
+            // Expose bug http://crbug.com/733172.
+            if (sourceSubRectangle) {
+                // Skip sub-rectangle tests for cube map textures for the moment.
+                if (bindingTarget == gl.TEXTURE_CUBE_MAP) {
+                    continue;
+                }
+                gl.texSubImage2D(targets[tt], 0, 0, 0,
+                                 sourceSubRectangle[2], sourceSubRectangle[3],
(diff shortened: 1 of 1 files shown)


qjia7:
@kenrussell @zhenyao PTAL. Thanks.

sdk/tests/js/tests/texture-corner-case-videos.js
```diff
@@ -187,6 +187,20 @@ function generateTest(desc,
...
+                if (bindingTarget == gl.TEXTURE_CUBE_MAP) {
+                    continue;
+                }
+                gl.texSubImage2D(targets[tt], 0, 0, 0,
+                                 sourceSubRectangle[2], sourceSubRectangle[3],
+                                 gl[pixelFormat], gl[pixelType], videoElement);
+            } else {
+                gl.texSubImage2D(targets[tt], 0, 0, 0, gl[pixelFormat], gl[pixelType], videoElement);
+            }
+            wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be no errors");
```
zhenyao: I am curious: if we add this error checking right after line 129, will it catch this bug? If yes, that seems a better place for this.

qjia7: No, it won't catch this bug. The error is because we use line 152  'gl.texImage2D(targets[tt], 0, gl[internalFormat], gl[pixelFormat], gl[pixelType], videoElement)' to initialize the texture. In chromium, that API will go gpu-gpu path which will upload the video visible size data. But TexSubImage will call GLContext()->TexSubImage which use video nature size.   But this video has natural size > visible size. So when call texture->ValidForTexture in TextureManager::ValidateTexSubImage, it will not meet the condition  max_x <= info.width and print bad dimentions.

line 128 TexImage2D and TexSubImage2D both use nature size. So it can't catch this error.

zhenyao: Thanks.  Then the current patch LGTM

PR #2465, opened 2017-07-25
Texture switch test compares to baseline

tsherif:
@kenrussell  Set up both the WebGL and WebGL 2 texture switch performance tests to first run for 2 seconds without switching textures to get a baseline frame count, then run again for 2 seconds while switching textures to get a test frame count to compare it to. 

I currently have it set up to fail if the test frame count drops below 80% of the baseline. This is well above what the affected AMD/Mac machine I have to test on can achieve (40-45%). The lowest I've seen from a passing machine is 88% on my Nexus 10, but there's still room to tweak the threshold if you need to.

The test run time and threshold are stored in RUNTIME and THRESHOLD variables near the start of the test so they're easy to tweak. The tests didn't become that much more complicated, so I didn't bother to refactor, but let me know if you still want to do that.

### DIFF
--- a/sdk/tests/conformance/rendering/texture-switch-performance.html
+++ b/sdk/tests/conformance/rendering/texture-switch-performance.html
@@ -99,7 +99,7 @@ if (!gl) {
     ++testFrameCount;
 
     if (Date.now() - testStartTime > RUNTIME) {
-      var perfString = " - achieved " + testFrameCount + " frames in " + ((Date.now() - testStartTime) / 1000.0) + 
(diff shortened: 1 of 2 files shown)


tsherif:
@kenrussell Ah, didn't realize there were linting errors in my PR. This should be good to go now.

kenrussell:
Thanks Tarek for this improvement. It looks good. On my unaffected laptop it routinely achieves 98 percent of the baseline score, so I think the thresholds are OK. I'll update the tests if it looks like they're flaky on our bots.

PR #2467, opened 2017-08-01
Moved WebGL 2.0.0 to final state per ratification vote by Khronos

kenrussell:
Promoters on June 30, 2017.

Synced with snapshotted version.

### DIFF
--- a/specs/2.0.0/index.html
+++ b/specs/2.0.0/index.html
@@ -4,7 +4,7 @@
 <head>
     <meta http-equiv="content-type" content="text/html; charset=utf-8">
     <title>WebGL 2 Specification</title>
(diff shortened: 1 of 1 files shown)


kenrussell:
Bookkeeping after ratification vote. Merging myself.

dmikis:
https://www.khronos.org/registry/webgl/specs/latest/2.0/ still says "Working Draft". Does it also need to be updated?

kenrussell:
Correct. https://www.khronos.org/registry/webgl/specs/2.0.0/ is the ratified version. https://www.khronos.org/registry/webgl/specs/latest/2.0/ is the ongoing editor's draft.

PR #2469, opened 2017-08-01
Test blits of R11F_G11F_B10F color framebuffers.

kenrussell:
This exposes a bug in Chrome, http://crbug.com/751206, affecting blits
between user framebuffers on macOS when the default back buffer is RGB
and non-antialiased.


kenrussell:
Note: also enhanced webgl-test-utils.js to support checking the color of canvases with floating-point back buffers.

This acts as a regression test for a bug affecting PlayCanvas' After the Flood demo. Note that this test already passes in Firefox.

Merging myself to get this necessary test into the repository sooner. Feedback after the fact is welcome.

PR #2470, opened 2017-08-03
Rename OES_EGL_image_external to WEBGL_video_texture

xingri:
Rename the extension so that it specifically covers the case of binding
video textures to WebGL textures.

### DIFF
--- a/extensions/proposals/WEBGL_video_texture/extension.xml
+++ b/extensions/proposals/WEBGL_video_texture/extension.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<proposal href="proposals/OES_EGL_image_external/">
-  <name>OES_EGL_image_external</name>
+<proposal href="proposals/WEBGL_video_texture/">
+  <name>WEBGL_video_texture</name>
 
   <contact> <a href="https://www.khronos.org/webgl/public-mailing-list/">WebGL
   working group</a> (public_webgl 'at' khronos.org) </contact>
@@ -20,21 +20,21 @@
   <overview>
     <mirrors href="https://www.khronos.org/registry/gles/extensions/OES/OES_EGL_image_external.txt"
              name="OES_EGL_image_external">
-      <addendum>Defines a new texture target <code>TEXTURE_EXTERNAL_OES</code>.</addendum>
-      <addendum>Provides a mechanism for binding <code>HTMLVideoElement</code>'s EGLImage to external texture targets.</addendum>
-      <addendum>Provides time of frame, texture width and height of <code>HTMLVideoElement</code>'s EGLImage.</addendum>
+      <addendum>Defines a new texture target <code>TEXTURE_VIDEO_IMAGE</code>.</addendum>
+      <addendum>Provides a mechanism for binding <code>HTMLVideoElement</code> stream to video texture targets.</addendum>
+      <addendum>Provides time of frame, texture width and height of <code>HTMLVideoElement</code>'s texture binding.</addendum>
     </mirrors>
 
     <features>
-      <feature>Add support for <code>OES_EGL_image_external</code> texture
+      <feature>Add support for <code>WEBGL_video_texture</code>
       binding of HTMLVideoElement.</feature>
 
-      <glsl extname="OES_EGL_image_external">
+      <glsl extname="WEBGL_video_texture">
         <stage type="fragment"/>
-        <type name="samplerExternalOES"/>
+        <type name="samplerVideoWEBGL"/>
 
         <function name="texture2D" type="vec4">
-          <param name="sampler" type="samplerExternalOES"/>
+          <param name="sampler" type="samplerVideoWEBGL"/>
 
           <param name="coord" type="vec2"/>
         </function>
@@ -52,13 +52,11 @@ interface WebGLVideoFrameInfo {
 };
 
 [NoInterfaceObject]
-interface OES_EGL_image_external {
-    const GLenum TEXTURE_EXTERNAL_OES             = 0x8D65;
-    const GLenum SAMPLER_EXTERNAL_OES             = 0x8D66;
-    const GLenum TEXTURE_BINDING_EXTERNAL_OES     = 0x8D67;
-    const GLenum REQUIRED_TEXTURE_IMAGE_UNITS_OES = 0x8D68;
+interface WEBGL_video_texture {
+    const GLenum TEXTURE_VIDEO_IMAGE             = 0x851D;
+    const GLenum SAMPLER_VIDEO_IMAGE             = 0x8B61;
 
-    [RaisesException] WebGLVideoFrameInfo EGLImageTargetTexture2DOES(
+    [RaisesException] WebGLVideoFrameInfo VideoElementTargetVideoTexture(
         GLenum target, HTMLVideoElement video);
 };
   </idl>
@@ -67,11 +65,11 @@ interface OES_EGL_image_external {
 
     <p> This a fragment shader that samples a video texture.</p>
     <pre>
-    #extension GL_OES_EGL_image_external : require
+    #extension GL_WEBGL_video_texture : require
     precision mediump float;
     varying vec2 v_texCoord;
 
-    uniform samplerExternalOES uSampler;
+    uniform samplerVideoWEBGL uSampler;
 
     void main(void) {
       gl_FragColor = texture2D(uSampler, v_texCoord);
@@ -84,11 +82,11 @@ interface OES_EGL_image_external {
     var videoTexture = gl.createTexture();
 
     function update() {
-        var ext = gl.getExtension('OES_EGL_image_external');
+        var ext = gl.getExtension('WEBGL_video_texture');
         if(ext !=== null){
-            gl.bindTexture(ext.TEXTURE_EXTERNAL_OES, videoTexture);
-            ext.EGLImageTargetTexture2DOES(ext.TEXTURE_EXTERNAL_OES, videoElement);
-            gl.bindTexture(ext.TEXTURE_EXTERNAL_OES, null);
+            gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE, videoTexture);
+            ext.VideoElementTargetVideoTexture(ext.TEXTURE_VIDEO_IMAGE, videoElement);
+            gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE, null);
         }
     }
 
@@ -100,7 +98,7 @@ interface OES_EGL_image_external {
         gl.vertexAttribPointer(vertexPositionAttribute, 3, gl.FLOAT, false, 0, 0);
 
         gl.activeTexture(gl.TEXTURE0);
-        gl.bindTexture(ext.TEXTURE_EXTERNAL_OES, videoTexture);
+        gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE, videoTexture);
         gl.uniform1i(gl.getUniformLocation(shaderProgram, "uSampler"), 0);
 
         gl.drawArrays(gl.TRIANGLE_STRIP, 0, 4);
@@ -131,6 +129,11 @@ interface OES_EGL_image_external {
       <change>Change EGLImageTargetTexture2DOES to be called at every WebGL rendering cycle.</change>
       <change>Add VideoFrameInfo interface.</change>
       <change>Change EGLImageTargetTexture2DOES to return VideoFrameInfo as a currently mapped frame.</change>
+    </revision>
+    <revision date="2017/08/03">
+      <change>Change Extension name to WEBGL_video_texture for abstracion of OES_EGL_image_external extension.</change>
+      <change>Define new sampler and texture type, TEXTURE_VIDEO_IMAGE and SAMPLER_VIDEO_IMAGE.</change>
+      <change>Change EGLImageTargetTexture2DOES to VideoElementTargetVideoTexture.</change>
     </revision>
   </history>
 </proposal>


extensions/proposals/WEBGL_video_texture/extension.xml
```diff
@@ -52,13 +52,11 @@ interface WebGLVideoFrameInfo {
...
 
 [NoInterfaceObject]
-interface OES_EGL_image_external {
-    const GLenum TEXTURE_EXTERNAL_OES             = 0x8D65;
-    const GLenum SAMPLER_EXTERNAL_OES             = 0x8D66;
-    const GLenum TEXTURE_BINDING_EXTERNAL_OES     = 0x8D67;
-    const GLenum REQUIRED_TEXTURE_IMAGE_UNITS_OES = 0x8D68;
+interface WEBGL_video_texture {
+    const GLenum TEXTURE_VIDEO_IMAGE             = 0x851D;
+    const GLenum SAMPLER_VIDEO_IMAGE             = 0x8B61;
```
kenrussell: Where did these enum values come from? I think that they need to be allocated out of a block that was given to the WebGL working group some time ago. This would imply that the names and values should be:

TEXTURE_VIDEO_IMAGE_WEBGL = 0x9248
SAMPLER_VIDEO_IMAGE_WEBGL = 0x9249

(The last enum allocated out of this range was MAX_CLIENT_WAIT_TIMEOUT_WEBGL in the WebGL 2.0 spec.)

extensions/proposals/WEBGL_video_texture/extension.xml
```diff
@@ -20,21 +20,21 @@
   <overview>
     <mirrors href="https://www.khronos.org/registry/gles/extensions/OES/OES_EGL_image_external.txt"
              name="OES_EGL_image_external">
-      <addendum>Defines a new texture target <code>TEXTURE_EXTERNAL_OES</code>.</addendum>
-      <addendum>Provides a mechanism for binding <code>HTMLVideoElement</code>'s EGLImage to external texture targets.</addendum>
-      <addendum>Provides time of frame, texture width and height of <code>HTMLVideoElement</code>'s EGLImage.</addendum>
+      <addendum>Defines a new texture target <code>TEXTURE_VIDEO_IMAGE</code>.</addendum>
```
kenrussell: I think that we should leave open the possibility that on some platforms, the TEXTURE_VIDEO_IMAGE(_WEBGL) target is actually aliased to TEXTURE_2D. At least, this will prevent developers from assuming that it is always a different target. What do you think? If you agree, could you update the spec to allow this possibility?

kenrussell:
@msc- @jdashg @kainino0x @RafaelCintron @grorg @zhenyao @Oletus Any feedback you have on this generalization is appreciated. This extension is important to improve the performance of accessing video textures in WebGL, so we should be careful to get it right.

extensions/proposals/WEBGL_video_texture/extension.xml
```diff
@@ -52,13 +52,11 @@ interface WebGLVideoFrameInfo {
...
-    const GLenum TEXTURE_EXTERNAL_OES             = 0x8D65;
-    const GLenum SAMPLER_EXTERNAL_OES             = 0x8D66;
-    const GLenum TEXTURE_BINDING_EXTERNAL_OES     = 0x8D67;
-    const GLenum REQUIRED_TEXTURE_IMAGE_UNITS_OES = 0x8D68;
+interface WEBGL_video_texture {
+    const GLenum TEXTURE_VIDEO_IMAGE             = 0x851D;
+    const GLenum SAMPLER_VIDEO_IMAGE             = 0x8B61;
 
-    [RaisesException] WebGLVideoFrameInfo EGLImageTargetTexture2DOES(
+    [RaisesException] WebGLVideoFrameInfo VideoElementTargetVideoTexture(
```
kenrussell: I think we need to allow the possibility that this call may fail. This would happen, for example, if the target video were not actually being decoded on the GPU. We don't want to emulate the semantics of this extension if it isn't actually getting a GPU texture from the video element. Instead the application should have to call texImage2D as they do now. Could you please think through how the fallback would work?

msc-:
I'm sick. I'll review when I've recovered enough.

There is one thing that I want to raise awareness of now though, having seen @zhenyao's Iframe (he needs a better name!!!) presentation at the BoF. Video texture and Iframe textures should use the same API for updating and latching the textures. They're both examples of dynamic textures.

kenrussell:
Thanks for your feedback @msc- . It's a fair point that the same API should be used for these two cases, but the texture targets and sampler types are going to have to be abstracted differently. They will use different texture types under the hood depending on the platform (TEXTURE_2D, TEXTURE_RECTANGLE_ARB, TEXTURE_EXTERNAL_OES).

extensions/proposals/WEBGL_video_texture/extension.xml
zhenyao: I think ultimately we need to specify on each platforms this new target is mapped to an user visible target, so developers know they need to rebind a previous texture to that target because bindTexture(TEXTURE_VIDEO_IMAGE) will destroy that binding. Could you add a TODO here to make sure we clarify this when we implement it on all platforms so we can finalize it?

extensions/proposals/WEBGL_video_texture/extension.xml
zhenyao: What about Promise<WebGLVideoFrameInfo> bindTextureSource(GLenum target, HTMLVideoElement video)?

This will align well with the IFrameToTexture extension proposal I am working on (Mark, feel free to suggest a better name).

So you can return the promise when the first frame of the video is ready, or reject the promise if something goes wrong or video isn't hardware decoded.

zhenyao: I think I misunderstood the proposal a bit (thanks to kbr to clarify it for me).

It sounds to me this is more like a texImage2D call, so here is my suggestion:
Promise WebGLVideoFrameInfo  texImageVideoImageWEBGL(GLenum target, HTMLVideoElement video)?

kenrussell: I think that for this update we should stick with this synchronous API, which latches in the current frame of the video (it doesn't request that it be produced). It's already hard enough to think about exactly how this is going to be ported to other platforms. We can consider merging your iframe-to-texture proposal with this one after we've prototyped this on more platforms. Does that sound OK?

zhenyao: I am a bit confused. I thought after this call returns, the texture is guaranteed to filled with the video frame as specified by the time in WebGLVideoFrameInfo. You are saying that's not the case? Then how do we know when to use the texture?

kenrussell: Yes, after this call returns, the texture is guaranteed to be filled with the video frame, and the WebGLVideoFrameInfo contains the metadata for that frame. It's a synchronous API.

The Promise-based API you suggested above would be asynchronous, requiring the developer to wait for the Promise to be resolved before they could work with the texture.

zhenyao: Talked with @kenrussell offline to understand why this can be synchronous in a multi-process situation like Chrome.  No objection from me to keep it synchronous then.  Still, I think a texImage* name is better than the current name.

extensions/proposals/WEBGL_video_texture/extension.xml
xingri: Thanks for the information. I will update the proposal.

extensions/proposals/WEBGL_video_texture/extension.xml
xingri: @zhenyao, @kenrussell  <  Thanks for the feedbacks.
We will consider new name starts with texImage and update it soon.

zhenyao:
Please wait for @msc- to take a look and provide feedback, so you can address all the raised issues together.

extensions/proposals/WEBGL_video_texture/extension.xml
msc-: I think it better that developers assume it is a different target than vice versa. For consistency across implementations and consistency of applications it is probably better that the video target is always treated as being different from texture 2D.

> I think ultimately we need to specify on each platforms this new target is mapped to an user visible target, so developers know they need to rebind a previous texture to that target because bindTexture(TEXTURE_VIDEO_IMAGE) will destroy that binding.

I'm not sure exactly what you are getting at here. bindTexture(...) of an existing texture to a target of a different  type would result in an INVALID_OPERATION error. Perhaps we are agreeing with each other.

zhenyao: For example, textureA is bound to TEXTURE_2D.

Now on a platform the underlying target for TEXTURE_VIDEO_IMAGE is TEXTURE_2D, so bindTexture(TEXTURE_VIDEO_IMAGE, textureB) will change TEXTURE_2D binding to textureB.

Now if a user tries to modify textureA by calling texImage2D(TEXTURE_2D, ...), then instead of modifying textureA, it's modifying textureB.

msc-:
@zhenyao, I hadn't realized you are using promises in your Iframe API. I would think timing is too tight for that when using video, if you are making any attempt to synchronize audio. Even in the example you showed, which was texturing a YouTube page, I wonder ...

Since we're now moving away from mirroring OES_EGL_image_external, I have to say that I like the model provided by Android's [SurfaceTexture](https://developer.android.com/reference/android/graphics/SurfaceTexture.html) for handling this stuff.  I'd like to see WebGL do something along those lines. I think it could work fine for WebGL with some minor changes to fit the Web platform.

For example we could create an ElementTexture object (I was going to suggest DynamicTexture but ...), having the same API as SurfaceTexture, and have ways to attach to it or create from it various types of HTML element (video, canvas, iframe, etc.) which would produce the texture content when updateTexImage is called.

xingri:
@msc- , thanks for the feedback. I would like to point out trade-off between simplification by abstraction and complexity by giving details. Please check following URL:
<https://yt-dash-mse-test.commondatastorage.googleapis.com/demo-player/dash-player-020416.html?manifest_url=assets/spherical_vp9.mpd&sphericaloestexture=true&disableScreenSizeCheck=true>

YouTube already has finished implementation by this proposal and will be published in a month.
I think it is possible because of simplicity of proposal. The developer at YouTube did not even ask a question when we shared the proposal and then just shared the above URL. 

We also can convert three.js library within an minutes. Also we have not get any more APIs request from content provider side.  I think providing find-grinned APIs also important but now we want to focus on simplification version first.

msc-:
I got these errors:

```
Missing webgl: undefined
ERROR: Shader compile reported
ERROR: 0:1: 'GL_OES_EGL_image_external' : extension is not supported
ERROR: 0:2: 'samplerExternalOES' : unsupported type 
ERROR: 0:4: 'texture2D' : no matching overloaded function found 
ERROR: 0:4: 'xyz' :  field selection requires structure or vector on left hand side 
ERROR: 0:4: 'constructor' : not enough data provided for construction 
```

I'm using Firefox on a Mac.

xingri:
@msc-  That's quite right. Since we have not apply our changes on upstream. We will do the work on chromium first by next month.

msc-:
I don't see how

```
gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE, videoTexture);
ext.VideoElementTargetVideoTexture(ext.TEXTURE_VIDEO_IMAGE, videoElement);
gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE, null);
```

is simpler than

```
    ext.updateTexImage();
```

plus I suspect that on some EGL implementations that `EGLImageTargetTexture2DOES` is quite a heavyweight operation.

The biggest issue is that there is no interlocking between producer and consumer of the video. How does the video decoder know it is not supposed to update the tex image? How does it know when the render has finished? Your API implies a copy at the time of the call to `VideoElementTargetVideoTexture` not true streaming video.

xingri:
@msc- If changing to updateTexImge is only your requirement, I would definitely accept it. 
Also,  EGLImageTargetTExture2DOES is not heavy since it just connect EGLImage Texture to WebGL Context which was done by copying bytes by TEXTURE_2D at every frame.

extensions/proposals/WEBGL_video_texture/extension.xml
xingri: I agree on @msc- 's idea. I would better to use as it as different from texture 2D.

msc-:
Sorry about the length of this. Phew!

What troubles me most about this proposal is that it does not specify the intended behavior for handling the video. This means implementations will vary and there is nothing to write conformance tests against. Some of the open questions are:

- Is the video single buffered or multi buffered?
- What exactly should happen when `VideoElementTargetVideoTexture` is called?
- What happens to the video between calls to `VideoElementTargetVideoTexture`?
    - Does the decoder go idle?
    - Does it keep running?
    - Do frames get thrown away?
- How do you ensure the video is not updated during rendering? Is it up to the application? This depends on the answers to some of the preceding questions.
- What is the orientation of the image? I.e. Does sampling at (u,v) = (0,0) return the logical lower left or upper left of the video image? 

Before the renaming one might have consulted GL_OES_EGL_image_external for answers. It would not though answer all the questions. One would find that what is now called `VideoElementTargetVideoTexture` is being overloaded and no explanation is provided of the additional behavior. I am talking here about its use in `update()`.

In GL_OES_EGL_image_external the call `EGLImageTargetTexture2DOES` is not used in the inner loop of an application. It is a one time call used to attach an external image to a GL texture object. Thereafter any changes made to that image by the external producer are usually visible when the texture is sampled. Updating the image requires invoking its producer, e.g, an OpenVG renderer. The application has to handle synchronization between production and sampling. In other words the image is single buffered. None of this helps with understanding how the call works as part of the update loop and with a video producer.

These concerns have nothing to do with the shape of the API.

> Also, EGLImageTargetTExture2DOES is not heavy since it just connect EGLImage Texture to WebGL Context which was done by copying bytes by TEXTURE_2D at every frame.

I think something like "previously" is missing from "which was done". If I read "was" as "is" then I am completely missing the point of this extension which, I thought, is to avoid copying. Here my concern is that "just connect" may indeed be a heavyweight operation on some implementations. Since the OES extension intends it to be a one-time thing there won't have been much pressure to optimize it.

Regarding my proposal about SurfaceTexture, it is almost the same API as proposed here, except
-  It has separate functions for making the connection between video and texture and for updating the texture image. IMO this makes the operation much clearer.
- It is hides, or does not require, the unbinding and binding that ensures changes are visible.  This makes it a bit simpler to use and makes it compatible with implementation on top of EGLStreams.
- It handles the video image orientation.

In addition it supports single-buffered use and frame available notifications. These functions could be put in an additional extension layered on a base extension that has just the functionality proposed here.

xingri:
@msc- , first of all,  thanks for the review. 

> - Is the video single buffered or multi buffered? 

It depends on browser implementation but we used multiple buffers.

> - What exactly should happen when VideoElementTargetVideoTexture is called? 

It passes EGLImage to WebGL Context

>3. What happens to the video between calls to VideoElementTargetVideoTexture?
>- Does the decoder go idle?
>- Does it keep running? 
>- Do frames get thrown away? 

- Yes.
- Frame is kept while we are rendering
- No, decoder keeps decoding

>4. How do you ensure the video is not updated during rendering? Is it up to the application? This depends on the answers to some of the preceding questions.

Unless application call VideoElementTargetVideoTexture, the application can keep the current frame.

>5. What is the orientation of the image? I.e. Does sampling at (u,v) = (0,0) return the logical lower left or upper left of the video image?  

Same as texImage2D of WebGL specification(The first pixel transferred from the source to the WebGL implementation corresponds to the upper left corner of the source. )

> In GL_OES_EGL_image_external the call EGLImageTargetTexture2DOES is not used in the inner loop of an application.  In GL_OES_EGL_image_external the call EGLImageTargetTexture2DOES is not used in the inner loop of an application. It is a one time call used to attach an external image to a GL texture object. Thereafter any changes made to that image by the external producer are usually visible when the texture is sampled. Updating the image requires invoking its producer, e.g, an OpenVG renderer. The application has to handle synchronization between production and sampling. In other words the image is single buffered. None of this helps with understanding how the call works as part of the update loop and with a video producer.

That's not true.  We modified proposal per your request in last January . At each time the API called, it would update the current Video Frame's resolution info & time stamp.  Also we can synchronize the EGLImage mapped to video frame.

_Revision 2, 2017/01/10_

_Change EGLImageTargetTexture2DOES to be called at every WebGL rendering cycle._
_Add VideoFrameInfo interface._
_Change EGLImageTargetTexture2DOES to return VideoFrameInfo as a currently mapped frame._

extensions/proposals/WEBGL_video_texture/extension.xml
tharkum: May be need also add the following enum values?

TEXTURE_BINDING_VIDEO_IMAGE_WEBGL = 0x924A
glGet() params returns a single value, the name of the texture currently bound to the target GL_TEXTURE_VIDEO_IMAGE_WEBGL

REQUIRED_TEXTURE_VIDEO_IMAGE_UNITS_WEBGL = 0x924B
For platforms which will use underlying TEXTURE_EXTERNAL_OES for new texture target we must know
how much texture image units needed for video texture.

According to https://www.khronos.org/registry/OpenGL/extensions/OES/OES_EGL_image_external.txt
Each TEXTURE_EXTERNAL_OES texture object may require up to 3 texture image units for each texture unit to which it is bound.

tharkum:
@msc- I will add some comments to @xingri answer.

Initial version of "OES_EGL_image_external" extension was expected to working in the following manner:

- create texture
- bind texture (TEXTURE_EXTERNAL_OES target)
- attach an video image stream to texture (EGLImageTargetTexture2DOES)

- rendering in loop

I agree "EGLImageTargetTexture2DOES" not perfect name for it, @zhenyao version looks much better - bindTextureSource(GLenum target, HTMLVideoElement video).

Once connection betweeen video and texture was established we can use this texture in internal WebGL operations.
In this case texture image update happens hiddenly for application level - it is responsibility of browser.

    var videoElement = document.getElementById("video");
    var videoTexture = gl.createTexture();

    function setup() {
        var ext = gl.getExtension('OES_EGL_image_external');
        if(ext !=== null){
            gl.bindTexture(ext.TEXTURE_EXTERNAL_OES, videoTexture);
            ext.EGLImageTargetTexture2DOES(ext.TEXTURE_EXTERNAL_OES, videoElement);
        }
    }

    function render() {
        ....
        gl.drawArrays(gl.TRIANGLE_STRIP, 0, 4);
    }

And rendering cycle:

    setup();
    while (true) {
       processInput();
       render();
    }


Before external texture wil be sampled browser will update (latching new frame from video stream which based on EGLStream or SurfaceTexture-like solution) - on gl.drawArrays() execution.

It is allowing to use one video decoder stream (which could be already existed for rendering video) for many attached texture and when they will be sampled they always will be up-to-date with current video stream state.

For implementation in chromium it was made through mailbox mechanism:
many client texture points to one service texture which connected to existed video stream (EGLStream/SurfaceTexture).

This idea was implemented on top of chromium and couldn't be applied to another browsers (firefox/...) in the same manner.

In summary, this extension works in the same way as your SurfaceTexture proposal but without "new frame" notification and "update" method on application level.

msc-:
@xingri, @tharkum thank you for the information. My point it that this kind of information needs to be in the spec.

>-  No, decoder keeps decoding
>-  Yes.
>-  Frame is kept while we are rendering

How many frames are kept? If every frame is kept but rendering takes a long time, then the video is going to fall behind the audio. How this is handled needs to be described in the spec.

I thought of another question. If an application does

```javascript
    videoElement.currentTime = someTime;
    gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE, videoTexture);
    ext.VideoElementTargetVideoTexture(ext.TEXTURE_VIDEO_IMAGE, videoElement)
    gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE, null);
    gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE, videoTexture);
    gl.drawArrays(gl.TRIANGLE_STRIP, 0, 4);
```

Is it guaranteed that the app wiil see the frame corresponding to "someTime"? One more example of the detail the spec needs.

> That's not true. We modified proposal per your request in last January . At each time the API called, it would update the current Video Frame's resolution info & time stamp. Also we can synchronize the EGLImage mapped to video frame.

What I wrote is true. I was describing GL_OES_EGL_image_external which does not describe the functionality added by the modified proposal. This spec does not describe it either, the issue I am raising.

> Before external texture wil be sampled browser will update (latching new frame from video stream which based on EGLStream or SurfaceTexture-like solution) - on gl.drawArrays() execution.

This is s non-starter. Browsers won't want to be in the business of figuring out which draw call is going to sample a video texture and applications won't want the video advanced multiple frames (1 per draw call) each time they draw their scene. This is one of the reasons I maintain that a per-GL-frame explicit update call is needed.

Here is my suggestion:

```javascript
    var videoElement = document.getElementById("video");
    var videoTexture = gl.createTexture();
    var elementTexture;

    function setup() {
        var ext = gl.getExtension('OES_EGL_image_external');
        if (ext !=== null){
            // Attaches videoElement to videoTexture and binds videoTexture to
            // TEXTURE_EXTERNAL_OES.
            elementTexture = new ext.ElementTexture(videoElement, videoTexture);
        } else {
            bail();
        }
    }

    function render() {
        ....
        gl.drawArrays(gl.TRIANGLE_STRIP, 0, 4);
    }
```

And rendering cycle:

```javascsript
    setup();
    while (true) {
       processInput();
       elementTexture.updateTexImage();
       render();
    }
```

The nice thing about having the ElementTexture object is that functions and properties can be added now or in the future to support applications that want to do more than simply playback a video. For example querying info about the currently latched frame and frame available notifications.  It can also be easily extended to support using other types of HTML element as the image producer.

```javascript
    var canvasElement = document.getElementById("canvas");
    var canvasTexture = gl.createTexture();
    elementTexture = new ext.ElementTexture(canvasElement, canvasTexture);
```

or

```javascript
    var iFrameElement = document.getElementById("myIFrame");
    var iFrameTexture = gl.createTexture();
    elementTexture = new ext.ElementTexture(iFrameElement, iFrameTexture);
```

I don't care for the name ElementTexture. The name is too similar but the object very different from, the other *Texture variables.

msc-:
>>  What is the orientation of the image? I.e. Does sampling at (u,v) = (0,0) return the logical lower left or upper left of the video image?
>
> Same as texImage2D of WebGL specification(The first pixel transferred from the source to the WebGL implementation corresponds to the upper left corner of the source. )

The WebGL spec does not say that. In OpenGL & OpenGL ES the first pixels transferred to the implementation will be what is returned by sampling at the UV coordinate (0,0). WebGL does not change that. Whether it is upper or lower left corner of an image depends entirely on the image data being uploaded. This spec. needs to say which it is for video, or provide a query like SurfaceTexture's `getTransformMatrix`.

WebGL has a flipY pixel unpack parameter that allows apps to request flipping during upload. Does that affect the result for external textures? It needs to be stated in the spec. It would be counter-intuitive if it did, because there is, in theory, no pixel unpacking going on here.

tharkum:
@msc - Thanks for reply.

According to your suggestions and using Android SurfaceTexture as basis  we could modify current extension to the following manner:

1. Rename it to "WEBGL_element_texture"

2. Define WebIDL 

```
[
] callback interface ElementTextureListener {
    void onFrameAvailable();
};

[ 
    NoInterfaceObject
] interface ElementTexture {
    double getTImestamp();
    Float32Array getTransformMatrix();
    void updateTexImage();
    void setElementTextureListener(ElementTextureListener? listener);
};
```
 + option 1

```
[ 
    NoInterfaceObject
] interface WEBGL_element_texture {
    typedef (HTMLCanvasElement or
                  HTMLVideoElement) TextureSource;

    const GLenum TEXTURE_ELEMENT_WEBGL = 0x9248;
    const GLenum SAMPLER_ELEMENT_WEBGL = 0x9249;
    const GLenum TEXTURE_BINDING_ELEMENT_WEBGL = 0x924A;
    const GLenum REQUIRED_TEXTURE_ELEMENT_IMAGE_UNITS_WEBGL = 0x924B;

    [RaisesException] ElementTexture? createElementTexture(WebGLTexture? texture, TextureSource source);
};
```
OR

+ option 2

```
[ 
    NoInterfaceObject
] interface WEBGL_element_texture {
    const GLenum TEXTURE_ELEMENT_WEBGL = 0x9248;
    const GLenum SAMPLER_ELEMENT_WEBGL = 0x9249;
    const GLenum TEXTURE_BINDING_ELEMENT_WEBGL = 0x924A;
    const GLenum REQUIRED_TEXTURE_ELEMENT_IMAGE_UNITS_WEBGL = 0x924B;
};
```
and also define new extension "WEBGL_video_texture" which will extend our base extension through partial interface (not sure that it is appropriate option)

```
[ 
    NoInterfaceObject
] interface VideoTexture {
    ... (add methods or members if needed)
};
VideoTexture implements ElementTexture;

partial interface WEBGL_element_texture {
    [RaisesException] VideoTexture? createVideoTexture(WebGLTexture? texture, HTMLVideoElement element);
};
```

At least option 2 will allow to add "WEBGL_iframe_texture" or "WEBGL_canvas_texture" easily in future.

Thanks for any advice.

msc-:
@tharkum thanks for putting together the IDL. It accurately reflects what I have in mind. Whatever the API, the intended operation needs to be completely specified, as I said above. Completing a detailed spec.  will help decide whether option 1 or 2 is better. I'm inclined toward option 2 at present but have no strong feeling.

I'm also inclined towards calling it `HTMLElementTexture` rather than `ElementTexture`.

In seeking an answer to this question

1. Does the timestamp have any relevance for an element not derived from HTMLMediaElement?

 I have discovered that there is draft for [Media Capture from DOM Elements](https://w3c.github.io/mediacapture-fromelement/#idl-def-CanvasCaptureMediaStream) that adds a `captureStream` method to HTMLMediaElements and HTMLCanvasElements. I don't know if there is any plan to support this on HTMLIFrameElements but it seems like the `MediaStream` returned by `captureStream` is the ideal thing to use as the source for these textures. It is probably too far out in the future for those seeking a video texture solution today so we should keep with the element as the texture source. A browser could start to use `captureStream` underneath when it implements that function.

Speaking of the timestamp, is it the desired presentation time of the frame? During discussions on WEBGL_dynamic_texture it emerged that on Apple systems video frames have their required presentation time associated with them.

Other questions

2. Do we need to add another query for the size of the video frame?
    - Several people have said they need to know this info. I do not know the intended use. If the use is so they can use integer texture coordinates, i.e. 0 to video{Height,Width}, then `getTransformMatrix` could provide a suitable transform, otherwise the additional query is needed. If we had a function returning an object with both timestamp and size, I'd be worried about the object generation.

3. How can we avoid `getTransformMatrix` generating objects?

4. How can the application find out if it is getting behind the video? Do we need another function for that?

5. Do we need a query for the frame rate so the application can budget its time?

msc-:
I am going on vacation on Monday and will be gone 3 weeks. If anyone has questions for me needing an immediate answer, best ask now.

tharkum:
@msc- Thanks for your comment and sorry for late response.

> I'm also inclined towards calling it HTMLElementTexture rather than ElementTexture.

Yeah, it sounds more accurate.

> 1. Does the timestamp have any relevance for an element not derived from HTMLMediaElement?

For ElementTexture delivered from Canvas/IFrame the timestamp is normally monotonically increasing
(starting point of the timestamp should be equal to zero), but it may be reset when the position is set
by Media.
Anyway we can move `getTimestamp` to partial interface for video texture, but in current version it could be useful for determining time offsets between subsequent frames for Canvas/IFrame.

> Speaking of the timestamp, is it the desired presentation time of the frame? 

The timestamp will be associated with the texture image set by the most recent call to `updateTexImage` (same as the transform matrix).

> 2. Do we need to add another query for the size of the video frame?

I don't know how it will be helpful. As you mentioned `video{Width,Height} + getTransformMatrix` will be enough. Although due some platform restrictions texture image size could be less then video size. 
So may we could keep it in mind and add it later if it will be really needed. 

> 3. How can we avoid getTransformMatrix generating objects?

Make `void getTransformMatrix(Float32Array)`

> 4. How can the application find out if it is getting behind the video? Do we need another function for that?

It could listen `onFrameAvailable()` callback  (non informative) or we can provide new function `long getAvailableTimestamp()` which will return timestamp for the latest available frame (decoded texture image).
Application can use it to determinate time offset and how it is getting behind the video.

> 5. Do we need a query for the frame rate so the application can budget its time?

Still looking at it, so not answer yet..

Thanks.

msc-:
Thanks @tharkum for your thoughtful answers.

> sorry for late response

No worries. My comment was not a rebuke merely a statement of fact so people won't wonder why I don't respond.

I think we should leave `getTimestamp` on the base interface. I agree the value should be the time associated with the frame whose image was set by the most recent call to `updateTexImage`.  The question is should that time be the time relative to the start of the media (the Media Stream Counter in OpenML speak) or the time when the frame wishes to be displayed. The former makes more sense to me.

`video{Width,Height}` give the size in CSS pixels. I suspect the people requesting size want it in device pixels.

Please update `getTransformMatrix` in the IDL in your comment to `void getTransformMatrix(Float32Array)`.

If we do add a query for the frame rate, it will have to deal with variable rate video. For example by returning the minimum inter-frame interval.

kenrussell:
@xingri @tharkum @msc- in order to make forward progress on this extension proposal I'm going to merge this pull request now. It can be updated further, but let's open issues regarding it and other proposals, and capture some of the results of this discussion in them. Also please feel free to put forth other extension proposals generalizing it; however, video elements are handled very differently than other DOM elements, and I think we are going to need special handling for them.

PR #2471, opened 2017-08-08
Roll back shader restrictions in WEBGL_multiview spec

Oletus:
From prototyping we have learned that most of the expected perfomance
gains from the extension can be realized without the shader
restrictions on NVIDIA OpenGL drivers, and particularly in the context
of the web the shader restrictions might only buy marginal additional
performance. As such it makes sense to leave the restrictions out to
make the API simpler and easier to program.

### DIFF
--- a/extensions/proposals/WEBGL_multiview/extension.xml
+++ b/extensions/proposals/WEBGL_multiview/extension.xml
@@ -62,34 +62,7 @@
         Likewise the shading language preprocessor <code>#define GL_OVR_multiview</code>, will be defined to 1 if the extension is supported.
       </addendum>
       <addendum>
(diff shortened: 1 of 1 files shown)


Oletus:
Please take a look! I'd expect this should be uncontroversial - I was the one pushing for these restrictions in the first place, but it's been demonstrated by our prototyping work that we'll get close enough to theoretical maximum performance even without them.

kenrussell:
This looks great - simpler is better. Any other comments @jdashg @RafaelCintron ?

Oletus:
I'm going ahead with merging this. There will be ample possibilities for more feedback later since this is still at the proposal stage. There's still one other change I'll do based on our prototyping results but after that I'd like to start the public review.

PR #2472, opened 2017-08-08
Non-normative note about poor performance of sampler uniform updates.

tsherif:
Adds a non-normative note about the performance problems with updating sampler uniforms on some systems (see https://bugs.chromium.org/p/chromium/issues/detail?id=735483).

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -2997,7 +2997,7 @@ WebGLRenderingContext implements WebGLRenderingContextBase;
             invalid length, an <code>INVALID_VALUE</code> error will be generated. The length is
             invalid if it is too short for or is not an integer multiple of the assigned type.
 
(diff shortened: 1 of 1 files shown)


specs/latest/1.0/index.html
```diff
@@ -2996,6 +2996,12 @@ <h2 class="no-toc">Table of contents</h2>
             If the array passed to any of the vector forms (those ending in <code>v</code>) has an
             invalid length, an <code>INVALID_VALUE</code> error will be generated. The length is
             invalid if it is too short for or is not an integer multiple of the assigned type.
+
+            <div class="note">Perfomance problems have been observed on some implementations when
```
grovesNL: `Perfomance` should be `Performance`

tsherif: Fixed, thanks.

PR #2475, opened 2017-08-10
tag mp4 file as REC601 content

FredrikHubinette:
This specifies how the content should be interpreted, and avoids problem where some systems will guess that the content is REC709, which it is not.

The file was created by adding color tags to the SAR entries of the previous file.


kenrussell:
Thanks @FredrikHubinette for this update. I tested it on Mac and Windows and it works fine. This will help reduce undefined behaviors in the conformance tests.

Also, for posterity: @FredrikHubinette pointed out that this tag was added using this tool, in order to avoid re-encoding the file:
https://forum.doom9.org/showthread.php?t=152419

jdashg:
Cool, we did this previously for vp9+webm in #2240.

PR #2476, opened 2017-08-11
Fix conformance2/rendering/blitframebuffer-r11f-g11f-b10f.html

zhenyao:


zhenyao:
This is a simple fix to unblock rolling tests into chromium. Merging myself.

kenrussell:
Thanks Mo for fixing this. I failed to test it inside the conformance harness.

PR #2477, opened 2017-08-11
Out-of-range vertex may not generate INVALID_OPERATION

qjia7:
This can make Chrome/ANGLE disable index range retrieving in validation.
See https://bugs.chromium.org/p/angleproject/issues/detail?id=1393

### DIFF
--- a/sdk/tests/conformance2/rendering/element-index-uint.html
+++ b/sdk/tests/conformance2/rendering/element-index-uint.html
@@ -50,6 +50,34 @@ void main() {
   gl_FragColor = color;
 }
 </script>
+<script id="vsCheckOutOfBounds" type="x-shader/x-vertex">
+    precision mediump float;
+    attribute vec2 position;
+    attribute vec4 vecRandom;
+    varying vec4 v_color;
+
+    // Per the spec, each component can either contain existing contents
+    // of the buffer or 0.
+    bool testFloatComponent(float component) {
+        return (component == 0.2 || component == 0.0);
+    }
+    // The last component is additionally allowed to be 1.0.
+    bool testLastFloatComponent(float component) {
+        return testFloatComponent(component) || component == 1.0;
+    }
+
+    void main() {
+        if (testFloatComponent(vecRandom.x) &&
+            testFloatComponent(vecRandom.y) &&
+            testFloatComponent(vecRandom.z) &&
+            testLastFloatComponent(vecRandom.w)) {
+            v_color = vec4(0.0, 1.0, 0.0, 1.0); // green -- Out of range
+        } else {
+            v_color = vec4(1.0, 0.0, 0.0, 1.0); // red -- Unexpected value
+        }
+        gl_Position = vec4(position, 0.0, 1.0);
+    }
+</script>
 
 </head>
 <body>
@@ -87,9 +115,8 @@ for (var ii = 0; ii < 2; ++ii) {
         // These tests are tweaked duplicates of the buffers/index-validation* tests
         // using unsigned int indices to ensure that behavior remains consistent
         runIndexValidationTests(drawType);
-        runCopiesIndicesTests(drawType);
+        runIndexOutOfRangeTests(drawType);
         runResizedBufferTests(drawType);
-        runVerifiesTooManyIndicesTests(drawType);
         runCrashWithBufferSubDataTests(drawType);
 
         wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no errors");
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/rendering/element-index-uint.html
```diff
@@ -89,7 +89,6 @@
         runIndexValidationTests(drawType);
         runCopiesIndicesTests(drawType);
         runResizedBufferTests(drawType);
-        runVerifiesTooManyIndicesTests(drawType);
```
qjia7: Removing runVerifiesTooManyIndicesTests is because that the body of it is almost same with runCopiesIndicesTests.

qjia7:
@kenrussell @zhenyao Please take a look. When I tried to remove the index range in ANGLE, this file failed. See https://chromium-review.googlesource.com/c/607413

sdk/tests/conformance2/rendering/element-index-uint.html
```diff
@@ -309,13 +308,10 @@
     var indices = new Uint32Array([ 10000, 0, 1, 2, 3, 10000 ]);
     gl.bufferData(gl.ELEMENT_ARRAY_BUFFER, indices, drawType);
     wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.drawElements(gl.TRIANGLE_STRIP, 4, gl.UNSIGNED_INT, 4)");
-    wtu.shouldGenerateGLError(gl, gl.INVALID_OPERATION, "gl.drawElements(gl.TRIANGLE_STRIP, 4, gl.UNSIGNED_INT, 0)");
-    wtu.shouldGenerateGLError(gl, gl.INVALID_OPERATION, "gl.drawElements(gl.TRIANGLE_STRIP, 4, gl.UNSIGNED_INT, 8)");
-    indices[0] = 2;
-    indices[5] = 1;
+    wtu.shouldGenerateGLError(gl, gl.INVALID_OPERATION, "gl.drawElements(gl.TRIANGLE_STRIP, 4, gl.UNSIGNED_INT, 28)");
+    indices = new Uint32Array([ 10000, 0, 1, 2, 3, 10000, 2, 1]);
```
zhenyao: I don't know how this can work because you didn't upload the data to a buffer at all after this.

sdk/tests/conformance2/rendering/element-index-uint.html
```diff
@@ -265,7 +264,7 @@
     shouldBeUndefined('gl.drawElements(gl.TRIANGLES, 3, gl.UNSIGNED_INT, 0)');
     wtu.glErrorShouldBe(gl, gl.NO_ERROR);
     debug("Enable normals, out-of-range");
-    gl.vertexAttribPointer(normalLoc, 3, gl.FLOAT, false, 7 * sizeInBytes(gl.FLOAT), 4 * sizeInBytes(gl.FLOAT));
+    gl.vertexAttribPointer(normalLoc, 3, gl.FLOAT, false, 7 * sizeInBytes(gl.FLOAT), 21 * sizeInBytes(gl.FLOAT));
```
kenrussell: Why is this change being made? It breaks the purpose of the original test, which is that the last fetched normal will be out-of-range. I think this should be reverted, and the test should essentially be made optional.

Microsoft contributed the test sdk/tests/conformance/rendering/out-of-bounds-index-buffers.html , which covers both the situation where the browser does index validation, and where it delegates index validation to the driver.

Could you consider looking at that other test and seeing whether it can be adapted to uint element indices?

sdk/tests/conformance2/rendering/element-index-uint.html
kenrussell: This was part of runCopiesIndicesTests. It looks like there was once a bug where the ArrayBuffer holding the indices was being held by reference, instead of copied.

sdk/tests/conformance2/rendering/element-index-uint.html
qjia7: @kenrussell Thanks for your suggestion. I have followed sdk/tests/conformance/rendering/out-of-bounds-index-buffers.html to rewrite the out of range test case.  Currently, I move all out of range related tests to function runIndexOutOfRangeTests.  runCopiesIndicesTests was also merged into this function.

qjia7:
@kenrussell @zhenyao  Please take another look. I notice that when the drawType is gl.DYNAMIC_DRAW, out of range test will fail as below (tested in modified chrome in which I removed the IndexRange validation and didn't generate an INVALID_OPERATION error) :
FAIL should be 0,255,0,255
at (1, 0) expected: 0,255,0,255 was 5,250,0,255

It's not about errorRange. Even I set it, it still fails at other position. However, all cases can pass if the drawType is gl.STATIC_DRAW. I think the test case should be no problem. But it may expose angle or driver bug in dynamic draw mode. I need to dig into it.

qjia7:
> I notice that when the drawType is gl.DYNAMIC_DRAW, out of range test will fail ...

@kenrussell  @zhenyao  @null77 The failure reason is because the angle's special implementation for d3d DYNAMIC_DRAW. With index range validation, it can ensure the translation is always right. But without index range validation, we should check whether we can reserve the required space. If you are interested for the changes that I modified in ANGLE, please see https://chromium-review.googlesource.com/c/607413

qjia7:
Ping @kenrussell @zhenyao

PR #2479, opened 2017-08-11
Clean up uniform buffer tests

Oletus:
Uniform buffer draw tests can check that the uniform buffer is in the
std140 layout, since the WebGL spec mandates std140 layout. This makes
the tests simpler and enables sharing some common functionality
between subtests as well.

This refactoring is done to make way for a few more uniform buffer
tests that are needed for bugs that recently got fixed in ANGLE
related to structs in uniform blocks.

The test was also expecting that the value of
UNIFORM_BUFFER_OFFSET_ALIGNMENT was divisible by 4, but that was
somewhat non-obvious. The test for it is also clearer now. This
requirement didn't seem to be enshrined in the spec previously, but it
is reasonable and the test has been there for a while so put it in the
spec now.

### DIFF
--- a/sdk/tests/conformance2/buffers/uniform-buffers.html
+++ b/sdk/tests/conformance2/buffers/uniform-buffers.html
@@ -137,6 +137,8 @@ if (!gl) {
     testPassed("WebGL context exists");
 
     runBindingTest();
+    runBadShaderTest();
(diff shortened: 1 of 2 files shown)


kenrussell:
Great updates. LGTM

PR #2482, opened 2017-08-14
Test initializing an array with another one that has the same name

Oletus:
It's possible to initialize variables with a variable with the same
name from an outer scope. This is broken in ANGLE's HLSL backend for
arrays, so add a separate test for the array case.


zhenyao:
LGTM
Thanks @Oletus to catch this corner case and add a conformance test.

PR #2483, opened 2017-08-14
Test passing a sampler array in a struct to a function

Oletus:
This is broken on ANGLE's HLSL backend.

Includes some cleanup in an older similar test.


zhenyao:
LGTM

PR #2484, opened 2017-08-14
tag more files as REC601

FredrikHubinette:


kenrussell:
Talked with Fredrik about recreating this pull request cleanly from a new branch. Let me own this and we'll probably close it in favor of a forthcoming one.

PR #2485, opened 2017-08-15
tagged more mp4 files as REC 601

FredrikHubinette:


kenrussell:
Thanks very much for fixing these. Follow-on to #2475 . Tested locally on macOS and Windows. Merging now.

PR #2486, opened 2017-08-15
Add diff spec for signed normalized vertex attributes in WebGL 1.0.

kenrussell:
Different versions of OpenGL and OpenGL ES on top of which WebGL 1.0
is implemented follow different rules. Since this does not affect the
majority of applications, document it as a difference in behavior.
This may be strengthened later.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -4109,6 +4109,33 @@ extensions.
     <code>0.0</code>, <code>vec4(0.0)</code>, <code>0</code>, <code>false</code>, etc.
 </p>
 
+<h3>Vertex attribute conversions from normalized signed integers to floating point</h3>
+
+<p>
+    The OpenGL ES 2.0 spec <a href="#refsGLES20">[GLES20]</a> section 2.1.2 "Data Conversions",
+    subsection "Conversion from Integer to Floating-Point" defines conversion from a normalized
+    signed integer <em>c</em>, where the bit width of the type is <em>b</em>, to floating-point
+    value <em>f</em> as:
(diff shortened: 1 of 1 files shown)


msc-:
Only WebGL 1..0 is affected by this. As WebGL 2 is based on either desktop or OpenGL ES 3. It should always use the zero-preserving mapping. Does D3D use the non-zero-preserving rule? If not, then there is no need to note a difference in the WebGL 2 spec.

dmikis:
@kenrussell 

The addition seems all right.

@msc- 

> As WebGL 2 is based on either desktop or OpenGL ES 3. It should always use the zero-preserving mapping.

That depends on version of desktop GL context WebGL 2's running on. If it's older than 4.2, then vertex attribs should be converted by non-zero-preserving rule (at least according to specs, on my macbook on GL 4.1 looks like the zero-preserving rule is used).

> Does D3D use the non-zero-preserving rule?

I can't find anything in MSDN, but testing on a windows machine gave results suggesting that it uses the zero-preserving rule.

kenrussell:
Kirill, can you please provide a WebGL 2.0 conformance test that verifies that the zero-preserving rule is obeyed? Per @msc- 's comment I would rather make only the change to the WebGL 1.0 spec and not the 2.0 spec. Thanks.

dmikis:
@kenrussell https://github.com/KhronosGroup/WebGL/pull/2488

kenrussell:
Removed the WebGL 2.0 spec change and updated the pull request's comment.

zhenyao:
LGTM

PR #2487, opened 2017-08-17
Require a multiview program to draw to a multiview framebuffer

Oletus:
Supporting drawing to multiview framebuffers with non-multiview
programs could require generating multiple versions of the program
object behind the scenes. This would require more complex code and
would result in worse and less predictable performance. Since this
kind of use has limited utility, it is better to make it to generate
an error.


kenrussell:
Sounds like a good restriction.

PR #2489, opened 2017-08-18
Require color_buffer exts if float textures are implicitly renderable.

jdashg:
Also fixes some indent levels.

### DIFF
--- a/sdk/tests/conformance/extensions/oes-texture-float.html
+++ b/sdk/tests/conformance/extensions/oes-texture-float.html
@@ -82,41 +82,76 @@ var canvas = document.getElementById("canvas");
 var gl = wtu.create3DContext(canvas);
 
 if (!gl) {
-  testFailed("WebGL context does not exist");
+    testFailed("WebGL context does not exist");
 } else {
-  testPassed("WebGL context exists");
+    testPassed("WebGL context exists");
 
-  var texturedShaders = [
-      wtu.simpleTextureVertexShader,
-      "testFragmentShader"
-  ];
-  var testProgram =
-      wtu.setupProgram(gl,
-                       texturedShaders,
-                       ['vPosition', 'texCoord0'],
-                       [0, 1]);
-  var quadParameters = wtu.setupUnitQuad(gl, 0, 1);
+    var texturedShaders = [
+        wtu.simpleTextureVertexShader,
+        "testFragmentShader"
+    ];
+    var testProgram =
+        wtu.setupProgram(gl,
+                         texturedShaders,
+                         ['vPosition', 'texCoord0'],
+                         [0, 1]);
+    var quadParameters = wtu.setupUnitQuad(gl, 0, 1);
 
-  // First verify that allocation of floating-point textures fails if
-  // the extension has not been enabled yet.
-  runTextureCreationTest(testProgram, false);
+    // First verify that allocation of floating-point textures fails if
+    // the extension has not been enabled yet.
+    runTextureCreationTest(testProgram, false);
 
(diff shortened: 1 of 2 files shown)


kenrussell:
Could I please ask what's the motivation for making this change right now?

We're behind on our bug backlog on Chromium and I'm not sure how long it will take us to start passing these tests again once this change goes in.

CC @zhenyao @kainino0x -- if one of you would be able to make this change for WebGL 1.0 contexts in Chromium, we could merge this without impact.

jdashg:
I can move these tests to separate files for the color_buffer_float extensions, if your concern is maintaining expected-pass markings for float-texture extensions. It was just more invasive to do this to begin with.

The acute motivation here is that people don't use the color_buffer extensions because they don't signal anything in at least Chrome, since they are never exposed there. This perpetuates confusion about float-rtt, and makes naive extension stat gathering (like webglstats) fail to represent how wide support for float/half-float-rtt is.

kenrussell:
OK. Let us implement the implicit enabling of these extensions in Chrome before we merge this pull request so we don't regress these tests. @zhenyao volunteered to pick this up but if he doesn't get to it this week I'll do it next week.

sdk/tests/conformance/extensions/oes-texture-float.html
```diff
@@ -82,41 +82,76 @@
...
+
+        (function() {
+            debug("");
+            var renderableExtName = "WEBGL_color_buffer_float";
+            if (isRenderable(gl)) {
+                var supported = gl.getSupportedExtensions();
+                assertMsg(supported.includes(renderableExtName),
+                          "Implicit renderability requires exposing support for " + renderableExtName);
+            } else {
+                var renderableExt = gl.getExtension(renderableExtName);
```
zhenyao: Should we just assert the extension isn't included in the supported extensions here?

sdk/tests/conformance/extensions/oes-texture-half-float.html
```diff
@@ -173,17 +174,49 @@
...
-        runRenderTest(gl.RGB, [10000, 10000, 10000, 1], null);
+        (function() {
+            debug("");
+            var renderableExtName = "EXT_color_buffer_half_float";
+            if (isRenderable(gl, ext)) {
+                var supported = gl.getSupportedExtensions();
+                assertMsg(supported.includes(renderableExtName),
+                          "Implicit renderability requires exposing support for " + renderableExtName);
+            } else {
+                var renderableExt = gl.getExtension(renderableExtName);
```
zhenyao: Should we just assert the extension isn't included in the supported extensions here?

sdk/tests/conformance/extensions/oes-texture-float.html
jdashg: Yeah, that would simplify this code.

jdashg: Actually no, because we have to call getExtension anyways in order to activate the extension.

zhenyao: My understanding is when we enable OES_texture_float, if these formats are color-renderable, we automatically enable WEBGL_color_buffer_float implicitly?

zhenyao: Per working group discussion, we are going to remove "Although this feature has historically been allowed, new implementations should not implicitly support float rendering and applications should be modified to explicitly enable WEBGL_color_buffer_float." from https://www.khronos.org/registry/webgl/extensions/OES_texture_float/. Let's merge this pull request as is and I'll make the change together with the spec change, so the test and spec are always consistent.

PR #2490, opened 2017-08-22
Fix offscreencanvas-resize test to comply with spec

junov:
This fix covers two spec compliance issues: 1) When an OffscreenCanvas
is resized, its size must propagate to the placeholder canvas's
width and height attributes; 2) Setting the width and height attributes
on a placeholder canvas throws an exception.

Related spec change: https://github.com/whatwg/html/pull/2951


junov:
cc @juj

zhenyao:
Thanks for the fix!

PR #2491, opened 2017-08-22
Regression test for attribute 0 being unconsumed and unallocated

kainino0x:
This tests for a driver crash on macOS (OpenGL Core Profile only).
Affects both Chrome (WebGL 1/2) and Firefox (WebGL 2 only).

http://crbug.com/756293


sdk/tests/conformance/attribs/gl-vertex-attrib-zero-issues.html
```diff
@@ -142,7 +156,65 @@
   gl.disableVertexAttribArray(3);
 }
 
+gl.disableVertexAttribArray(3);
```
zhenyao: nit: this seems unnecessary?

sdk/tests/conformance/attribs/gl-vertex-attrib-zero-issues.html
```diff
@@ -142,7 +156,65 @@
...
+}
+for (var i = 0; i < draws.length; ++i) {
+  var draw = draws[i];
+  gl.bindBuffer(gl.ARRAY_BUFFER, unallocatedBuffer);
+  gl.useProgram(pNoAttribs);
+  gl.enableVertexAttribArray(0);
+  gl.vertexAttribPointer(0, 3, gl.FLOAT, false, 0, 0);
+  draw();
+  // clear errors if any
+  gl.getError();
```
zhenyao: Should we make sure there is no GL error? Otherwise you won't be able to catch a bug in your testing code.

If needed, you can move vertex attribs set up code into the draw functions.

sdk/tests/conformance/attribs/gl-vertex-attrib-zero-issues.html
kainino0x: yup, removed.

kainino0x:
@zhenyao this is ready again. PTAL

sdk/tests/conformance/attribs/00_test_list.txt
```diff
@@ -10,3 +10,4 @@ gl-vertexattribpointer.html
 gl-vertexattribpointer-offsets.html
 --min-version 1.0.2 gl-vertex-attrib-render.html
 gl-vertex-attrib-zero-issues.html
+--min-version 1.0.4 gl-vertex-attrib-zero-unconsumed-out-of-bounds.html
```
zhenyao: get rid of "zero" in the file name

PR #2493, opened 2017-08-24
Allow array length() method to be called on arbitrary expressions

Oletus:
Reverse the expectations on a test that was based on an incorrect
interpretation of the spec. While the GLSL ES 3.00 spec says that
length is only allowed on "array names", it was always the intent to
allow it on arbitrary expressions that return an array. This has been
clarified in GLSL ES 3.20.

Since the revised test adds new requirements, it makes sense to
remove the test entirely from the 2.0.0 conformance suite.


kenrussell:
Thanks Olli for researching and clarifying this. What is needed to make conformance2/glsl3/array-length-side-effects.html start passing again after this change?

Oletus:
A fix in ANGLE's shader translator is needed. Since HLSL doesn't have an array length operation, any expressions with side effects that have .length() called on them need to be transformed so that the expression with side effects is separated, and length() is replaced with a constant. It can probably be done something like this:

(foo()).length()

becomes

(foo()[0], 5)

where foo() is a function returning an array of length 5.

kenrussell:
Is an ANGLE bug already filed about this? What's the timeframe for getting the test passing again? It would be good to not leave tests failing for an indefinite period of time. Thanks for your help.

Oletus:
I filed an ANGLE bug now: https://bugs.chromium.org/p/angleproject/issues/detail?id=2142

It should not be that much work to fix, I'll try to take care of it during the next few weeks.

PR #2494, opened 2017-08-25
New tests for texImage* with visible webgl canvas

junov:
The new tests cover cases when the source image for a texture upload
is a visible WebGL canvas.  When the canvas is visible (in the DOM)
it gets presented to screen, which alters internal state, which can be a
source of buggy behavior.

Related chrome bug: crbug.com/759180

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
@@ -118,11 +118,26 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
       setCanvasToRedGreen(ctx);
     }
 
-    function runOneIteration(canvas, useTexSubImage2D, flipY, program, bindingTarget, opt_texture)
+    function setCanvasTo32x32(ctx, bindingTarget) {
+      ctx.canvas.width = ctx.canvas.height = 32;
+      setCanvasToRedGreen(ctx);
+    }
+
+    function runOneIteration(canvas, useTexSubImage2D, flipY, visible, program, bindingTarget, opt_texture)
     {
         debug('Testing ' + (useTexSubImage2D ? 'texSubImage2D' : 'texImage2D') + ' with flipY=' +
-              flipY + ' bindingTarget=' + (bindingTarget == gl.TEXTURE_2D ? 'TEXTURE_2D' : 'TEXTURE_CUBE_MAP') +
+              flipY + ' visible=' + visible +
+              ' bindingTarget=' + (bindingTarget == gl.TEXTURE_2D ? 'TEXTURE_2D' : 'TEXTURE_CUBE_MAP') +
               ' canvas size: ' + canvas.width + 'x' + canvas.height + ' with red-green');
+
+        if (visible && !canvas.parentNode) {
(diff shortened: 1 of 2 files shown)


zhenyao:
Is it possible to modify the existing tex-image-and-sub-image-2d-with-webgl-canvas.js and tex-image-and-sub-image-3d-with-webgl-canvas.js to cover the visible webgl canvas case? So they cover both invisible and visible webgl canvas. That could avoid adding a large set of new tests being added.

kenrussell:
+1 to Mo's suggestion. Could you please modify the existing tex-image-and-sub-image-*-with-webgl-canvas harnesses to cover this case (run the tests twice, or run them twice for a subset of formats) instead of duplicating both the harnesses and the tests? Thanks.

junov:
Done.  I modified existing tests instead of creating new ones.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
```diff
@@ -213,18 +228,24 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
 
     function runTest()
     {
-        var ctx = wtu.create3DContext();
-        var canvas = ctx.canvas;
+        // Note: We use preserveDrawingBuffer:true to prevent canvas
+        // visibility from interfering with the tests.
+        var ctx = wtu.create3DContext(null, { preserveDrawingBuffer:true });
```
zhenyao: Can we only set preserveDrawingBuffer:true for the test cases where visible: true? This can be easily done by creating another canvas.

Otherwise we lose test coverage.

sdk/tests/js/tests/tex-image-and-sub-image-3d-with-webgl-canvas.js
```diff
@@ -145,14 +159,18 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
 
     function runTest()
     {
-        var ctx = wtu.create3DContext();
+        // Note: We use preserveDrawingBuffer:true to prevent canvas
+        // visibility from interfering with the tests.
+        var ctx = wtu.create3DContext(null, { preserveDrawingBuffer:true });
```
zhenyao: Same here.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
junov: Done.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
```diff
@@ -214,17 +216,29 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
     function runTest()
     {
         var ctx = wtu.create3DContext();
-        var canvas = ctx.canvas;
+        canvas = ctx.canvas;
```
zhenyao: Sorry, but you removed the var here.

junov: Whoops.  Fixed.

zhenyao:
Thank you!

PR #2495, opened 2017-08-30
Add test for using an input with interpolation qualifier as l-value

Jiawei-Shao:
This patch adds tests to expose a compiler bug in ANGLE. The compiler
should always report a compile error when handling an input with
interpolation qualifier as l-value in shaders.

### DIFF
--- a/sdk/tests/conformance2/glsl3/00_test_list.txt
+++ b/sdk/tests/conformance2/glsl3/00_test_list.txt
@@ -16,6 +16,7 @@ const-array-init.html
 forbidden-operators.html
 frag-depth.html
 --min-version 2.0.1 gradient-in-discontinuous-loop.html
+--min-version 2.0.1 input-with-interpotaion-as-lvalue.html
 invalid-default-precision.html
 invalid-invariant.html
 loops-with-side-effects.html
--- a/sdk/tests/conformance2/glsl3/input-with-interpotaion-as-lvalue.html
+++ b/sdk/tests/conformance2/glsl3/input-with-interpotaion-as-lvalue.html
@@ -1,0 +1,104 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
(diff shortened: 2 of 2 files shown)


Jiawei-Shao:
@qjia7 @Richard-Yunchao @gyagp  PTAL, thanks!

Jiawei-Shao:
@Oletus PTAL, thanks!

sdk/tests/conformance2/glsl3/00_test_list.txt
```diff
@@ -16,6 +16,7 @@ const-array-init.html
 forbidden-operators.html
 frag-depth.html
 --min-version 2.0.1 gradient-in-discontinuous-loop.html
+input-with-interpotaion-as-lvalue.html
```
Oletus: Add --min-version 2.0.1

sdk/tests/conformance2/glsl3/input-with-interpotaion-as-lvalue.html
```diff
@@ -0,0 +1,99 @@
+<!--
+
+/*
+** Copyright (c) 2016 The Khronos Group Inc.
```
Oletus: Fix date to 2017

sdk/tests/conformance2/glsl3/input-with-interpotaion-as-lvalue.html
```diff
@@ -0,0 +1,99 @@
...
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+<script src="../../js/glsl-conformance-test.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<script type="application/javascript">
+"use strict";
+description();
```
zhenyao: Can you add a comment here pointing to the spec text that's related to this behavior? As far as I can tell from a quick check, these qualifiers can't be applied to input vars in vertex shaders or output vars in fragment shaders (GLSL ES 3.0.0, section 4.3). So this rule isn't violated here.  If there is another rule, please document it here so readers in the future don't have to search it.  Also, maybe also add the angle bug link here.

kenrussell: Agreed, this test looks incorrect. GLSL 3.00.6 sections 4.3.4 "Input variables", section 4.3.6 "Output Variables" and section 4.3.9 "Interpolation" all indicate that it's legal (and actually required) for the qualifiers to match on the vertex shader's outputs and fragment shader's inputs.

We've had problems in this area in the past where the "invariant" qualifier had to be erased on some platforms because of inherent mismatches between desktop GLSL and GLSL ES's shader semantics. See #1924 and #1926 . Perhaps this is the root cause of this issue?

Oletus: ESSL 3.00.6 section 4.3.4: "Variables declared as in or centroid in may not be written to during shader execution." This test is not about qualifier matching between shader stages, maybe you misunderstood the intent of the test?

zhenyao: I see.  I was distracted by iterating all three interpolation qualifiers and thought this test was about that.

Then can we add here this citation from spec @Oletus quoted?

kenrussell: Sorry for misunderstanding the test.

Richard-Yunchao:
Agreed that the description for this bug is too simple. And there is no link to the ANGLE bug or fix. Please add necessary quotation and description, and take care of this, Jiawei. @Jiawei-Shao

sdk/tests/conformance2/glsl3/input-with-interpotaion-as-lvalue.html
Jiawei-Shao: @zhenyao  Done

Jiawei-Shao:
Thanks for all comments from @Oletus  @zhenyao @kenrussell  @Richard-Yunchao. I have added the missing SPEC quotation and link to the ANGLE bug. PTAL, thanks!

zhenyao:
LGTM
Thank you

PR #2496, opened 2017-08-30
Add OffscreenCanvas cases to texImage tests

junov:


kenrussell:
Thanks, nice addition. Could you check to see if these new tests increase the run time significantly of in particular the WebGL 2.0 versions of these tests? There are a lot of texture formats in that version of the spec and we've had to be careful to not add tons of new tests. Thanks.

CC @zhenyao @jdashg @RafaelCintron @grorg

junov:
I tried to strike a reasonable balance between runtime and coverage
The increase in run time only affects the canvas and webgl-canvas variants of the texImage tests/
I tried to be minimalist by not repeating **all** the cases that we have for canvas elements, but I think it is important to cover various permutations format, flipy, subImage, 2d/3d targets, so there is still a bit of a combinatorial explosion there.

The new case take about the same time to run as the old cases. So based on the number of cases, this change increases the run time by 33%-50%.

Also, the new cases are skipped in browsers that do not expose OffscreenCanvas.

kenrussell:
Understand the need to test the various permutations.

We should rebalance the WebGL 2.0 conformance test shards [1] after landing this and rolling it in to Chromium.

Any feedback from other browser vendors before merging this addition?

[1] https://cs.chromium.org/chromium/src/content/test/gpu/gather_swarming_json_results.py

PR #2497, opened 2017-08-31
Add a missing include of js-test-post

Oletus:
The sampler-array-struct-function-arg test did not signal completion
and so could time out in the test harness.


zhenyao:
LGTM

PR #2498, opened 2017-08-31
Add a test for non-existent transform feedback varyings

Oletus:
This covers the following ANGLE bug:
https://bugs.chromium.org/p/angleproject/issues/detail?id=2141


zhenyao:
LGTM

PR #2499, opened 2017-08-31
Test texStorage and texSubImage with ASTC

kainino0x:


sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
```diff
@@ -1846,7 +1850,8 @@
     canvas.height = height;
     gl.viewport(0, 0, width, height);
     debug("");
-    debug("Testing " + formatToString(format) + " " + width + "x" + height + " (" + test.mode + ")");
```
zhenyao: can we still keep the "mode"?

kainino0x: Done

PR #2505, opened 2017-09-14
Clarify WebGL1 float color buffer extensions behaviors

zhenyao:
and testing against them

### DIFF
--- a/extensions/EXT_color_buffer_half_float/extension.xml
+++ b/extensions/EXT_color_buffer_half_float/extension.xml
@@ -37,9 +37,10 @@
         Renderbuffers can be created in these formats. These and textures
         created with <code>type = HALF_FLOAT_OES</code>, which will have one
         of these internal formats, can be attached to framebuffer object color
-        attachments for rendering. <code>RGBA16F</code> textures are guaranteed
-	to be color-renderable. <code>RGB16F</code> textures are color-renderable
-        if <code>checkFramebufferStatus()</code> returns <code>FRAMEBUFFER_COMPLETE</code>.
+        attachments for rendering. Implementations supporting this extension is
+	required to support rendering to <code>RGBA16F</code> format.
+	Applications must check framebuffer completeness to determine if
+	<code>RGA16F</code> is supported.
       </feature>
 
       <feature>
--- a/extensions/EXT_color_buffer_half_float/extension.xml
+++ b/extensions/EXT_color_buffer_half_float/extension.xml
@@ -39,14 +39,14 @@
         of these internal formats, can be attached to framebuffer object color
         attachments for rendering. <code>RGBA16F</code> textures are guaranteed
 	to be color-renderable. <code>RGB16F</code> textures are color-renderable
-        if <code>getFramebufferStatus()</code> returns <code>FRAMEBUFFER_COMPLETE</code>.
+        if <code>checkFramebufferStatus()</code> returns <code>FRAMEBUFFER_COMPLETE</code>.
       </feature>
 
       <feature>
         <p><span style="color: red">NOTE:</span> fragment shaders outputs
         gl_FragColor and gl_FragData[0] will only be clamped and converted
-        when the color buffer is fixed-point and <code>BlendColor()</code> and
-        <code>ClearColor()</code> will no longer clamp their parameter values
+        when the color buffer is fixed-point and <code>blendColor()</code> and
+        <code>clearColor()</code> will no longer clamp their parameter values
         on input. Clamping will be applied as necessary at draw time according
         to the type of color buffer in use.</p>
       </feature>
--- a/extensions/EXT_color_buffer_half_float/extension.xml
+++ b/extensions/EXT_color_buffer_half_float/extension.xml
@@ -37,7 +37,9 @@
         Renderbuffers can be created in these formats. These and textures
         created with <code>type = HALF_FLOAT_OES</code>, which will have one
         of these internal formats, can be attached to framebuffer object color
-        attachments for rendering.</p>
+        attachments for rendering. <code>RGBA16F</code> textures are guaranteed
+	to be color-renderable. <code>RGB16F</code> textures are color-renderable
+        if <code>getFramebufferStatus()</code> returns <code>FRAMEBUFFER_COMPLETE</code>.
       </feature>
 
       <feature>
@@ -139,6 +141,10 @@ interface EXT_color_buffer_half_float {
 
     <revision date="2016/05/05">
       <change>Subsumed in WebGL 2.0 by EXT_color_buffer_float.</change>
+    </revision>
+
+    <revision date="2017/09/14">
+      <change>Require RGBA16F to be color-renderable and RGB16F to be optionally color-renderable.</change>
     </revision>
   </history>
 </extension>
--- a/extensions/WEBGL_color_buffer_float/extension.xml
+++ b/extensions/WEBGL_color_buffer_float/extension.xml
@@ -39,10 +39,9 @@
 
       <feature>
         <p>The 32-bit floating-point type <code>RGB32F</code> may also optionally
-	become available as a color-renderable format. Renderbuffers can be created
-	in this format. These and textures created with <code>format = RGB</code>
-	and <code>type = FLOAT</code> as specified in <a
-        href="http://www.khronos.org/registry/webgl/extensions/OES_texture_float/">OES_texture_float</a>,
+	become available as a color-renderable format. These and textures created
+	with <code>format = RGB</code> and <code>type = FLOAT</code> as specified in
+	<a href="http://www.khronos.org/registry/webgl/extensions/OES_texture_float/">OES_texture_float</a>,
         can be attached to framebuffer object color attachments for rendering.
 	Applications must check framebuffer completeness to determine if an
 	implementation actually supports this format.
--- a/extensions/WEBGL_color_buffer_float/extension.xml
+++ b/extensions/WEBGL_color_buffer_float/extension.xml
@@ -44,7 +44,7 @@
         <code>type = FLOAT</code> as specified in <a
         href="http://www.khronos.org/registry/webgl/extensions/OES_texture_float/">OES_texture_float</a>,
         can be attached to framebuffer object color attachments for rendering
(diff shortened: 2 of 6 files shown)


zhenyao:
@kenrussell @jdashg Please review
@RafaelCintron @grorg @kainino0x FYI

extensions/WEBGL_color_buffer_float/extension.xml
```diff
@@ -38,10 +38,21 @@
       </feature>
 
       <feature>
+        <p>The 32-bit floating-point type <code>RGB32F</code> may become available
+        as a color-renderable format. Renderbuffers can be created in this
+        format. These and textures created with <code>format = RGB</code> and
+        <code>type = FLOAT</code> as specified in <a
+        href="http://www.khronos.org/registry/webgl/extensions/OES_texture_float/">OES_texture_float</a>,
+        can be attached to framebuffer object color attachments for rendering
+	if <code>getFrameBufferStatus()</code> retruns <code>FRAMEBUFFER_COMPLETE</code>.
```
kenrussell: typos:
  getFrameBufferStatus() -> checkFramebufferStatus()
  retruns -> returns

extensions/EXT_color_buffer_half_float/extension.xml
```diff
@@ -37,7 +37,9 @@
         Renderbuffers can be created in these formats. These and textures
         created with <code>type = HALF_FLOAT_OES</code>, which will have one
         of these internal formats, can be attached to framebuffer object color
-        attachments for rendering.</p>
+        attachments for rendering. <code>RGBA16F</code> textures are guaranteed
+	to be color-renderable. <code>RGB16F</code> textures are color-renderable
+        if <code>getFramebufferStatus()</code> returns <code>FRAMEBUFFER_COMPLETE</code>.
```
kenrussell: getFramebufferStatus() -> checkFramebufferStatus()

zhenyao: Done

extensions/WEBGL_color_buffer_float/extension.xml
zhenyao: Done

zhenyao:
@msc- FYI

extensions/WEBGL_color_buffer_float/extension.xml
```diff
@@ -38,10 +38,21 @@
       </feature>
 
       <feature>
+        <p>The 32-bit floating-point type <code>RGB32F</code> may become available
+        as a color-renderable format. Renderbuffers can be created in this
+        format. These and textures created with <code>format = RGB</code> and
+        <code>type = FLOAT</code> as specified in <a
+        href="http://www.khronos.org/registry/webgl/extensions/OES_texture_float/">OES_texture_float</a>,
+        can be attached to framebuffer object color attachments for rendering
+	if <code>checkFrameBufferStatus()</code> returns <code>FRAMEBUFFER_COMPLETE</code>.
```
kenrussell: Capitalization should be:
checkFramebufferStatus()

zhenyao: Thanks for catching it. Updated.

extensions/EXT_color_buffer_half_float/extension.xml
```diff
@@ -37,14 +37,16 @@
         Renderbuffers can be created in these formats. These and textures
         created with <code>type = HALF_FLOAT_OES</code>, which will have one
         of these internal formats, can be attached to framebuffer object color
-        attachments for rendering.</p>
+        attachments for rendering. <code>RGBA16F</code> textures are guaranteed
+	to be color-renderable. <code>RGB16F</code> textures are color-renderable
```
jdashg: I would say:
RGB16F and RGBA16F effective internal formats become color-renderable, but only RGBA16F is guaranteed to return FRAMEBUFFER_COMPLETE from checkFramebufferStatus().

extensions/EXT_color_buffer_half_float/extension.xml
```diff
@@ -37,14 +37,16 @@
...
+        if <code>checkFramebufferStatus()</code> returns <code>FRAMEBUFFER_COMPLETE</code>.
```
jdashg: I think the most correct way to do this is to make both RGB16F and RGBA16F formats color-renderable, but only guarantee not to be FRAMEBUFFER_UNSUPPORTED for RGBA16F. If the implementation doesn't support RGB16F, it would generate FRAMEBUFFER_UNSUPPORTED.

I do think we should deprecate RGB16F, since it's not actually renderable in ES3+ even with extensions.

extensions/WEBGL_color_buffer_float/extension.xml
```diff
@@ -65,6 +76,7 @@
 [NoInterfaceObject]
 interface WEBGL_color_buffer_float {
   const GLenum RGBA32F_EXT = 0x8814;
+  const GLenum RGB32F_EXT = 0x8815;
```
jdashg: Don't add this, at least not in this bug. Supporting renderbuffers here is scope creep.

extensions/WEBGL_color_buffer_float/extension.xml
```diff
@@ -38,10 +38,21 @@
       </feature>
 
       <feature>
+        <p>The 32-bit floating-point type <code>RGB32F</code> may become available
+        as a color-renderable format. Renderbuffers can be created in this
+        format. These and textures created with <code>format = RGB</code> and
+        <code>type = FLOAT</code> as specified in <a
+        href="http://www.khronos.org/registry/webgl/extensions/OES_texture_float/">OES_texture_float</a>,
+        can be attached to framebuffer object color attachments for rendering
+	if <code>checkFramebufferStatus()</code> returns <code>FRAMEBUFFER_COMPLETE</code>.
```
jdashg: They can always be attached to framebuffers, they just can't always be rendered to.
I would say "may also optionally become color-renderable".

sdk/tests/conformance/extensions/oes-texture-float.html
```diff
@@ -114,24 +114,20 @@
...
-                    return;
-                }
-                testPassed(renderableExtName + " supported. Proceeding to its tests.");
+            var supported = gl.getSupportedExtensions().includes(renderableExtName);
+            if (renderable && !supported) {
+                testFailed("RGBA/FLOAT is color renderable but " + renderableExtName + " not exposed");
+            } else if (supported && !renderable) {
+                testFailed(renderableExtName + " is exposed but RGBA/FLOAT is not color renderable");
+            }
+            if (supported) {
```
jdashg: This should be run even if it's not claimed-supported if it is (incorrectly) implicitly renderable.
You introduce a bug here by not calling getExtension(), which means this will rely on implicit activation. IIRC the idea is that we drop recommendation that new implementations require explicit activation here, but I think we stopped short of requiring implicit activation. If we change this we *must* change the specs to reflect this.

kenrussell: I thought the idea was that the current change to the OES_texture_float and OES_texture_half_float extension specs removing the sentence "Although this feature has historically been allowed..." mandated explicit activation of the renderability extensions if they're supported. If the language isn't strong enough then perhaps this sentence:

"Implementations supporting float rendering via this extension will implicitly enable..."

should also change "will" to "must".

jdashg: I propose going straight to:
Upon activation of this extension, implementations supporting WEBGL_color_buffer_float shall implicitly enable it.
NB: This requirement maintains the historical behavior prior to the differentiation of float renderability from float textures, so as to not break existing content.

kenrussell: Sounds good to me.

extensions/EXT_color_buffer_half_float/extension.xml
zhenyao: Revised. We can discuss RGB16F removal but I will leave that out of this pull request.

extensions/WEBGL_color_buffer_float/extension.xml
zhenyao: Revised

extensions/WEBGL_color_buffer_float/extension.xml
zhenyao: Removed

sdk/tests/conformance/extensions/oes-texture-float.html
zhenyao: Addressed in the spec and leave the test as is.

zhenyao:
@jdashg Revised. Please take another look.

extensions/WEBGL_color_buffer_float/extension.xml
```diff
@@ -38,13 +38,14 @@
       </feature>
 
       <feature>
-        <p>The 32-bit floating-point type <code>RGB32F</code> may become available
-        as a color-renderable format. Renderbuffers can be created in this
-        format. These and textures created with <code>format = RGB</code> and
-        <code>type = FLOAT</code> as specified in <a
+        <p>The 32-bit floating-point type <code>RGB32F</code> may also optionally
+	become available as a color-renderable format. Renderbuffers can be created
```
jdashg: Renderbuffers cannot be created in this format. Just remove that line from here.

zhenyao: Removed

zhenyao:
Merging myself since both @kenrussell  and @jdashg are both OK'ed.
Others, if you have more comments, feel free to let me know and I'll address in a followup.

extensions/EXT_color_buffer_half_float/extension.xml
```diff
@@ -37,14 +37,17 @@
         Renderbuffers can be created in these formats. These and textures
         created with <code>type = HALF_FLOAT_OES</code>, which will have one
         of these internal formats, can be attached to framebuffer object color
-        attachments for rendering.</p>
+        attachments for rendering. Implementations supporting this extension is
+	required to support rendering to <code>RGBA16F</code> format.
+	Applications must check framebuffer completeness to determine if
+	<code>RGA16F</code> is supported.
```
dmikis: RGB?

PR #2506, opened 2017-09-15
Test for clearBufferfv against back buffer of alpha:false canvas.

kenrussell:
Regression test for http://crbug.com/765729 .


kenrussell:

Test passes in Firefox. Fails in current Chrome, but passes with the bug fix.

@grorg FYI. If Safari switches to using an IOSurface for its back buffer it will need the associated workaround making this test pass, as well as others.

zhenyao:
LGTM

PR #2507, opened 2017-09-15
Improve/fix comments in multisampled-depth-renderbuffer-initialization

kainino0x:
Small fixes, follow up to #2418


zhenyao:
LGTM

PR #2510, opened 2017-09-19
Fix a typo in extension spec.

zhenyao:
@msc-


zhenyao:
Thanks for spotting this typo. Merging myself.

Issue #2508, opened 2017-09-16
Issues needing to be solved in WEBGL_video_texture proposal


kenrussell:
In #2470 many issues were raised by @msc- @tharkum @zhenyao and others. Opening this issue to capture them separately from that pull request.

In discussions with people working on Chrome Media and color-correct rendering in Chrome, the point has been made that the implicit conversion to an "RGB" color space, which will occur while sampling the video, should define that the result is extended sRGB. This handles wide gamut and HDR videos. Canvases are increasingly color managed in browsers, so the conversion from sRGB to the display color profile will be done by the browser's compositor afterward.

There are many more issues to be discussed. Please add to this discussion and propose pull requests addressing portions of them.

msc-:
OES_EGL_image_external says that the RGB samples are in whatever color space the video is in giving the examples of sRGB and Rec. 601. Do you really want to convert everything to extended sRGB or is it better to have a query so the app can determine the color space?

If you're willing to do conversion, maybe it is better to return samples in linear RGB space. That would make life easier for applications.

When I have time I'll copy my list of issues from #2470 to here. As for the HTMLElementTexture proposal, I'm on the fence about adding that here also or opening another issue. Opinions?

Speaking of that, @kenrussell wrote

> however, video elements are handled very differently than other DOM elements, and I think we are going to need special handling for them.

I think at the level of OES_EGL_image_external and related extensions, which is very generic, there is no difference. It gives you a texture object which is an opaque block of memory whose content is written by an external agent and could be updated at any time. To avoid tearing you need either (a) multiple buffers, e.g. an EGLStream, or (b) the application has to explicitly synchronize using locks and controlling the operation of the external agent. (a) is completely generic. Controlling the agent in (b) is specific to the agent type but is outside the scope of the external texture spec. Locking is also outside the scope.

Please give some examples of how the differences between these different DOM element types leak into the external texture seen by WebGL?

RafaelCintron:
Edge currently uses the [MediaEngine](https://msdn.microsoft.com/en-us/library/windows/desktop/hh447918(v=vs.85).aspx) APIs for video playback. Specifically, for Canvas (2D and WebGL), it uses the [TransferVideoFrame](https://msdn.microsoft.com/en-us/library/windows/desktop/hh448021(v=vs.85).aspx) API to extract video frames.

TransferVideoFrame does more than just "copy" the video into the output texture. Depending on the video, it also performs one or more of the following operations:
[...]

PR #2511, opened 2017-09-19
Add WEBGL_element_texture (#2508)

tharkum:
* Add new HTMLElementTexture object and new tokens for
       TEXTURE_ELEMENT_WEBGL texture target.


extensions/proposals/WEBGL_element_texture/extension.xml
```diff
@@ -0,0 +1,298 @@
...
+
+  <contact> <a href="https://www.khronos.org/webgl/public-mailing-list/">WebGL
+  working group</a> (public_webgl 'at' khronos.org) </contact>
+
+  <contributors>
+    <contributor>Byungseon Shin, LG Electronics</contributor>
+
+    <contributor>Andrei Volykhin, LG Electronics</contributor>
+
+    <contributor>Mark Callow, HI Corporation</contributor>
```
msc-: Please change to Edgewise Consulting. I left HI a couple of years ago.

extensions/proposals/WEBGL_element_texture/extension.xml
```diff
@@ -0,0 +1,298 @@
...
+  <number>NN</number>
+
+  <depends>
+    <api version="1.0"/>
+  </depends>
+
+  <overview>
+      <p>This extension defines a new object, the <code>HTMLElementTexture</code>,
+    that can be used to efficiently transfer a sequence of image frames from
+    a producer which out of control the WebGL into a WebGL texture.
```
msc-: "the WebGL" -> "of WebGL".

extensions/proposals/WEBGL_element_texture/extension.xml
```diff
@@ -0,0 +1,298 @@
...
+
+    <p>Sampling a WebGL texture will return an RGBA vector in the same
+    colorspace as the source image. If the source image is stored in YUV
+    (or some other basis) then the YUV values will be transformed to RGB values.
+    (But these RGB values will be in the same colorspace as the original image.
+    Colorspace here includes the linear or non-linear encoding of the samples.
+    For example, if the original image is in the sRGB color space then the RGB value
+    returned by the sampler will also be sRGB, and if the original image is stored
+    in ITU-R Rec. 601 YV12 then the RGB value returned by the sampler will be
+    an RGB value in the ITU-R Rec. 601 colorspace)</p>
```
msc-: As I wrote in #2508, this decision was made before sRGB support in OpenGL ES. If we continue with this approach a query for either the colorspace name or the actual transfer function will be needed so apps can convert to linear data for blending and can convert properly to the canvas color space.

We should consider having this extension return samples in linear space instead. It should only be a few more instructions in the injected shader code.

extensions/proposals/WEBGL_element_texture/extension.xml
```diff
@@ -0,0 +1,298 @@
...
+    <code>REQUIRED_TEXTURE_ELEMENT_IMAGE_UNITS_WEBGL</code>, and ActiveTexture
+    set to the texture unit to which the texture object is bound.</p>
+
+    <p><code>HTMLElementTexture</code> provides a commands for <em>latching</em> an
+    image frame into the consuming texture as its contents and retrieving additional
+    information as a timestamp or a transformation matrix.</p>
+
+    <p>This extension does not cover the details of how a <code>HTMLELementTexture</code>
+    object could be created from a producer element.
+    Different kinds of producers works differently and described in additional externsion
```
msc-: "works" -> "work".

"and described" -> "so object creation is described"

extensions/proposals/WEBGL_element_texture/extension.xml
```diff
@@ -0,0 +1,298 @@
...
+      <param name="target" type="GLenum"/>
+      <param name="pname" type="GLenum"/>
+      <code>REQUIRED_TEXTURE_ELEMENT_IMAGE_UNITS_WEBGL</code> is accepted
+      as the <code>pname</code> parameter of <code>getTexParameter*()</code>
+    </function>
+  </newtok>
+
+  <samplecode xml:space="preserve">
+
+    <p> This a fragment shader that samples a element texture.</p>
```
msc-: "a element" -> "an element"

extensions/proposals/WEBGL_element_texture/extension.xml
```diff
@@ -0,0 +1,298 @@
...
+  </idl>
+
+  <!-- new functions -->
+
+  <newfun>
+    <p>On <code>HTMLElementTexture</code>:</p>
+
+    <function name="updateTexImage" type="void">
+      Update the texture image to the most recent frame from the image stream
+      and release the previously-held frame.
```
msc-: Change the order. Put release first. E.g, "release the previously-held frame, if any, and update the texture image to the most recent frame."

extensions/proposals/WEBGL_element_texture/extension.xml
```diff
@@ -0,0 +1,298 @@
...
+    varying vec2 v_texCoord;
+
+    uniform samplerElementWEBGL uSampler;
+
+    void main(void) {
+      gl_FragColor = texture2D(uSampler, v_texCoord);
+    }
+    </pre>
+
+    <p>This shows application that renders HTML*Element using proposed extension.</p>
```
msc-: "renders HTML*Element" to "renders an HTML*Element"

extensions/proposals/WEBGL_element_texture/extension.xml
```diff
@@ -0,0 +1,298 @@
...
+    var customElement = document.getElementById("...");
+    var customTexture = gl.createTexture();
+    var elementTexture;
+
+    function initialize() {
+      var ext = gl.getExtension('WEBGL_element_texture');
+      if (ext === null)
+        return;
+
+      elementTexture = ext.create*Texture(customElement, customTexture);
```
msc-: Do we really need other extensions? What would create*Texture need to do differently for different types of HTMLElement? If we really need them, this example should show calling getExtension for one of them and using a create function from the extension.

extensions/proposals/WEBGL_element_texture/extension.xml
```diff
@@ -0,0 +1,298 @@
...
+      gl.bindTexture(ext.TEXTURE_ELEMENT_WEBGL, customTexture);
+      elementTexture.updateTexImage();
+      gl.bindTexture(ext.TEXTURE_ELEMENT_WEBGL, null);
+    }
+
+    function render() {
+        .....
+
+        gl.activeTexture(gl.TEXTURE0);
+        gl.bindTexture(ext.TEXTURE_ELEMENT_WEBGL, customTexture);
```
msc-: It seems unnecessary to keep binding and unbinding this customTexture. Unless the underlying implementation requires it, we shouldn't do it.

kenrussell:
Attempting to abstract between handling videos, and any other kind of rendered HTML element, is in my opinion the wrong direction to go with these extensions.

Videos are treated, and rendered, very differently than other elements. The hardware-accelerated video decoding paths tend to produce YUV format textures. The rendering of arbitrary other DOM elements does not; that rendering process produces RGB textures. Further, DOM layout and rendering are expensive operations, and an API designed to latch in a rendered texture from the DOM should be structured differently – more asynchronously – than one designed to work with videos.

Additionally, if a user's single shader is intended to abstract over sampling a YUV or RGB texture, that's going to introduce either significant dynamic branching in the shader – costing run-time performance – or require dynamic shader recompilation by the browser depending on the type of texture being sampled, which will add significant complexity to WebGL implementations.

I suggest that you make the improvements to the precision of the WEBGL_video_texture extension proposal directly on that extension proposal, and abandon this one. I don't think we should rewrite WEBGL_video_texture on top of this proposal.

xingri:
@kenrussell < Thanks for the comments. We will discuss it internally and update soon.

msc-:
@kenrussell lists 2 issues:

1. YUV vs RGB
2. Synchronous vs asynchronous

The obvious solution to no. 1 is to use a regular sampler for non-video elements so we should move the special texture target and sampler to the video texture extension.

Note the dynamic branching or recompilation issue can arise with videos of different formats.  Therefore it is important for the spec. to explain this as, in the proposed element_texture spec. We might even consider a stronger recommendation to use a separate shader per video.

On no. 2, the only possible synchronous part of this spec. is `updateTexImage()` which might have to wait for a frame to finish decoding. There is an API for asynchronous notification of frame available.

With all the browsers seemingly moving to a multi-process architecture isn't there a good chance that 2d canvases and iframes might be rendered asynchronously to the WebGL canvas? In that case you'll need the same buffering and possible wait-for-completion as video and an `updateTexImage()` function.

I still think it is possible to find a useful common base extension.

msc-:
> I suggest that you make the improvements to the precision of the WEBGL_video_texture extension proposal directly on that extension proposal, and abandon this one. I don't think we should rewrite WEBGL_video_texture on top of this proposal.

If we do go ahead with a video-only solution, I suggest replacing the content of WEBGL_video_texture with this proposal with the generic HTMLElement parts removed. This is a far more complete proposal than WEBGL_video_texture.

extensions/proposals/WEBGL_element_texture/extension.xml
tharkum: I used information from old extension, i will change it.

extensions/proposals/WEBGL_element_texture/extension.xml
tharkum: Ok, i will remove this part from proposal and instead add issue for linear vs non-linear color space.

extensions/proposals/WEBGL_element_texture/extension.xml
tharkum: Unfortunately i forgot to add information how we should use bindTexture + updateTexImage().

I think updateTexImage() shouldn't latch the most recent frame from image stream if  our linked WebGL texture isn't bound to new texture target in active texture unit currently or need to use the SurfaceTexture way (it will implicitly bind its texture to the TEXTURE_VIDEO_WEBGL texture target.).

extensions/proposals/WEBGL_element_texture/extension.xml
tharkum: I added this code to mention how we will use this texture with new target in the rendering stage.
No point to keep binding and unbinding this texture all time. I will remove these lines.

tharkum:
Althrough common extension  for arbitary HTML elements will allow to make common interface for them it will make everything much complex.
Canvas and Iframe (in future) wouldn't use new texture target, so WEBGL_element_texture extension is not needed so much.

I will upload new WEBGL_video_texture on base of this proposal and we could focus on async and color space issues.

PR #2512, opened 2017-09-19
Improve EXT_color_buffer_half_float extension spec.

zhenyao:
Per Mark Callow suggestions.


zhenyao:
@msc- Please take a look.

msc-:
Thanks @zhenyao.

PR #2513, opened 2017-09-19
Adding back runFramebufferTest() in oes-texture-half-float.html

zhenyao:


zhenyao:
@kenrussell Can you review? This test was spec compliant so get filtered out in the last pull request. However, it still has values, like ReadPixels, wasn't tested anywhere else.  So fix it and add it back.

kenrussell:
Looks good Mo.

Note that this caught a bug in Chrome on the Nexus 5X phone.

Merging now. @jdashg FYI.

PR #2515, opened 2017-09-20
Add test case for texture fetches inside conditionals.

kenrussell:
This is a reduced test case for a bug reported by Figma. It produces an HLSL compiler error on ANGLE's Direct3D 9 backend, but doesn't (yet) produce incorrect rendering results.


kenrussell:
I'd like to merge this now and refine it if Figma can suggest changes which will make it start failing on ANGLE's D3D9 backend. @zhenyao @kainino0x @jdashg @RafaelCintron in particular.

zhenyao:
LGTM after the fact

zhenyao:
Didn't realize it's not merged yet. Merging now.

PR #2518, opened 2017-09-21
Add extension WEBGL_texture_source_iframe proposal

zhenyao:


zhenyao:
Wrote on the plane to Chicago.

@kenrussell Please include it in tomorrow's discussion if it's not too late

greggman:
There's also still the issue of being able to read visited links by their color either directly (gl.readPixels, etc...) or indirectly through shader timing.

I don't know if this would be a solution and it would be a non-WebGL thing but what if you could create a "safe" iframe like

      <iframe safe="true" ...>

Where `safe=true` means

* src must be same origin
* cross origin non-CORS images just are not allowed period. 
* the CSS :visited selector becomes a no-op
* etc...

In other words, rather than having to check everything about iframe fits the requirements instead don't allow the iframe to have issues in the first place.

zhenyao:
I think more along the lines of limiting read back to iframe textures and anything these textures may contaminate.

extensions/proposals/WEBGL_texture_source_iframe/extension.xml
```diff
@@ -0,0 +1,113 @@
...
+  void setEventForwarding(function(Event));
+};
+  </idl>
+  <newfun>
+    <function name="bindTextureSource" type="Promise">
+      <param name="target" type="GLenum"/>
+      <param name="iframe" type="HTMLIFrameElement"/>
+      <p>
+	Connect an iframe to the texture currently bound to <code>target</code>. Once the iframe
+	rendering is ready to be blitted to the texture, return a Promise. If <code>iframe</code>
```
msc-: I thought we wanted to avoid blitting and get the browser to render directly into the GPU memory that will be sampled.

zhenyao: I didn't mean for the spec to be specific about implementation.  If blitting is understood as one copy, then we should re-word it.  Chrome implemented it as one copy and is efficient enough.  Of course we will think about zero copy implementation, but the spec should not get into that.

zhenyao:
Merging this now for easier discusssion

extensions/proposals/WEBGL_texture_source_iframe/extension.xml
msc-: How about "once the iframe rendering is ready for sampling"?

I agree the spec should not be specific about implementation and yes blitting is understood as one copy.

msc-:
You also need to document the texture origin or offer a getTransform function like the HTMLElement proposal. That is when you sample at (s,t) = (0,0) do you see the logical lower left or upper left of the iframe rendering?

zhenyao:
During today's khronos f2f, we decided to rid all transformations of the iframe through the dom tree (CSS, scales, etc) when it's attached to a texture. So (s,t)=(0,0) will be the same for all iframes.

Issue #2514, opened 2017-09-20
conformance/glsl/bugs/sampler-array-struct-function-arg.html in wrong directory?


kenrussell:
This test:
https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/conformance/glsl/bugs/sampler-array-struct-function-arg.html

calls:
wtu.create3DContext(canvas, undefined, 2);

but it's in the WebGL 1.0 test directory and marked --min-version 1.0.4.

Is it supposed to be in conformance2/glsl3/ and marked --min-version 2.0.1?

zhenyao:
I don't think this is WebGL2 specific.  We can just get rid of the version in wtu.create3DContext() call.

PR #2521, opened 2017-09-25
Remove incorrect context version from sampler array test

Oletus:
This test is not intended to be specific to WebGL 2.0.

Fixes issue #2514.


kenrussell:
Thanks @Oletus for fixing this.

PR #2523, opened 2017-09-26
Add test case performing repeated blitFramebuffer calls without clears.

kenrussell:
Test case for crbug.com/768969 .

### DIFF
--- a/sdk/tests/conformance2/rendering/blitframebuffer-test.html
+++ b/sdk/tests/conformance2/rendering/blitframebuffer-test.html
@@ -52,15 +52,56 @@ if (!gl) {
     testFailed("WebGL context does not exist");
 } else {
     testPassed("WebGL context exists");
+    blit_framebuffer_repeated();
     blit_framebuffer_feedback_loop();
     blit_framebuffer_multisampling_srgb();
 }
 
+function blit_framebuffer_repeated() {
+    debug("");
+    debug("This test verifies repeated calls to blitFramebuffer.");
+
+    // Create offscreen fbo and its color attachment.
+    var tex_2d = gl.createTexture();
+    gl.bindTexture(gl.TEXTURE_2D, tex_2d);
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA8, width, height, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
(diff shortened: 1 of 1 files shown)


kenrussell:
Note: this test indicates there's no bug in Chromium.

kenrussell:
@zhenyao @jdashg in particular please take a look.

I suspect this is already covered by other conformance tests but didn't confirm that.

zhenyao:
LGTM

sdk/tests/conformance2/rendering/blitframebuffer-test.html
```diff
@@ -52,15 +52,56 @@
...
 }
 
+function blit_framebuffer_repeated() {
+    debug("");
+    debug("This test verifies repeated calls to blitFramebuffer.");
+
+    // Create offscreen fbo and its color attachment.
+    var tex_2d = gl.createTexture();
+    gl.bindTexture(gl.TEXTURE_2D, tex_2d);
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA8, width, height, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
```
jdashg: Slight preference for texStorage in webgl2-only code.

sdk/tests/conformance2/rendering/blitframebuffer-test.html
```diff
@@ -52,15 +52,56 @@
...
+    wtu.setFloatDrawColor(gl, [ 1.0, 0.0, 0.0, 1.0 ]);
+    wtu.drawUnitQuad(gl);
+    gl.bindFramebuffer(gl.READ_FRAMEBUFFER, fb);
+    gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, null);
+    gl.blitFramebuffer(0, 0, width, height, 0, 0, width, height, gl.COLOR_BUFFER_BIT, gl.NEAREST);
+    gl.bindFramebuffer(gl.READ_FRAMEBUFFER, null);
+    wtu.checkCanvas(gl, [ 255, 0, 0, 255 ], "should be red at first");
+
+    gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, fb);
+    gl.bindFramebuffer(gl.READ_FRAMEBUFFER, null);
```
jdashg: Skip setting READ_FB if we aren't going to read from it.
Generally, choose to either default to using FRAMEBUFFER except for calls to BlitFramebuffer, or only set the READ_/DRAW_FRAMEBUFFER that you need for the next command.

sdk/tests/conformance2/rendering/blitframebuffer-test.html
kenrussell: Fixing in #2569.

sdk/tests/conformance2/rendering/blitframebuffer-test.html
kenrussell: Fixed throughout this test in #2569.

PR #2529, opened 2017-10-12
Add test for vector/scalar arithmetic inside a for loop

Oletus:
This exposes a bug in the NVIDIA GL driver.


zhenyao:
Thanks @Oletus

PR #2530, opened 2017-10-14
WEBGL_texture_source_iframe: some rewording about promises

dmikis:
I've done a bit of rewording regarding promises to be a bit closer to TAG's guide, i.e.:

- added `<void>` to signify that no additional data's passed with promise;
- replaces "returns promise" with "promise will be fulfilled" as it's a bit more accurate.

/cc @zhenyao


zhenyao:
Thanks!

PR #2531, opened 2017-10-17
Add more tests for vector/scalar arithmetic inside a loop

Oletus:
Add more variations of shaders that reproduce the same bug. This adds
better coverage for a workaround under development for ANGLE.

Some more complex variations are put to a separate test file - the
workaround in ANGLE will not be sophisticated enough to handle them,
but they should be fixed by a future driver update.


zhenyao:
LGTM
Thank you

PR #2533, opened 2017-10-21
Make test harness result reporter log success message to console

gofr:
The wrapped logging function it uses only takes one argument.
Concatenate the message to the success state string rather
than passing it as an unused second argument.


zhenyao:
LGTM
Thanks for catching this bug

PR #2534, opened 2017-10-26
Clarify WEBGL_multiview spec

Oletus:
The multiview spec is now written against the latest revision of the
native OVR_multiview spec. The native OVR_multiview spec has been
tightened to specify more error conditions, so some of the addendums
in the WebGL spec are not necessary anymore.

Formatting and examples are also improved.

This doesn't introduce any major functional changes.


Oletus:
Can we merge this? The changes here are mostly orthogonal to what we decide to do with opaque framebuffers.

jdashg:
Yep, sorry for the delay.

PR #2535, opened 2017-10-27
Add tests for switch/case validation corner cases

Oletus:
These tests expose bugs in ANGLE, most of which have been already
fixed in ANGLE TOT.


zhenyao:
LGTM

PR #2536, opened 2017-10-27
Added a test to ensure preserveDrawingBuffer isn't leaking resources

toji:
This test provoks a crash on Samsung Galaxy S6 devices running Android 6.0.1. Other devices seem to handle it fine.

### DIFF
--- a/sdk/tests/conformance/rendering/00_test_list.txt
+++ b/sdk/tests/conformance/rendering/00_test_list.txt
@@ -23,6 +23,7 @@ point-size.html
 --min-version 1.0.4 point-specific-shader-variables.html
 --min-version 1.0.3 point-with-gl-pointcoord-in-fragment-shader.html
 --min-version 1.0.3 polygon-offset.html
+preservedrawingbuffer-leak.html
 --min-version 1.0.2 simple.html
 triangle.html
 line-loop-tri-fan.html
--- a/sdk/tests/conformance/rendering/00_test_list.txt
+++ b/sdk/tests/conformance/rendering/00_test_list.txt
@@ -23,7 +23,7 @@ point-size.html
 --min-version 1.0.4 point-specific-shader-variables.html
 --min-version 1.0.3 point-with-gl-pointcoord-in-fragment-shader.html
 --min-version 1.0.3 polygon-offset.html
-preservedrawingbuffer-leak.html
+--min-version 1.0.4 preservedrawingbuffer-leak.html
 --min-version 1.0.2 simple.html
 triangle.html
 line-loop-tri-fan.html
--- a/sdk/tests/conformance/rendering/preservedrawingbuffer-leak.html
+++ b/sdk/tests/conformance/rendering/preservedrawingbuffer-leak.html
@@ -1,7 +1,7 @@
 <!--
 
 /*
-** Copyright (c) 2012 The Khronos Group Inc.
+** Copyright (c) 2017 The Khronos Group Inc.
 **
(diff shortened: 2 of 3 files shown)


sdk/tests/conformance/rendering/00_test_list.txt
```diff
@@ -23,6 +23,7 @@ point-size.html
 --min-version 1.0.4 point-specific-shader-variables.html
 --min-version 1.0.3 point-with-gl-pointcoord-in-fragment-shader.html
 --min-version 1.0.3 polygon-offset.html
+preservedrawingbuffer-leak.html
```
zhenyao: please add --min-version 1.0.4

sdk/tests/conformance/rendering/preservedrawingbuffer-leak.html
```diff
@@ -0,0 +1,60 @@
+<!--
+
+/*
+** Copyright (c) 2012 The Khronos Group Inc.
```
zhenyao: nit:2017

sdk/tests/conformance/rendering/preservedrawingbuffer-leak.html
```diff
@@ -0,0 +1,60 @@
...
+<head>
+<meta charset="utf-8">
+<title>WebGL PreserveDrawingBuffer Leak Test</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"> </script>
+<script src="../../js/tests/iterable-test.js"> </script>
+</head>
+<body>
+<canvas id="example" width="4096" height="4096" style="width: 128px; height: 128px;"></canvas>
```
kenrussell: Was it possible to provoke the crash with a smaller canvas?

sdk/tests/js/tests/iterable-test.js
```diff
@@ -162,12 +162,40 @@ IterableTest = (function() {
...
+    function test() {
+      var x = Math.random() * gl.drawingBufferWidth;
+      var y = Math.random() * gl.drawingBufferHeight;
+      var width = Math.random() * (gl.drawingBufferWidth - x);
+      var height = Math.random() * (gl.drawingBufferHeight - y);
+
+      gl.enable(gl.SCISSOR_TEST);
+      gl.scissor(x, y, width, height);
+      gl.clearColor(Math.random(), Math.random(), Math.random(), 1.0);
+      gl.clear(gl.COLOR_BUFFER_BIT);
```
kenrussell: This is great. Was it possible to provoke the crash with fully user-level operations rather than relying on preserveDrawingBuffer:true ? It would be good to know which exact operation seems to be buggy.

sdk/tests/conformance/rendering/preservedrawingbuffer-leak.html
toji: Yes, but the test took more iterations to provoke the failure. It would fail with a 2k canvas but at 100 iterations it was a flaky failure that may have side effects on other tests if it passed on the problematic device. At 4k it fails reliably in the 20-30 iteration range, so I set the test iteration count to 50 to be safe.

We can lower the canvas size and up the iteration count if you'd prefer.

sdk/tests/js/tests/iterable-test.js
toji: Haven't yet tried, sorry.

PR #2541, opened 2017-11-07
Add a test for varying struct with an inline definition

Oletus:
This test exposes a bug that was recently fixed in ANGLE.


zhenyao:
LGTM

PR #2542, opened 2017-11-12
Add test for scissor rectangle maintenance bug during implicit clears.

kenrussell:
Regression test for http://crbug.com/778770 .


kenrussell:
@zhenyao @kainino0x please review.

Test already passes in Safari and Firefox. Fails in Chrome without the fix for the bug.

PR #2543, opened 2017-11-13
Adds a test for rendering directly after resizing the canvas

greggman:
This test passes on Firefox on my 2014 MBP macOS 10.13 but fails on Chrome 62-64


kenrussell:
Thanks for this contribution. It turns out that drawingbuffer-static-canvas-test.html in the same directory also caught this bug, but it seems fine to add this targeted test too.

A couple of tiny issues, the copyright date is off and ideally the test wouldn't rely on support for large points, but I'll look into doing an update after merging this.

Merging now. Thanks again for the test.

FYI @grorg @jdashg @zhenyao @kainino0x @RafaelCintron

kenrussell:
Note that this was filed as http://crbug.com/775202 a while back.

PR #2544, opened 2017-11-13
Simplify test: eliminate 2D canvas, draw triangles instead of point.

kenrussell:
Fix latent bug in wtu.setupSimpleColorProgram. Fix copyright date.

Follow-on to #2543.

Acts as regression test for http://crbug.com/775202 .

### DIFF
--- a/sdk/tests/conformance/canvas/render-after-resize-test.html
+++ b/sdk/tests/conformance/canvas/render-after-resize-test.html
@@ -1,7 +1,7 @@
 <!--
 
 /*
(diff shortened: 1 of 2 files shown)


kenrussell:
@zhenyao @kainino0x @jdashg @grorg @RafaelCintron @Oletus FYI. Tested this locally and verified that it still reproduces the bug on macOS 10.13 on NVIDIA GPUs.

@greggman FYI. Hopefully this doesn't break your intent with the test, but it simplifies the OpenGL calls being made.

Merging myself.

PR #2545, opened 2017-11-14
Add test that initializes an array inside a loop

Oletus:
This attempts to reproduce a driver bug seen on Adreno 418.


Oletus:
@zhenyao , could you try to see if this test fails on Adreno 418 as intended (for example Nexus 5X)? On Adreno 505 that I had available to me the test passed.

zhenyao:
OK, let's merge this to make testing easier. Thanks Olli.

zhenyao:
I tested on Nexus 5. Unfortunately the added test passed.

PR #2546, opened 2017-11-14
Fix fencepost errors covered up by wtu.checkCanvasRect.

kenrussell:
Follow-on to #2544 and #2543.


kenrussell:
Verified locally that the bug's still reproduced in Chrome and Safari on the affected hardware (Macs with NVIDIA GPUs running 10.13). Merging myself.

PR #2549, opened 2017-11-20
Simplify render-after-resize-test. Draw single unit quad.

kenrussell:
Exposes the issue more simply. Follow-on to #2546.

Still passes in Firefox, fails in Chrome and Safari, on macOS 10.13 on
MacBook Pro with NVIDIA GPU.


kenrussell:
FYI @grorg @jdashg @zhenyao @kainino0x @RafaelCintron @greggman . Merging myself.

greggman:
LGTM. Thank you for making it simpler

PR #2550, opened 2017-11-22
Accept INVALID_OPERATION for texSubImage2D when format/type differs

derry19:
INVALID_OPERATION should also be an acceptable result because texSubImage2d could be called with format and type that differ from format and type that was set previously by texImage2D.


jdashg:
It's true that it could generate either error, and GL says that that whichever error is hit first is reported, dropping all subsequent errors. Order of validation would change which error is returned here.

kenrussell:
Thanks, this looks fine -- could you please apply this change to the conformance suite snapshots too, under conformance-suites/ ? The ones that need updating are 1.0.1, 1.0.2, 1.0.3 and 2.0.0. (These are supposed to keep running if there was a semantic bug in earlier releases.)

Thanks.

kenrussell:
Thanks for this update and thanks @jdashg for pinging me. Looks good.

PR #2552, opened 2017-11-29
Tests should accept multiple valid error conditions

derry19:
These tests should also accept INVALID_OPERATION for the same reason we discussed in https://github.com/KhronosGroup/WebGL/pull/2550

### DIFF
--- a/conformance-suites/1.0.1/conformance/textures/tex-input-validation.html
+++ b/conformance-suites/1.0.1/conformance/textures/tex-input-validation.html
@@ -221,11 +221,11 @@ testCases =
   [ {target: gl.TEXTURE_2D,
      format: 0x1903, // GL_RED
      type: gl.UNSIGNED_BYTE,
-     expectedError: gl.INVALID_ENUM},
+     expectedError: [gl.INVALID_ENUM, gl.INVALID_OPERATION] },
     {target: gl.TEXTURE_2D,
      format: gl.RGBA,
      type: gl.BYTE,
-     expectedError: gl.INVALID_ENUM},
+     expectedError: [gl.INVALID_ENUM, gl.INVALID_OPERATION] },
     {target: gl.TEXTURE_2D,
      format: gl.RGBA,
      type: gl.UNSIGNED_BYTE,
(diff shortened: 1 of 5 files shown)


zhenyao:
I am not sure I agree to this change. My understanding is, INVALID_OPERATION is only generated if format and type are both valid, but their combination is invalid. So INVALID_OPERATION isn't applicable here because format is invalid.

derry19:
Yes, you are right. INVALID_OPERATION should not be an acceptable result in those cases. Thanks for the help.

jdashg:
I still don't really agree. I really hate quibbling over which error to return, since we give text errors now, which basically obsoletes the need for error enums.

Internally in Firefox, we found it's fastest to first look up the pair. If the pair isn't valid, we then check if either is INVALID_ENUM, otherwise we return INVALID_OP.

While this is ok, I guess, it does mean we have to maintain an 'is this enum valid' path, which is unnecessary otherwise. I would prefer to drop this codepath.

zhenyao:
I double checked the spec, the INVALID_ENUM error isn't explicitly specified under TexImage*. The text describing the error is, "Specifying a combination of values for format, type, and internalformat that is not listed as a valid combination in tables 3.2 or 3.3 generates the error INVALID_OPERATION."

If (and only if) Firefox's behavior doesn't break any dEQP tests, then I can agree to making this change.

RafaelCintron:
I agree with re-opening this pull request and accepting this change. 

In this test case, there are two errors.  One error is that the type and format pass to texSubImage2D do not match what was given to texImage2D.  The second is that invalid enums are passed to texSubImage2D.  If a test is doing two things wrong, we should accept either error condition and not mandate the order of error checking.

FYI, in Edge, we let texImage2D do all of the calculus involved with rationalizing format, types and enabled extensions.  texSubImage2D merely checks whether the format and the type match the previous successful texImage2D calls (if any) and forwards the rest of the logic to ANGLE.  So for this particular test, the enums never got to be checked for validity.

RafaelCintron:
@jdashg and @kenrussell , what are your thoughts?

zhenyao:
Let's merge this to unblock Firefox/Edge conformance

PR #2557, opened 2017-12-06
Fix a bug in tcuFuzzyImageCompare.bilinearSample

jchen10:
This is a typo compared with the native dEQP source code.


jchen10:
The native code:
https://android.googlesource.com/platform/external/deqp/+/master/framework/common/tcuFuzzyImageCompare.cpp#140

jchen10:
@kenrussell  A minor fix, PTAL!

kenrussell:
@zhenyao @kainino0x @jdarpinian @jdashg FYI.

jchen10:
Thanks for reviewing!
I just caught it by chance. Not sure how bad it affected.

PR #2559, opened 2017-12-08
Add a regression test for ANGLE bug (crbug.com/793307).

zhenyao:
@kainino0x @kenrussell @jdashg PTAL
@jdarpinian FYI

### DIFF
--- a/sdk/tests/conformance2/uniforms/draw-with-uniform-blocks.html
+++ b/sdk/tests/conformance2/uniforms/draw-with-uniform-blocks.html
@@ -89,12 +89,20 @@ function runTest() {
     gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, indexBuf);
     gl.bufferData(gl.ELEMENT_ARRAY_BUFFER, new Int16Array([ 0, 1, 2, 2, 1, 3 ]), gl.STATIC_DRAW);
 
+    var uboDataSize = gl.getActiveUniformBlockParameter(
+        program, uboIndex, gl.UNIFORM_BLOCK_DATA_SIZE);
+    if (uboDataSize == 0) {
+        testFailed("uniform block data size invalid");
+        return;
+    }
+
     wtu.glErrorShouldBe(gl, gl.NO_ERROR, "No errors from setup");
 
     debug("draw lower triangle - should be red");
     var uboBuf1 = gl.createBuffer();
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/uniforms/draw-with-uniform-blocks.html
```diff
@@ -0,0 +1,132 @@
...
+    var indexBuf = gl.createBuffer();
+    gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, indexBuf);
+    gl.bufferData(gl.ELEMENT_ARRAY_BUFFER, new Int16Array([ 0, 1, 2, 2, 1, 3 ]), gl.STATIC_DRAW);
+
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "No errors from setup");
+
+    debug("draw lower triangle - should be red");
+    var uboBuf1 = gl.createBuffer();
+    gl.bindBufferBase(gl.UNIFORM_BUFFER, 0, uboBuf1);
+    gl.bufferData(gl.UNIFORM_BUFFER, new Float32Array([ 1, 0, 0, 1 ]), gl.STATIC_DRAW);
```
kenrussell: Section 2.12.6.5 "Uniform Buffer Object Bindings" of the ES 3.0 spec says:

"When executing shaders that access uniform blocks, the binding point corre- sponding to each active uniform block must be populated with a buffer object with a size no smaller than the minimum required size of the uniform block (the value of UNIFORM_BLOCK_DATA_SIZE)."

I don't see alignment of these buffer objects up to UNIFORM_BLOCK_DATA_SIZE. Am I missing something or is the test case not correct on all platforms?

zhenyao: You are right. We should use the queried size to be on the safe side, although usually if the data is aligned to 4 and fit the UB in the shader, that's fine.

zhenyao: Revised. PTAL again.

kenrussell:
Looks good – does this still reproduce the bug on the affected platform?

Could you please update the pull request's description with a link to the Chromium or ANGLE bug?

zhenyao:
Yes, the latest test fails on Windows M63, but pass on Mac.

kenrussell:
Thanks Mo for producing and fixing up the test. Merged.

PR #2560, opened 2017-12-15
 Relax restrictions on transform feedback buffer use

### LINKED ISSUE #2346: WebGL2 Buffer restrictions overly restrictive?

jdarpinian:
Instead of generating an error at buffer binding time, generate an error only when a buffer is actually used simultaneously for transform feedback and other purposes, such as in a vertex attrib for a draw call. Fixes #2346


sdk/tests/conformance2/buffers/buffer-type-restrictions.html
```diff
@@ -88,16 +88,6 @@
...
-
-  if ((firstTarget == gl.TRANSFORM_FEEDBACK_BUFFER && secondTarget != gl.TRANSFORM_FEEDBACK_BUFFER) ||
-      (firstTarget != gl.TRANSFORM_FEEDBACK_BUFFER && secondTarget == gl.TRANSFORM_FEEDBACK_BUFFER)) {
-    bind(firstBindFn, firstTarget, buffer);
-    bind(secondBindFn, secondTarget, buffer);
-
-    var message = "Binding buffer first with " + firstBindFn + " to gl." + firstTargetStr
-    + " and simultaneously binding buffer with " + secondBindFn + " to gl." + secondTargetStr + " should FAIL";
-    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, message);
-  }
```
kenrussell: To avoid having the WebGL 2.0.0 conformance tests break on newer browsers, can you make this same change in the conformance-suites/2.0.0/ directory?

sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
```diff
@@ -0,0 +1,137 @@
...
+<script id="fshader" type="x-shader/x-fragment">#version 300 es
+precision mediump float;
+out vec4 dummy;
+void main() {
+  dummy = vec4(1);
+}
+</script>
+<script>
+"use strict";
+description("This test verifies using the default transform feedback object");
```
kenrussell: Do browsers' WebGL 2.0 implementations currently allow using the default transform feedback object, or do they fail?

The reason I ask is that WebGL doesn't allow modification of the default texture object, and we had considered doing the same for transform feedback.

If it's possible to make all of the queries against the default transform feedback object (like getIndexedParameter), seems OK to allow the default transform feedback to be used, rather than requiring one to be bound by the user.

sdk/tests/conformance2/transform_feedback/simultaneous_access.html
```diff
@@ -0,0 +1,201 @@
...
+  gl.bufferData(gl.COPY_WRITE_BUFFER, 16, gl.STATIC_DRAW);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "transform feedback is active");
+  gl.endTransformFeedback();
+}
+
+function runTexImageTest() {
+  var tex = gl.createTexture();
+  gl.bindTexture(gl.TEXTURE_2D, tex);
+
+  gl.texImage2D
```
kenrussell: Incomplete test.

jdarpinian: Good catch, removed. I hadn't noticed because the code still runs like this :open_mouth:

sdk/tests/conformance2/transform_feedback/default_transform_feedback.html
jdarpinian: Oh, I pulled this test in from #2347 because it was a pending test that was failing because of this bug, but on further consideration I should wait for #2347 to be resolved before committing this.

sdk/tests/conformance2/buffers/buffer-type-restrictions.html
jdarpinian: Done.

sdk/tests/conformance2/transform_feedback/simultaneous_access.html
```diff
@@ -0,0 +1,201 @@
...
+
+  // this should fail because the transform feedback's buffer #0 and the
+  // srcVAO2's buffer #0 are the same buffer
+  runFeedback(gl, prog, srcVAO2, tf);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "buffer used as vertex attrib and tf simultaneously");
+
+  gl.bindBuffer(gl.ARRAY_BUFFER, dstBuffer);
+  wtu.checkFloatBuffer(gl, gl.ARRAY_BUFFER, expected, "should be the same as before as nothing has executed");
+
+  gl.bindBufferBase(gl.UNIFORM_BUFFER, 0, dstBuffer);
```
kainino0x: It's still bound as `ARRAY_BUFFER` at this point, right? Shouldn't it be unbound from `ARRAY_BUFFER` so that doesn't interfere with the test of `UNIFORM_BUFFER`? Same in a few places below.

kainino0x: Not sure how fragile this test is (for false positives), but to make it less fragile you could possibly make new buffers every time.

specs/latest/2.0/index.html
```diff
@@ -3292,14 +3292,20 @@ <h3><a name="BUFFER_OBJECT_BINDING">Buffer Object Binding</a></h3>
...
+        Attempting to read from or write to a buffer which is bound to a transform feedback buffer binding
+        point, via any mechanism other than transform feedback itself (e.g. as a vertex array in a draw
+        call), fails with an INVALID OPERATION error. There is one exception to this rule: bufferData,
+        bufferSubData, getBufferSubData, and copyBufferSubData are allowed as long as transform feedback
+        is not active.
     </p>
 
     <div class="note rationale">
-        This restriction has been added to prevent undefined values when reading from or writing to a buffer object that is simultaneously bound to a transform feedback buffer binding point and elsewhere in the GL. An example of such a situation is <code>ReadPixels</code> to a pixel buffer object binding point. See <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.15.2">OpenGL ES 3.0.4 &sect;2.15.2</a>)</span>.
+        If an error were not generated, the values read or written would be undefined. <span class="gl-spec">
```
kainino0x: Not necessary, but: could be nice to keep some of the rationale that was there before (like the example) (updated to the new semantics).

sdk/tests/conformance2/transform_feedback/simultaneous_access.html
```diff
@@ -0,0 +1,201 @@
...
+  gl.bindBufferBase(gl.TRANSFORM_FEEDBACK_BUFFER, outLoc, dstBuffer);
+  // this binds the default (id = 0) TRANSFORM_FEEBACK buffer
+  gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, null);
+
+  runFeedback(gl, prog, srcVAO1, tf);
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "draw should be successful");
+
+  const expected = [2, 4, 6, 8];
+  gl.bindBuffer(gl.ARRAY_BUFFER, dstBuffer);
+  wtu.checkFloatBuffer(gl, gl.ARRAY_BUFFER, expected);
```
zhenyao: As I mentioned below, this should generate INVALID_OPERATION.

specs/latest/2.0/index.html
```diff
@@ -3292,14 +3292,20 @@ <h3><a name="BUFFER_OBJECT_BINDING">Buffer Object Binding</a></h3>
...
+
     <p>
-        In addition, a buffer can not be simultaneously bound to a transform feedback buffer binding point
-        and another buffer binding point in the WebGL 2.0 API. Attempting to violate this rule generates an
-        <code>INVALID_OPERATION</code> error, and the state of the binding point will remain untouched.
+        Attempting to read from or write to a buffer which is bound to a transform feedback buffer binding
+        point, via any mechanism other than transform feedback itself (e.g. as a vertex array in a draw
+        call), fails with an INVALID OPERATION error. There is one exception to this rule: bufferData,
+        bufferSubData, getBufferSubData, and copyBufferSubData are allowed as long as transform feedback
+        is not active.
```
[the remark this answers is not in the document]
jdarpinian: You're right, I didn't think about it this way before. The language in the GL spec is a little ambiguous, but this may be the most reasonable interpretation. I will rewrite the spec language and update the tests for this.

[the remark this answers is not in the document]
jdashg: Specifically the way we validate this in Firefox is each buffer has these two members:
    size_t mTFBindCount;
    size_t mNonTFBindCount;
Any operation read or write to the buffer while `mTFBindCount && mNonTFBindCount` should error.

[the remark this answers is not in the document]
jdarpinian: @jdashg Do you update the counts on each buffer when VAOs or TF objects are bound/unbound, or only when the buffer is added/removed from a VAO or TF object?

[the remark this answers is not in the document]
jdashg: A 'background' TFO does not count towards buffer TF bindings currently, so TFO rebinds update the counts.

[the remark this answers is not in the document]
jdarpinian: Are VAOs treated the same way?

jdarpinian:
OK, I've updated the spec language and tests to match Firefox's behavior. PTAL.

specs/latest/2.0/index.html
```diff
@@ -3320,6 +3304,22 @@ <h3><a name="COPYING_BUFFERS">Copying Buffers</a></h3>
         Same as with <a href="#BUFFER_OBJECT_BINDING">Buffer Object Binding</a> restrictions above.
     </div>
 
+    <h3>Preventing undefined behavior with Transform Feedback</h3>
+
+    <p>
+        A buffer which is simultaneously bound to a TRANSFORM_FEEDBACK_BUFFER binding point and any other
```
kainino0x: The example seems to say this, but maybe clarify: The TRANSFORM_FEEDBACK_BUFFER binding point of the currently bound TF object only, right, not any other unbound ones??

jdarpinian: Yes, I'll clarify that.

specs/latest/2.0/index.html
```diff
@@ -3314,6 +3304,22 @@ <h3><a name="COPYING_BUFFERS">Copying Buffers</a></h3>
         Same as with <a href="#BUFFER_OBJECT_BINDING">Buffer Object Binding</a> restrictions above.
     </div>
 
+    <h3>Preventing undefined behavior with Transform Feedback</h3>
+
+    <p>
+        A buffer which is simultaneously bound to a TRANSFORM_FEEDBACK_BUFFER binding point in the currently
+        bound transform feedback object and any other binding point in the WebGL API cannot be used. Any
+        attempted use of such a double bound buffer fails with an <code>INVALID_OPERATION</code> error,
+        regardless of whether transform feedback is enabled. For example, readPixels to a PIXEL_PACK_BUFFER
```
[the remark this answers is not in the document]
jdashg: If it's dual-bound, it's undefined for *all* writes, including via a transform feedback binding point.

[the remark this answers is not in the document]
zhenyao: OK, I re-read the ES3 spec and I agree with Jeff. Ignore my comment then.

Let's merge.

sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
```diff
@@ -0,0 +1,216 @@
...
+gl.bindBufferBase(gl.UNIFORM_BUFFER, 0, uniformBuffer);
+drawWithFeedbackBound(gl, prog, vao, tf, false);
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "tf buffer not used as uniform buffer anymore");
+drawWithFeedbackBound(gl, prog, vao, tf, true);
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "tf buffer not used as uniform buffer anymore");
+
+gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, tf);
+const tfBuffer2 = createBuffer(gl, Float32Array.BYTES_PER_ELEMENT * 4);
+gl.bindBufferBase(gl.TRANSFORM_FEEDBACK_BUFFER, 0, tfBuffer2);
+drawWithFeedbackBound(gl, prog, badVao, tf);
```
kenrussell: Did you mean to leave off the last argument, making it implicitly false?

sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
```diff
@@ -0,0 +1,216 @@
...
+const expected = [2, 4, 6, 8];
+gl.bindBuffer(gl.TRANSFORM_FEEDBACK_BUFFER, tfBuffer);
+wtu.checkFloatBuffer(gl, gl.TRANSFORM_FEEDBACK_BUFFER, expected);
+
+debug("<hr/>Test ARRAY_BUFFER");
+// this should fail because the transform feedback's buffer #0 and the
+// badVao's buffer #0 are the same buffer
+const badVao = createVAO(gl, tfBuffer);
+drawWithFeedbackBound(gl, prog, badVao, tf, false);
+wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "buffer used as vertex attrib and tf simultaneously");
```
kenrussell: Most call sites of drawWithFeedbackBound in this test make two calls, with "true" and "false" as the last argument. Suggest adding another function which makes the two calls internally, so you can avoid duplicating the detail strings. This would also make it easier to add a suffix like "(with tf enabled)" or similar.

sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
```diff
@@ -0,0 +1,216 @@
...
+
+debug("<hr/>Test PIXEL_UNPACK_BUFFER");
+const tex = gl.createTexture();
+gl.bindTexture(gl.TEXTURE_2D, tex);
+gl.bindBuffer(gl.PIXEL_UNPACK_BUFFER, vertexBuffer);
+gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA8, 1, 1, 0, gl.RGBA, gl.UNSIGNED_BYTE, 0);
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "PIXEL_UNPACK_BUFFER is not bound for transform feedback");
+
+gl.bindBuffer(gl.PIXEL_UNPACK_BUFFER, tfBuffer);
+gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA8, 1, 1, 0, gl.RGBA, gl.UNSIGNED_BYTE, 0);
```
kenrussell: Maybe add a helper function which takes these operations in an anonymous function, calling them twice, and the second time, calls bindTransformFeedback(gl.TRANSFORM_FEEDBACK, null) first, and assert that the operation succeeds? This will test the code path which does the deref of the buffer's state that it's bound to the active transform feedback object.

kenrussell:
One more case that needs testing (is a different validation path in Chrome, at least) is indexed draw calls: drawElements. Ideally all of the cases that use drawWithFeedbackBound would try both non-indexed and indexed draw calls, and verify that either an error or NO_ERROR is generated in both cases. I'll comment on the Chromium code review too ( https://chromium-review.googlesource.com/826491 ).

PR #2561, opened 2017-12-18
Fix possible race condition in image-bitmap-from-blob tests.

kenrussell:
Possible cause of http://crbug.com/795945 .


kenrussell:

Note that this test doesn't exist in the 1.0.3 snapshot.

zhenyao:
LGTM

jdashg:
Oops, yep!

PR #2562, opened 2017-12-19
Propose ARB_texture_storage_multisample.

jdashg:


extensions/proposals/ARB_texture_storage_multisample/extension.xml
```diff
@@ -0,0 +1,56 @@
...
+
+  <number>XYZ</number>
+
+  <depends>
+    <api version="2.0"/>
+  </depends>
+
+  <overview>
+    <mirrors href="https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_texture_storage_multisample.txt"
+             name="ARB_texture_storage_multisample">
```
kenrussell: This is problematic because this is an OpenGL extension, not an OpenGL ES extension. Previously when subsetting OpenGL functionality like https://www.khronos.org/registry/webgl/extensions/WEBGL_depth_texture/ and https://www.khronos.org/registry/webgl/extensions/ANGLE_instanced_arrays/ did, first an OpenGL ES extension was written (under the ANGLE_ prefix, for example), and then the WebGL extension wrapped it.

Is this intended to expose ES 3.1's multisampled texture support? If so, can this explicitly be stated?

If you really want to reference the ARB_ version, can you add some addendum that references to the 1D and rectangular texture types are dropped (because they don't exist in ES 3.0)? Are there other references to OpenGL-specific features or functionality that don't exist in ES?

extensions/proposals/ARB_texture_storage_multisample/extension.xml
```diff
@@ -0,0 +1,56 @@
...
+    </mirrors>
+
+    <features>
+      <feature>
+        Adds support multisampled textures.
```
kenrussell: support -> support for

jdashg:
Ok, I made it more directly a 'pull in from es31'.

kenrussell:
Thanks for clarifying. This seems OK though without a specific ES extension underneath it I'm not 100% sure how we'll expose this in ANGLE. Let's merge as a proposal and request further comment from @vonture @null77 in particular.

PR #2563, opened 2017-12-19
Propose WEBGL_sync and WEBGL_es3_buffers.

jdashg:
Together these allow for efficient async readback, as well as supporting PBO-based GPU feedback iteration.

### DIFF
--- a/extensions/proposals/WEBGL_es3_buffers/extension.xml
+++ b/extensions/proposals/WEBGL_es3_buffers/extension.xml
@@ -1,0 +1,108 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- vi:set sw=2 ts=4: -->
+<?xml-stylesheet href="../../extension.xsl" type="text/xsl"?>
+<proposal href="proposals/WEBGL_es3_buffers/">
+  <name>WEBGL_es3_buffers</name>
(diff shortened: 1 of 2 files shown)


kenrussell:
This is a lot of functionality to pull into WebGL 1.0. @kainino0x @jdarpinian and I discussed this more today and we remain concerned that trying to optimize the synchronous API to do efficient readbacks is going to lead to performance pitfalls and a lot of implementation complexity.

We would rather pursue the more explicit asynchronous readback API that @kainino0x has been working on, because it is a step toward a primitive that we would collectively want in the GPUWeb API. I think we can all agree that we would not want to design GPUWeb's readbacks to emulate OpenGL's style, but instead be explicitly asynchronous.

@grorg @RafaelCintron do you have any comments on this pull request?

jdashg:
I don't think the async API is what we necessarily want in GPUWeb, and I don't think this is a good place to prototype that.

kenrussell:
Per discussion in the working group: we are going to postpone making these proposals, and for the moment, try to focus our engineering efforts on making it possible for Edge and Safari to ship WebGL 2.0.

jdashg:
Closing in favor of recommending WebGL 2.

PR #2565, opened 2017-12-19
Add a test for weird name conflict on NVIDIA GL driver

Oletus:
Some recent versions of NVIDIA GL drivers have an issue with a
function in parameter being passed as an inout argument while there is
a global variable with the same name.

The issue will be fixed in a future NVIDIA driver update.

http://crbug.com/792210


zhenyao:
LGTM

PR #2569, opened 2017-12-30
Fix review comments from @jdashg on #2523.

kenrussell:


kenrussell:
@jdashg @zhenyao @kainino0x in particular please review.

PR #2570, opened 2018-01-02
Add a test for assigning twice to a swizzled vec inside a function

Oletus:
This reproduces a crash at least in Chromium running on recent NVIDIA
OpenGL drivers.

Chrome bug report that this test case is based on is here:
http://crbug.com/798117


kenrussell:
Nice reduction. LGTM

PR #2573, opened 2018-01-04
Regression tests for ANGLE UBO dirty bit bug

tsherif:
Regression tests for: https://bugs.chromium.org/p/chromium/issues/detail?id=792966

Ported the two ANGLE UBO dirty bit regression tests, SimpleBufferChange and DependentBufferChange, from https://github.com/google/angle/blob/master/src/tests/gl_tests/UniformBufferTest.cpp

I haven't been able to test that this actually catches the bug, since Chrome and Firefox are both updated for me, so this should be verified by someone. (Is there an easy way to get older versions of Chrome to test things like this?)

### DIFF
--- a/sdk/tests/conformance2/uniforms/00_test_list.txt
+++ b/sdk/tests/conformance2/uniforms/00_test_list.txt
@@ -2,3 +2,5 @@
 --min-version 2.0.1 gl-uniform-arrays-sub-source.html
 --min-version 2.0.1 query-uniform-blocks-after-shader-detach.html
 --min-version 2.0.1 uniform-blocks-with-arrays.html
+--min-version 2.0.1 simple-buffer-change.html
+--min-version 2.0.1 dependent-buffer-change.html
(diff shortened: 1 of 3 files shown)


null77:
You can find older versions of Chrome for testing here:

https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Win/

Note these are Chromium, not Chrome. So they are missing some closed-source components.

I might have put these tests in the same file, but seems fine really. There's also a couple tests from src/tests/gl_tests/StateChangeTest.cpp that would be nice to be ported, but generally these tests LGTM.

tsherif:
I had the impression that the conformance tests always just did one thing, but could definitely change this if it would be preferable to have them all in one file.

I've taken a look at the state change and the vertex attribute ANGLE tests you mentioned, and should able to put together a PR for those in a few days.

null77:
Awesome! Thanks for the contributions. I don't know if there's a specific policy for file contents.

zhenyao:
LGTM as well
Thanks for the contribution

PR #2566, opened 2017-12-21
Use Web IDL's new-ish interface mixins concept

romandev:
WebIDL recently introduced dedicated syntax for mixins [1]. This
replaces the existing [NoInterfaceObject] and "implements" syntax with
"interface mixin" and "includes" in the appropriate places.

This fixes #2538, fixes #2539 issues.

[1] https://github.com/heycam/webidl/commit/45e8173d40ddff8dcf81697326e094bcf8b92920

PR #2575, opened 2018-01-04
Fix misuse of mixin syntax

### LINKED ISSUE #2574: [WebIDL] introduced incorrect IDL syntax in #2566
Thanks for updating both WebGL specs to use the mixin syntax.

Unfortunately, #2566 introduced syntax that is not supported by WebIDL: a mixin that itself includes another mixin.

Currently the spec reads:

```webidl
interface mixin WebGL2RenderingContextBase { };
interface WebGL2RenderingContext { };
WebGL2RenderingContextBase includes WebGLRenderingContextBase; // NOT SUPPORTED!
WebGL2RenderingContext includes WebGL2RenderingContextBase;
```

That is, the `WebGL2RenderingContextBase` mixin includes the `WebGLRenderingContextBase` (that's the part that is not supported).

Instead the  `WebGL2RenderingContext` interface should include both the `WebGLRenderingContextBase` and `WebGL2RenderingContextBase` mixins, like so:

```webidl
interface mixin WebGL2RenderingContextBase { };
interface WebGL2RenderingContext { };
WebGL2RenderingContext includes WebGLRenderingContextBase;
WebGL2RenderingContext includes WebGL2RenderingContextBase;
```

Allowing support for mixins to include other mixins is actually quite involved, and `WebGL2RenderingContextBase` is the only such case across the platform, afaik. It's also easily solved by having `WebGL2RenderingContext` include both mixins instead.

/cc @kainino0x @kenrussell @romandev

romandev:
Introduced `interface mixin` syntax in #2566 but it's incorrect.
A `interface mixin` SHOULD NOT includes another mixin.

This fixes #2574.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -967,11 +967,11 @@ interface mixin <dfn id="WebGL2RenderingContextBase">WebGL2RenderingContextBase<
   [WebGLHandlesContextLoss] GLboolean isVertexArray(WebGLVertexArrayObject? vertexArray);
   void bindVertexArray(WebGLVertexArrayObject? array);
 };
-WebGL2RenderingContextBase includes WebGLRenderingContextBase;
 
 interface <dfn id="WebGL2RenderingContext">WebGL2RenderingContext</dfn>
 {
 };
+WebGL2RenderingContext includes WebGLRenderingContextBase;
 WebGL2RenderingContext includes WebGL2RenderingContextBase;
 
 </pre>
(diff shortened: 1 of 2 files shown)


specs/latest/2.0/index.html
```diff
@@ -967,7 +967,7 @@ <h3><a name="WEBGLRENDERINGCONTEXT">The WebGL context</a></h3>
   [WebGLHandlesContextLoss] GLboolean isVertexArray(WebGLVertexArrayObject? vertexArray);
   void bindVertexArray(WebGLVertexArrayObject? array);
 };
-WebGL2RenderingContextBase includes WebGLRenderingContextBase;
+WebGL2RenderingContext includes WebGLRenderingContextBase;
```
tobie: There's nothing that prevents you from using an includes statement before you actually declare the interface that includes it, but it looks weird.

tobie: I suggest you move it right above the line 975 (`WebGL2RenderingContext includes WebGL2RenderingContextBase;`).

romandev: Ah, nice review. Done.
Thank you!

kenrussell:
Thanks for the fix.

PR #2576, opened 2018-01-05
Add a test case for CopyTexImage2D behavior when source/target images

zhenyao:
are from the same texture.

This is a regression test for crbug.com/797235

### DIFF
--- a/sdk/tests/conformance2/textures/misc/copy-texture-image-same-texture.html
+++ b/sdk/tests/conformance2/textures/misc/copy-texture-image-same-texture.html
@@ -70,25 +70,51 @@ function checkFramebuffer(expected) {
     }
 }
 
+function checkTextureLevelColor(texture, level, level_width, level_height, x, y, width, height, color) {
+    var fbo = gl.createFramebuffer();
+    gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+    var color_renderable_tex = gl.createTexture();
+    gl.bindTexture(gl.TEXTURE_2D, color_renderable_tex);
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA8, level_width, level_height, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, color_renderable_tex, 0);
+    checkFramebuffer([gl.FRAMEBUFFER_COMPLETE]);
+
+    var program = wtu.setupTexturedQuad(gl);
+    gl.bindTexture(gl.TEXTURE_2D, texture);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_BASE_LEVEL, level);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAX_LEVEL, level);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.REPEAT);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.REPEAT);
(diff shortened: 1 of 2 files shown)


zhenyao:
@kenrussell @kainino0x @jdarpinian @vonture one of you please review

kenrussell:
This test doesn't actually verify the contents of the destination texture level. Could you please add that checking? Otherwise the test is not that strict.

sdk/tests/conformance2/textures/misc/copy-texture-image-same-texture.html
```diff
@@ -0,0 +1,157 @@
...
+    wtu.fillTexture(gl, texture, 64, 64, [255, 0, 0, 255], 0, gl.RGBA, gl.UNSIGNED_BYTE, gl.RGBA8);
+    gl.generateMipmap(gl.TEXTURE_2D);
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture, 1);
+    checkFramebuffer([gl.FRAMEBUFFER_COMPLETE]);
+
+    gl.copyTexImage2D(gl.TEXTURE_2D, 2, gl.LUMINANCE_ALPHA, 0, 0, 16, 16, 0);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "copyTexImage2D should succeed.");
+    checkFramebuffer([gl.FRAMEBUFFER_INCOMPLETE_ATTACHMENT]);
+
+    checkTextureLevelColor(texture, 2, 16, 16, 0, 0, 16, 16, [255, 255, 255, 255]);
```
zhenyao: On Linux and Mac, this fails with [255, 255, 0, 255]. On Windows or Linux with --use-gl=angle, this passes.

I think the test is correct, but we have a bug in Chrome command buffer implementation.

I'll file a bug for this.

zhenyao:
@kainino0x Ken is out of office. Can you review this?

sdk/tests/conformance2/textures/misc/copy-texture-image-same-texture.html
```diff
@@ -0,0 +1,157 @@
...
+    var fbo = gl.createFramebuffer();
+    gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+
+    var texture = gl.createTexture();
+    wtu.fillTexture(gl, texture, 64, 64, [255, 0, 0, 255], 0, gl.RGBA, gl.UNSIGNED_BYTE, gl.RGBA8);
+    gl.generateMipmap(gl.TEXTURE_2D);
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture, 1);
+    checkFramebuffer([gl.FRAMEBUFFER_COMPLETE]);
+
+    gl.copyTexImage2D(gl.TEXTURE_2D, 2, gl.RGBA8, 0, 0, 64, 64, 0);
```
kainino0x: It took me a while to understand this, can you add a comment just to clarify?

    // Level 0 is 64x64. Reallocate level 2 at 64x64, so the framebuffer is no longer complete.

sdk/tests/conformance2/textures/misc/copy-texture-image-same-texture.html
```diff
@@ -0,0 +1,157 @@
...
+    var fbo = gl.createFramebuffer();
+    gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+
+    var texture = gl.createTexture();
+    wtu.fillTexture(gl, texture, 64, 64, [255, 0, 0, 255], 0, gl.RGBA, gl.UNSIGNED_BYTE, gl.RGBA8);
+    gl.generateMipmap(gl.TEXTURE_2D);
+    gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture, 1);
+    checkFramebuffer([gl.FRAMEBUFFER_COMPLETE]);
+
+    gl.copyTexImage2D(gl.TEXTURE_2D, 2, gl.LUMINANCE_ALPHA, 0, 0, 16, 16, 0);
```
kainino0x: // Level 0 is gl.RGBA. Reallocate level 2 with LUMINANCE_ALPHA, so the framebuffer is no longer complete.

zhenyao:
Detailed comments added.

kenrussell:
Great work. Thanks Mo for strengthening this test. It looks like it uncovers some longstanding bugs.

PR #2579, opened 2018-01-12
Probe for max texture level for each format separately.

jdashg:
Different formats have different effective max texture sizes/max mip levels.

### DIFF
--- a/sdk/tests/conformance/textures/misc/texture-size-limit.html
+++ b/sdk/tests/conformance/textures/misc/texture-size-limit.html
@@ -150,7 +150,7 @@ function testFormatType(t, test) {
     // a maxLevel L8 texture, but only a maxLevel-1 RGB8 texture.
     var maxLevelsForFormat = numLevels;
     while (true) {
-      gl.texImage2D(target, maxLevelsForFormat-1, test.format, 1, 1, 0, test.format, test.type, pixels);
+      gl.texImage2D(target, maxLevelsForFormat-1, test.format, 1, 1, 0, test.format, test.type, null);
       var err = gl.getError();
       if (err == gl.OUT_OF_MEMORY) {
         debug("Probe failed for level=" + (maxLevelsForFormat-1) + ", reducing...");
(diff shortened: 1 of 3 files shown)


sdk/tests/conformance/textures/misc/texture-size-limit.html
```diff
@@ -125,43 +125,63 @@
...
       gl.texImage2D(target, level, test.format, size, otherDimension, 0, test.format, test.type, pixels);
       wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no error for level: " + level + " " + size + "x" + otherDimension);
-      gl.texImage2D(target, level, test.format, otherDimension, size, 0, test.format, test.type, pixels);
-      wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no error for level: " + level + " " + otherDimension + "x" + size);
-      gl.texImage2D(target, level, test.format, badSize, badOtherDimension, 0, test.format, test.type, pixels);
-      wtu.glErrorShouldBe(gl, gl.INVALID_VALUE, "should generate INVALID_VALUE for level: " + level + " " + badSize + "x" + badOtherDimension);
-      gl.texImage2D(target, level, test.format, badOtherDimension, badSize, 0, test.format, test.type, pixels);
-      wtu.glErrorShouldBe(gl, gl.INVALID_VALUE, "should generate INVALID_VALUE for level: " + level + " " + badOtherDimension + "x" + badSize);
+      if (badOtherDimension != badSize) {
+        gl.texImage2D(target, level, test.format, otherDimension, size, 0, test.format, test.type, pixels);
```
kainino0x: Did you mean to use badOtherDimension here?

kainino0x: Well, maybe not. I was misled by the `badOtherDimension != badSize` condition. Maybe if it's correct, the condition could be refactored? (maybe `t.target == gl.TEXTURE_2D`, because that seems to be equivalent)

kainino0x: If you're going to remove the two tests that were below this one, I actually don't think badSize and badOtherDimension need to exist at all; it seems like `pixels` can be changed to `size * otherDimension * test.size` and the `bad` variables can be removed.

kainino0x: I see you resolved this. :)

jdashg:
Fixed. Sometimes it shocking how broken of JS code will run without error.

kenrussell:
In general this looks good though I'm concerned about the portability of the probing code and in particular its checking of out-of-memory errors. On what browsers and operating systems has the new test been run?

jdashg:
It has been run on Linux Mesa+Intel. I can run it against all our conformance candidates, but the probing code should be a strict improvement over the current approach.
There exists a hypothetical implementation that deals in mip levels sparsely, and therefore may only have issues with large level=0 images, but we should handle that only when we run into it.

jdashg:
This works for me on Firefox Nightly Win+Intel, Win+NV, Lin+Intel, Lin+AMD, Lin+NV, Mac+Intel, Mac+NV, as well as Chrome Canary Win+Intel.

jdashg:
@kenrussell Anything else you need?

kenrussell:
@jdashg I'm very sorry for letting this sit so long. Tested on a few more platforms with Chrome (Mac+NVIDIA, Linux+NVIDIA, Windows+AMD); looking good. Merging now.

PR #2580, opened 2018-01-13
Add GLSL 3.00 test for forward declarations of functions.

kenrussell:
Fails on Pixel C with Chrome because of flattening of precision
qualifiers into variables' types, and because of a bug in the GLSL
compiler on this device.

Thanks to Robert Muth of art.muth.org for reporting.

### DIFF
--- a/sdk/tests/conformance2/glsl3/00_test_list.txt
+++ b/sdk/tests/conformance2/glsl3/00_test_list.txt
@@ -14,6 +14,7 @@ compound-assignment-type-combination.html
 const-array-init.html
 --min-version 2.0.1 float-parsing.html
 forbidden-operators.html
+--min-version 2.0.1 forward-declaration.html
 frag-depth.html
 --min-version 2.0.1 gradient-in-discontinuous-loop.html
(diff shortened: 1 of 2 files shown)


kenrussell:
Please review. @kainino0x @zhenyao @jdashg 

@Oletus FYI. I don't think it's worth trying to work around this bug.

kenrussell:
Note: also tested on a Pixel 2 (Qualcomm GPU) and NVIDIA SHIELD tablet, as well as on desktop macOS and Linux; none have this bug.

kainino0x:
The Pixel C uses the open-source Nouveau driver rather than the proprietary NVIDIA drivers used on SHIELD and most desktop machines. So this may reproduce on Linux/Nouveau but otherwise should be pretty specific to Pixel C and maybe a handful of similar devices.

kenrussell:
Ah, thanks for that information about the Pixel C's driver. Since Nouveau uses Mesa as its infrastructure I will hope that this has since been fixed in their top-of-tree driver. Merging now.

PR #2581, opened 2018-01-15
Remove unnecessary GLenums about coverage

Richard-Yunchao:
These 2 GLenums have been added into WebGL 1.0 spec: https://www.khronos.org/registry/webgl/specs/latest/1.0/. So it is not necessary to add them again in WebGL 2.0 spec. You know, WebGL 2.0 spec only show the new added GLenums.

### DIFF
--- a/specs/2.0.0/index.html
+++ b/specs/2.0.0/index.html
@@ -1085,8 +1085,6 @@ WebGL2RenderingContext implements WebGL2RenderingContextBase;
                 <tr><td>RASTERIZER_DISCARD</td><td>GLboolean</td></tr>
                 <tr><td>READ_BUFFER</td><td>GLenum</td></tr>
                 <tr><td>READ_FRAMEBUFFER_BINDING</td><td>WebGLFramebuffer</td></tr>
-                <tr><td>SAMPLE_ALPHA_TO_COVERAGE</td><td>GLboolean</td></tr>
-                <tr><td>SAMPLE_COVERAGE</td><td>GLboolean</td></tr>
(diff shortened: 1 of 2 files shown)


Richard-Yunchao:
One question here. SAMPLE_ALPHA_TO_COVERAGE and SAMPLE_COVERAGE are only added into latest WebGL 1.0 spec into getParameter. Why not add them into the old specs through 1.0.0 to 1.0.3? These two GLenums have been defined in all old WebGL 1.0 specs, but they are not added into the list for getParameter API in old specs through 1.0.0 to 1.0.3. 

I suppose we should add these two GLenums into old specs for getParameter too, but there might be some compatible issues which I don't know.

zhenyao:
I think it's negligence that they are missing from earlier version and they should be added back. I don't really see a compatibility issue here.

zhenyao:
LGTM

Richard-Yunchao:
Thank you, @zhenyao . @kainino0x added SAMPLE_ALPHA_TO_COVERAGE and SAMPLE_COVERAGE into getParameter into latest WebGL 1.0 spec at https://github.com/KhronosGroup/WebGL/pull/2403. 

Maybe @kainino0x  can add them into 1.0.0 ~ 1.0.3 specs and conformance tests too if necessary.

kainino0x:
IMO we can't merge those changes back into the 1.0.0~1.0.3 snapshots, as they spec and test previously unspecified behavior of getParameter. We can only make the old snapshots less strict, not more strict.

jdashg:
Agreed, we don't take new restrictions in old spec snapshots.

Richard-Yunchao:
OK. Got it.

PR #2582, opened 2018-01-16
Add all multisample texture related features in GLES 3.1 to WEBGL_texture_multisample.

Richard-Yunchao:
PATL. @kenrussell , @jdashg , @zhenyao , @JiangYizhou 

This is based on ANGLE_texture_multisample at https://chromium-review.googlesource.com/c/angle/angle/+/859917. I am not sure whether I should wait for that extension to merge at first (reviewed by ANGLE team). But I would like to let you (members from Khronos WebGL WG) know in case you have some good suggestions.

FYI. @null77 , @vonture , @Kangz , @kainino0x , @qjia7 , @Jiawei-Shao , @gyagp

### DIFF
--- a/extensions/proposals/WEBGL_texture_multisample/extension.xml
+++ b/extensions/proposals/WEBGL_texture_multisample/extension.xml
@@ -9,6 +9,9 @@
 
   <contributors>
     <contributor>Jeff Gilbert, Mozilla</contributor>
+    <contributor>Yunchao He, Intel</contributor>
+    <contributor>Yizhou Jiang, Intel</contributor>
+    <contributor>Members of the WebGL working group</contributor>
   </contributors>
 
   <number>XYZ</number>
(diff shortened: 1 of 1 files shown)


Richard-Yunchao:
A couple of explanation and open specific for this WEBGL extension:
1） I rewrite the <overview> section, although it is based on ANGLE_texture_multisample. Because there are still some differences in detail like: 1) no ANGLE suffix for all APIs. 2) getInterger/getInteger_v/getBoolean/... are all combined to getParameter in WebGL. 

2) Some thoughts about the multisample texture extension and the big extension to expose all features in GLES 3.1 (compute shader is the most important feature, among many other features include multisample texture). The big extension, which can be named as WEBGL_es31, would include all features of multisample texture. But WEBGL_es31 is based on GLES 3.1 interface in ANGLE, not GLES 3.0 extension. In addition, WEBGL_es31 can not be supported by MacOS, while multisample texture can be supported by MacOS. Because many features in GLES 3.1 can not be supported by MacOS. But multisample texture can be supported since OpenGL 3.2 and latest MacOS can support OpenGL 4.1.  
BTW, I began to write WEBGL_es31 extension recently too, but it would take some time.

Richard-Yunchao:
Unfortunately, we need to close this one according to WebGL WG's discussion. 

(I'd like to paste the comments from Ken according to the WebGL WG's discussion for clarification)

The situation is as follows:

1. The EXT_multisampled_render_to_texture extension is the only variant of multisampled textures that allows the MSAA buffer to live entirely in tile memory on tiled architectures.
2. Unreal Engine is – with 99% certainty – only using the EXT_multisampled_render_to_texture extension on Android, and its WebAssembly build target is based on this platform.
3. Using ES 3.1's multisampled textures provides no performance advantage over basic multisampled renderbuffers and glBlitFramebuffer to resolve them.

Unfortunately, if ES 3.1's multisampled textures are added as an extension, Unreal Engine won't use them. Even if the engine did use them, resolving them to a single-sampled texture is almost surely going to be slower than doing it with glBlitFramebuffer, since the resolve would be being done by a user's shader.

The working group decided on the call that the best thing to do is not invest the time to separate out ES 3.1's multisampled textures into an extension. Instead, we should expose the EXT_multisampled_render_to_texture extension on those platforms that actually support it. This will provide efficient antialiasing to many mobile devices. (We also discussed the possibility of emulating this extension on all platforms, but decided it's not a good idea, because it would add even more per-draw-call overhead.)

We are also planning to collaborate with Epic Games to add code to their WebAssembly export path to use regular MSAA renderbuffers if EXT_multisampled_render_to_texture isn't available.

PR #2583, opened 2018-01-16
Only complain on stencil writemask/value_mask/ref front/back mismatch…

jdashg:
… for the `s` stencil bits of the draw framebuffer.

No one seems to implement this yet. Is this what we want?

### DIFF
--- a/sdk/tests/conformance/misc/webgl-specific-stencil-settings.html
+++ b/sdk/tests/conformance/misc/webgl-specific-stencil-settings.html
@@ -1,0 +1,188 @@
+<!--
+
+/*
+** Copyright (c) 2018 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL stencil mask/func front-state-back-state equality test</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+
+<script>
+"use strict";
+var wtu = WebGLTestUtils;
+description("Tests that stencil mask/func are validated correctly when the front state and back state differ.");
+
+var gl = wtu.create3DContext();
+
+function testStencilMaskCase(mask, error) {
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.stencilMaskSeparate(gl.FRONT, " + mask[0] + ")");
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.stencilMaskSeparate(gl.BACK,  " + mask[1] + ")");
+    // If an error is generated, it should be at draw time.
+    wtu.shouldGenerateGLError(gl, error, "wtu.dummySetProgramAndDrawNothing(gl)");
+}
+
+function testStencilMask(errIfMismatch) {
+    testStencilMaskCase([0, 256], gl.NO_ERROR);
+    testStencilMaskCase([1, 256], errIfMismatch);
+    testStencilMaskCase([1, 257], gl.NO_ERROR);
+    testStencilMaskCase([1, 258], errIfMismatch);
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.stencilMask(1023)", "resetting stencilMask");
+}
+
+function testStencilFuncCase(ref, mask, error) {
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.stencilFuncSeparate(gl.FRONT, gl.ALWAYS, " + ref[0] + ", " + mask[0] + ")");
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.stencilFuncSeparate(gl.BACK,  gl.ALWAYS, " + ref[1] + ", " + mask[1] + ")");
+    // If an error is generated, it should be at draw time.
+    wtu.shouldGenerateGLError(gl, error, "wtu.dummySetProgramAndDrawNothing(gl)");
+}
+
+function testStencilFunc(errIfMismatch) {
+    testStencilFuncCase([ 256,  257], [1023, 1023], gl.NO_ERROR);
+    testStencilFuncCase([ 256,  254], [1023, 1023], errIfMismatch);
+
+    testStencilFuncCase([  -1,    0], [1023, 1023], gl.NO_ERROR);
+    testStencilFuncCase([  -1,  254], [1023, 1023], errIfMismatch);
+
+    testStencilFuncCase([   0,    0], [   1,  257], gl.NO_ERROR);
+    testStencilFuncCase([   0,    0], [   1,  258], errIfMismatch);
+
+    testStencilFuncCase([   1,    1], [1024, 2048], gl.NO_ERROR);
+    testStencilFuncCase([   1,    1], [2048, 1024], gl.NO_ERROR);
+
+    testStencilFuncCase([  -1,   -1], [1023, 1023], gl.NO_ERROR);
+    testStencilFuncCase([  -1,    0], [1023, 1023], gl.NO_ERROR);
+    testStencilFuncCase([   0,   -1], [1023, 1023], gl.NO_ERROR);
+    testStencilFuncCase([   0,    0], [1023, 1023], gl.NO_ERROR);
+
+    testStencilFuncCase([  -1,  255], [1023, 1023], errIfMismatch);
+    testStencilFuncCase([   0,  256], [1023, 1023], errIfMismatch);
+    testStencilFuncCase([   0, 1024], [1023, 1023], errIfMismatch);
+    testStencilFuncCase([   1,  257], [1023, 1023], errIfMismatch);
+    testStencilFuncCase([ 255,   -1], [1023, 1023], errIfMismatch);
+    testStencilFuncCase([ 256,    0], [1023, 1023], errIfMismatch);
+    testStencilFuncCase([1024,    0], [1023, 1023], errIfMismatch);
+    testStencilFuncCase([ 257,    1], [1023, 1023], errIfMismatch);
+
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.stencilFunc(gl.ALWAYS, 0, 1023)", "resetting stencilFunc");
+}
+
+const fb = gl.createFramebuffer();
+gl.bindFramebuffer(gl.FRAMEBUFFER, fb);
+const colorRB = gl.createRenderbuffer();
+gl.bindRenderbuffer(gl.RENDERBUFFER, colorRB);
+gl.renderbufferStorage(gl.RENDERBUFFER, gl.RGBA4, 1, 1);
+gl.framebufferRenderbuffer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.RENDERBUFFER, colorRB);
+
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "initial framebuffer setup")
+
+function testStencilSettings(haveDepthBuffer, haveStencilBuffer, enableStencilTest, errIfMismatch) {
+    debug("");
+    debug("With depthbuffer=" + haveDepthBuffer +
+            ", stencilbuffer=" + haveStencilBuffer +
+            ", stencilTest=" + enableStencilTest +
+            ", expecting error=" + wtu.glEnumToString(errIfMismatch) +
+            " for mismatching mask or func settings.");
+
+    let rbo = null;
+    let attachment;
+    if (haveDepthBuffer || haveStencilBuffer) {
+        rbo = gl.createRenderbuffer();
+        gl.bindRenderbuffer(gl.RENDERBUFFER, rbo);
+
+        let internalformat;
+        if (haveDepthBuffer && haveStencilBuffer) {
+            internalformat = gl.DEPTH_STENCIL;
+            attachment = gl.DEPTH_STENCIL_ATTACHMENT;
+        } else if (haveDepthBuffer) {
+            internalformat = gl.DEPTH_COMPONENT16;
+            attachment = gl.DEPTH_ATTACHMENT;
+        } else if (haveStencilBuffer) {
+            internalformat = gl.STENCIL_INDEX8;
+            attachment = gl.STENCIL_ATTACHMENT;
+        }
+        gl.renderbufferStorage(gl.RENDERBUFFER, internalformat, 1, 1);
+        gl.framebufferRenderbuffer(gl.FRAMEBUFFER, attachment, gl.RENDERBUFFER, rbo);
+    }
+    shouldBe("gl.checkFramebufferStatus(gl.FRAMEBUFFER)", "gl.FRAMEBUFFER_COMPLETE");
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "depth/stencil renderbuffer setup")
+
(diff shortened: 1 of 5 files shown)


jdashg:
- Add test for DEPTH w/o STENCIL, vice versa
- Use effective format RGBA8 (RGBA/UBYTE) to guarantee framebuffer completeness
- Do we need to backport?

RafaelCintron:
@jdashg , I tried running your new test with the ANGLE running in webglCompatibility mode and received several failures.  

The fourth call to `drawArrays(gl.POINTS, 0, 0)` in `testForStencilBits` fails with INVALID_OPERATION in the following ANGLE code in ValidateDrawCommon

````C++
// Note: these separate values are not supported in WebGL, due to D3D's limitations. See
// Section 6.10 of the WebGL 1.0 spec.
Framebuffer *framebuffer = state.getDrawFramebuffer();
if (context->getLimitations().noSeparateStencilRefsAndMasks || extensions.webglCompatibility)
{
    const FramebufferAttachment *dsAttachment =
        framebuffer->getStencilOrDepthStencilAttachment();
    GLuint stencilBits                = dsAttachment ? dsAttachment->getStencilSize() : 0;
    GLuint minimumRequiredStencilMask = (1 << stencilBits) - 1;
    const DepthStencilState &depthStencilState = state.getDepthStencilState();

    bool differentRefs = state.getStencilRef() != state.getStencilBackRef();
    bool differentWritemasks =
            (depthStencilState.stencilWritemask & minimumRequiredStencilMask) !=
            (depthStencilState.stencilBackWritemask & minimumRequiredStencilMask);
    bool differentMasks = (depthStencilState.stencilMask & minimumRequiredStencilMask) !=
                              (depthStencilState.stencilBackMask & minimumRequiredStencilMask);

    if (differentRefs || differentWritemasks || differentMasks)
    {
        if (!extensions.webglCompatibility)
        {
            ERR() << "This ANGLE implementation does not support separate front/back stencil "
                     "writemasks, reference values, or stencil mask values.";
        }
        ANGLE_VALIDATION_ERR(context, InvalidOperation(), StencilReferenceMaskOrMismatch);
        return false;
    }
}
````

Though `differentWritemasks` and `differentMasks` are false, `differentRefs` is true.  The stencil ref is 1 and the stencil back ref is 257.  Looks like the minimum required bits only applies to the masks, not the reference value.

File: sdk/tests/conformance/misc/webgl-specific.html [hunk 1]
```diff
@@ -78,23 +78,59 @@
...
+const depthStencilRB = gl.createRenderbuffer();
+gl.bindRenderbuffer(gl.RENDERBUFFER, depthStencilRB);
+gl.renderbufferStorage(gl.RENDERBUFFER, gl.DEPTH_STENCIL, 1, 1);
+gl.framebufferRenderbuffer(gl.FRAMEBUFFER, gl.DEPTH_STENCIL_ATTACHMENT, gl.RENDERBUFFER,
+                           depthStencilRB);
+
+debug("...with 8 stencil bits in the draw framebuffer");
+testForStencilBits(gl.INVALID_OPERATION);
+
+gl.deleteRenderbuffer(fb);
```
RafaelCintron: Nit: Should be `deleteFramebuffer` since `fb` is a framebuffer, not a renderbuffer.

sdk/tests/conformance/misc/webgl-specific.html
```diff
@@ -78,23 +78,59 @@
...
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.drawArrays(gl.POINTS, 0, 0)");
+
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.stencilMaskSeparate(gl.BACK, 258)");
+    wtu.shouldGenerateGLError(gl, errIfMismatch, "gl.drawArrays(gl.POINTS, 0, 0)");
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.stencilMask(1023)");
+
+
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.stencilFuncSeparate(gl.FRONT, gl.ALWAYS, 1, 1023)");
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.stencilFuncSeparate(gl.BACK, gl.ALWAYS, 257, 1023)");
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.drawArrays(gl.POINTS, 0, 0)");
```
RafaelCintron: @jdashg , I tried running your new test with the ANGLE running in webglCompatibility mode and received several failures.
This fourth call to `drawArrays` fails with INVALID_OPERATION in the following ANGLE code in `ValidateDrawBase`

````C++
// Note: these separate values are not supported in WebGL, due to D3D's limitations. See
// Section 6.10 of the WebGL 1.0 spec.
Framebuffer *framebuffer = state.getDrawFramebuffer();
if (context->getLimitations().noSeparateStencilRefsAndMasks || extensions.webglCompatibility)
{
    const FramebufferAttachment *dsAttachment =
        framebuffer->getStencilOrDepthStencilAttachment();
    GLuint stencilBits                = dsAttachment ? dsAttachment->getStencilSize() : 0;
    GLuint minimumRequiredStencilMask = (1 << stencilBits) - 1;
    const DepthStencilState &depthStencilState = state.getDepthStencilState();

    bool differentRefs = state.getStencilRef() != state.getStencilBackRef();
    bool differentWritemasks =
            (depthStencilState.stencilWritemask & minimumRequiredStencilMask) !=
            (depthStencilState.stencilBackWritemask & minimumRequiredStencilMask);
    bool differentMasks = (depthStencilState.stencilMask & minimumRequiredStencilMask) !=
                              (depthStencilState.stencilBackMask & minimumRequiredStencilMask);

    if (differentRefs || differentWritemasks || differentMasks)
    {
        if (!extensions.webglCompatibility)
        {
            ERR() << "This ANGLE implementation does not support separate front/back stencil "
                     "writemasks, reference values, or stencil mask values.";
        }
        ANGLE_VALIDATION_ERR(context, InvalidOperation(), StencilReferenceMaskOrMismatch);
        return false;
    }
}
````
Though `differentWritemasks` and `differentMasks` are false, `differentRefs` is true. The stencil ref is 1 and the stencil back ref is 257. Looks like the minimum required bits only applies to the masks, not the reference value.

**Comment by jdashg** [comment]
Do we want to require the reference values to be the same? The reference values are all masked by their respectively-masked masks, so refs at least always behave as if they are masked to the available bits.
We can continue to require that the unmasked values match if we want, if it's too much to ask ANGLE to deal with resolving this difference.

File: sdk/tests/conformance/misc/webgl-specific.html [hunk 3]
```diff
@@ -78,23 +78,59 @@
...
+
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.stencilFuncSeparate(gl.BACK, gl.ALWAYS, 258, 1023)");
+    wtu.shouldGenerateGLError(gl, errIfMismatch, "gl.drawArrays(gl.POINTS, 0, 0)");
```
RafaelCintron: This call to `drawArrays` fails for the same reason as above.

sdk/tests/conformance/misc/webgl-specific.html
kenrussell: It seems like a bug in ANGLE that the minimumRequiredStencilMask isn't applied to the reference values. @RafaelCintron, do you agree? Is applying this mask compatible with Direct3D's semantics?

RafaelCintron:
Adding @Kangz @null77 and @vonture.

For reference, ANGLE has two tests for this functionality. One is [RequiresSameStencilMaskAndRef](https://github.com/google/angle/blob/master/src/tests/gl_tests/WebGLCompatibilityTest.cpp#L1183) added by @Kangz as part of the WebGL compatibility test. The other is [DifferentStencilMasksTests](https://github.com/google/angle/blob/master/src/tests/gl_tests/DifferentStencilMasksTest.cpp) added by Jinyoung Hur.

kenrussell:
Looking at those tests it seems to me that if we change WebGL's validation semantics so that the stencil reference value is masked by minimumRequiredStencilMask, too, then things should still work fine on Direct3D. I think we should make that change to the validation, and test it.

sdk/tests/conformance/misc/webgl-specific.html
RafaelCintron: @kenrussell , D3D11 does not support creating textures with more than 8bits of stencil data.  Hence, the read and write masks are only limited to 8 bits.  See the [DEPTH_STENCIL_DESC](https://msdn.microsoft.com/en-us/library/windows/desktop/ff476110(v=vs.85).aspx) documentation.  The reference value is an unsigned integer. See the [OMSetDepthStencilState](https://msdn.microsoft.com/en-us/library/windows/desktop/ff476463(v=vs.85).aspx) for documentation.

In [`StateManager11::syncDepthStencilState`](https://github.com/google/angle/blob/master/src/libANGLE/renderer/d3d/d3d11/StateManager11.cpp#L1084), ANGLE clamps the reference value before calling D3D11 with a comment stating this is because of the GL spec.  

I suppose It would be nice to inform the web developer when bits of their reference value is not taking effect but I do not feel strongly about that.

RafaelCintron:
@jdashg , since D3D11 only supports setting one reference value for depth stencil state, we need to enforce the reference values for read and write be equal.

jdashg:
@RafaelCintron Is it ok if they are equal only /after/ masking? That's what I've proposed here.

jdashg:
Should we just deprecate stencilFuncSeparate and stencilMaskSeparate? stencilOpSeparate seems to be the only aspect that D3D11 supports.

jdashg:
It's easier/clearer if we just forbid separate values outright, instead of being tricky and only forbidding them when active.

kenrussell:
@jdashg @RafaelCintron a couple of questions:

1. @RafaelCintron you mentioned "the reference values for read and write must be equal". I don't understand that -- did you mean "the reference values for front and back must be equal"? In that case I support @jdashg 's desire to compare them *only* after masking with the minimumRequiredStencilMask in ANGLE. (Or, in other words, a mask that includes the number of bits in the stencil buffer -- this change to WebGL's validation would be done on all platforms.)

2. @jdashg D3D11 supports different stencil functions for front and back faces. D3D11_DEPTH_STENCIL_DESC [3] has FrontFace and BackFace members of type D3D11_DEPTH_STENCILOP_DESC, and those contain the functions. I think this change and associated ANGLE/WebGL validation changes should just focus on the issue at hand of masking various values like the reference value by the number of bits actually in the stencil buffer.

3. @jdashg above you mentioned "instead of being tricky and only forbidding them when active" -- did you mean, if the STENCIL_TEST were active? I agree that we should stick with the current WebGL semantics in https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.11 and reject all draw calls made while these illegal states are set. However, I don't see a way to reject *setting* the illegal states.

Some references:
[1] ANGLE's current stencil validation in ValidateDrawBase:
https://cs.chromium.org/chromium/src/third_party/angle/src/libANGLE/validationES.cpp?l=2479

[2] ANGLE's RenderStateCache::getDepthStencilState , which populates the D3D11_DEPTH_STENCIL_DESC structure:
https://cs.chromium.org/chromium/src/third_party/angle/src/libANGLE/renderer/d3d/d3d11/RenderStateCache.cpp?l=185

[3] Documentation for D3D11_DEPTH_STENCIL_DESC:
https://msdn.microsoft.com/en-us/library/windows/desktop/ff476110(v=vs.85).aspx

kenrussell:
After attempting to make this change in ANGLE, I discovered that masking the stencil reference value by the number of stencil bits is fundamentally incompatible with the OpenGL ES spec. OpenGL ES already specifies that the stencil reference value is clamped to the range [0, 2^s-1], where s is the number of bits in the current framebuffer's stencil buffer.

The rule that the WebGL spec will need to implement is that the draw call is rejected if the stencil front and back reference values are different, after *clamping* to the range [0, 2^s-1]. Can we agree on these new semantics? It is going to be impossible to update ANGLE otherwise.

I attempted to update the test in:
https://github.com/KhronosGroup/WebGL/pull/2592

and implement the new checks in ANGLE here:
https://chromium-review.googlesource.com/890605

In Chromium, this required this patch to be applied:
https://chromium-review.googlesource.com/890606

It doesn't work yet. Errors aren't being generated where they should be, and some assertion is being hit when the -1 reference value is sent. I don't know how to see the stack traces for those assertion failures on Windows.

I think we need to get this test passing 100% somewhere before we merge this pull request, so we know whether the new semantics are workable.

jdashg:
Can we just require that they're all the same? That's way simpler, and ANGLE wouldn't have to do any crazy stuff. Since it's required in D3D11, apps that were targeting d3d11 natively would have to deal with not having this. It is definitely simpler if we just standardize on embracing d3d11's restrictions directly.

kenrussell:
WebGL's current rules are that everything must match. An external developer found while running the drawElements tests on top of ANGLE that this was too restrictive, and @null77 and @vonture reviewed https://chromium-review.googlesource.com/302703 to make these checks more compatible with the dEQP.

I think I can debug the clamping of the reference values. If we get that working in ANGLE then @jdashg are you OK with changing the WebGL spec validation to include that?

It would be best to not revert https://chromium-review.googlesource.com/302703 . Because of masking and clamping, the Direct3D restrictions are actually that these writemasks and reference values only have to be equal up to the number of bits in the stencil buffer.

jdashg:
I would actually prefer to retain the explicit d3d11 restriction, if deqp is the only place we've run into problems with it. I think that would be simpler, and shouldn't change how actual content is written, since it's not useful to rely on masks/values outside the number of stencil bits.

I don't think we should put in extra work just so we happen to pass deqp. I'm actually surprised that restricting things to the number of stencil bits is enough to pass deqp, since restricting to the number of stencil bits doesn't bring back differing front/back masks and references.

jdashg:
Updated. I'll check this against my local patches tomorrow.
@RafaelCintron Please re-review.

kenrussell:
Looks great! Great work @jdashg and @kainino0x!

sdk/tests/conformance/misc/webgl-specific-stencil-settings.html
```diff
@@ -0,0 +1,189 @@
...
+gl.framebufferRenderbuffer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.RENDERBUFFER, colorRB);
+
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "initial framebuffer setup")
+
+function testStencilSettings(haveDepthBuffer, haveStencilBuffer, enableStencilTest, errIfMismatch) {
+    debug("");
+    debug("With depthbuffer=" + haveDepthBuffer +
+            ", stencilbuffer=" + haveStencilBuffer +
+            ", stencilTest=" + enableStencilTest +
+            ", expecting error=" + errIfMismatch +
```
RafaelCintron: Nit: Consider using `glEnumToString(errIfMismatch)` to make the output read a bit easier.  `wtu.shouldGenerateGLError` already uses this helper function.

sdk/tests/conformance/misc/webgl-specific-stencil-settings.html
```diff
@@ -0,0 +1,189 @@
...
+var wtu = WebGLTestUtils;
+description("Tests that stencil mask/func are validated correctly when the front state and back state differ.");
+
+var gl = wtu.create3DContext();
+var program = wtu.loadStandardProgram(gl);
+gl.useProgram(program);
+var vertexObject = gl.createBuffer();
+gl.bindBuffer(gl.ARRAY_BUFFER, vertexObject);
+gl.enableVertexAttribArray(0);
+gl.vertexAttribPointer(0, 4, gl.FLOAT, false, 0, 0);
```
RafaelCintron: Nit: Since the draw primitive is not relevant for the test, consider using `wtu.setupQuad` and `wtu.drawQuad` to save lines of code.

sdk/tests/conformance/misc/webgl-specific-stencil-settings.html
```diff
@@ -0,0 +1,189 @@
...
+    wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.stencilMaskSeparate(gl.BACK,  " + mask[1] + ")");
+    // If an error is generated, it should be at draw time.
+    wtu.shouldGenerateGLError(gl, error, "gl.drawArrays(gl.POINTS, 0, 0)", "drawArrays");
+}
+
+function testStencilMask(errIfMismatch) {
+    testStencilMaskCase([0, 256], gl.NO_ERROR);
+    testStencilMaskCase([1, 256], errIfMismatch);
+    testStencilMaskCase([1, 257], gl.NO_ERROR);
+    testStencilMaskCase([1, 258], errIfMismatch);
```
RafaelCintron: If the max stencil value is 255 and we're ANDing that with the front and back masks, I don't understand how any of these could pass without error.  The front and back masks have to match.   

I must be missing something.

sdk/tests/conformance/misc/webgl-specific-stencil-settings.html
kainino0x: Thanks, I forgot about that helper.

sdk/tests/conformance/misc/webgl-specific-stencil-settings.html
kainino0x: Good point, I copied this from webgl-specific without thinking about whether it was all necessary.

sdk/tests/conformance/misc/webgl-specific-stencil-settings.html
kainino0x: As written, the front and back mask settings don't have to match - just the mask bits that actually take effect. On an 8 bit stencil buffer, only the least significant 8 bits take effect. So the check:
`(0 & 255) == (256 & 255)`
passes (NO_ERROR).

kainino0x:
Here:
https://github.com/jdashg/WebGL/pull/5

sdk/tests/conformance/misc/webgl-specific-stencil-settings.html
RafaelCintron: Ah, right.  I misread the code.

jdashg:
Just waiting on CI.

kenrussell:
Great. Let's make sure to squash and merge this one and compose a good commit message, to keep the commit history clearer.

kenrussell:
@jdashg and I synced up and he wants to finish testing this with Firefox before merging, so let's wait for confirmation. Or @jdashg feel free to merge yourself. Thanks.

jdashg:
Confirmed fixed by my pending Gecko patches too, fwiw.

PR #2584, opened 2018-01-19
Update new test harness's use of Telemetry API.

kenrussell:
It seems this WebDriver based harness also uses the Chrome-specific
Telemetry API when running on ChromeOS.

Fixes http://crbug.com/803843 . Thanks to nednguyen@ for the patch.


kenrussell:
Simple change. Merging myself.

Note haven't tested this because I don't have a ChromeOS setup.

kenrussell:
Note: the bug above is unfortunately and unnecessarily restricted view. The API change in question happened in http://crbug.com/801578 .

Issue #2585, opened 2018-01-24
Not returning INVALID_OPERATION when using incompatible texture types?


greggman:
Section 5.3 of the spec says

> ## 5.23 A sampler type must match the internal texture format
>
> Texture lookup functions return values as floating point, unsigned integer or signed integer, depending on the sampler type passed to the lookup function. If the wrong sampler type is used for texture access, i.e., the sampler type does not match the texture internal format, the returned values are undefined in OpenGL ES Shading Language 3.00.6 (OpenGL ES Shading Language 3.00.6 §8.8). In WebGL, generates an INVALID_OPERATION error in the corresponding draw call, including drawArrays, drawElements, drawArraysInstanced, drawElementsInstanced , and drawRangeElements.
>
> If the sampler type is floating point and the internal texture format is normalized integer, it is considered as a match and the returned values are converted to floating point in the range [0, 1].

Section 3.8 of the OpenGL ES 3.0 spec says

> The internal data type of a texture may be signed or unsigned normalized fixedpoint,
signed or unsigned integer, or floating-point, depending on the internal format
of the texture. The correspondence between the internal format and the internal
data type is given in tables 3.13-3.14. Fixed-point and floating-point textures return
a floating-point value and **integer textures return signed or unsigned integer values**.
The fragment shader is responsible for interpreting the result of a texture lookup as
the correct data type, otherwise the result is undefined.

Which suggests that `R16I` is not a normalized integer. 

And yet neither Firefox nor Chrome is generating `INVALID_OPERATION` when using an `R16I` texture with a `sampler3D`.

https://jsfiddle.net/greggman/vt01t0xt/

Am I incorrect?

zhenyao:
You are correct. I just checked chromium implementation. In GLES2DecoderImpl::PrepareTexturesForRender() we fail to check if the sampler type matches the texture format.

zhenyao:
@jdashg FYI

zhenyao:
Gregg, do you mind adding your test into conformance suite?

zhenyao:
I filed crbug.com/805518 for chrome

jdashg:
Yep, I noticed we don't do this right now.

jdashg:
Resolved by #2587.

PR #2587, opened 2018-01-25
add test for incompatible texture types for samplers

greggman:
See #2585

note: shadow sampler types are not tested here.
they probably need a separate test as they have
more requirements related to texture parameters


zhenyao:
Thanks for your contribution. This is a very thorough test!

PR #2592, opened 2018-01-28
NOT FOR COMMIT: showing changes to #2583.

kenrussell:
Showing work-in-progress changes against @jdashg 's #2583. Still don't have this version of the test passing yet in ANGLE.


kainino0x:
I merged these changes into jdashg's PR #2583

PR #2594, opened 2018-01-30
Add a test for a HLSL bug with if, return and else if

Oletus:
An if statement containing a return together with an else if produces
wrong output on Chrome and Firefox on Windows.

This is a minimal test case for this bug: http://anglebug.com/2325


zhenyao:
LGTM

PR #2595, opened 2018-02-01
Reset vertex attrib divisors between instancing tests

Oletus:
The newest version of the ANGLE_instanced_arrays test runs different
subtests with different shaders, which may have overlapping attrib
locations. Make sure that the different subtests don't interfere with
each other by resetting vertex attrib divisors after each subtest.

Previously the test could pass if attrib locations were automatically
assigned in a certain way, but changes to attribute location
assignment in the WebGL implementation could make it fail when it
should have passed.

This change does not need to be backported to earlier conformance suite
versions nor the WebGL 2 test for instancing, as they only use one
shader and always bind vertex attrib locations explicitly, so the
locations can only be set one way.


zhenyao:
Thanks for catching this issue and fixing it. LGTM.

PR #2597, opened 2018-02-06
Assign attribute locations explicitly in luma texture copy test

Oletus:
The test is not querying attribute locations, so use the test
utilities to bind the attributes to locations 0 and 1. This ensures
that the test runs consistently.

This change is backported also to conformance suite version 2.0.0.


Oletus:
@zhenyao this will also be needed in Chromium to be able to merge the symbol table optimizations into ANGLE. Could you do another WebGL roll with this change? It should be fairly easy since no expectations changes should be needed.

zhenyao:
LGTM. I will take care of the roll.

Oletus:
Thanks a lot!

zhenyao:
Thank you!

PR #2598, opened 2018-02-07
Define that sync objects do not become signaled until control returns to the event loop.

kenrussell:
Follow the same semantics for sync objects as were previously used for
queries. Add a test for the new behavior.

### DIFF
--- a/sdk/tests/conformance2/sync/sync-webgl-specific.html
+++ b/sdk/tests/conformance2/sync/sync-webgl-specific.html
@@ -51,6 +51,7 @@ var sync = null;
 
 if (!gl) {
   testFailed("context does not exist");
+  finishTest();
 } else {
   testPassed("context exists");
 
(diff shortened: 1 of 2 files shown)


kenrussell:
@jdashg please review.

@zhenyao @kainino0x @jdarpinian FYI.

kainino0x:
Tests LGTM as well.

sdk/tests/conformance2/sync/sync-webgl-specific.html
```diff
@@ -77,11 +78,99 @@
...
+}
+
+function runGetSyncParameterTest() {
+  debug("");
+  debug("Verifying sync object isn't signaled too early");
+  sync = gl.fenceSync(gl.SYNC_GPU_COMMANDS_COMPLETE, 0);
+  // Verify as best as possible that the implementation doesn't allow a sync
+  // object to become signaled in the same frame, by spin-looping for some time.
+  var startTime = Date.now();
+  while (Date.now() - startTime < 2000) {
```
jdashg: 300ms-1000ms is fine. 2000 seems unnecessarily high.

sdk/tests/conformance2/sync/sync-webgl-specific.html
```diff
@@ -77,11 +78,99 @@
...
+  requestAnimationFrame(finishSyncParameterTest);
+}
+
+var startTimeOfFinish = 0;
+
+function finishSyncParameterTest() {
+  if (startTimeOfFinish == 0) {
+    startTimeOfFinish = Date.now();
+  }
+  if (Date.now() - startTimeOfFinish > 2000) {
```
jdashg: 2000 used without being a named constant again. Please name and define one.

jdashg:
Ok, this works for me. The title here is inaccurate though, since this matches the query behavior that the app must return to the event loop, not necessarily the next frame. It's worth correcting so the changelog shows what we actually did.

sdk/tests/conformance2/sync/sync-webgl-specific.html
kenrussell: Changed to 1000 ms.

sdk/tests/conformance2/sync/sync-webgl-specific.html
kenrussell: Thanks for your feedback. Named and defined one.

kenrussell:
Thanks @jdashg for your review. I rebased, squashed and edited the commit message to be accurate. Will update the pull request title here too. Merging.

PR #2600, opened 2018-02-19
Dirty flag tests

tsherif:
@null77 @kenrussell Finally got around to writing these. Was a bit swamped last month. Wasn't exactly sure how to organize them, so let me know if you'd like anything moved around. 

I made two versions of the tests that apply to both WebGL 1 and 2. If the WebGL 1 versions are enough, I can remove the WebGL 2 ports.

### DIFF
--- a/sdk/tests/conformance/attribs/gl-disabled-vertex-attrib-update.html
+++ b/sdk/tests/conformance/attribs/gl-disabled-vertex-attrib-update.html
@@ -87,8 +87,8 @@ for (var i = 0; i < testValues.length; ++i) {
   gl.vertexAttrib1f(attribLoc, testValue);
   wtu.clearAndDrawIndexedQuad(gl, gridRes);
   wtu.checkCanvas(gl, [0, 255, 0, 255], "should be green");
-} 
-  
+}
+
 wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be no errors");
 
 var successfullyParsed = true;
--- a/sdk/tests/conformance/attribs/gl-disabled-vertex-attrib-update.html
+++ b/sdk/tests/conformance/attribs/gl-disabled-vertex-attrib-update.html
@@ -61,7 +61,7 @@ void main() {
 <script>
 // Tests that repeatedly updating a disabled vertex attribute works as expected.
 // This covers an ANGLE bug where dirty bits for current values were ignoring repeated updates.
-// Based on: https://github.com/google/angle/blob/master/src/tests/gl_tests/VertexAttributeTest.cpp#L1454-L1500
+// Based on ANGLE test (VertexAttributeTest, DisabledAttribUpdates) from https://github.com/google/angle/blob/master/src/tests/gl_tests/VertexAttributeTest.cpp
 "use strict";
 var wtu = WebGLTestUtils;
 description();
--- a/sdk/tests/conformance/attribs/gl-disabled-vertex-attrib-update.html
+++ b/sdk/tests/conformance/attribs/gl-disabled-vertex-attrib-update.html
@@ -1,0 +1,100 @@
+<!--
+
+/*
+** Copyright (c) 2012 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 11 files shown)


kainino0x:
The WebGL 1.0 tests are run as part of the WebGL 2.0 conformance, so if the tests have the same content then we only need one copy.

tsherif:
Didn't realize that, thanks. The only difference was that I used GLSL 3 for the ports, so I've removed them.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -661,6 +661,7 @@ var setupIndexedQuadWithOptions = function (gl, options) {
 
   if (options.colorLocation !== undefined) {
     var colors = new Float32Array(numVerts * 4);
+    poffset = 0;
```
kainino0x: This change seems right. But how did it work before? Do you think this code path was just not used in any other tests?

sdk/tests/conformance/attribs/gl-disabled-vertex-attrib-update.html
```diff
@@ -0,0 +1,100 @@
+<!--
+
+/*
+** Copyright (c) 2012 The Khronos Group Inc.
```
kainino0x: 2012 -> 2018 here and the other new files

sdk/tests/conformance/attribs/gl-disabled-vertex-attrib-update.html
```diff
@@ -0,0 +1,100 @@
...
+varying float v_result;
+void main() {
+  gl_FragColor = v_result > 0.0 ? vec4(0.0, 1.0, 0.0, 1.0) : vec4(1.0, 0.0, 0.0, 1.0);
+}
+</script>
+
+<script>
+// Tests that repeatedly updating a disabled vertex attribute works as expected.
+// This covers an ANGLE bug where dirty bits for current values were ignoring repeated updates.
+// Based on ANGLE test (VertexAttributeTest, DisabledAttribUpdates) from https://github.com/google/angle/blob/master/src/tests/gl_tests/VertexAttributeTest.cpp
```
kainino0x: Just in case this changes, can you use a git hash instead of `master` in these URLs? (here and below)

sdk/tests/js/webgl-test-utils.js
tsherif: The `colorLocation` option is never used in the code base. I used it for a first iteration of one of the tests and found it broken. I switched to setting things manually so that option's gone back to being unused.

tsherif:
@kainino0x I addressed the issues with the copyright and links. Let me know if there's anything else.

kenrussell:
@tsherif *thank you* for putting together these tests. They look excellent. Were you able to verify (perhaps with an older Chromium build from https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Win/ , linked from https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html and https://build.chromium.org ) that these tests actually caught the bugs?

@kainino0x thank you for doing the first reviews!

tsherif:
@kenrussell Didn't get a chance to check because I'm not sure in which versions the bug was present. If you can point me to one, I'd be happy to test it out.

kainino0x:
The earliest of these tests seems to have been added on 2017-08-30.
An ANGLE roll before that was https://crrev.com/fa614d15, commit position 498282.
So here's a build before that point:
https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Win/488518/
(which was just before the branch for Chrome 61.)

tsherif:
The tests are all passing in that build. I'll take a look and see if I made any obvious porting errors.

The ANGLE commit ID I get from chrome://gpu is 0d2ecb4ea992. Could you guys confirm that the original ANGLE tests fail on that commit?

tsherif:
I can confirm that that build does not demonstrate the bug, and I'm having trouble finding one that the does. I hunted down the original bug tickets that @null77 pointed me to for the tests: 
https://bugs.chromium.org/p/chromium/issues/detail?id=779675
https://bugs.chromium.org/p/chromium/issues/detail?id=778689
They provide two test cases:
https://cad.onshape.com/documents/3bd144e9d07128b23daaa194/w/dfd7445ab6dce84f2b2b621c/e/fd9047327da21caec41aae82
https://jsfiddle.net/wy3cra8g/

I tried the test cases in the build Kai provided as well as the first bad builds mentioned in the bug tickets, 503308 and 502908, and they work fine in all three.

I'm not sure how to proceed. Maybe someone on the ANGLE team can provide some insight?

null77:
@tsherif , I haven't had a chance yet to check this out, but will put it on the list of things to look at. Are you sure your D3D11 back-end is getting initialized correctly? you could share the contents of about:gpu.

tsherif:
@null77 Find the Chrome GPU dumps attached for the builds I tested. Let me know if anything looks suspicious. They all have the renderer string as `ANGLE (NVIDIA GeForce GT 650M Direct3D11 vs_5_0 ps_5_0)` and there aren't any errors being logged.

[chrome-gpu-498282.txt](https://github.com/KhronosGroup/WebGL/files/1762628/chrome-gpu-498282.txt)
[chrome-gpu-502908.txt](https://github.com/KhronosGroup/WebGL/files/1762629/chrome-gpu-502908.txt)
[chrome-gpu-503308.txt](https://github.com/KhronosGroup/WebGL/files/1762631/chrome-gpu-503308.txt)

jdashg:
@tsherif Can you sign the CLA so I can merge this?

PR #2602, opened 2018-03-02
Expanded transform feedback tests

jdarpinian:
Adds a new test switching-objects.html which tests various consequences of switching transform feedback objects. This includes a test for the new behavior where TRANSFORM_FEEDBACK_BUFFER is not modified when switching transform feedback objects. The OpenGL/ES working group just approved the spec change for this new behavior.

Also adds testing of drawElements cases in simultaneous_binding.html, in addition to existing drawArrays cases.

### DIFF
--- a/sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
+++ b/sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
@@ -72,13 +72,14 @@ if (!gl) {
     testPassed("WebGL context exists");
 }
 
-function drawWithFeedbackBound(gl, prog, vao, tf, enableFeedback) {
+function drawWithFeedbackBound(gl, drawFunction, prog, vao, tf, enableFeedback) {
   gl.useProgram(prog);
   gl.bindVertexArray(vao);
   gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, tf);
   if (enableFeedback) gl.beginTransformFeedback(gl.POINTS);
-  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "Should not be any errors before drawing");
-  gl.drawArrays(gl.POINTS, 0, 4);
+  let error = gl.getError();
+  if (error != gl.NO_ERROR) testFailed("Unexpected error before drawing: " +  error)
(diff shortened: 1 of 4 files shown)


sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
```diff
@@ -125,8 +132,14 @@
 gl.uniformBlockBinding(prog, ubi, 0);
 gl.bindBufferBase(gl.UNIFORM_BUFFER, 0, uniformBuffer);
 
-drawWithFeedbackBound(gl, prog, vao, tf, true);
+let drawArrays = ()=>gl.drawArrays(gl.POINTS, 0, 4);
+let drawElements = ()=>gl.drawElements(gl.POINTS, 4, gl.UNSIGNED_SHORT, 0);
```
zhenyao: What about draw*Instanced?

sdk/tests/conformance2/transform_feedback/switching-objects.html
```diff
@@ -0,0 +1,245 @@
+<!--
+
+/*
+** Copyright (c) 2017 The Khronos Group Inc.
```
zhenyao: nit: 2018

zhenyao:
LGTM

sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
jdarpinian: Done. In the process I discovered some problems with draw*Instanced buffer size checking. New test here: https://github.com/KhronosGroup/WebGL/pull/2604

sdk/tests/conformance2/transform_feedback/switching-objects.html
jdarpinian: Done

zhenyao:
Still LGTM
Feel free to merge yourself

jdashg:
Please supply a new PR that includes a link (even if khronos-private) to the resolution for new behavior. None of the ES specs have been rev'd, so we need to errata this in the spec text for now.

PR #2603, opened 2018-03-02
Include file path fix for a test.

vikaschromie:
This test was passing when using python SimpleHTTPServer even if the include
paths were incorrect. This is because python smartly searched for these include
files in other folders. The test failed when command line was used to run test
by bots or when running test directly on chrome without using the simplehttpserver.


zhenyao:
LGTM

PR #2604, opened 2018-03-03
Test transform feedback into too small buffers.

jdarpinian:
The ES spec says that drawing should always fail with GL_INVALID_OPERATION when the bound transform feedback buffers are too small to hold the result. Chrome and Firefox both fail some of these tests (on Linux Nvidia at least).

### DIFF
--- a/sdk/tests/conformance2/transform_feedback/too-small-buffers.html
+++ b/sdk/tests/conformance2/transform_feedback/too-small-buffers.html
@@ -69,27 +69,6 @@ if (!gl) {
     testPassed("WebGL context exists");
 }
 
-function drawWithFeedbackBound(gl, drawFunction, prog, vao, tf, enableFeedback) {
-  gl.useProgram(prog);
-  gl.bindVertexArray(vao);
-  gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, tf);
-  if (enableFeedback) gl.beginTransformFeedback(gl.POINTS);
-  let error = gl.getError();
-  if (error != gl.NO_ERROR) testFailed("Unexpected error before drawing: " +  error)
-  drawFunction();
-  if (enableFeedback) gl.endTransformFeedback();
-  gl.bindVertexArray(null);
-  gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, null);
-}
-
-function createBuffer(gl, dataOrSize) {
-  const buf = gl.createBuffer();
-  gl.bindBuffer(gl.ARRAY_BUFFER, buf);
-  gl.bufferData(gl.ARRAY_BUFFER, dataOrSize, gl.STATIC_DRAW);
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/transform_feedback/too-small-buffers.html
```diff
@@ -0,0 +1,242 @@
...
+    drawFunction();
+    wtu.glErrorShouldBe(gl, error, "draw");
+    gl.endTransformFeedback();
+  }
+
+  gl.useProgram(progInterleaved);
+
+  debug("<h3>interleaved - Baseline success case</h3>")
+  gl.bindBufferBase(gl.TRANSFORM_FEEDBACK_BUFFER, 0, tfBuffer0);
+  gl.bufferData(gl.TRANSFORM_FEEDBACK_BUFFER, 8*4, gl.STREAM_READ);
```
zhenyao: nit: if you use a const sizeOfFloat instead of 4, it's easier to read. Here and below.

sdk/tests/conformance2/transform_feedback/too-small-buffers.html
```diff
@@ -0,0 +1,242 @@
...
+
+  debug("<h3>interleaved - Buffer too small</h3>")
+  gl.bufferData(gl.TRANSFORM_FEEDBACK_BUFFER, 8*4-1, gl.STREAM_READ);
+  doTransformFeedback(drawFunction, gl.INVALID_OPERATION);
+  wtu.checkFloatBuffer(gl, gl.TRANSFORM_FEEDBACK_BUFFER,
+      [0, 0, 0, 0, 0, 0, 0]);
+
+  debug("<h3>interleaved - Multiple draws success case</h3>")
+  gl.bindBufferBase(gl.TRANSFORM_FEEDBACK_BUFFER, 0, tfBuffer0);
+  gl.bufferData(gl.TRANSFORM_FEEDBACK_BUFFER, 8*4*2, gl.STREAM_READ);
```
zhenyao: Maybe I misunderstood something, but why do we need twice the data store? Won't the second draw simply overwrite the results from the first draw?

Same question for next case also.

zhenyao: Per offline discussion with @jdarpinian , ES spec 3.0.4, page 89,

"When transform feedback is paused, no vertices are recorded. When transform
feedback is resumed, subsequent vertices are appended to the bound buffer objects
immediately following the last vertex written before transform feedback was
paused."

So it seems between draw calls within a transformfeedback session (between Begin* and End* calls), buffer pointer does advance and carry over to the next draw call.

sdk/tests/conformance2/transform_feedback/too-small-buffers.html
jdarpinian: Done.

kenrussell:
Great work @jdarpinian on these new tests. LGTM after the fact.

PR #2606, opened 2018-03-09
Support robust_buffer_access_behavior implementations.

jdashg:
In particular, don't try to draw $LARGE numbers of triangles. An RBAB
implementation will not reject these 'invalid' draws, and will try to draw
all the triangles, which, for N=0x7fffffff trips the TDR and incurs context
loss.


kenrussell:
As discussed in the working group meetings, looks good.

Issue #2500, opened 2017-09-07
Angle-instanced-arrays-out-of-bounds.html contains ambigous error condition check


RafaelCintron:
The WebGL 1.0 test [angle-instanced-arrays-out-of-bounds.html ](https://www.khronos.org/registry/webgl/sdk/tests/conformance/extensions/angle-instanced-arrays-out-of-bounds.html?webglVersion=1&quiet=0) contains the following section:

````
Test with an empty buffer for the instanced attribute
...
PASS getError was expected value: NO_ERROR : after evaluating: ext.drawElementsInstancedANGLE(gl.TRIANGLES, 10000, gl.UNSIGNED_BYTE, 0, 0)
...
````

Note that we pass a primCount of zero but a count of 10000.  This call to drawElementsInstancedANGLE is technically incorrect since the element array buffer only has 9 items and we are passing a count of 10000.  However, many implementations will no-op the call since the primCount is 0 and not bother to check for other errors.  ANGLE, on the other hand, performs the no-op check after checking for element array buffer out of bounds. Hence, the test fails.  

````C++
bool ValidateDrawElementsInstancedBase(ValidationContext *context,
                                       GLenum mode,
                                       GLsizei count,
                                       GLenum type,
                                       const GLvoid *indices,
                                       GLsizei primcount)
{
    if (primcount < 0)
    {
        context->handleError(InvalidValue() << "primcount cannot be negative.");
        return false;
    }

    if (!ValidateDrawElementsCommon(context, mode, count, type, indices, primcount))
    {
        return false;
    }

    // No-op zero primitive count
    return (primcount > 0);
}
````

What is the right answer here?  Should early outs win or should error checking win?  I'm inclined to go with the latter for this particular case. If you call drawElementsInstanced and pass a primitive count of zero, you're doing it wrong.

greggman:
I'm not sure I'm conviced

it is perfectly reasonable to be able to pass a primcount of 0.

let's say you have an interactive demo with a slider that goes from 0 to
1000 primcount. You shouldn't have to special case zero.

whether or not count can be larger then your buffers when primcount is zero
is a separate issue from allowing primcount to be zero.

erroring when count is too large and primcount is zero seems to be
inconsistent with other parts of the spec. Youre not asking to draw so
nothing is out of bounds.

I also dont see how it's technically incorrect
real GL doesn't check so it would run just fine

kainino0x:
> Youre not asking to draw so nothing is out of bounds.

[...]

PR #2607, opened 2018-03-13
Spec drawElements behavior with null ELEMENT_ARRAY_BUFFER binding.

jdashg:


jdashg:
This should resolve issue #2500.

jdashg:
@kenrussell @RafaelCintron bump

sdk/tests/js/tests/out-of-bounds-test.js
```diff
@@ -136,11 +136,15 @@ var runDrawElementsTest = function(callTemplate, gl, wtu, ext) {
     gl.bufferData(gl.ARRAY_BUFFER, new Float32Array([ 0,0.5,0, -0.5,-0.5,0, 0.5,-0.5,0 ]), gl.STATIC_DRAW);
     gl.vertexAttribPointer(0, 3, gl.FLOAT, false, 0, 0);
 
-    var indexObject = gl.createBuffer();
 
+    debug('');
+    debug('Test null index buffer');
+    gl.bindBuffer(gl.ELEMENT_ARRAY_BUFFER, null);
+    wtu.shouldGenerateGLError(gl, gl.INVALID_OPERATION, wtu.replaceParams(callTemplate, {count: 0, type: 'gl.UNSIGNED_BYTE', offset: 0}));
```
kenrussell: I didn't remember us deciding to make this specific change for zero-length draw calls. Is this needed because this is part of the validation done for draw calls with length > 0? If so, OK – what are we going to do about older versions of the conformance suite? Can you supply changes back to version 1.0.1 that allow both the old and new behaviors?

specs/latest/1.0/index.html
```diff
@@ -3462,12 +3462,22 @@ <h3><a name="BUFFER_OBJECT_BINDING">Buffer Object Binding</a></h3>
...
+<p>
+If a vertex attribute is enabled as an array via <code>enableVertexAttribArray</code> but no buffer
+is bound to that attribute (generally via <code>bindBuffer</code> and
+<code>vertexAttribPointer</code>), then draw commands (<code>drawArrays</code> or
+<code>drawElements</code>) will generate an <code>INVALID_OPERATION</code> error.
+
+<p>
+If an indexed draw command (<code>drawElements</code>) is called and no <code>WebGLBuffer</code> is
+bound to the <code>ELEMENT_ARRAY_BUFFER</code> binding point, an <code>INVALID_OPERATION</code>
+error is generated.
```
kenrussell: Per above – I'm guessing it's not possible to bypass this one check only for zero-length draw calls?

specs/latest/1.0/index.html
```diff
@@ -3590,7 +3589,11 @@ <h3><a name="ATTRIBS_AND_RANGE_CHECKING">Enabled Vertex Attributes and Range Che
 cause any error to be generated during a call to <code>drawArrays</code>
 or <code>drawElements</code>.
 
-</p>
+    <h3>Out-of-bounds fetches from the index buffer</h3>
+
+<p>
+Calling an indexed drawing command (<code>drawElements</code>) that fetches index elements outside
+the bounds of ELEMENT_ARRAY_BUFFER will result in an INVALID_OPERATION error.
```
RafaelCintron: I presume we'll need to modify @null77 's [Don't no-op draw calls for zero count in validation](https://github.com/google/angle/commit/9fdaa497b489b7e7739e29812c85da2125b6a681) change to take these new rules into account?

specs/latest/1.0/index.html
jdashg: We could, but it breaks pattern with how we handle null-buffer vertexAttribArrays. IMO it's cleaner to just always do it?

specs/latest/1.0/index.html
jdashg: Yes, probably. Firefox does this validation before we call into ANGLE, fwiw.

sdk/tests/js/tests/out-of-bounds-test.js
jdashg: Validation should be consistent [0,N], instead of special-casing 0. If the validation code usually requires an index buffer, then we should check it even if it's technically fine to omit. (particularly because there's no performance incentive to make no-op draw calls fast)

jdashg:
Is there anything else outstanding to address here?

sdk/tests/js/tests/out-of-bounds-test.js
kenrussell: OK.

jdashg:
Running my patched build of Firefox against the current top-of-tree /without/ this change passes, so previous snapshots should be fine.

I think we're good to merge.

PR #2608, opened 2018-03-13
constant-precision-qualifier.html is finishing early due to an extra finishTest()

patrto:
finishTest() is called right after test() returns, ending the test prematurely before all the test cases complete. This is because test() has a call to GLSLConformanceTester.runRenderTests(), which runs the tests asynchronously. runRenderTests() also already calls finishTest() after the last test completes.

### DIFF
--- a/sdk/tests/conformance/glsl/bugs/constant-precision-qualifier.html
+++ b/sdk/tests/conformance/glsl/bugs/constant-precision-qualifier.html
@@ -95,10 +95,12 @@ function test() {
   var gl = wtu.create3DContext();
   if (!gl) {
     testFailed("context does not exist");
+    finishTest();
     return;
   }
   if (gl.getShaderPrecisionFormat(gl.FRAGMENT_SHADER, gl.HIGH_FLOAT).precision == 0) {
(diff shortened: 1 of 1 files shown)


zhenyao:
I agree we don't need the finishTest() in the end because GLSLConformanceTester.runRenderTests() already call it in all its paths. However, I don't think GLSLConformanceTester.runRenderTests() is asynchronous so the current code should not cause any problem, only it's less elegant. Just want to clarify for the record. Still LGTM.

patrto:
Thank you for the review @zhenyao

GLSLConformanceTester.runRenderTests() uses setTimeout() between tests, so with the current code, GLSLConformanceTester.runRenderTests() returns after the first test case and calls finishTest() before the second test case starts.

zhenyao:
I didn't realize we are using setTimeout(). There is really no reason to use setTimeout(..., 1) and make it async. Instead, we should just execute tests in a for loop. Are you willing to upload another pull request to fix that?

patrto:
Sure, I can try it and make sure all the test still pass

PR #2610, opened 2018-03-13
Add test to verify DYNAMIC_DRAW bufferData functionality

patrto:
This test verifies that vertex attributes with DYNAMIC_DRAW buffers are updated correctly if the buffer is not always updated on every draw.

This is a test for the following ANGLE bug: https://bugs.chromium.org/p/angleproject/issues/detail?id=2381

### DIFF
--- a/sdk/tests/conformance/buffers/buffer-data-dynamic-delay.html
+++ b/sdk/tests/conformance/buffers/buffer-data-dynamic-delay.html
@@ -1,0 +1,114 @@
+<!--
+/*
+** Copyright (c) 2018 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/buffers/buffer-data-dynamic-delay.html
```diff
@@ -0,0 +1,114 @@
...
+gl.bufferData(gl.ARRAY_BUFFER, new Float32Array([
+     1.0,  1.0, 0.0, 1.0,
+    -1.0,  1.0, 0.0, 1.0,
+    -1.0, -1.0, 0.0, 1.0,
+     1.0,  1.0, 0.0, 1.0,
+    -1.0, -1.0, 0.0, 1.0,
+     1.0, -1.0, 0.0, 1.0
+    ]), gl.DYNAMIC_DRAW);
+gl.enableVertexAttribArray(0);
+gl.vertexAttribPointer(0, 4, gl.FLOAT, false, 0, 0);
```
jdashg: Please use vec2 and TRIANGLE_STRIP for simplicity unless you have reason to do otherwise.

PR #2611, opened 2018-03-13
Test queueing multiple context restored event after context is lost

derry19:
When calling restoreContext() multiple times after context is lost, each function call will cause a context restored event queued. All queued context restored events should be merged before a context restored event is fired. 
Calling restoreContext() multiple times should not cause error or crash. Added a test to test the behavior.


RafaelCintron:
sldkjfslkjgdlkfgjdlkgj

slfdkjd

RafaelCintron:
Apologies, I didn't mean to close the pull request.

kenrussell:
Sorry for the long delay reviewing.

I re-studied the WEBGL_lose_context specification and looked at Chrome's implementation. Agree that the two calls to restoreContext() should not cause any errors since context restoration is asynchronous. Checked in Chrome, Firefox and Safari and verified that the new check works in all browsers.

Merging now.

PR #2612, opened 2018-03-14
Remove setTimeout() from GLSLConformanceTester.runTests()

patrto:
GLSLConformanceTester.runTests() uses setTimeout(.., 1) in between tests, making the tests asynchronous. runOneTest(), the function called in the setTimeout callback, doesn't do intensive work that takes a long time, and it does a checkCanvas at the end of the function that would reset the harness timeout timer. Changing this to a synchronous for loop instead.

This PR is a followup from #2608

### DIFF
--- a/sdk/tests/js/glsl-conformance-test.js
+++ b/sdk/tests/js/glsl-conformance-test.js
@@ -311,17 +311,11 @@ function runTests(shaderInfos, opt_contextVersion) {
     return;
   }
 
-  var testIndex = 0;
-  var runNextTest = function() {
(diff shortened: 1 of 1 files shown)


patrto:
@zhenyao, please review when you get a chance

zhenyao:
LGTM
Thank you!

kenrussell:
For what it's worth, I think this was made asynchronous to allow the web browser to update its output while these potentially long-running tests were running. But as long as the tests continue to work reasonably well in all harnesses, simplifying it sounds good. Sorry for the long delay in commenting on this.

PR #2613, opened 2018-03-15
Test that transform feedback calculations don't overflow.

jdarpinian:


sdk/tests/conformance2/transform_feedback/too-small-buffers.html
```diff
@@ -235,6 +236,25 @@
...
+gl.bindBufferBase(gl.TRANSFORM_FEEDBACK_BUFFER, 0, tfBuffer0);
+gl.bufferData(gl.TRANSFORM_FEEDBACK_BUFFER, (1<<16)*sizeOfFloat*2, gl.STREAM_READ);
+
+gl.beginTransformFeedback(gl.POINTS);
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "before draw");
+// If count and primcount are stored in 32-bit signed integers and then
+// multiplied to calculate the number of transform feedback vertices, the
+// calculation will overflow to 0.
+gl.drawArraysInstanced(gl.POINTS, 0, 1<<16, 1<<16);
+wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "integer overflow");
```
zhenyao: I think this is either integer overflow or buffer not big enough. 32 bit buffer size limit is Chrome command buffer limit, it's not a limit from WebGL spec.

jdarpinian: True, though this bug could still exist in an implementation that supports >32 bit buffer sizes. The type of the arguments to drawArraysInstanced is GLsizei which is 32 bit on all platforms. If the multiplication is done before converting to 64 bit then this bug will still happen.

I've added a note that the buffer is too small.

zhenyao: True, that could an implementation bug that they fail to use a 64 bit variable to hold the minimal target buffer size.

PR #2614, opened 2018-03-21
Add a regression test for Chrome's DeleteBuffers behavior.

zhenyao:


zhenyao:
@kenrussell @kainino0x @jdarpinian Can one of you please take a look?

kenrussell:
Sorry for not reviewing this until now. LGTM after the fact.

PR #2615, opened 2018-03-26
Explicitly specify uniform and attribute name collisions

Oletus:
WebGL has long enforced behavior that goes against GLSL ES 3.00.6
section 12.47. WebGL requires that uniform and attribute names don't
collide, whereas GLSL ES specifies that there's no collision in this
case. The WebGL test that enforces this is:

conformance\glsl\misc\shaders-with-name-conflicts.html

The test refers to discussion in OpenGL ES working group, but it seems
that the results of the discussion were not put into writing outside
of the test case.

Explicitly specify the WebGL behavior to clear confusion around this.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -4148,6 +4148,22 @@ extensions.
     </div>
 </p>
 
+<h3>Uniform and attribute name collisions</h3>
+
+<p>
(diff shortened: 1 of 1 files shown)


zhenyao:
LGTM
Thanks Olli for clarifying this

Oletus:
Thanks for the quick review!

PR #2616, opened 2018-03-26
Clarify WebGL 2 spec for attribute aliasing

Oletus:
GLSL ES 3.00.6 already has strict rules when it comes to attribute
aliasing, so WebGL 2.0 can simply refer to those rather than spell out
its own.

A new test is added to fully cover the attribute aliasing rules of
GLSL ES 3.00.6. It shares some code with the older attribute aliasing
test for WebGL 1.0.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -3527,13 +3527,10 @@ WebGL2RenderingContext includes WebGL2RenderingContextBase;
 
     <h3><a name="VERTEX_ATTRIBUTE_ALIASING">Vertex Attribute Aliasing</a></h3>
     <p>
-        In the WebGL 2.0 API, if more than one active attribute name is bound to the same location, it is considered
-        aliased, regardless whether there exists a path through the shader that consumes more than of these
-        attributes. Such definition of aliasing is broader than what is defined in
-        <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.11.5">OpenGL ES 3.0.4 &sect;2.11.5</a>)</span>.
-    </p>
-    <p>
-        A link error is required when aliasing exists.
+        WebGL 2.0 API implementations must strictly follow GLSL ES 3.00.6 section 12.46, that
+        specifies that any vertex attribute aliasing is disallowed. In addition, WebGL 1.0 spec
+        section <a href="../1.0/index.html#ATTRIBUTE_ALIASING">Attribute aliasing</a> applies to
+        GLSL ES 1.00 shaders used with the WebGL 2.0 API.
     </p>
 
     <h3><a name="NO_PRIMITIVE_RESTART_FIXED_INDEX">PRIMITIVE_RESTART_FIXED_INDEX is always enabled</a></h3>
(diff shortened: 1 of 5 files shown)


Oletus:
This is one possible way to resolve conflicts between attribute aliasing rules in GLSL ES 3.00.6 vs WebGL 2.0.

On the other hand, if we want to ensure better WebGL backwards compatibility, we could opt for alternative, relaxed rules for WebGL 2.0 that would be more in line with WebGL 1.0 rules. This would also make the implementation simpler, since the set of attributes to validate would not be dependent on shader version. Any thoughts?

zhenyao:
I like the alternative where we allow aliasing but enforcing the WebGL 1.0 rule, a link error must occur when more than one of the aliased attributes is active in the executable program.

Oletus:
Here's an alternative pull request making the WebGL 2.0 rules consistent with WebGL 1.0 instead: https://github.com/KhronosGroup/WebGL/pull/2619

Oletus:
(also fixed missing JS file in this patch, in case we after all decide to go with this alternative)

zhenyao:
LGTM, but let's wait to hear others' opinions on this.

Oletus:
@zhenyao To be clear - do you mean that the other alternative looks good? Maybe further discussion should be on the other pull request.

zhenyao:
Yes, I meant the alternative. Sorry about the confusion.

Oletus:
So, are we going forward with this alternative as discussed last Thursday? @jdashg @kenrussell ?

specs/latest/2.0/index.html
```diff
@@ -3527,13 +3527,10 @@ <h3><a name="CLIENT_WAIT_SYNC">clientWaitSync</a></h3>
...
     <h3><a name="VERTEX_ATTRIBUTE_ALIASING">Vertex Attribute Aliasing</a></h3>
     <p>
-        In the WebGL 2.0 API, if more than one active attribute name is bound to the same location, it is considered
-        aliased, regardless whether there exists a path through the shader that consumes more than of these
-        attributes. Such definition of aliasing is broader than what is defined in
-        <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.11.5">OpenGL ES 3.0.4 &sect;2.11.5</a>)</span>.
-    </p>
-    <p>
-        A link error is required when aliasing exists.
+        WebGL 2.0 API implementations must strictly follow GLSL ES 3.00.6 section 12.46, that
```
jdashg: s/that/which/

specs/latest/2.0/index.html
```diff
@@ -3527,13 +3527,10 @@ <h3><a name="CLIENT_WAIT_SYNC">clientWaitSync</a></h3>
...
+        specifies that any vertex attribute aliasing is disallowed. In addition, WebGL 1.0 spec
+        section <a href="../1.0/index.html#ATTRIBUTE_ALIASING">Attribute aliasing</a> applies to
+        GLSL ES 1.00 shaders used with the WebGL 2.0 API.
```
jdashg: "As stated in ES 3.0.5 p57, ESSL1 shaders may still alias, as allowed by the WebGL 1.0 spec."

Oletus:
Addressed comments, merging this.

kenrussell:
LGTM after the fact. The expanded test looks good. Sorry for the delay reviewing.

PR #2617, opened 2018-03-26
Revamp text results, include json block for result details.

jdashg:


jdashg:
@kenrussell Need anything else from me here?

kenrussell:
Thanks Jeff. I didn't test this patch but it looks good. Merging now.

CC'ing @kainino0x @jdarpinian @zhenyao @grorg @RafaelCintron .

PR #2619, opened 2018-03-27
Clarify WebGL specs for attribute aliasing

Oletus:
The intent with WebGL attribute aliasing rules has been to ensure
interoperability and that the rules are consistent between WebGL 1.0
and WebGL 2.0.

The WebGL 2.0 attribute aliasing chapter did not previously take
into account that GLSL ES 3.00.6 has strict rules when it comes to
attribute aliasing. To make the rules match with WebGL 1.0, WebGL 2.0
needs to explicitly relax the rules from GLSL ES.

The spec sections on attribute aliasing are also clarified to talk
about statically used attributes rather than "active attributes". This
makes WebGL more testable, since which attributes are active can
depend on compiler optimizations, whereas static use has a clear
definition.

New tests are added to enforce the rule that statically used
attributes are validated for aliasing.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -3527,13 +3527,16 @@ WebGL2RenderingContext includes WebGL2RenderingContextBase;
 
     <h3><a name="VERTEX_ATTRIBUTE_ALIASING">Vertex Attribute Aliasing</a></h3>
     <p>
-        In the WebGL 2.0 API, if more than one active attribute name is bound to the same location, it is considered
-        aliased, regardless whether there exists a path through the shader that consumes more than of these
-        attributes. Such definition of aliasing is broader than what is defined in
-        <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.11.5">OpenGL ES 3.0.4 &sect;2.11.5</a>)</span>.
+        The WebGL 2.0 API relaxes the restrictions in GLSL ES 3.00.6 section 12.46. In the WebGL 2.0
+        API, two attributes that are bound to the same location are only considered to be aliased in
+        case they are both statically used. Otherwise the resolution of GLSL ES 3.00.6 section 12.46
+        applies to all shader programs used with the WebGL 2.0 API, so aliasing of statically used
+        attributes does cause linking to fail.
     </p>
     <p>
-        A link error is required when aliasing exists.
+        This rule is consistent with WebGL 1.0 spec section
+        <a href="../1.0/index.html#ATTRIBUTE_ALIASING">Attribute Aliasing</a>, so GLSL ES 1.00
+        shaders behave the same in WebGL 2.0 and WebGL 1.0 with respect to attribute aliasing.
     </p>
 
     <h3><a name="NO_PRIMITIVE_RESTART_FIXED_INDEX">PRIMITIVE_RESTART_FIXED_INDEX is always enabled</a></h3>
--- a/sdk/tests/conformance/attribs/00_test_list.txt
+++ b/sdk/tests/conformance/attribs/00_test_list.txt
@@ -1,4 +1,5 @@
 --min-version 1.0.3 gl-bindAttribLocation-aliasing.html
+--min-version 1.0.4 gl-bindAttribLocation-aliasing-inactive.html
 --min-version 1.0.3 gl-bindAttribLocation-matrix.html
 --min-version 1.0.4 gl-bindAttribLocation-nonexistent-attribute.html
 --min-version 1.0.4 gl-bindAttribLocation-repeated.html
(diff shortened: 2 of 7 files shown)


zhenyao:
This looks good to me.

specs/latest/2.0/index.html
```diff
@@ -3527,13 +3527,16 @@ <h3><a name="CLIENT_WAIT_SYNC">clientWaitSync</a></h3>
 
     <h3><a name="VERTEX_ATTRIBUTE_ALIASING">Vertex Attribute Aliasing</a></h3>
     <p>
-        In the WebGL 2.0 API, if more than one active attribute name is bound to the same location, it is considered
-        aliased, regardless whether there exists a path through the shader that consumes more than of these
-        attributes. Such definition of aliasing is broader than what is defined in
-        <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.11.5">OpenGL ES 3.0.4 &sect;2.11.5</a>)</span>.
+        The WebGL 2.0 API relaxes the restrictions in GLSL ES 3.00.6 section 12.46. In the WebGL 2.0
+        API, two attributes that are bound to the same location are only considered to be aliased in
+        case they are both statically used. Otherwise the resolution of GLSL ES 3.00.6 section 12.46
```
jdashg: We don't want to be more forgiving than our sources spec here. This won't be the only place where webglsl3 diverges from webglsl1.

zhenyao: It's one less thing that may break WebGL1 code when upgrading to WebGL2.

We already have the code for implementing WebGL1 on top of ES3 drivers. So extending that to WebGL2 should be trivia.

jdashg: No, I don't want to do this. Pulling forward restrictions is something we should be avoiding.

zhenyao: It's a little tiny effort on browser vendors' side, but could make upgrading WebGL 1 contents to WebGL 2 experience smoother, which is an important goal.

jdashg:
This requires more rather than less from our translator, and I don't buy
that it represents a commensurate improvement to porting ergonomics. I am
against such a relaxation.

On Mar 28, 2018 6:51 PM, "zhenyao" <notifications@github.com> wrote:

*@zhenyao* commented on this pull request.
------------------------------

In specs/latest/2.0/index.html
<https://github.com/KhronosGroup/WebGL/pull/2619#discussion_r177935824>:

> @@ -3527,13 +3527,16 @@ <h3><a name="CLIENT_WAIT_SYNC">clientWaitSync</a></h3>

     <h3><a name="VERTEX_ATTRIBUTE_ALIASING">Vertex Attribute Aliasing</a></h3>
     <p>
-        In the WebGL 2.0 API, if more than one active attribute name
is bound to the same location, it is considered
-        aliased, regardless whether there exists a path through the
shader that consumes more than of these
-        attributes. Such definition of aliasing is broader than what
is defined in
-        <span class="gl-spec">(<a
href="http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf#nameddest=section-2.11.5">OpenGL
ES 3.0.4 &sect;2.11.5</a>)</span>.
+        The WebGL 2.0 API relaxes the restrictions in GLSL ES 3.00.6
section 12.46. In the WebGL 2.0
+        API, two attributes that are bound to the same location are
only considered to be aliased in
+        case they are both statically used. Otherwise the resolution
of GLSL ES 3.00.6 section 12.46

It's a little tiny effort on browser vendors' side, but could make
upgrading WebGL 1 contents to WebGL 2 experience smoother, which is an
important goal.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/KhronosGroup/WebGL/pull/2619#discussion_r177935824>,
or mute
the thread

jdashg:
These are the sort of deviations that can bite us later down the road, especially in aggregate. Sticking closer to the source API should be the default, and I don't see this as relieving any serious pain, or at least no pain anyone who's already been working with es3 hasn't encountered. The use-case for aliased locations in general is low, and I wish we didn't support them at all.

zhenyao:
As Olli pointed out, this simplifies code path, i.e., our implementation behaves the same when dealing with GLSL ES 1.0 shaders, rather than one way with 1.0 shaders in WebGL 1.0, and another way in WebGL 2.0. It's confusing that the same shaders behave differently at LinkProgram() in WebGL 1.0 and 2.0.

I have no data how many contents will be affected by this aliasing rule, but WebGL 2.0 adoption is still low since only Firefox/Chrome supports it. More bug reports are likely to come in when WebGL 2.0 is universally supported.

Oletus:
This is the simplest alternative to implement, since here the rules are the same for WebGL 1.0 and 2.0.

This also preserves compatibility with WebGL 2.0 content that may have already been written against the current implementation.

However, I get Jeff's point that we don't want to deviate from the source specification without a very good reason. And we might want to implement the stricter rules for GLSL ES 3.10 either way, in which case this doesn't really save us the implementation trouble. The alternative pull request that changes the WebGL 2.0 rules to refer to GLSL ES 3.00.6 and adds a test for it is here: https://github.com/KhronosGroup/WebGL/pull/2616 I'm happy with either alternative myself.

PR #2620, opened 2018-03-28
Add a test case in for loop condition

qjia7:
The test in for loop expression can't expose an Intel driver bug. We
add another test in for loop condition to conver it.

### DIFF
--- a/sdk/tests/conformance2/glsl3/short-circuiting-in-loop-condition.html
+++ b/sdk/tests/conformance2/glsl3/short-circuiting-in-loop-condition.html
@@ -64,6 +64,30 @@ void main() {
   result = success ? vec4(0, 1.0, 0, 1.0) : vec4(0, 1.0, 0, 0);
 }
 </script>
+<script id="fshaderForCondition" type="x-shader/x-fragment">#version 300 es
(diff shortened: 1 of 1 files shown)


qjia7:
PTAL Thanks.

zhenyao:
What's wrong with the original code? It looks OK to me.

qjia7:
The problem is that the original code can't expose the Intel driver bug. The corresponding test is 'Pass' when u is false. However, if we use 'u && foo()' as the for condition like other 'while' shaders, the same error can be reproduced. Since this file is mainly to test the loop condition, I think it's better to put 'u && foo()'  as a condition. Obviously, Intel driver has bug when processing 'u && foo()' as for and while loop condition. I will report this bug to our driver team. Thanks.

zhenyao:
The description for this test case is "in for loop expression", which is accurate.

I think the original test is still valuable.

Can you just add another test case with description "in for loop condition"?

qjia7:
>Can you just add another test case with description "in for loop condition"?

Done. Thanks for your suggestion.

zhenyao:
LGTM
Thanks

PR #2623, opened 2018-03-30
Replaced invalid "enum" with "GLenum" in disjoint timer query extensi…

### LINKED ISSUE #2596: Invalid WebIDL for extensions
The IDL for some of the extensions is invalid.

For example, here: https://github.com/KhronosGroup/WebGL/blob/master/extensions/EXT_disjoint_timer_query/extension.xml#L68

The parameter type should be `GLenum`, not `enum`.

Similarly, here:
https://github.com/KhronosGroup/WebGL/blob/master/extensions/WEBGL_shared_resources/extension.xml#L343

There are missing semicolons.

And here:
https://github.com/KhronosGroup/WebGL/blob/master/extensions/WEBGL_shared_resources/extension.xml#L333

The callback has the wrong name.

kenrussell:
…ons.

Verified that these are the only places in the extensions' IDL which
have this mistake.

Fixes #2596.


kenrussell:
Please review. @zhenyao @kainino0x in particular. Thanks.

zhenyao:
Thanks for the fix. LGTM

PR #2624, opened 2018-04-03
Report skipped tests as both skipped and failed rather than just failed.

jdarpinian:
Skipped tests also show as gray instead of red.


jdashg:
The way I restructured it, it should never say "PASS" unless you actually pass every known test file. Skips aren't counted as failures, but they don't count as PASS either.

I like skipped tests showing as something other than red.

jdarpinian:
As far as I can tell, both before and after your change, skipped tests are counted as failures, and colored red. I think it's good to count them as failures, but it's useful to have a count of skipped tests too.

kenrussell:
Looks good James.

PR #2625, opened 2018-04-03
Fix usage of glEnumToString in webgl-specific-stencil-settings

kainino0x:
and add a safeguard against future glEnumToString usage mistakes

(follow-up to #2583)


kenrussell:
LGTM

PR #2626, opened 2018-04-04
Change dummySetProgramAndDrawNothing to draw >0 elements

kainino0x:
* the program now sets `gl_Position` and `gl_FragColor` to `vec4(0.0)`
* drawArrays now draws one triangle instead of 0
  (but that triangle is of zero size)

This is necessary to make sure the draw call doesn't early out before it
does validation.

(follow-up to #2583)


kenrussell:
LGTM

PR #2627, opened 2018-04-05
Use fixed number of iterations in sync-webgl-specific.

kenrussell:
Like the conformance2/query/occlusion-query test, use a fixed number
of iterations, as on slow automated test machines, a fixed duration
based on Date.now() leads to test flakiness.

Improve the synchronization with the GL server by using readPixels
instead of flush or finish.

### DIFF
--- a/sdk/tests/conformance2/sync/sync-webgl-specific.html
+++ b/sdk/tests/conformance2/sync/sync-webgl-specific.html
@@ -48,6 +48,7 @@ debug("Canvas.getContext");
 var wtu = WebGLTestUtils;
 var gl = wtu.create3DContext("canvas", null, 2);
 var sync = null;
+var pixel = new Uint8Array(4);
 
 if (!gl) {
   testFailed("context does not exist");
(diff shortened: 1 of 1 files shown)


kenrussell:
@jdashg @kainino0x @jdarpinian @zhenyao please review.

This test has been flaky on our test bots since it was introduced, though it always passes locally (on much faster hardware). See http://crbug.com/818336 .

This changes the form of the test to be more like the preexisting occlusion-query test, which is tick-based rather than time-based.

sdk/tests/conformance2/sync/sync-webgl-specific.html
```diff
@@ -91,28 +97,23 @@
...
 
-var startTimeOfFinish = 0;
 
 function finishSyncParameterTest() {
-  if (startTimeOfFinish == 0) {
-    startTimeOfFinish = Date.now();
-  }
-  if (Date.now() - startTimeOfFinish > signaling_retry_ms) {
+  if (--iteration == 0) {
     testFailed("Sync object wasn't signaled in a reasonable timeframe (" + (Date.now() - startTimeOfFinish) + " ms)");
```
zhenyao: This startTimeOfFinish is no longer defined

kenrussell: Thanks for catching this. Fixed.

kenrussell:
Merging myself to try to get our bots reliably green again.

PR #2628, opened 2018-04-05
Test combinations of clearBuffer*(), clear() and draws.

kenrussell:
With the preserveDrawingBuffer:false context creation option, Chromium
was incorrectly tracking the effects of clearBuffer on the default
back buffer.

Modified various quad setup functions in webgl-test-utils to allow the
quad to be more easily scaled.

Regression test for http://crbug.com/828262 .

### DIFF
--- a/sdk/tests/conformance2/rendering/clearbuffer-and-draw.html
+++ b/sdk/tests/conformance2/rendering/clearbuffer-and-draw.html
@@ -1,0 +1,181 @@
+<!--
+
+/*
+** Copyright (c) 2018 The Khronos Group Inc.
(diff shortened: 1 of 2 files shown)


kenrussell:
@kainino0x or @jdashg please review.

@zhenyao @jdarpinian any feedback also welcome.

Test already passes in Firefox Nightly. Passes in Chrome with fix for http://crbug.com/828262 .

zhenyao:
Looks good to me.

kenrussell:
Thanks for reviewing. Merging myself.

PR #2629, opened 2018-04-06
Add a test for using non-constant texture offset

Oletus:
The existing texture offset tests did not seem to cover this invalid
usage of the texture offset functions yet.


zhenyao:
LGTM

kenrussell:
LGTM too.

Are there already sufficient tests of constant texture offsets? It would be great to have this test cover both this case, and the case where the various "ivec2 offset" variables are const, thereby making the shaders valid.

Oletus:
Yes, there's pretty good coverage for valid texture offset function calls in dEQP. This was the one thing I didn't find preexisting tests for.

PR #2630, opened 2018-04-06
Add a test for having a case statement inside a block

Oletus:
This was not being validated correctly by ANGLE until recently, though
this would not cause issues if the native drivers were working
correctly.


zhenyao:
LGTM

kenrussell:
Nice new test. Better to catch this in ANGLE rather than rely on the drivers.

PR #2631, opened 2018-04-07
wtu.loadShader's gl.shaderSource should be treated as infallible.

jdashg:
loadShader thinks shaderSource generated an error, but really it's an error from a previous call. We can either assert before calling shaderSource that there's no current GL error, OR we can just treat shaderSource as infallible, which it really should be here.


kenrussell:
Looking through the file, seems fine, though in the other function in this file which necessarily calls gl.getError (checkTextureSize) we should probably change it so it calls gl.getError upon entry to the function.

PR #2632, opened 2018-04-07
Light amendment of http to https in the readme

etisdew:
I'm just getting ready to try digest everything I can about using webGL and thought I'd make a small contribution if its welcome.

Thank you for your work and making this available.


kenrussell:
Thanks, looks good.

PR #2633, opened 2018-04-07
Fix IDL of WEBGL_shared_resources draft extension.

### LINKED ISSUE #2596: Invalid WebIDL for extensions

kenrussell:
Checked with https://www.w3.org/2009/07/webidl-check (plugging in some
needed definitions above this extension's IDL).

Fixes #2596.

### DIFF
--- a/extensions/WEBGL_shared_resources/extension.xml
+++ b/extensions/WEBGL_shared_resources/extension.xml
@@ -330,7 +330,7 @@ interface WEBGL_shared_resources {
      long acquireSharedResource(
          WebGLSharedObject resource,
          GLenum mode,
-         AcquireResourcesCallback callback);
+         AcquireSharedResourcesCallback callback);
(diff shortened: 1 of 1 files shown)


kenrussell:
@kainino0x please review – thanks for catching the fact that I only did half the job with my earlier pull request.

PR #2635, opened 2018-04-12
Add more stencil validation tests

kainino0x:
Add new tests testing
- various transitions from valid to invalid states (in order to test invalidation of the cached validation in Chrome)
- drawing to the default framebuffer

Also modifies dummySetProgramAndDrawNothing to keep one WebGLProgram _per WebGL context_.

https://crbug.com/806557


kainino0x:
@kenrussell: PTAL (since Jeff is OOO)

kainino0x:
Updated commit message + description.

kenrussell:
Great work. I think this covers all of the cases where a cache of the stencil validation checks – if implemented by the browser – needs to be invalidated.

PR #2637, opened 2018-04-13
Add test verifying minimal quality of rendered lines.

kenrussell:
Regression test for http://crbug.com/820461 and
http://crbug.com/swiftshader/99 .


kenrussell:
Verified locally on Linux that the test passes normally, and fails with an older build of Chrome and --use-gl=swiftshader.

kenrussell:
@kainino0x @jdashg @RafaelCintron @grorg please take a look.

PR #2639, opened 2018-04-19
Limit blitFramebuffer source and destination rectangle size

Oletus:
BlitFramebuffer bugs that trigger when using large rectangle
dimensions seem to be widespread on all desktop OpenGL drivers. They
have been seen on at least:

AMD Windows OpenGL
Intel Windows OpenGL
Intel Linux OpenGL
NVIDIA OpenGL (all platforms)

In addition there are some issues with handling large dimensions in
existing blitFramebuffer workarounds inside ANGLE.

In some cases these issues can be worked around by clipping the
blitFramebuffer rectangles so that their width and height fit a signed
32-bit integer. However, this approach doesn't work in case
blitFramebuffer is used to scale the framebuffer and the greatest
common divisor of the source and destination dimensions is 1.

Since these issues likely don't have any impact on real-world
applications, it's reasonable to limit the dimensions in the spec so
that rectangles taller or wider than the maximum GLint value are not
allowed.


kainino0x:
The new tests LGTM.

However it seems like we could keep the old tests too - just for slightly smaller rectangles that are exactly 2^31 in size.

Oletus:
Added a few more simple tests. We could also add more tests that check the blitFramebuffer results from the destination framebuffer, but I'd rather do that separately.

kenrussell:
LGTM after the fact.

PR #826, opened 2015-01-24
rejecting WEBGL_shared_resources

pyalot:
proposing to reject WEBGL_shared_resources

PR #2642, opened 2018-05-03
Reject WEBGL_shared_resources.

kenrussell:
At this point in the WebGL API's development it is not profitable to
invest the time to implement this extension. There are other higher
priorities, and all browsers are committed to making it possible to
draw WebGL-rendered canvases into other canvases quickly, which
essentially solves this problem.

Provides a bit more description than #826.

Removes test case added in #300.


kenrussell:
@jdashg @grorg @RafaelCintron please take a look.

@kainino0x @jdarpinian Comments welcome. Would like one of the other WebGL implementers to merge this.

greggman:
I 100% agree this should be rejected. 

That said, where can I read more about this? I haven't seen the proposed solution that makes this possible. I've commented else where and what I think the proposed solutions are and why they are not quick. I maybe have missed the solution or mis-understood it but in any case I'd love to know what the proposed solution is and where it's being discussed. Thanks!

> and all browsers are committed to making it possible to draw WebGL-rendered canvases into other canvases quickly

kenrussell:
The one-copy solution is to allocate one or more canvases using the 2D rendering context, and then draw a WebGL-rendered canvas into them via CanvasRenderingContext2D.drawImage. When this draw operation is GPU-accelerated, it works well. This code path is already GPU accelerated at least in Chrome, and other browsers are prioritizing accelerating it.

The zero-copy solution uses multiple canvases each of which have the ImageBitmapRenderingContext allocated:
https://html.spec.whatwg.org/multipage/canvas.html#the-imagebitmap-rendering-context

An OffscreenCanvas is allocated and a WebGL rendering context created for it. Each frame, OffscreenCanvas.transferToImageBitmap is used to get a handle to the rendered buffer:
https://html.spec.whatwg.org/multipage/canvas.html#the-offscreencanvas-interface

and it is drawn into one of the ImageBitmapRenderingContexts using transferFromImageBitmap.

The zero-copy path is implemented in Chrome behind a flag and we hope that it will ship soon, including in other browsers. CC'ing @fserb for any further comments.

extensions/rejected/WEBGL_shared_resources/extension.xml
```diff
@@ -370,5 +370,10 @@ interface WebGLTexture : WebGLSharedObject {
     <revision date="2014/07/15">
       <change>Added NoInterfaceObject extended attribute to extension interface and WebGLShareGroup.</change>
     </revision>
+    <revision date="2018/05/02">
+      <change>Rejected after discussion on public_webgl and no strong objections. At this point in
+      the WebGL API's development it is not profitable to invest the time to implement this
+      extension.</change>
```
kainino0x: It may be worth mentioning the alternative(s) here.

kenrussell: Thanks for the suggestion. Described the alternatives.

jdarpinian:
LGTM

greggman:
The problem with the OffscreenCanvas solution is that AFAICT there is no way to use it preformantly with 2 or more canvases of different sizes.

I imagine it like this. With 1 OffscreenScreen canvas and 1 ImageBitmapRenderingContext there are 2 drawingbuffers of the same size. The one the WebGL context is rendering into and the one the ImageBitmapRenderingContext is displaying in the canvas. When you call `transferFromImageBitmap` those 2 buffers get swapped. It's fast an efficent

When you have 2 canvases that no longer works because each canvas has a different sized drawingBuffer so you can no longer swap the drawingbuffers and everytime you want to draw to a different canvas you have to change the width and height of the context to match. Changing the width and height of a context allocates a new drawingbuffer which is a slow operation.

Is there a way to avoid that slow allocation or am I mis-understanding the how the OffscreenCanvas solution can stay performance for multiple canvases?

jdashg:
Firefox lazily allocates backbuffers, so everything should work as expected. The only thing not handled would be recycling of differently-sized surfaces, but so far, the allocations this would cause aren't problematic.

jdashg:
We commit to satisfying drawing to multiple canvases, but this extension is not the path forward there.

PR #2643, opened 2018-05-07
Add a test for declaring both gl_FragColor and gl_FragData invariant

Oletus:
Declaring a variable invariant doesn't access it, so it should not
really count as using the variable. Earlier versions of ANGLE would
treat declaring both gl_FragColor and gl_FragData invariant an error,
but recent versions have been fixed to allow this usage.

### DIFF
--- a/sdk/tests/conformance/glsl/misc/00_test_list.txt
+++ b/sdk/tests/conformance/glsl/misc/00_test_list.txt
@@ -5,7 +5,7 @@ embedded-struct-definitions-forbidden.html
 --min-version 1.0.4 empty-declaration.html
 empty_main.vert.html
 --min-version 1.0.3 expression-list-in-declarator-initializer.html
-fragcolor-fragdata-invariant.html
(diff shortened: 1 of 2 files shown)


zhenyao:
Looks good to me.

sdk/tests/conformance/glsl/misc/00_test_list.txt
```diff
@@ -5,6 +5,7 @@ embedded-struct-definitions-forbidden.html
 --min-version 1.0.4 empty-declaration.html
 empty_main.vert.html
 --min-version 1.0.3 expression-list-in-declarator-initializer.html
+fragcolor-fragdata-invariant.html
```
kenrussell: Could you please add --min-version 1.0.4 for bookkeeping purposes?

Oletus: Of course, done.

PR #2646, opened 2018-05-10
 Add a test for texImage3D from ImageData with unpack params

kainino0x:
Regression test for https://crbug.com/804123

This is a pretty minimal test covering the case we saw - it certainly could be expanded.


kenrussell:
Very nice. A small targeted test sounds good.

PR #2647, opened 2018-05-15
Test texStorage[2,3]D with non-square sizes.

jdashg:
Regression test from Firefox where e.g. level=9, width=512, height=1
fails.

https://bugzilla.mozilla.org/show_bug.cgi?id=1461293


kenrussell:
Nice and simple. LGTM

jdashg:
Oops, cubemaps have to be square. Will fix tomorrow.

kenrussell:
Oops. Let me revert this so the tests on top-of-tree are correct and let's resubmit it tomorrow with the fix.

PR #2649, opened 2018-05-16
Add correctness tests to tex-unpack-params-imagedata.html

kainino0x:
Follow-up to https://github.com/KhronosGroup/WebGL/pull/2646

### DIFF
--- a/sdk/tests/conformance2/textures/misc/tex-unpack-params-imagedata.html
+++ b/sdk/tests/conformance2/textures/misc/tex-unpack-params-imagedata.html
@@ -40,14 +40,14 @@
 <script>
 var wtu = WebGLTestUtils;
 var gl = wtu.create3DContext(undefined, undefined, 2);
-var actual;
+let actual;
 
 description("TexImage3D from ImageData with unpack params");
 
 debug("TexImage3D from ImageData with UNPACK_IMAGE_HEIGHT set (crbug.com/804123)");
 
 // framebuffer for readback
-var fbo = gl.createFramebuffer();
+const fbo = gl.createFramebuffer();
 gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
 
 function makeTestData(size, start) {
--- a/sdk/tests/conformance2/textures/misc/tex-unpack-params-imagedata.html
+++ b/sdk/tests/conformance2/textures/misc/tex-unpack-params-imagedata.html
@@ -40,22 +40,103 @@
 <script>
 var wtu = WebGLTestUtils;
 var gl = wtu.create3DContext(undefined, undefined, 2);
+var actual;
 
 description("TexImage3D from ImageData with unpack params");
 
 debug("TexImage3D from ImageData with UNPACK_IMAGE_HEIGHT set (crbug.com/804123)");
 
-let imageData = new ImageData(64, 64);
-let texture = gl.createTexture();
+// framebuffer for readback
+var fbo = gl.createFramebuffer();
+gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+
+function makeTestData(size, start) {
+  const data = new Uint8ClampedArray(size);
+  for (let i = 0; i < size; ++i) {
+    data[i] = (start + i) % 256;
+  }
+  return data;
+}
+
+// source data
+//
+//   dstWidth = 4
+// <-->
+// xxxx ^ ^ ^ ^
+// xxxx |   | |
+// xxxx |   | |
+// xxxx |   | v dstHeight = 4
+// ---- |   |
+// ---- |   |
+// ---- |   v unpackImageHeight = 7
+// xxxx | |
+// xxxx |
+// xxxx |
+// xxxx |
+// ---- |
+// ---- |
+// ---- |
+// xxxx | |
+// xxxx |
+// xxxx |
+// xxxx |
+// ---- |
+// ---- |
+// ---- |
+// xxxx | v dstDepth = 4
+// xxxx |
+// xxxx |
+// xxxx v srcHeight = 25
+// <-->
+//   srcWidth = 4
+const unpackImageHeight = 7;
+const dstWidth = 4;
+const dstHeight = 4;
+const dstDepth = 4;
+const srcWidth = dstWidth;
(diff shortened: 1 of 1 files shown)


kainino0x:
PTAL, @jdashg in particular, because:
It seems that Firefox is not respecting UNPACK_IMAGE_HEIGHT when uploading ImageData. (Once you pass `tex-unpack-params.html` you may pass this as well.)

There was one other issued I saw, where the result of `readPixels` seemed to depend on UNPACK_IMAGE_HEIGHT in Firefox. However I haven't been able to reproduce it after an "edit code and also happen to update Firefox" cycle, so I don't know whether I lost the repro case or it was fixed by the Firefox update. Just a heads up.

kenrussell:
The new test looks good to me, but deferring review to Jeff.

jdashg:
Will look tomorrow.

sdk/tests/conformance2/textures/misc/tex-unpack-params-imagedata.html
```diff
@@ -40,24 +40,77 @@
...
 
 debug("TexImage3D from ImageData with UNPACK_IMAGE_HEIGHT set (crbug.com/804123)");
 
-let imageData = new ImageData(64, 64);
+// framebuffer for readback
+var fbo = gl.createFramebuffer();
+gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+
+function makeTestData(size, start) {
+  let data = new Uint8ClampedArray(size);
```
jdashg: You should be able to use const here.

sdk/tests/conformance2/textures/misc/tex-unpack-params-imagedata.html
```diff
@@ -40,24 +40,77 @@
...
+for (let z = 0; z < depth; ++z) {
+  gl.framebufferTextureLayer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, texture, 0, z);
+  if (gl.checkFramebufferStatus(gl.FRAMEBUFFER) == gl.FRAMEBUFFER_COMPLETE) {
+    gl.readPixels(0, 0, width, height, gl.RGBA, gl.UNSIGNED_BYTE, actual);
+    debug(`for z = ${z}`);
+    shouldBeTrue(`areArraysEqual(actual,
+      makeTestData(/*  size = */ width * height * sizeofRGBA8,
+                   /* start = */ 1 + ${z} * width * unpack_image_height * sizeofRGBA8))`);
+  } else {
+    debug("framebuffer incomplete: skipped");
```
jdashg: I believe we require RGBA8 to be renderable, so you can just assume (or assert) this.

sdk/tests/conformance2/textures/misc/tex-unpack-params-imagedata.html
```diff
@@ -40,24 +40,77 @@
...
+  }
+}
+wtu.glErrorShouldBe(gl, gl.NO_ERROR);
 
 gl.deleteTexture(texture);
+gl.deleteFramebuffer(fbo);
```
jdashg: That's nice of you, but you don't need to. :)

sdk/tests/conformance2/textures/misc/tex-unpack-params-imagedata.html
```diff
@@ -40,24 +40,77 @@
...
+  return data;
+}
+
+// source data
+const unpack_image_height = 7;
+const width = 4;
+const height = 4;
+const depth = 4;
+const srcWidth = width;
+const srcHeight = (height - 1) * unpack_image_height + height;
```
jdashg: Can you leave a comment about the shape of the thing you're trying to create? This equation is hard to verify.

jdashg: I think you want (depth-1)*unpack_height+height.

sdk/tests/conformance2/textures/misc/tex-unpack-params-imagedata.html
```diff
@@ -40,24 +40,77 @@
...
 wtu.glErrorShouldBe(gl, gl.NO_ERROR, "larger upload");
+// check
+actual = new Uint8Array(width * height * sizeofRGBA8);
+for (let z = 0; z < depth; ++z) {
+  gl.framebufferTextureLayer(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, texture, 0, z);
+  if (gl.checkFramebufferStatus(gl.FRAMEBUFFER) == gl.FRAMEBUFFER_COMPLETE) {
+    gl.readPixels(0, 0, width, height, gl.RGBA, gl.UNSIGNED_BYTE, actual);
+    debug(`for z = ${z}`);
+    shouldBeTrue(`areArraysEqual(actual,
+      makeTestData(/*  size = */ width * height * sizeofRGBA8,
```
jdashg: Don't embed makeTestData in the eval.

sdk/tests/conformance2/textures/misc/tex-unpack-params-imagedata.html
kainino0x: I think you're right, thanks. I'll comment it.

kainino0x:
Ok, done.

On Firefox 61.0b8 I'm seeing it fail on Mac/Intel but pass on Linux/NVIDIA. So apparently it is Mac or Mac/Intel specific.

sdk/tests/conformance2/textures/misc/tex-unpack-params-imagedata.html
```diff
@@ -40,22 +40,103 @@
...
+var actual;
 
 description("TexImage3D from ImageData with unpack params");
 
 debug("TexImage3D from ImageData with UNPACK_IMAGE_HEIGHT set (crbug.com/804123)");
 
-let imageData = new ImageData(64, 64);
-let texture = gl.createTexture();
+// framebuffer for readback
+var fbo = gl.createFramebuffer();
```
dmikis: I'd say it sould be `const` if we're using ES6 (and it seems we're).

kenrussell:
What's the status of this pull request? Can it be moved forward?

PR #2650, opened 2018-05-16
Add tests for blitting to cover width/height overflow.

Richard-Yunchao:
The range of an integer is [-0x7fffffff - 1, 0x7fffffff].
We restrict the width/height to the max value that can be
store in an integer in latest WebGL 2 spec. If we
reverse the data order for blitting, an invalid value like
0x7fffffff + 1 would become a valid value for an integer:
-0x7fffffff - 1. But it is invalid for blitting.

The new added tests cover this corner case to make it clear.

### DIFF
--- a/sdk/tests/conformance2/rendering/blitframebuffer-size-overflow.html
+++ b/sdk/tests/conformance2/rendering/blitframebuffer-size-overflow.html
@@ -91,7 +91,11 @@ function blit_region_test() {
 
     gl.blitFramebuffer(-1, -1, max, max, 0, 0, width, height, gl.COLOR_BUFFER_BIT, gl.NEAREST);
     wtu.glErrorShouldBe(gl, gl.INVALID_VALUE, "Using source width/height greater than max 32-bit integer should fail.");
+    gl.blitFramebuffer(max, max, -1, -1, 0, 0, width, height, gl.COLOR_BUFFER_BIT, gl.NEAREST);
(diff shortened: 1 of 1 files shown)


jdashg:
I thought we forbade out-of-range blits?

kainino0x:
For the record: this change follows on to https://github.com/KhronosGroup/WebGL/pull/2639

jdashg:
I wish we would just forbid out-of-range blits, but ok.

sdk/tests/conformance2/rendering/blitframebuffer-size-overflow.html
```diff
@@ -91,8 +91,12 @@
 
     gl.blitFramebuffer(-1, -1, max, max, 0, 0, width, height, gl.COLOR_BUFFER_BIT, gl.NEAREST);
     wtu.glErrorShouldBe(gl, gl.INVALID_VALUE, "Using source width/height greater than max 32-bit integer should fail.");
+    gl.blitFramebuffer(max, max, -1, -1, 0, 0, width, height, gl.COLOR_BUFFER_BIT, gl.NEAREST);
+    wtu.glErrorShouldBe(gl, gl.INVALID_VALUE, "Using source width/height greater than max 32-bit integer should fail.");
```
jdashg: I would prefer "greater than INT32_MAX".

Richard-Yunchao:
Thanks for your review. Jeff and Kai.

I saw failures about these tests on Intel platforms after you changed the spec about blitting src/dst rect size at #2639 , so I revisited the tests and added a couple of tests here in order to make it clear.

PR #2652, opened 2018-05-18
Add WebGL workload simulator

jdarpinian:
A tool that provides a UI to select various kinds of WebGL work to perform each frame. Useful to test the effect of various different workloads on performance.

Try it out here: https://rawcdn.githack.com/jdarpinian/WebGL/a1a585e2227f5c17a1c05cc906564b03cd04d679/sdk/tests/extra/workload-simulator.html?continuous


sdk/tests/extra/workload-simulator.html
```diff
@@ -0,0 +1,604 @@
...
+  let scissorEnabled = false;
+  for (let i = 1; i < numDrawCalls; i++) {
+    if (instancesDrawn > instances && !scissorEnabled) {
+      // If we've drawn all of the requested pixels already, enable the scissor
+      // test so we only draw one pixel per draw call for the rest of the calls.
+      scissorEnabled = true;
+      gl.enable(gl.SCISSOR_TEST);
+    }
+    gl.drawArraysInstanced(gl.TRIANGLE_STRIP, 0, 4, instancesPerCall);
+    instancesDrawn += instancesPerCall;
```
kenrussell: Have you observed on some platforms that the additional pixels drawn are having an effect? On my MacBook Pro they seem to be being short-circuited by the driver or something similar, and they don't cause a slowdown.

greggman:
hrm, I set it draw 50m pixels, picked postmessage, watched my Mac reboot 😓

sdk/tests/extra/workload-simulator.html
jdarpinian: I think the machine is just too fast. Try "Multiply the above slider values by: 10"

I put the default maximum values pretty low for testing on low-end Android phones. Also because on some platforms it's pretty easy to hose your machine, as Gregg just discovered :neutral_face:

PR #2653, opened 2018-05-31
Readback correctness tests (for nonblocking readback)

kainino0x:
* Remove WEBGL_get_buffer_sub_data_async.
* Add tests for getBufferSubData correctness and performance.

With the new "nonblocking getBufferSubData" optimizations, getbufferSubData needs these additional correctness tests to make sure it doesn't return data that is too old or too new.

The performance tests are in extra/. They may not pass consistently.

All of the tests pass in Firefox (including the performance tests, though of course by a smaller margin.)

PR #2655, opened 2018-06-05
Remove getBufferSubDataAsync tests from 00_test_list.txt

kainino0x:
See #2653.

I missed these in the previous PR.


kenrussell:
LGTM after the fact.

PR #2656, opened 2018-06-07
WebKit (now) supports requestAnimationFrame

JusSn:
LayoutTests have supported window.requestAnimationFrame for a while. I am removing this WebKit-specific code so that window.requestAnimationFrame calls will not fail on WebKit's test framework.


grorg:
Looks good. I can't imagine any issues on other browsers either.

PR #2657, opened 2018-06-12
Test constructing a struct with a single member

Oletus:
This is a regression test for a Swiftshader bug.


kenrussell:
Looks great. Sorry for the delay reviewing.

PR #2658, opened 2018-06-19
Specify binding one buffer to two transform feedback outputs

Oletus:
This case used to be underspecified, so behavior varies between
platforms. Particularly on D3D having two stream outputs written to
the same buffer is disallowed, and on OpenGL ES there's some undefined
behavior. Prohibit having the same buffer bound to two transform
feedback outputs when calling BeginTransformFeeback to avoid
non-portable behaviors from now on.

Currently the test doesn't yet generate the expected error on any of
the platforms tested. On Chromium or Firefox running on ANGLE's D3D
backend, the buffer is not touched by transform feedback.
On Chromium's GL backend, only one of the outputs is written to the
buffer by transform feedback.

The behavior is the same on recent HW/drivers from both NVIDIA and
Intel.

### DIFF
--- a/sdk/tests/conformance2/transform_feedback/same-buffer-two-binding-points.html
+++ b/sdk/tests/conformance2/transform_feedback/same-buffer-two-binding-points.html
@@ -1,0 +1,131 @@
+<!--
+
+/*
+** Copyright (c) 2018 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL Transform Feedback Conformance Tests - one buffer bound to two binding points</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<script id="vshader" type="x-shader/x-vertex">#version 300 es
+
+in vec4 in_data;
+out vec4 out_add;
+out vec4 out_mul;
+void main(void) {
+    out_add = in_data + vec4(2.0, 3.0, 4.0, 5.0);
+    out_mul = in_data * vec4(2.0, 3.0, 4.0, 5.0);
+}
+</script>
+<script>
+"use strict";
+description();
+
+debug("");
+
+var wtu = WebGLTestUtils;
+var gl = wtu.create3DContext(null, null, 2);
+var program = null;
+
+if (!gl) {
+    testFailed("WebGL context does not exist");
+} else {
+    testPassed("WebGL context exists");
+    runTwoOutFeedbackTest();
+}
+
+function runTwoOutFeedbackTest() {
+    debug("");
+    debug("Test binding the same buffer to two transform feedback binding points. Buffer should be untouched and an error should be generated.")
+
+    // Build the input and output buffers
+    var in_data = [
+        1.0, 2.0, 3.0, 4.0,
+        2.0, 4.0, 8.0, 16.0,
+        0.75, 0.5, 0.25, 0.0
+    ];
+
+    var in_buffer = gl.createBuffer();
+    gl.bindBuffer(gl.ARRAY_BUFFER, in_buffer);
+    gl.bufferData(gl.ARRAY_BUFFER, new Float32Array(in_data), gl.STATIC_DRAW);
+
+    // Create the transform feedback shader
+    program = wtu.setupTransformFeedbackProgram(gl, ["vshader", wtu.simpleColorFragmentShaderESSL300],
+        ["out_add", "out_mul"], gl.SEPARATE_ATTRIBS,
+        ["in_data"]);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "linking transform feedback shader should not set an error");
+    shouldBeNonNull("program");
+
+    // Draw the the transform feedback buffers
+    var tf = gl.createTransformFeedback();
+
+    gl.enableVertexAttribArray(0);
+    gl.bindBuffer(gl.ARRAY_BUFFER, in_buffer);
+    gl.vertexAttribPointer(0, 4, gl.FLOAT, false, 16, 0);
+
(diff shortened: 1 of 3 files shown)


Oletus:
I also opened a bug against the native GL specs about this and proposed the same error there.

specs/latest/2.0/index.html
```diff
@@ -4075,6 +4075,23 @@ <h3>blitFramebuffer rectangle width/height limitations</h3>
         blitFramebuffer is used to scale the framebuffer.
     </div>
 
+    <h3>Transform feedback buffer binding limitations</h3>
```
jdarpinian: There is already a section farther up about transform feedback buffer binding restrictions. I'd suggest moving this section next to the other one, or merging the sections together.

Oletus: There might be references to section numbers floating around, so it's better to add new sections to the end of the document. But it seems reasonable to merge the sections, I added another commit to this pull request that does just that.

jdashg:
Looking at ES3, I don't see this as being underspecified. It looks like just bad drivers. If everyone's wrong the same way, we should consider standardizing on it though. It's too bad: Partially-separate attribs seems like a real usecase. We might be able to efficiently polyfill this if we want, though we would have to guess at the temporary buffer size. Let's wait to hear what the GL WG has to say.

Running the test locally on Mac+NV, it seems to only write TF[1] if TF[0] and TF[1] are the same buffer.

specs/latest/2.0/index.html
```diff
@@ -3443,6 +3443,21 @@ <h3>Preventing undefined behavior with Transform Feedback</h3>
         OpenGL ES 3.0.5 &sect;2.15.2</a>)</span>
     </div>
 
+    <p>
+        In case the same buffer is bound to more than one transform feedback binding point that will
+        be written to, <code>BeginTransformFeedback</code> generates an
```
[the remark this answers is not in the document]
Oletus: I did some more experiments, it seems like ANGLE's DX backend fails to write the outputs correctly even when using two completely disjoint regions of the same output buffer for different varyings. On NVIDIA's GL driver it works. Don't know if it is something fundamental in DX or if it's an ANGLE bug. But it would be easy enough to amend the spec so that binding disjoint regions with bindBufferRange are allowed.

Oletus:
I pushed one more commit to this pull request that allows disjoint regions. It's a bit unclear whether we can actually do so though - it might be hitting either some bug in ANGLE's D3D backend or a fundamental limitation in D3D that would be hard to work around.

specs/latest/2.0/index.html
```diff
@@ -3443,6 +3443,22 @@ <h3>Preventing undefined behavior with Transform Feedback</h3>
...
+        in the buffer chosen by <code>bindBufferRange</code> overlap, <code>BeginTransformFeedback</code>
+        generates an <code>INVALID_OPERATION</code> error.
+    </p>
+
+    <div class="note">
+        This can happen only in <code>SEPARATE_ATTRIBS</code> mode.
+    </div>
+
+    <div class="note rationale">
+        Behavior when the same buffer is bound to two outputs is not specified in the OpenGL ES spec
```
kenrussell: Instead of "same buffer" should we write "same region of the same buffer"?

kenrussell:
Thanks @Oletus for updating this. Studying some general D3D documentation on stream output:
https://msdn.microsoft.com/en-us/library/windows/desktop/bb205122(v=vs.85).aspx

the specific docs for ID3D11DeviceContext::SOSetTargets:
https://msdn.microsoft.com/en-us/library/windows/desktop/ff476484(v=vs.85).aspx

the documentation of ID3D11Buffer, which states "A buffer can also be bound to a single pipeline stage for writing; however, the same buffer cannot be bound for reading and writing simultaneously":
https://msdn.microsoft.com/en-us/library/windows/desktop/ff476351(v=vs.85).aspx

and ANGLE's computation of the buffers and offsets to pass to this method:
https://cs.chromium.org/chromium/src/third_party/angle/src/libANGLE/renderer/d3d/d3d11/StateManager11.cpp?type=cs&q=sosettargets&sq=package:chromium&g=0&l=3337
https://cs.chromium.org/chromium/src/third_party/angle/src/libANGLE/renderer/d3d/d3d11/TransformFeedback11.cpp?type=cs&q=mBufferOffsets&sq=package:chromium&g=0&l=42

it looks like ANGLE's doing everything correctly.

@RafaelCintron could you please help here? Could you find out whether it's legal to use the same ID3D11Buffer with different offsets as multiple different stream output targets? Thanks.

jdashg:
My testcase:
https://github.com/jdashg/misc/commit/0eb237787778687dd9aec308d600f5289fdf5a56

kenrussell:
@jdashg good test case.

@null77 @vonture any opinions on handling this automatically inside ANGLE and perhaps issuing a performance warning, vs. forbidding it? I'm slightly concerned about compatibility with existing OpenGL ES code bases if WebGL forbids it, but also think transform feedback is probably not that widely used.

jdashg:
We can always ban it for now an renege on it later.

vonture:
I would prefer to ban it.  It is surprising that dEQP has no test of this (or we're suppressing it and I didn't notice).

It's not trivial in ANGLE's internals to create shadow buffers and then trigger a copy later.

sdk/tests/conformance2/transform_feedback/same-buffer-two-binding-points.html
```diff
@@ -0,0 +1,179 @@
...
+    wtu.checkFloatBuffer(gl, gl.TRANSFORM_FEEDBACK_BUFFER, expectedZeroes);
+
+    // Test non-overlapping ranges. This should work.
+    debug("");
+    gl.enable(gl.RASTERIZER_DISCARD);
+    gl.bufferData(gl.TRANSFORM_FEEDBACK_BUFFER, Float32Array.BYTES_PER_ELEMENT * in_data.length * 2, gl.STATIC_DRAW);
+    gl.bindBufferRange(gl.TRANSFORM_FEEDBACK_BUFFER, 0, out_buffer, 0, in_data.length * 4);
+    gl.bindBufferRange(gl.TRANSFORM_FEEDBACK_BUFFER, 1, out_buffer, in_data.length * 4, in_data.length * 4);
+
+    doDrawWithTransformFeedback(gl.NO_ERROR, "same buffer bound to two transform feedback varyings with bindBufferRange, non-overlapping ranges");
```
jdashg: Let's forbid this.

specs/latest/2.0/index.html
```diff
@@ -3443,6 +3443,22 @@ <h3>Preventing undefined behavior with Transform Feedback</h3>
         OpenGL ES 3.0.5 &sect;2.15.2</a>)</span>
     </div>
 
+    <p>
+        In case the same buffer is bound to more than one transform feedback binding point that will
```
jdashg: "active transform feedback binding index, generate INVALID_OPERATION."

specs/latest/2.0/index.html
jdashg: Change the rationale to:
This is a restriction from D3D11, where this behavior is not allowed.

RafaelCintron:
Apologies for not responding sooner.  I was out of office on vacation last week.

As @jdashg discovered by using the Direct3D SDK layers, data from multiple streams cannot go to a single buffer.  I confirmed this with my contacts on the Direct3D team.

I agree with @vonture to ban this use case for WebGL. Creating shadow buffers and making buffer copies can have a high performance penalty (both processing and memory) for large data sets.  In the past, we've gotten into trouble with features that have "surprise" emulation costs on certain platforms.

I would prefer that developers structure their buffer streams to avoid these costs on all platforms.

sdk/tests/conformance2/transform_feedback/same-buffer-two-binding-points.html
Oletus: Done.

specs/latest/2.0/index.html
Oletus: Used a slightly different wording, but I believe it captures the intent unambiguously.

Oletus:
I got back from a short  vacation and addressed the review comments.

Oletus:
For some reason GitHub doesn't let me reply to one of the review comments, but all comments have been addressed. Can we get this merged? Khronos also seems to be leaning on making the reads/writes undefined in this case in native GL specs.

specs/latest/2.0/index.html
```diff
@@ -3443,6 +3443,20 @@ <h3>Preventing undefined behavior with Transform Feedback</h3>
         OpenGL ES 3.0.5 &sect;2.15.2</a>)</span>
     </div>
 
+    <p>
+        In case the same buffer is bound to more than one binding point in the active transform feedback,
+        <code>BeginTransformFeedback</code> generates an <code>INVALID_OPERATION</code> error.
```
jdashg: beginTransformFeedback starts lowercase.

jdashg:
Thanks for pushing this through.

specs/latest/2.0/index.html
Oletus: Fixed.

PR #2659, opened 2018-06-23
Transform feedback generic binding point should not block buffer usage

jdarpinian:
The transform feedback generic binding point is not part of the transform feedback object and is not used for transform feedback. Only the indexed binding points are used. A buffer that is bound to the generic binding point should be usable for either transform feedback or non-transform-feedback purposes.

I realized this was a problem in [crbug.com/853978](http://crbug.com/853978). The previous behavior was confusing because unbinding the transform feedback object was not enough to enable drawing from a transform feedback buffer if it was still bound to the generic binding point.

### DIFF
--- a/sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
+++ b/sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
@@ -59,7 +59,7 @@ void main() {
 </script>
 <script>
 "use strict";
-description("This test verifies that access to a buffer simultaneously bound to TRANSFORM_FEEDBACK_BUFFER and another binding point is forbidden.");
+description("This test verifies that access to a buffer simultaneously bound to a transform feedback object and a non-transform-feedback binding point is forbidden.");
 
 debug("");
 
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
```diff
@@ -242,50 +266,82 @@
...
-wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "copyBufferSubData with double bound buffer");
+gl.bindBuffer(gl.COPY_WRITE_BUFFER, null);
+
+// The value of the TRANSFORM_FEEDBACK_BUFFER generic bind point should not
+// affect the legality of any operation.
+let genericBindPointValues = [()=>null, ()=>tfBuffer, ()=>vertexBuffer];
+
+for (let genericBindPointValue of genericBindPointValues) {
+  gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, tf);
+  debug("<h3>With TRANSFORM_FEEDBACK_BUFFER generic bind point value " + genericBindPointValue + "</h3>");
```
kenrussell: Nice trick to print the arrow functions. :D

PR #2661, opened 2018-06-25
Add wtu.dispatchTask for use instead of waitForComposite.

jdashg:
Texture upload tests in particular take a long time because they run a max of
60 tests per second, because they rely on waitForComposite to yield execution.
waitForComposite should only be used when necessary. Use dispatchTask if we
want to yield to the event loop.

Before:
all/conformance/textures/canvas (Passed: 3880/3880 in 48.68 seconds)
conformance/textures/canvas/tex-2d-rgb-rgb-unsigned_byte.html (Passed: 1122/1122 in 14113.0 ms)

After:
all/conformance/textures/canvas (Passed: 3880/3880 in 8.90 seconds)
conformance/textures/canvas/tex-2d-rgb-rgb-unsigned_byte.html (Passed: 1122/1122 in 2097.0 ms)

### DIFF
--- a/sdk/tests/js/webgl-test-utils.js
+++ b/sdk/tests/js/webgl-test-utils.js
@@ -2865,6 +2865,35 @@ var waitForComposite = function(callback) {
   countDown();
 };
 
+var setZeroTimeout = (function() {
+  // See https://dbaron.org/log/20100309-faster-timeouts
+
+  var timeouts = [];
+  var messageName = "zero-timeout-message";
+
+  // Like setTimeout, but only takes a function argument.  There's
+  // no time argument (always zero) and no arguments (you have to
+  // use a closure).
+  function setZeroTimeout(fn) {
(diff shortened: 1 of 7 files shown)


sdk/tests/js/webgl-test-utils.js
```diff
@@ -2865,6 +2865,35 @@ var waitForComposite = function(callback) {
...
+
+  var timeouts = [];
+  var messageName = "zero-timeout-message";
+
+  // Like setTimeout, but only takes a function argument.  There's
+  // no time argument (always zero) and no arguments (you have to
+  // use a closure).
+  function setZeroTimeout(fn) {
+      timeouts.push(fn);
+      window.postMessage(messageName, "*");
```
kenrussell: Waiting with bated breath for the inevitable fallout from using this trick in the conformance suite. :D

PR #2662, opened 2018-06-28
Fix WebGL 2.0 spec about non-zero divisor requirement

Oletus:
The conformance tests were previously fixed to remove the requirement
that there's at least one non-zero divisor, as it is not needed in
D3D11-backed implementations of WebGL. This change brings the spec in
line with the tests.

The test is also expanded a bit to cover drawArrays and drawElements,
not just their instanced variants.

Currently Chromium passes the test, but Firefox follows the old 2.0.0
spec rather than the newer version of the test.


kenrussell:
Mo's out this week – could you edit the pull request's description to link either to a bug or to the previous pull request in this area? Thanks.

jdashg:
This was #2419.

specs/latest/2.0/index.html
```diff
@@ -3508,8 +3508,8 @@ <h3><a name="ENABLED_ATTRIBUTE">Enabled Attribute</a></h3>
     <p>
```
jdashg: Remove this whole p and div. It's totally valid to draw without any enabled arrays about to active attribs.

Oletus: Updated the pull request to remove the whole section and also added a new test covering this. The test passes in Chrome and Firefox on Windows/NVIDIA at least.

PR #2663, opened 2018-06-30
Added test cases for structs as out and inout parameters.

kenrussell:
The highp variant of these tests fails on certain GPUs.

Regression tests for the Chromium bugs http://crbug.com/851870 and
http://crbug.com/851873.

Test cases originally from
https://github.com/mrdoob/three.js/issues/14137 .


kenrussell:
Verified that these tests fail on one of the affected phones, and pass on a different one as well as on desktop.

Please review. @Oletus @jdashg @kainino0x @RafaelCintron @grorg in particular. Thanks.

PR #2665, opened 2018-07-03
Add a regression test for crbug.com/859400

zhenyao:
@kainino0x PTAL

### DIFF
--- a/sdk/tests/conformance2/textures/misc/tex-subimage3d-canvas-bug.html
+++ b/sdk/tests/conformance2/textures/misc/tex-subimage3d-canvas-bug.html
@@ -1,0 +1,77 @@
+<!--
+
+/*
+** Copyright (c) 2018 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/textures/misc/tex-subimage3d-canvas-bug.html
```diff
@@ -0,0 +1,75 @@
...
+
+var wtu = WebGLTestUtils;
+var gl = wtu.create3DContext("example", undefined, 2);
+
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "Should be no errors from setup.");
+
+function runTest() {
+  var ctx = document.createElement('canvas').getContext("2d");
+  var maxTexSize = gl.getParameter(gl.MAX_TEXTURE_SIZE);
+  var width = Math.ceil(Math.sqrt(maxTexSize));
```
kainino0x: LGTM with extra comment explaining this math / why this is the right test case

zhenyao: Done

kenrussell:
LGTM too though I wonder whether it's guaranteed in all browsers' 2D canvas implementations that resizing the canvas that tall will work.

PR #2669, opened 2018-07-17
Add a basic WEBGL_multiview test for WebGL 2

Oletus:
The test tests multiview framebuffers based on 2D array textures.

The test covers basic framebuffer queries as well as rendering into
a multiview framebuffer. The test has been tested against an unmerged
prototype implementation for Chromium.

Separate tests could still be added for:
1. drawing into a multiview framebuffer with a non-multiview shader
(with num_views both 1 and > 1).
2. Interaction with timer queries.
3. Instanced draw to a multiview framebuffer.
4. Multiple color render targets in a multiview framebuffer.


Oletus:
Could I get a review on this? I'm planning to add more tests for multiview as a follow-up to this one, but this already has quite a bit of code to review.

sdk/tests/js/tests/webgl_multiview_util.js
```diff
@@ -0,0 +1,109 @@
...
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+** CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+** TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+** MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
+*/
+
+"use strict";
+
+function setupTexture(target)
```
toji: This function, with it's very generic name, seems out of place in a multiview utils file. I suggest either moving it to a different, more generic location (because it it pretty general) or giving it a name that makes it less prone to conflicts.

sdk/tests/js/tests/webgl_multiview_util.js
```diff
@@ -0,0 +1,109 @@
...
+    '        default:',
+    '            my_FragColor = vec4(0.0, 0.0, 0.0, 1.0);',
+    '            break;',
+    '    }',
+    '}'].join('\n');
+}
+
+function getExpectedColor(view) {
+    var mask = (view + 1);
+    return [(mask & 4) ? 255 : 0, (mask & 2) ? 255 : 0, (mask & 1) ? 255 : 0, 255];
```
toji: Feels like the switch statement above and this should use the same logic in both places, to make it easier to identify the relationship and update in the future if needed.

sdk/tests/conformance2/extensions/webgl_multiview.html
```diff
@@ -0,0 +1,429 @@
...
+        testFailed('Unexpected FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE ' + wtu.glEnumToString(gl, objectType) + ' after detach, should be NONE');
+    } else {
+        testPassed('FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE was NONE after detach');
+    }
+}
+
+function runClearTest()
+{
+    debug("");
+    debug("Testing clear");
```
toji: Would appreciate a slightly more informative comment, such as "Testing that glClear clears all views"

sdk/tests/js/tests/webgl_multiview_util.js
```diff
@@ -0,0 +1,109 @@
...
+    '            break;',
+    '        case 6u:',
+    '            my_FragColor = vec4(1.0, 1.0, 1.0, 1.0);',
+    '            break;',
+    '        default:',
+    '            my_FragColor = vec4(0.0, 0.0, 0.0, 1.0);',
+    '            break;',
+    '    }',
+    '}'].join('\n');
+}
```
toji: I think the existing shaders that you have are great, but I would like to see at least one example that shows the commonly expected code path of using the gl_ViewID_OVR to index into an uniform array. There's enough silly driver bugs out there, might as well be careful. 😛

sdk/tests/conformance2/extensions/webgl_multiview.html
Oletus: Done.

sdk/tests/js/tests/webgl_multiview_util.js
Oletus: Done.

sdk/tests/js/tests/webgl_multiview_util.js
Oletus: Renamed this to createTextureWithNearestFiltering.

sdk/tests/js/tests/webgl_multiview_util.js
Oletus: Done, added a test case that transforms the quad with a different transformation matrix according to the view.

sdk/tests/conformance2/extensions/webgl_multiview.html
```diff
@@ -0,0 +1,505 @@
...
+    wtu.drawUnitQuad(gl);
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be no errors from draw");
+
+    let readFb = gl.createFramebuffer();
+    gl.bindFramebuffer(gl.READ_FRAMEBUFFER, readFb);
+    for (let viewIndex = 0; viewIndex < views; ++viewIndex) {
+        gl.framebufferTextureLayer(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, colorTex, 0, viewIndex);
+        let expectedColor = getExpectedColor(viewIndex);
+        let colorRegionLeftEdge = (width / views) * viewIndex;
+        let colorRegionRightEdge = (width / views) * (viewIndex + 1);
```
kenrussell: The exact definition of the thin vertical strip should ideally be better described, perhaps in the MultiviewOffsetVertexShader.

extensions/WEBGL_multiview/extension.xml
```diff
@@ -164,7 +164,7 @@ interface WEBGL_multiview {
       <ul>
         <li>if <code>numViews</code> is less than one</li>
         <li>if <code>numViews</code> is more than <code>MAX_VIEWS_OVR</code></li>
-        <li>with the parameters set so that <code>baseViewIndex</code> + <code>numViews</code> is larger than the value of <code>MAX_ARRAY_TEXTURE_LAYERS</code> minus one</li>
+        <li>with the parameters set so that <code>baseViewIndex</code> + <code>numViews</code> is larger than the value of <code>MAX_ARRAY_TEXTURE_LAYERS</code></li>
```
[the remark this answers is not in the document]
Oletus: I don't expect that this spec change will give us a lot of issues. The issue never existed in the multiview implementation that's in the desktop OpenGL and D3D11 backends of ANGLE. OVR_multiview implementations in drivers might have the earlier, tighter spec in place but that will just mean they'll report an incorrect error.

When it comes to native OVR_multiview implementations that's probably the least of our worries - the first revisions of the extension were massively underspecified, and implementations seemed rushed out of the door with a lot of bugs. I'd expect that older OVR_multiview implementations will need to be blacklisted in browsers once support is added for it.

sdk/tests/conformance2/extensions/webgl_multiview.html
Oletus: I added a few comments describing the test shaders.

PR #2671, opened 2018-07-18
Add a test for draw framebuffer completeness

Oletus:
The test checks that the draw framebuffer can still be rendered to
even if the read framebuffer is incomplete.

The new test exposes a bug in Chrome 67. It seems to be specific to
ANGLE's D3D11 backend. Bug link: http://anglebug.com/2737

The latest stable version of Firefox is not affected by the bug.

### DIFF
--- a/sdk/tests/conformance2/rendering/framebuffer-completeness-draw-framebuffer.html
+++ b/sdk/tests/conformance2/rendering/framebuffer-completeness-draw-framebuffer.html
@@ -39,6 +39,7 @@
 <div id="console"></div>
 <script>
 // This exposes a bug in Chrome 67: If the read framebuffer is incomplete, then draw can fail even if the draw framebuffer is complete.
+// http://anglebug.com/2737
 
 "use strict";
 description();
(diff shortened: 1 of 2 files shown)


Oletus:
ANGLE was patched to fix the bug, I checked that latest version of Canary passes the added test. Should be OK to merge this now?

sdk/tests/conformance2/rendering/framebuffer-completeness-draw-framebuffer.html
```diff
@@ -0,0 +1,94 @@
...
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<script>
+// This exposes a bug in Chrome 67: If the read framebuffer is incomplete, then draw can fail even if the draw framebuffer is complete.
+
```
kenrussell: Could you add the ANGLE or Chromium bug ID in a comment?

Oletus: Done.

PR #2672, opened 2018-07-19
Improve logging for tex-video-using-tex-unit-non-zero.html .

kenrussell:
Use WebGLTestUtils.comparePixels, and include failing video type in
error messages.

Intended to diagnose intermittent failures of this test in
http://crbug.com/830901 .


kenrussell:

kenrussell:
Per discussion on today's conference call: this pull request just changes this test to use logging facilities that are better than the hand-crafted ones originally written here. If it's ever necessary to add more logging that's detrimental to the project overall then we'll figure out a way to do this without modifying the master branch of this repository.

PR #2673, opened 2018-07-19
Add [Exposed] Web IDL attribute to all WebGL interfaces.

### LINKED ISSUE #2571: Accessing WebGL types in a worker?
I'm probably just missing this somewhere but, I have some semi-generic WebGL code that needs to check if an object is `WebGLTexture` or a `WebGLRenderbuffer`. For the last several years I've had code like this

      isTexture = object instanceOf WebGLTexture;

But now I'm trying to run this code in a worker and AFAICT `WebGLTexture` does not exist inside a worker even though WebGL itself is running via `OffscreenCanvas`.

* Is this a known issue? 
* Will the WebGL types be exposed inside a worker?
* Are they exposed somewhere not the global object? 

As it is a few workarounds which seem unclean (some wishy washy feeling I know but ...)

1.  call `gl.isTexture` which requires a `WebGLTexture` but throws if it's not. Seems pretty heavy to generate an exception just to check this stuff.  In other words

         function isTexture(gl, obj) {
           let result = false;
           try {
             result = gl.isTexture(obj);
           } catch (e) {
           }
           return result;
         }

2. call `representative = gl.createTexture()` somewhere then use `objectInQuestion instanceOf representative.constructor`. ugh but it works and at least it's fast and doesn't generate garbage every check as 1 above does but it does generate a bogus object.

3. ???

kenrussell:
This extended attribute has been required by Web IDL since
heycam/webidl#423 . In order for WebGL to be accessible by web workers
using OffscreenCanvas, the attribute being added to all WebGL
interfaces is:
[Exposed=(Window,Worker)]

Fixes #2571 .

### DIFF
--- a/specs/latest/2.0/webgl2.idl
+++ b/specs/latest/2.0/webgl2.idl
@@ -28,19 +28,24 @@ typedef long long GLint64;
 typedef unsigned long long GLuint64;
 
 
-[Exposed=(Window,Worker)] interface WebGLQuery : WebGLObject {
+[Exposed=(Window,Worker)]
+interface WebGLQuery : WebGLObject {
 };
 
-[Exposed=(Window,Worker)] interface WebGLSampler : WebGLObject {
+[Exposed=(Window,Worker)]
+interface WebGLSampler : WebGLObject {
 };
 
-[Exposed=(Window,Worker)] interface WebGLSync : WebGLObject {
+[Exposed=(Window,Worker)]
+interface WebGLSync : WebGLObject {
 };
 
-[Exposed=(Window,Worker)] interface WebGLTransformFeedback : WebGLObject {
+[Exposed=(Window,Worker)]
+interface WebGLTransformFeedback : WebGLObject {
 };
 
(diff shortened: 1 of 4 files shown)


kenrussell:

specs/latest/2.0/webgl2.idl
```diff
@@ -10,19 +10,19 @@ typedef long long GLint64;
...
-interface WebGLSync : WebGLObject {
+[Exposed=(Window,Worker)] interface WebGLSync : WebGLObject {
 };
 
-interface WebGLTransformFeedback : WebGLObject {
+[Exposed=(Window,Worker)] interface WebGLTransformFeedback : WebGLObject {
 };
 
-interface WebGLVertexArrayObject : WebGLObject {
+[Exposed=(Window,Worker)] interface WebGLVertexArrayObject : WebGLObject {
```
jdashg: It would be nice the `[Exposed...` bit was always on the line before `interface...`.

kenrussell: Good point. Done throughout.

PR #2674, opened 2018-07-19
Add license to webgl.idl and webgl2.idl.

### LINKED ISSUE #2668: Add a license on the spec document
We in [TSJS-lib-generator](https://github.com/Microsoft/TSJS-lib-generator) wants to collect WebGL 1.0/2.0 IDLs to generate TypeScript library file, but it seems the spec does not mention anything about the license. Can you add one so that the TypeScript team can use the spec?

See also https://github.com/Microsoft/TSJS-lib-generator/pull/533#discussion_r202093015

kenrussell:
Use the same MIT license that's used throughout the WebGL SDK, as it
seems a better choice than Khronos' specification license.

Remove obsolete comments referring to Khronos' typed array spec; it
was folded into the ECMAScript specification some time ago.

Fixes #2668 .


kenrussell:
Please review. @jdashg @grorg @kainino0x @RafaelCintron

jdashg:
Please confirm with Neil.

kenrussell:
@jdashg Confirmed with Neil that this is OK. Merging.

PR #2676, opened 2018-07-20
Fix the ERROR_EPSILON for GL struct tests

### LINKED ISSUE #2675: GL struct tests use medium precision but with an EPSILON that is too strict
This issue affects the following tests:  conformance/ogles/GL/struct/struct_049_to_056.html

These tests make use of shader [nestedstructcomb_various_frag.frag ](https://github.com/KhronosGroup/WebGL/blob/6a1722fa0ed15b52563da09989115c729dca21c9/conformance-suites/1.0.3/conformance/ogles/GL/struct/nestedstructcomb_various_frag.frag#L31) which uses an ERROR_EPSILON of 0.1.    There is a corresponding vertex shader that uses the same value for ERROR_EPSILON.

The above shaders use an if-check similar to below (where "sum" is mediump):

> if ((sum > 130.0-ERROR_EPSILON) &&  (sum < 130.0+ERROR_EPSILON) ) 

but since 130.1 cannot be exactly represented in FP16 format, where 1 ULP is 0.125, an implementation that uses FP16 and uses round-to-zero rounding rules wont evaluate the above expression as expected.   The solution is to relax ERROR_EPSILON to 0.125 such that it equals 1 ULP at FP16.  

I'll create a pull request to fix the two affected shaders.

jackohound:
Affects the ogles/struct/struct_049_to_56.html tests.  The ERROR_EPSILON
value of 0.1 was too strict for FP16 implementation of medium precision.
This change relaxes the epsilon to 0.125.

Fixes #2675

### DIFF
--- a/conformance-suites/1.0.3/conformance/ogles/GL/struct/nestedstructcomb_various_frag.frag
+++ b/conformance-suites/1.0.3/conformance/ogles/GL/struct/nestedstructcomb_various_frag.frag
@@ -28,7 +28,7 @@ precision mediump float;
 #endif
 varying vec4 color;
 
-#define ERROR_EPSILON 0.1
+#define ERROR_EPSILON 0.125
 
 void main (void){
       struct second_nest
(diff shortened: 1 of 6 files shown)


jackohound:
@kenrussell, I wasn't sure which branches I should include in the PR.  
I've updated the PR to update 1.0.3, 2.0.0, and /sdk/conformance areas.

PR #2680, opened 2018-07-30
Explicitly set image's src in tex-video-using-tex-unit-non-zero.html .

kenrussell:
It's not clear whether there's any guarantee that an img tag whose
source is a data: URL will be fully loaded before inline scripts below
it execute. Instead, set an onload handler for the image and set its
src in script.

Intended to address intermittent failures in this test per
http://crbug.com/830901 .


kenrussell:
@kainino0x @jdashg please review. Thanks.

kenrussell:
Note that upon further thought this is probably not the problem; it's more likely that the video is occasionally displaying a blank frame. Still, if this particular issue seems to potentially be a problem then it should be resolved.

kenrussell:
I think this change should be non-controversial so merging it now.

PR #2682, opened 2018-07-31
Add WEBGL_multiview tests for single view operations

Oletus:
This adds two test files. One concentrates on drawing to multiview
framebuffers with shaders that don't contain a GL_OVR_multiview
extension directive. The other one does different operations on
multiview framebuffers containing just a single view.

All of the added tests pass with the latest version of ANGLE, however
latest Chromium still needs to integrate the latest ANGLE changes with
fixes to multiview read functionality.


Oletus:
Chrome now has the newer ANGLE so that these tests should all pass, merging.

PR #2684, opened 2018-08-01
Test WEBGL_multiview when the extension is disabled

Oletus:
The added tests make sure that none of the WEBGL_multiview
functionality is enabled if the extension is disabled.


Oletus:
This covers a bug in the initial Chromium implementation fixed by: https://chromium-review.googlesource.com/c/chromium/src/+/1158569

PR #2685, opened 2018-08-01
Expose "canvas" attribute only to Window context.

### LINKED ISSUE #2683: WebGLRenderingContextBase incorrectly exposes canvas attribute
#2673 added `[Exposed=(Window, Worker)]` to `WebGLRenderingContext` but it has `readonly attribute HTMLCanvasElement canvas` that cannot be exposed to Worker. Maybe the attribute needs its own `[Exposed=Worker]`?

kenrussell:
HTMLCanvasElement doesn't exist on workers.

Fixes #2683.


kenrussell:
Please review. @jdashg @kainino0x @RafaelCintron @grorg 

Note I didn't include OffscreenCanvas even though this is how the WebGL context is fetched on workers, because it seems to not be defined in the w3c's version of the spec here:
http://w3c.github.io/html/semantics-scripting.html#the-canvas-element

though it is defined in the whatwg version of the spec:
https://html.spec.whatwg.org/multipage/canvas.html#the-offscreencanvas-interface

Will follow up with @fserb on this.

PR #2687, opened 2018-08-03
Update Chrome troubleshooting URL.

kenrussell:
Make it point to the top-level help for Chrome. The previous URL has
been redirected to a "tips and tricks" page which is not very useful.

Thanks to webmaster James Riordon for raising the issue.


kenrussell:
FYI. Merging myself.

Issue #2690, opened 2018-08-07
Expose KHR_parallel_shader_compile to WebGL


jchen10:
More information at https://bugs.chromium.org/p/chromium/issues/detail?id=849576.

greggman:
Based on other async stuff in WebGL does this need to have extra restrictions?

For example `getQueryParameter` says 

> In order to ensure consistent behavior across platforms, queries' results must only be made available when the user agent's event loop is not executing a task. In other words:

> * A query's result must not be made available until control has returned to the user agent's main loop.
> * Repeatedly fetching a query's QUERY_RESULT_AVAILABLE parameter in a loop, without returning control to the user agent, must always return the same value.

Similarly I wonder if this spec should change to do the same thing.  Should it require not working in a spin loop?

Also, why do I care how many threads at all? It seems to me there's no reason for `MaxShaderCompilerThreadsKHR` nor `MAX_SHADER_COMPILER_THREADS_KHR`. 

All that's needed is being able to pass `COMPLETION_STATUS_KHR` to `getShaderParameter` and `getProgramParameter`.  An implementation that doesn't support parallel compilation will just always return `true` and the next call dealing with info about that shader/program will block (as it does now). Why so I have to opt in by calling `MaxShaderCompilerThreadsKHR`? Basically `COMPLETION_STATUS_KHR` is just asking *"when I check for success or I query locations will my call block, yes or no?"* so what's the point in specifying or querying a number of threads?

Also querying a number of threads is yet more data leak. Not that it matters much at this point for WebGL

jchen10:
@greggman Thank you for the comment.
For the extra async restriction, I am not quite sure about the across platforms consistency.  @kenrussell Maybe you can comment on this.

For your thread number concern, my thoughts are:
    We should keep consistent with the original spec if no strong incompatibility with Web.
    Querying thread number only returns what you have set, or 0xFFFFFFFF.  There is no risk of data leak.
    Setting thread number can offer a fine gained control on CPU resource to be engaged in compilation and linking.

kenrussell:
@greggman We'll consider amending the spec once we have a prototype working and see how well it works in practice. From the standpoints of security and interoperability between browsers and with the rest of the web platform, I think it's OK to let these programs compile and link as quickly as possible.

@jchen10 in Chrome I had been planning to enforce an implementation-specific maximum of 1 background thread per WebGL context. More than that is probably excessive.

greggman:
[...]

PR #2691, opened 2018-08-07
Expose KHR_parallel_shader_compile. (#2690)

jchen10:
Add the extension proposal.


jchen10:
@Richard-Yunchao, Please take a look firstly. Thanks!
cc @gyagp

extensions/proposals/KHR_parallel_shader_compile/extension.xml
```diff
@@ -0,0 +1,87 @@
...
+      void maxShaderCompilerThreadsKHR(GLuint count);
+    };
+  </idl>
+
+  <samplecode xml:space="preserve">
+    <pre>
+    var canvas = document.createElement("canvas");
+    var gl = canvas.getContext("webgl");
+    var ext = gl.getExtension('KHR_parallel_shader_compile');
+    // Don't do this unless you really want to override the implementation-specific maximum.
```
gyagp: do -> change?

extensions/proposals/KHR_parallel_shader_compile/extension.xml
```diff
@@ -0,0 +1,87 @@
...
+<!-- vi:set sw=2 ts=4: -->
+<?xml-stylesheet href="../../extension.xsl" type="text/xsl"?>
+<proposal href="proposals/KHR_parallel_shader_compile/">
+  <name>KHR_parallel_shader_compile</name>
+
+  <contact> <a href="https://www.khronos.org/webgl/public-mailing-list/">WebGL
+  working group</a> (public_webgl 'at' khronos.org) </contact>
+
+  <contributors>
+    <contributor>Members of the WebGL working group</contributor>
```
Richard-Yunchao: Add yourself here.

extensions/proposals/KHR_parallel_shader_compile/extension.xml
```diff
@@ -0,0 +1,87 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- vi:set sw=2 ts=4: -->
+<?xml-stylesheet href="../../extension.xsl" type="text/xsl"?>
+<proposal href="proposals/KHR_parallel_shader_compile/">
```
Richard-Yunchao: I think the name should be WEBGL_parallel_shader_compile, instead of KHR_parallel_shader_compile.

extensions/proposals/KHR_parallel_shader_compile/extension.xml
```diff
@@ -0,0 +1,87 @@
...
+  </depends>
+
+  <overview>
+    <mirrors href="https://www.khronos.org/registry/OpenGL/extensions/KHR/KHR_parallel_shader_compile.txt"
+             name="KHR_parallel_shader_compile">
+    </mirrors>
+
+    <features>
+      <feature>
+        Shader compilation and program linking may be performed in a separate CPU thread.
```
Richard-Yunchao: Can you describe this extension a little bit more, just like what it says at KHR_parallel_shader_compile?

extensions/proposals/KHR_parallel_shader_compile/extension.xml
kenrussell: KHR_parallel_shader_compile is preferred. The WebGL working group prefers to only rename extensions into the WEBGL_ namespace when there is a significant functionality difference compared to the underlying extension. In this case, the extension is being exposed exactly as specified.

extensions/proposals/KHR_parallel_shader_compile/extension.xml
```diff
@@ -0,0 +1,87 @@
...
+    };
+  </idl>
+
+  <samplecode xml:space="preserve">
+    <pre>
+    var canvas = document.createElement("canvas");
+    var gl = canvas.getContext("webgl");
+    var ext = gl.getExtension('KHR_parallel_shader_compile');
+    // Don't do this unless you really want to override the implementation-specific maximum.
+    ext.maxShaderCompilerThreadsKHR(3);
```
kenrussell: Let's update this code sample to show the best practices. Please:
 - Put the code which uses the extension object in "if (ext) { ... }" blocks, to make it clearer that this extension might not be supported
 - Query MAX_SHADER_COMPILER_THREADS_KHR
 - Set it to the maximum of (2, getParameter(MAX_SHADER_COMPILER_THREADS_KHR))

extensions/proposals/KHR_parallel_shader_compile/extension.xml
```diff
@@ -0,0 +1,87 @@
...
+    gl.shaderSource(fShader, fSource);
+    gl.compileShader(fShader);
+
+    var program = gl.createProgram();
+    gl.attachShader(program, vShader);
+    gl.attachShader(program, fShader);
+    gl.linkProgram(program);
+
+    function checkCompletion() {
+      if (gl.getProgramParameter(program, ext.COMPLETION_STATUS_KHR) == true) {
```
kenrussell: Let's make this code work both with and without the extension. Please start with "if (ext) { ... }" and have the fallback path (synchronous linking) do the error checks too.

extensions/proposals/KHR_parallel_shader_compile/extension.xml
jchen10: The initial value of MAX_SHADER_COMPILER_THREADS_KHR is 0xFFFFFFFF according to the spec. If users didn't set it, the query returns 0xFFFFFFFF always.

jchen10: Done

extensions/proposals/KHR_parallel_shader_compile/extension.xml
jchen10: Done.

extensions/proposals/KHR_parallel_shader_compile/extension.xml
jchen10: Done.

extensions/proposals/KHR_parallel_shader_compile/extension.xml
jchen10: Done.

jchen10:
@Richard-Yunchao Thanks for the suggestions.
@kenrussell  I checked with make. The generated html can work correctly.

PR #2693, opened 2018-08-08
Add a test for passing a struct from a const array

Oletus:
This covers bug http://crbug.com/871434 . The root cause seems to be
in the NVIDIA OpenGL driver.

### DIFF
--- a/sdk/tests/conformance2/glsl3/const-struct-from-array-as-function-parameter.html
+++ b/sdk/tests/conformance2/glsl3/const-struct-from-array-as-function-parameter.html
@@ -1,0 +1,80 @@
+<!--
+
+/*
+** Copyright (c) 2018 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/glsl3/const-struct-from-array-as-function-parameter.html
```diff
@@ -0,0 +1,80 @@
...
+
+// Covers bug:
+// http://crbug.com/871434
+
+GLSLConformanceTester.runTests([
+{
+  fShaderId: 'fShaderTest',
+  fShaderSuccess: true,
+  linkSuccess: true,
+  passMsg: 'Passing a struct from a dynamically indexed const array as a non-const function parameter should work.',
```
jdashg: s/as/into/
The spec VM treats structs passed into functions as copies, but 'as' sounds like it's a reference. While under the covers, the copy is certainly possible to optimize out, from the spec's standpoint, structs are always copied in. (and out, when marked as such)

Oletus: Done.

jdashg:
Thanks!

sdk/tests/conformance2/glsl3/const-struct-from-array-as-function-parameter.html
```diff
@@ -0,0 +1,80 @@
...
+bool useStruct( S s) { return s.member > 0.0; }
+
+out vec4 outColor;
+
+void main( void )
+{
+    outColor = vec4(0.0, 0.0, 0.0, 1.0);
+    for (int i = 0; i < 2; ++i) {
+        if (useStruct(s[i])) {
+            outColor.g++;
```
kenrussell: Is it intended in this test that g can evaluate to a value > 1.0?

PR #2695, opened 2018-08-09
Add WEBGL_multiview tests for depth buffer and transform feedback

Oletus:
These tests pass on the latest prototype implementation in Chromium.

### DIFF
--- a/sdk/tests/js/tests/webgl_multiview_util.js
+++ b/sdk/tests/js/tests/webgl_multiview_util.js
@@ -172,6 +172,39 @@ function getMultiviewColorFragmentShaderForDrawBuffers(drawBuffers) {
     return wtu.replaceParams(shaderCode, {'drawBuffers' : drawBuffers});
 }
 
+function getMultiviewVaryingVertexShader(views) {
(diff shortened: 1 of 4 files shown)


sdk/tests/js/tests/webgl_multiview_util.js
```diff
@@ -172,6 +172,39 @@ function getMultiviewColorFragmentShaderForDrawBuffers(drawBuffers) {
...
+    'void main() {',
+    '    gl_Position = a_position;',
+    '    testVarying = float(gl_ViewID_OVR);',
+    '}'].join('\n');
+    return wtu.replaceParams(shaderCode, {'num_views': views});
+}
+
+function getMultiviewVaryingFragmentShader() {
+    return ['#version 300 es',
+    '#extension GL_OVR_multiview : require',
```
jdashg: My preference for formatting of these arrays is:
return [
    'a',
    'b',
].join('\n');

I don't mind this either way, but consider it!

jdashg:
Thanks for writing these.

kenrussell:
LGTM after the fact.

PR #2696, opened 2018-08-11
Document non-blocking getBufferSubData

kainino0x:
I'm intending to use a derivative of this text in an MDN update as well.


kenrussell:
Looks good!

PR #2697, opened 2018-08-21
Propose WEBGL_blend_equation_advanced

AshleyScirra:
Proposal to expose KHR_blend_equation_advanced to WebGL.


kenrussell:
Thanks for this contribution. Could you please follow the pattern for the other extension proposals – create a subdirectory, create an extension.xml inside it (look at WEBGL_video_texture for an example), and after doing so, ensure it builds by running "make" in the extensions/ directory? Thanks.

AshleyScirra:
Is there some getting started guide somewhere? 'make' fails for me:

```
: not found: 2: ./find-exts:
find: missing argument to `-exec'
xsltproc --stringparam extensions  \
         summary.xsl summary.xml > registry.xml
compilation error: file summary.xml line 3 element registry
xsltParseStylesheetProcess : document is not a stylesheet
Makefile:33: recipe for target 'registry.xml' failed
make: *** [registry.xml] Error 5
```

kenrussell:
On Windows the Makefile needs to be run either from Cygwin or, likely, the Linux subsystem for Windows (though I haven't tried the latter).

AshleyScirra:
I was using the Linux subsystem for Windows. Is that not supported after all?

kenrussell:
Sorry, I've tried but can't test on it. Our corporate policy is to not allow the Windows Subsystem for Linux.

It looks like something's going wrong in the find-exts shell script; it's pretty tiny so if you can debug the issue we'll gladly accept a patch.

kenrussell:
Thanks, merged this locally and it looks good. We'll have to figure out what to do with the coherent version of the extension. Merging now.

AshleyScirra:
FWIW to demonstrate the performance impact of having to render via an intermediate texture, consider these measurements with Construct's WebGL engine measuring how many quads can be rendered at 30 FPS:

Additive blend (using blendFunc): 207000
Multiply blend (using shader): 1500

So when rendering with lots of quads (e.g. in a game engine with lots of blending sprites), the shader is over 100x slower than using the built-in blends.

kenrussell:
The WebGL working group discussed this last week and the main concern is whether the non-coherent version of this extension should be exposed at all. The thought is that only the coherent version should be exposed. It should provide the majority of the available performance benefits without introducing more undefined behavior into the API. #2704 has been filed about making this revision.

AshleyScirra:
According to gpuinfo.org support for non-coherent vs. coherent:
OpenGL ES: ~32% vs. ~28%
OpenGL (KHR variant only): ~47% vs. ~21%
Shame to lose the extra support on desktop but totally get the point about avoiding undefined behavior. Hopefully hardware support increases over time.

PR #2698, opened 2018-08-22
Fix href of rejected WEBGL_get_buffer_sub_data_async extension.

kenrussell:
This was causing the extension registry to fail to rebuild
successfully.


kenrussell:
FYI @kainino0x @jdashg . Merging myself to unblock others' work.

PR #2702, opened 2018-08-31
Make some tests succeed to fail more gracefully

nox:
Those tests error out when run in Servo because of missing support for a bunch of features, such as preserving the drawing buffer or loading SVG images. Instead, make the tests fail, and finish the tests
so that the next ones are started.

### DIFF
--- a/sdk/tests/conformance/buffers/buffer-data-array-buffer-delete.html
+++ b/sdk/tests/conformance/buffers/buffer-data-array-buffer-delete.html
@@ -54,24 +54,28 @@ canvas.addEventListener(
 
 var wtu = WebGLTestUtils;
 var gl = wtu.create3DContext(canvas, {preserveDrawingBuffer: true});
-shouldBeNonNull("gl");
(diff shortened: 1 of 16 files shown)


kenrussell:
The description of this pull request is confusing. Would this describe it more properly?

"Make some tests fail more gracefully and cooperatively.

Those tests time out when run in Servo because of missing support for a bunch of features, such as preserving the drawing buffer or loading SVG images. Instead, make the tests fail, and finish the tests
so that the next ones are started."

If that sounds good, the changes look okay, though some of them are quite extensive re-indentation (in particular, for tex-video-using-tex-unit-non-zero.html). Wish it had been possible to make the tests more robust with fewer changes.

nox:
I updated the description. I tried to make the changes minimal, but there is not much I can do when I need to return early and the code is directly in a `script` tag.

kenrussell:
Thanks. I edited the description manually while squashing and merging, too.

PR #2703, opened 2018-08-31
Better specify errors for vertexAttrib*v entry points.

### LINKED ISSUE #2700: Methods vertexAttrib* are underspecified
Contrary to their `uniform*` counterpart, it's not specified what should happen when the array passed to one of the vector forms (those ending with `v`) has an invalid length. Citing the spec about `uniform*`:

> If the array passed to any of the vector forms (those ending in `v`) has an invalid length, an `INVALID_VALUE` error will be generated. The length is invalid if it is too short for (...) the assigned type.

kenrussell:
This behavior is already tested by gl-vertex-attrib.html.

Fixes #2700.


kenrussell:

PR #2705, opened 2018-09-05
Generate errors with non active TF bindings.

null77:
This simplifies the logic for transform feedback simultaneous use.
Instead of conditionally checking if the TF is active, always check for
simultaneously bound array buffers. This simplifies the validation logic
since we don't need as many separate checks.

PTAL @kenrussell @jdarpinian @jdashg 

The only change to the usage here is making simultaneous use invalid when the user omits the call to glBeginTransformFeedback and calls a draw. In my mind there isn't a difference between validating a pack buffer's simultaneous use when the XFB is disabled vs an array buffer. It also cleans up ANGLE's validation logic as it reduces three separate checks to a single call.

### DIFF
--- a/sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
+++ b/sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
@@ -241,9 +241,9 @@ for (let [drawArrays, drawElements] of drawFunctions) {
   drawWithFeedbackBound(gl, drawArrays, prog, vao, tf, true);
   wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "tf enabled");
   drawWithFeedbackBound(gl, drawArrays, prog, vao, tf, false);
-  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "drawArrays: tf disabled");
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "drawArrays: tf disabled");
   drawWithFeedbackBound(gl, drawElements, prog, vao, tf, false);
-  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "drawElements: tf disabled");
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "drawElements: tf disabled");
   gl.bindBuffer(gl.COPY_READ_BUFFER, null);
 
   debug("<hr/>Test TF buffer bound to disabled vertex attrib");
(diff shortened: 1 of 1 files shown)


kenrussell:
@null77 , per @jdashg 's feedback, it would help to either have a concrete example of how this will simplify ANGLE's validation logic and/or enable caching of state. From previous discussions, it sounds like ANGLE's validation of BindBuffer is currently expensive, but it's not clear how this will help. Pointers to bugs or CLs would be helpful.

Note that WebGL's behavior in this area is an attempt to formalize the overly vague definitions in the ES spec. From ES 3.0.4 section 2.15.2 "Transform Feedback Primitive Capture":

"Buffers should not be bound or in use for both transform feedback and other purposes in the GL. Specifically, if a buffer object is simultaneously bound to a transform feedback buffer binding point and elsewhere in the GL, any writes to or reads from the buffer generate undefined values. Examples of such bindings include ReadPixels to a pixel buffer object binding point and client access to a buffer mapped with MapBufferRange."

By one reading of that spec, this pull request is valid.

WebGL attempted to be permissive in this area by limiting its restrictions to the active transform feedback object, though in my opinion it would also be OK to do this for the currently bound transform feedback regardless of its active status. I do agree with @jdashg that if it's possible to maintain the current more permissive behavior and still reach ANGLE's performance goals, that would be better.

null77:
Thanks for the feedback. I think I agree we probably shouldn't land this. But I'll try to give some context and motivation.

ANGLE's validation is quite complicated for this check and affects both BindBuffer and all draw call performance. Currently on a draw we not only need to check the transform feedback object for a binding conflict as Jeff's code snippet suggested. We also need to check every uniform buffer object for a transform feedback conflict with a non-active transform feedback. e.g.

unfiormBuffer = GenBuffers
BindBuffer(UNIFORM_BUFFER, 0, uniformBuffer)
BindTransformFeedback(xfb1)
BindBuffer(XFB_BUFFER, 0, uniformBuffer)
BindTransformFeedback(xfb2)
DrawArrays -> error

We also need to do the same check for all of the buffers bound to the vertex array. This kind of check is very expensive and even with caching adds overhead. Without caching is can be quite slow. Jeff do you know what Firefox does here? Note that the check can't be cached in the TF object, since the unbound TF is the one that owns the error.

I had a WIP CL that I abandoned here that attempted to remove some of these checks. It can give you a brief idea of where the interlinked caching is in ANGLE:

https://chromium-review.googlesource.com/c/angle/angle/+/1207432

jdashg:
UBO buffers not used for TF:
https://dxr.mozilla.org/mozilla-central/rev/703546ab6d0cb643028a1ab4fda997b38f38a2e6/dom/canvas/WebGLContextDraw.cpp#334
TF buffers not used for non-TF:
https://dxr.mozilla.org/mozilla-central/rev/703546ab6d0cb643028a1ab4fda997b38f38a2e6/dom/canvas/WebGLContextDraw.cpp#361
Attribs not used for TF:
https://dxr.mozilla.org/mozilla-central/rev/703546ab6d0cb643028a1ab4fda997b38f38a2e6/dom/canvas/WebGLProgram.cpp#509

The required overhead of these checks is probably not terrible? (traverse all ubo/tfo/vao buffers and check the tf/non-tf refcounts) 

Naive caching might help when models are drawn across multiple draw calls. Firefox does not cache this presently, but a naive cache that's invalidated by BindBuffer* might be useful for some use cases.

jdashg:
Tentatively closing this, since it sounds like we're okay without it.

null77:
Yeah, fine with me.

PR #2706, opened 2018-09-11
Add canvas-teximage-after-multiple-drawimages.html test.

kenrussell:
Add test case uploading a 2D-rendered canvas which has had a series of
drawImage operations (from both images and other 2D-rendered canvases)
applied to it.

Regression test for http://crbug.com/878545 .


kenrussell:
@kainino0x @jdashg @grorg @RafaelCintron please review.

This is the simplest test I could create which still reproduced the bug in Chrome. I'm pretty sure this test already passes in all the other browsers, though only tested on Firefox so far.

jdashg:
Yep, merge when ready. const/let would be nice, but not a blocker.

kenrussell:
Thanks @kainino0x and @jdashg for your reviews. Used const/let instead of var, and added a couple more comments. Merging now.

PR #2707, opened 2018-09-11
Add ComputeBoids example

qjia7:
This example is ported from
https://github.com/google/nxt-standalone/blob/master/examples/ComputeBoids.cpp.

Try this example using latest chrome canary with below commands:
Chrome.exe --use-cmd-decoder=passthrough --enable-webgl2-compute-context
           --use-angle=gl  --use-gl=angle

Currently, it still can't get correct result on d3d backend since SSBO
for d3d is still under development in ANGLE.

### DIFF
--- a/sdk/demos/common/webgl-utils.js
+++ b/sdk/demos/common/webgl-utils.js
@@ -92,6 +92,14 @@ var OTHER_PROBLEM = '' +
   '<a href="http://get.webgl.org/troubleshooting/">Click here for more information.</a>';
 
 /**
+ * Mesasge for need better hardware
+ * @type {string}
+ */
+var WEBGL2_COMPUTE_NOT_SUPPORTED = '' +
+  "It doesn't appear your computer can support WebGL 2.0 Compute.<br/>" +
+  '<a href="http://get.webgl.org/troubleshooting/">Click here for more information.</a>';
+
+/**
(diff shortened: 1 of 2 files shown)


qjia7:
@kenrussell The latest chrome canary has regression. https://chromium-review.googlesource.com/c/angle/angle/+/1185743 caused that regression.  If you want to test this example, please build your own libGLESv2.so  without https://chromium-review.googlesource.com/c/angle/angle/+/1185743 's changes (revert 337bd698407850d17522a61a2ed3bb881a563e5c) using ANGLE source code. And replace it with the current one in chrome canary. Then you can get the right result.

I put one libGLESv2.so that I built here https://drive.google.com/drive/folders/1y58cD3hNX94Mf5oD2An0ghv_VszpO-9P in case you want to have a try.

kainino0x:
Have you reported a bug about that regression? If so could you link it here?

qjia7:
@kainino0x Report the bug here https://bugs.chromium.org/p/angleproject/issues/detail?id=2810. Also add it in code description. Thanks.

kenrussell:
Great work porting this sample to WebGL 2.0 Compute and thanks for filing the bug! I assume it will be fixed soon as you've already identified the cause, so let's merge this so others can more easily take a look.

PR #2708, opened 2018-09-11
Avoid using window's named getter on a couple of tests

nox:
Servo doesn't implement the named getter yet and thus those tests fail for
reasons unrelated to our WebGL support.

### DIFF
--- a/sdk/tests/conformance/rendering/framebuffer-switch.html
+++ b/sdk/tests/conformance/rendering/framebuffer-switch.html
@@ -42,6 +42,7 @@
 "use strict";
 description("Test framebuffer switching. The test switches between two framebuffers, copying rendering results from one to the other.");
 var wtu = WebGLTestUtils;
+var canvas = document.getElementById("canvas");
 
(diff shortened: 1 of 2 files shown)


jdashg:
Caaaan you just fix Servo?

nox:
I wish I had the bandwidth to do that but I don't so I would rather just land a quick change on those tests, most other tests do just like in my patch anyway.

sdk/tests/conformance/rendering/framebuffer-switch.html
```diff
@@ -42,6 +42,7 @@
 "use strict";
 description("Test framebuffer switching. The test switches between two framebuffers, copying rendering results from one to the other.");
 var wtu = WebGLTestUtils;
+var canvas = document.getElementById("canvas");
 
 var gl = wtu.create3DContext("canvas");
```
kenrussell: Could this have been changed to:
create3DContext(canvas)

instead of "canvas", at this point?

Only a minor question.

Issue #2704, opened 2018-09-04
Revise WEBGL_blend_equation_advanced proposal


kenrussell:
Per discussion in the WebGL working group, the WEBGL_blend_equation_advanced extension proposal should be revised. The working group's opinion is that it should require the coherent version of the extension and not expose the non-coherent behavior. This should provide the majority of the performance benefits without exposing more undefined behavior to the developer – and without exposing two extensions, which would otherwise likely be needed.

AshleyScirra:
Should the extension be renamed to WEBGL_blend_equation_advanced_coherent as well?

kenrussell:
Yes, it should be.

AshleyScirra:
This should be all done now: WEBGL_blend_equation_advanced_coherent has been proposed and WEBGL_blend_equation_advanced has been deleted.

kenrussell:
Thanks @AshleyScirra . Verified the coherent extension is showing up in the extension registry https://www.khronos.org/registry/webgl/extensions/ .

PR #2709, opened 2018-09-13
Propose WEBGL_blend_equation_advanced_coherent

AshleyScirra:
Proposal to expose KHR_blend_equation_advanced_coherent to WebGL. Forked from the previous proposal for KHR_blend_equation_advanced according to #2704.


kenrussell:
Thanks, looks fine - assume you've built this locally.

Will you put up a follow-on pull request to delete the non-coherent version of this from the repository?

AshleyScirra:
Wasn't sure if the procedure was to move it to rejected or just delete it. Can delete the non-coherent version if that's the way to go.

PR #2710, opened 2018-09-13
Add a proposal for a BPTC compressed texture extension

Oletus:
The proposed extension simply exposes the native compressed texture
extension ARB_texture_compression_bptc.

The extension should be supportable on both OpenGL and D3D11. In D3D11
the compressed texture formats exposed here are called BC7 and BC6H.
D3D11 also has signed and unsigned variants of the float format and
linear and SRGB variants of the fixed point format.


jdashg:
I'm happy with moving this straight to Draft.

jdashg:
@Oletus Do you want to update this PR to put it in Drafts, or have this merged first into Proposals?

Oletus:
Let's merge this first to proposals. It can be moved to draft when some tests are ready.

I added a mention that sampling from the SRGB format introduced in the extension performs a color space conversion.

PR #2712, opened 2018-09-17
Improve compressed texture tests

Oletus:
This refactors code to share some common utilities for testing compressed textures, improves negative test coverage and adds some tests for WEBGL_compressed_texture_etc1.

### DIFF
--- a/sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
+++ b/sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
@@ -30,6 +30,7 @@
 <link rel="stylesheet" href="../../resources/js-test-style.css"/>
 <script src="../../js/js-test-pre.js"></script>
 <script src="../../js/webgl-test-utils.js"></script>
+<script src="../../js/tests/compressed-texture-utils.js"></script>
 <title>WebGL WEBGL_compressed_texture_astc Conformance Tests</title>
 <style>
 img {
@@ -1601,6 +1602,7 @@ var decoded_12x12_argb_hdr = new Uint8Array([
 ]);
 
 var wtu = WebGLTestUtils;
+var ctu = CompressedTextureUtils;
 var contextVersion = wtu.getDefault3DContextVersion();
 var canvas = document.getElementById("canvas");
 var gl = wtu.create3DContext(canvas, {antialias: false});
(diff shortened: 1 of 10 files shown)


Oletus:
I didn't refactor the PVRTC tests since I didn't have a test device where the extension would be supported. All other changes have been tested with the extensions both supported and not supported.

Oletus:
Could I get a review on this? After this refactoring it will be easier to add tests for the new BPTC and RGTC texture compression extensions.

sdk/tests/conformance/textures/misc/00_test_list.txt
```diff
@@ -1,5 +1,5 @@
 --min-version 1.0.4 canvas-teximage-after-multiple-drawimages.html
---max-version 1.9.9 compressed-tex-image.html
+compressed-tex-image.html
```
kenrussell: Good catch on this test. It looks like the WebGL 2.0 version of the test has been deleted since it was originally integrated.

sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
```diff
@@ -1975,74 +1894,51 @@
...
-            return p;
-        }
-    }
-    return "0x" + format.toString(16);
-}
-
 function expectedByteLength(w, h, format) {
 
-    if (format == ext.COMPRESSED_RGBA_ASTC_4x4_KHR)
+    if (format == validFormats.COMPRESSED_RGBA_ASTC_4x4_KHR || format == validFormats.COMPRESSED_SRGB8_ALPHA8_ASTC_4x4_KHR)
```
kenrussell: I guess we weren't exercising the sRGB versions of these formats previously?

sdk/tests/conformance/textures/misc/compressed-tex-image.html
```diff
@@ -63,10 +59,6 @@
   var tex = gl.createTexture();
   gl.bindTexture(gl.TEXTURE_2D, tex);
 
-  wtu.shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.compressedTexImage2D(gl.TEXTURE_2D, 0, COMPRESSED_RGB_S3TC_DXT1_EXT, 4, 4, 0, new Uint8Array(8))");
-  wtu.shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.compressedTexImage2D(gl.TEXTURE_2D, 0, COMPRESSED_RGBA_S3TC_DXT1_EXT, 4, 4, 0, new Uint8Array(8))");
-  wtu.shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.compressedTexImage2D(gl.TEXTURE_2D, 0, COMPRESSED_RGBA_S3TC_DXT5_EXT, 4, 4, 0, new Uint8Array(16))");
-  wtu.shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.compressedTexImage2D(gl.TEXTURE_2D, 0, ETC1_RGB8_OES, 4, 4, 0, new Uint8Array(8))");
   wtu.shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.compressedTexImage2D(gl.TEXTURE_2D, 0, COMPRESSED_RGB_PVRTC_4BPPV1_IMG, 8, 8, 0, new Uint8Array(8))");
   wtu.shouldGenerateGLError(gl, gl.INVALID_ENUM, "gl.compressedTexImage2D(gl.TEXTURE_2D, 0, COMPRESSED_RGBA_PVRTC_4BPPV1_IMG, 8, 8, 0, new Uint8Array(8))");
```
kenrussell: It looks like if the webgl-compressed-texture-pvrtc.html test is similarly refactored in terms of your new helpers, that this test is redundant and can just be deleted. Do you agree?

kenrussell: Oops, looks like you already addressed this in your pull request comment.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -2619,6 +2619,28 @@ var getSupportedExtensionWithKnownPrefixes = function(gl, name) {
   }
 };
 
+/**
+ * @param {WebGLRenderingContext} gl The WebGLRenderingContext to use.
+ * @param {string} name Name of extension to look for.
+ * @param {boolean} extensionEnabled True if the extension was enabled successfully via gl.getExtension().
+ */
+var runExtensionSupportedTest = function(gl, name, extensionEnabled) {
```
jdashg: I would sooort of rather we do this test for all extensions in a single test file, rather than spreading this boilerplate out across the extension test files, but centralizing it here at least helps.

kenrussell: Agreed, that would help.

Additionally, in Chromium there have been regressions where certain extensions suddenly stopped being supported. In order to catch those, we had to add tests outside this harness which assert that certain extensions must be supported on certain platforms.

jdashg: FWIW our tests for this are here:
https://dxr.mozilla.org/mozilla-central/source/dom/canvas/test/webgl-mochitest/ensure-exts

PR #2713, opened 2018-09-18
Delete WebGL_blend_equation_advanced proposal

AshleyScirra:
Delete WebGL_blend_equation_advanced in favour of the new WEBGL_blend_equation_advanced_coherent proposal.


AshleyScirra:
Reference: https://github.com/KhronosGroup/WebGL/pull/2709#issuecomment-421439318

kenrussell:
Thanks, looks fine. No need to clutter the "rejected extensions" section.

PR #2714, opened 2018-09-21
Add a test for WEBGL_compressed_texture_bptc

Oletus:
The test is not very thorough yet, as it's not testing for correct
decoding of the compressed data. However, more complete tests will be
easier to develop once there is a prototype implementation available.

The extension is moved to draft so that prototyping can proceed.


kenrussell:
Looks good so far Olli. Merging now.

PR #2715, opened 2018-09-21
Add a test for instanced rendering with large divisors

Oletus:
This test covers a recently fixed ANGLE bug http://anglebug.com/2832
that was present on the D3D11 backend.

### DIFF
--- a/sdk/tests/conformance2/rendering/instanced-rendering-large-divisor.html
+++ b/sdk/tests/conformance2/rendering/instanced-rendering-large-divisor.html
@@ -1,0 +1,165 @@
+﻿<!--
+
+/*
+** Copyright (c) 2018 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+** IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/rendering/instanced-rendering-large-divisor.html
```diff
@@ -0,0 +1,165 @@
...
+<div id="console"></div>
+<script id="outputVertexShader" type="x-shader/x-vertex">#version 300 es
+layout(location = 0) in vec2 a_position;
+layout(location = 1) in float a_positionOffset;
+layout(location = 2) in vec4 a_color;
+out vec4 v_color;
+void main()
+{
+    gl_Position = vec4(a_position.x * 0.05 + a_positionOffset, a_position.y * 0.05, 0.0, 1.0);
+    gl_PointSize = 4.0f;
```
kenrussell: Why does the vertex shader set gl_PointSize when drawing triangles? It's undefined behavior if drawing points and gl_PointSize isn't set – and it's hard to fix this because we didn't want to inject gl_PointSize initialization arbitrarily into vertex shaders. Is it also undefined behavior if drawing triangles and setting gl_PointSize? I thought we ran into issues in this area in the past due to driver bugs - see http://crbug.com/872292 .

sdk/tests/conformance2/rendering/instanced-rendering-large-divisor.html
```diff
@@ -0,0 +1,165 @@
...
+    var program = wtu.setupProgram(gl, ["outputVertexShader", "outputFragmentShader"]);
+
+    gl.clearColor(0.0, 0.0, 1.0, 1.0);
+
+    wtu.setupIndexedQuadWithOptions(gl, {positionLocation: 0});
+
+    var offsetBuf = gl.createBuffer();
+    gl.bindBuffer(gl.ARRAY_BUFFER, offsetBuf);
+    var offsetData = [];
+    for (var i = 0; i < 4; ++i) {
```
kenrussell: A bit confusing that 4 offsets are produced because only the first 2 are consumed by the draw calls below, if I understand the test correctly. Can this be reduced to 2? How about a comment indicating that the last offset is used by the last quad drawn?

sdk/tests/conformance2/rendering/instanced-rendering-large-divisor.html
Oletus: Whoops, I copied this from a slightly different test I made for ANGLE, I'll fix this.

sdk/tests/conformance2/rendering/instanced-rendering-large-divisor.html
Oletus: It's good to have extra offsets since that will produce recognizable artifacts in case the divisor used on the backend is wrong. I can add a comment about this.

kenrussell: Sounds good.

sdk/tests/conformance2/rendering/instanced-rendering-large-divisor.html
kenrussell: Aha, thanks.

PR #2717, opened 2018-09-25
Add EXT_multi_draw_arrays extension proposal.

kenrussell:
Upon request from multiple CAD customers.

### DIFF
--- a/extensions/proposals/EXT_multi_draw_arrays/extension.xml
+++ b/extensions/proposals/EXT_multi_draw_arrays/extension.xml
@@ -38,13 +38,13 @@
 [NoInterfaceObject]
 interface EXT_multi_draw_arrays  {
   void multiDrawArraysEXT(GLenum mode, 
-                          (Int32Array or sequence&lt;GLint&gt;) firstList, GLuint firstOffset,
-                          (Int32Array or sequence&lt;GLsizei&gt;) countList, GLuint countOffset,
+                          (Int32Array or sequence&lt;GLint&gt;) firstsList, GLuint firstsOffset,
+                          (Int32Array or sequence&lt;GLsizei&gt;) countsList, GLuint countsOffset,
                           GLsizei drawcount);
   void multiDrawElementsEXT(GLenum mode,
-                            (Int32Array or sequence&lt;GLint&gt;) countList, GLuint countOffset,
+                            (Int32Array or sequence&lt;GLint&gt;) countsList, GLuint countsOffset,
                             GLenum type,
-                            (Int32Array or sequence&lt;GLsizei&gt;) offsetList, GLuint offsetOffset,
+                            (Int32Array or sequence&lt;GLsizei&gt;) offsetsList, GLuint offsetsOffset,
                             GLsizei drawcount);
 };
   </idl>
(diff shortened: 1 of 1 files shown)


kenrussell:
Please review. @kainino0x @jdashg @Oletus in particular. Thanks.

(This can certainly be fleshed out a bit more, but putting a skeleton proposal in place sooner rather than later.)

Oletus:
I think I'd rather add an entry point that would also allow for instanced attributes - so first, count and instanced attributes could all be different per "instance" drawn. The regular multidraw calls in OpenGL don't allow instanced attributes which significantly reduces their usefulness. It forces the application to bake the transformation into the geometry.

The variant I'm proposing would have to be implemented in software on the backend before going into the driver, but it would still save a lot of command buffer and validation overhead, and this way many more applications could adopt it.

kenrussell:
@Oletus we can certainly add other variants. What would the IDL look like for the variants you propose?

I would like to add these two because they're relatively easy to understand, based on an existing OpenGL extension, and I'm pretty sure will be  useful to the two CAD vendors we're working with. Will confirm with at least one of them. Do you have any objections to adding these?

Oletus:
As discussed in the F2F, I'm fine with the proposal as it stands if we have proof that these do in fact benefit at least some applications. I also found that in GL 4.6 there's a gl_DrawID that's set in shaders when using multi-draw calls and could satisfy the need of per-geometry shader parameters. Is this something that could be added to this extension? The high GL version requirement of course limits its availability, so maybe it would need to be a separate extension or an optional feature.

dmikis:
Since the calls don't rely on lengths of `count` and `firsts` arrays (judging by `drawCount` arguments), maybe we also can add on offset parameter into them which would be 0 by default?

jdashg:
> Since the calls don't rely on lengths of `count` and `firsts` arrays (judging by `drawCount` arguments), maybe we also can add on offset parameter into them which would be 0 by default?
I like this.

jdashg:
@Oletus Good find, but I think we shouldn't tie this extension to GL 4.6, and I think this extension is useful without gl_DrawID.

extensions/proposals/EXT_multi_draw_arrays/extension.xml
```diff
@@ -0,0 +1,102 @@
...
+
+      <feature>The <code>offset</code> arguments to <code>multiDrawArraysEXT</code> and <code>multiDrawElementsEXT</code> choose the starting offset into their respective typed arrays or sequences. This primarily avoids allocation of temporary typed array views.</feature>
+    </features>
+  </overview>
+
+  <idl xml:space="preserve">
+[NoInterfaceObject]
+interface EXT_multi_draw_arrays  {
+  void multiDrawArraysEXT(GLenum mode, 
+                          (Int32Array or sequence&lt;GLint&gt;) firstList, GLuint firstOffset,
```
jdashg: Consider `firstsList`, `firstsOffset`, `countsList`, `countsOffset`, which I think is clearer about them being lists of firsts and counts. `firstOffset` and `countOffset` in particular give the wrong impression.

kenrussell: Thanks, good idea - renamed.

kenrussell:
Merging now. @kainino0x and @austinEng already have some pending changes and a rename for this extension from their prototyping work.

jdashg:
Thanks for writing this proposal.

PR #2718, opened 2018-09-26
Bikeshedding on WEBGL_video_texture

kainino0x:
This is a placeholder PR for bikeshedding on WEBGL_video_texture. I've suggested a rename here, because I think the old one was confusing.

I like something like this because it retains the "texImage2D" heritage. Without it, it's not clear that this behaves much like an upload call. However I'm open to other ideas.

### DIFF
--- a/extensions/proposals/WEBGL_video_texture/extension.xml
+++ b/extensions/proposals/WEBGL_video_texture/extension.xml
@@ -56,7 +56,7 @@ interface WEBGL_video_texture {
     const GLenum TEXTURE_VIDEO_IMAGE             = 0x851D;
     const GLenum SAMPLER_VIDEO_IMAGE             = 0x8B61;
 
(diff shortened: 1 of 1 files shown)


kenrussell:
Good suggestion. I like it, but one concern is that with the texImage2D naming, users may think it acts like a copy. Maybe wording involving "latch video texture" or similar could help.

Ultimately we'll also need to respecify this so that the reference to the video texture is only live during the current requestAnimationFrame callback, or similar.

MarkCallow:
> Without it, it's not clear that this behaves much like an upload call

> Maybe wording involving "latch video texture" or similar could help.

Because of the latch video behavior, this function really does *not* behave much like an upload call.

kainino0x:
With the original name, I was concerned with making sure it doesn't sound like the texture *always* tracks the video, latching rather than always magically "referring to" the most recent frame.

kainino0x:
(But yes, it's important to distinguish it from an "upload"/copy as well.)

PR #2724, opened 2018-10-03
Clarify that all depth resources are cleared to 1.0 by default.

jdashg:
Not just renderbuffers.


kenrussell:
LGTM after the fact - I'm sorry for not reviewing this promptly.

Issue #2670, opened 2018-07-17
 feedback loops with framebuffers where NONE is specified by drawBuffers


greggman:
This part of the spec seems like it might need revisiting?

The issue is if you create a framebuffer, attach 2 or more textures as color attachments, then call `drawBuffers` with either `[gl.NONE, gl.COLOR_ATTACHMENT0 + 1]` or `[gl.COLOR_ATTACHMENT0, gl.NONE]` while reading one of the textures whether or not you should get `INVALID_OPERATION` for a feedback loop as it mentions in [section 6.25 of the spec](https://www.khronos.org/registry/webgl/specs/1.0/#6.25)

There's a test for WebGL1 [here](https://www.khronos.org/registry/webgl/sdk/tests/conformance/extensions/webgl-draw-buffers-feedback-loop.html) and one for WebGL2 [here](https://www.khronos.org/registry/webgl/sdk/tests/conformance2/rendering/rendering-sampling-feedback-loop.html)

Which Firefox passes and Chrome fails and a 2yr old bug filed [here](https://bugs.chromium.org/p/chromium/issues/detail?id=660844)

But it's not clear these tests are correct.

Looking at both the OpenGL ES 2.0 spec and the OpenGL ES 3.0 spec neither say anything about `drawBuffers`. They only say if the texture is attached to the framebuffer and is read in the shader then the results are undefined. They don't stay anything about whether or not setting `drawBuffers` to `NONE` prevents a feedback loop.

This came up from [this SO answer](https://stackoverflow.com/questions/50014874/opengl-and-webgl-sampling-from-a-texture-attached-to-current-framebuffer/51385346#) which links to the [Khronos wiki](https://www.khronos.org/opengl/wiki/Memory_Model#Framebuffer_objects) which suggests that that is still considered a feedback loop and therefore the test is wrong and both Firefox and Chrome are wrong. All 3 cases should generate `INVALID_OPERATION`?

jdashg:
That makes sense but we should stop calling it feedback loops, because
that's not the issue here. In this case it's effectively a simultaneous
usage issue, like our transform feedback woes. I'll reread the spec on this
because I remember being happy with our implementation. (It wouldn't be the
first time the wiki is wrong)

jdashg:
Yep, looks like the deceptive name fooled us. All three subtests in should fail, as drawBuffers has no interaction with "feedback loops".

jdashg:
This is solved by #2727.

PR #2727, opened 2018-10-09
DrawBuffers has no effect on sampling feedback.

jdashg:
Incomplete textures should be ignored by feedback detection also.


jdashg:
This should resolve #2670.

jdashg:
Actually since cubemap attachments must be cube-complete, we should be able to always do this detection properly, even on mipmap-incomplete textures.
That said, it would be easy enough to excuse incomplete textures as impossible "to sample from the texture object T bound to texture unit U", and therefore skipped for the purposes of feedback.

sdk/tests/conformance/rendering/rendering-sampling-feedback-loop.html
```diff
@@ -62,82 +83,102 @@
...
+    }
+
     init();
 
     // The sampling texture is bound to COLOR_ATTACHMENT1 during resource allocation
     allocate_resource();
 
-    rendering_sampling_feedback_loop([gl.NONE, gl.COLOR_ATTACHMENT1], gl.INVALID_OPERATION);
-    rendering_sampling_feedback_loop([gl.COLOR_ATTACHMENT0, gl.COLOR_ATTACHMENT1], gl.INVALID_OPERATION);
-    rendering_sampling_feedback_loop([gl.COLOR_ATTACHMENT0, gl.NONE], gl.NO_ERROR);
```
kainino0x: Is it possible to keep a [gl.COLOR_ATTACHMENT0] test case too? I know it's not supposed to make a difference, but just for coverage.

Or maybe keep all of the original cases (with two attachments), if there's no particular need to simplify.

sdk/tests/conformance/rendering/rendering-sampling-feedback-loop.html
```diff
@@ -62,82 +83,102 @@
...
+        rendering_sampling_feedback_loop([gl.NONE]);
+    }
 }
 
 function init() {
-    program = wtu.setupProgram(gl, ['vshader', 'fshader'], ['aPosition', 'aTexCoord'], [0, 1]);
-    positionLoc = gl.getAttribLocation(program, "aPosition");
-    texCoordLoc = gl.getAttribLocation(program, "aTexCoord");
+    let shaders = ["vs100", "fs100"];
+    if (gl.texStorage2D) {
```
kainino0x: Is this a version check? Don't we have a wtu thing for "is webgl2"? `getDefault3DContextVersion() === 2` I guess

jdashg: Yeah, but it's easier to do it this way, and it always works directly on whatever `gl` is passed in. TBH a gl.version() would be nice to inject when we create a context, but this works.

sdk/tests/conformance/rendering/rendering-sampling-feedback-loop.html
jdashg: Sure.

kenrussell:
LGTM after the fact. Thanks @kainino0x for reviewing this.

PR #2729, opened 2018-10-16
Test readPixels from multisampled framebuffers.

jdarpinian:
The readPixels tests were all being run with {antialias: false}. There's a bug in Qualcomm's Adreno drivers (http://crbug.com/890002) that only manifests with {antialias: true}. This change tests both configurations.


kenrussell:
Nice work James. LGTM

We discussed this offline and the large diff is only a re-indent; also have asked that the crbug link be included in the pull request description. Will merge after that's done.

PR #2730, opened 2018-10-17
*TexSubImage on an undefined texture image is INVALID_OPERATION.

jdashg:
Since this can happen before we check all our enums, this subtest
should accept either error.

### DIFF
--- a/sdk/tests/conformance/textures/misc/compressed-tex-image.html
+++ b/sdk/tests/conformance/textures/misc/compressed-tex-image.html
@@ -58,7 +58,7 @@ if (!gl) {
 
   var tex = gl.createTexture();
   gl.bindTexture(gl.TEXTURE_2D, tex);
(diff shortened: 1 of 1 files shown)


jdashg:
Maybe @kainino0x has a sec to check my work here?

kainino0x:
I think this looks good, but not exactly sure what the error case is. What texture image is undefined here?

jdashg:
@kainino0x The image at level 0 of the texture is undefined:
https://jsfiddle.net/bwr1zofk/2/

Chrome and Firefox both validate it this way, but actually, it looks like the spec allows this. The 'default' tex image is unsized RGBA w/h/d 0/0/0.

kainino0x:
(Is it worth updating snapshots? I think usually we do for cases like this.)

jdashg:
No snapshot update needed, this was introduced post-snapshot in #1320.

kenrussell:
I'm sorry for dropping this on the floor. Looks good after the fact.

Issue #2456, opened 2017-07-13
Rethink object deletion semantics and container objects


kenrussell:
In #2178 the semantics of deleting objects were rethought and clarified. There are still some concerns about the portability of the new semantics, especially around deletion of objects which are still attached to container objects.

Taking Buffers and one of their container objects (VAOs) as an example, the following behavior is definitely portable:

1) If Buffer B1 is attached to VAOs V1 and V2, V1 is currently bound, and B1 is deleted, it should be unbound from V1 but not V2.

2) If V2 is then bound and B1 is deleted again, it should not be unbound from V2.

3) IsBuffer should probably start returning false for B1 as soon as DeleteBuffer is called on B1; see comments on #2178.

greggman:
Will this be back ported to WebGL1? I notice the conformance tests for WebGL2 test this but not for OES_vertex_array_object in WebGL1

RafaelCintron:
I agree with @kenrussell 's concerns as well as those raised by @zhenyao in the comments for #2178.  

As I understand it, @kenrussell is proposing we do the following:
1. WebGL implementations must reference count all resources.  Attaching a resource to another resource will increment the reference count.  Detaching a resource from another resource will decrement the reference count. 
2. Deleting a resource will mark it as “pending for deletion”.   
3. When a resource is pending for deletion, the underlying hardware resource is kept alive so long as the reference count is greater than zero.  Once the reference count reaches zero, the underlying hardware resource is destroyed.
4. A resource continues to function as a valid attachment so long as its reference count is greater than zero, even if it is marked as pending for deletion.
5. A resource that is pending for deletion cannot be added as a new attachment to another resource.
6. A resource will return false to isXXXX when it is marked as pending for deletion, regardless of the existing reference count.
7. When a resource is first marked as pending for deletion, it will detach itself from all resources that are currently bound to the pipeline regardless of the outstanding reference count. This includes: 
a. The active framebuffer
b. The active vertex array object
c. The active program
d. The active transform feedback buffer
e. The texture binding points
8. Once a resource is marked as pending for deletion, subsequent calls to delete it are ignored.  This includes the unbinding behavior described in 7.  For instance, if you delete a buffer, it will become detached for the current vertex array object.  However, if you later attach a different vertex array object that contains the buffer, deleting the buffer a second time will have no effect on the current vertex array object.  

[...]

PR #2731, opened 2018-10-17
DEQP lifetimes test should detach (not delete) buffer from VAO.

jdashg:
The spec used to forbid detaching using vertexAttribPointer.

This matches the behavior from #2456, and replaces #2227.


jdashg:
I updated the 2.0.0 snapshot also. I don't see a way to let this subtest pass either way without degrading coverage for other subtests.
@kenrussell is modifying the 2.0.0 snapshot in this way ok? It should not change your conformance results.

PR #2733, opened 2018-10-23
Split up shaderoperator/common_functions.html.

kenrussell:
This test suite was taking close to the maximum timeout in Chromium's
CI system.

Addresses http://crbug.com/820225 .


kenrussell:
Please review. @kainino0x @jdashg in particular.

Verified this locally by running all of the new HTML files as well as some of the ones that came after these tests and confirming that the expected tests run in the expected HTML files.

sdk/tests/deqp/functional/gles3/es3fShaderOperatorTests.js
```diff
@@ -2663,6 +2664,10 @@ es3fShaderOperatorTests.ShaderOperatorTests.prototype.init = function() {
         all, es3fShaderOperatorTests.unaryGenTypeFuncs(Math.trunc)));
     comm.push(op("round", "round", GT, [v(GT, 2.5, 2.5)], f(0.2), f(0.7),
         all, es3fShaderOperatorTests.unaryGenTypeFuncs(roundToEven)));
+
+    comm = new es3fShaderOperatorTests.BuiltinFuncGroup("common_functions", "Common function tests.");
```
kainino0x: I'm not really familiar with this stuff, but should these be numbered? `_01`, etc

kenrussell: By convention, higher up in this file, they aren't numbered. Doing so changes the dEQP's notion of the test name, and global suppressions wouldn't work any more

kainino0x: Ah, I see. LGTM

PR #2736, opened 2018-10-24
Follow up fixes for lifetime changes from #2731.

jdashg:
Test that isBuffer always returns false after deletion.


jdashg:
#2731 was incomplete, as Firefox's CI was only too happy to tell me!

PR #2741, opened 2018-11-09
Test repeated rendering of WebGL canvas to 2D canvas.

kenrussell:
Each draw should show only the most recently rendered content from the
WebGL canvas. This was not the case when rendering via SwiftShader.

Regression test for http://crbug.com/894021 .


kenrussell:
Merging this now so others can run this test case more easily.

PR #2742, opened 2018-11-10
Clean up unused variable.

kenrussell:
Small follow-up to #2741.


kenrussell:
Tiny change. Merging myself. @kainino0x FYI

PR #2743, opened 2018-11-10
Clear background of 2D canvas as originally intended.

kenrussell:
Another small follow-on to #2741 .


kenrussell:
Small fix; merging myself. @kainino0x @jdashg FYI

PR #2744, opened 2018-11-15
Run canvas-teximage-after-multiple-drawimages test in onload handler.

kenrussell:
The JavaScript execution was racing with document parsing, causing the
test to fail intermittently. Follow-on to #2706.

Chromium bug: http://crbug.com/905682 .


kenrussell:
Any of @jdarpinian @kainino0x @austinEng @jdashg in particular please review. Thanks.

kenrussell:
Thanks Kai.

PR #2750, opened 2018-12-06
Test shadow samplers and TEXTURE_COMPARE_MODE

jdarpinian:
Add checks for shadow samplers and add consideration of normalization for integer textures. The rules are:

Float samplers accept only float textures and normalized integer textures (signed or unsigned).

Signed samplers accept only signed integer textures (normalized or unnormalized).

Unsigned samplers accept only unsigned integer textures (normalized or unnormalized).

Shadow samplers accept only depth textures, and only if their TEXTURE_COMPARE_MODE state is set.

As a special extra restriction, depth textures cannot be sampled by non-shadow samplers if their TEXTURE_COMPARE_MODE state is set. Non-depth textures are not affected by this restriction.

### DIFF
--- a/sdk/tests/conformance2/uniforms/incompatible-texture-type-for-sampler.html
+++ b/sdk/tests/conformance2/uniforms/incompatible-texture-type-for-sampler.html
@@ -65,73 +65,84 @@ function makeFragmentShader(samplerType, uvType) {
    `;
 }
 
+// Sampler types.
+const FLOAT = 1;
+const SIGNED = 2;
+const UNSIGNED = 3;
+const SHADOW = 4;
+
 const textureInternalFormatInfo = {};
 {
   const t = textureInternalFormatInfo;
   // unsized formats
-  t[gl.ALPHA]              = { textureFormat: gl.ALPHA,           colorRenderable: true,  textureFilterable: true,  bytesPerElement: [1, 2, 2, 4],        type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT], };
-  t[gl.LUMINANCE]          = { textureFormat: gl.LUMINANCE,       colorRenderable: true,  textureFilterable: true,  bytesPerElement: [1, 2, 2, 4],        type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT], };
-  t[gl.LUMINANCE_ALPHA]    = { textureFormat: gl.LUMINANCE_ALPHA, colorRenderable: true,  textureFilterable: true,  bytesPerElement: [2, 4, 4, 8],        type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT], };
-  t[gl.RGB]                = { textureFormat: gl.RGB,             colorRenderable: true,  textureFilterable: true,  bytesPerElement: [3, 6, 6, 12, 2],    type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT, gl.UNSIGNED_SHORT_5_6_5], };
-  t[gl.RGBA]               = { textureFormat: gl.RGBA,            colorRenderable: true,  textureFilterable: true,  bytesPerElement: [4, 8, 8, 16, 2, 2], type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT, gl.UNSIGNED_SHORT_4_4_4_4, gl.UNSIGNED_SHORT_5_5_5_1], };
+  t[gl.ALPHA]              = { textureFormat: gl.ALPHA,           samplerType: FLOAT,    depth: false, bytesPerElement: [1, 2, 2, 4],        type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT], };
+  t[gl.LUMINANCE]          = { textureFormat: gl.LUMINANCE,       samplerType: FLOAT,    depth: false, bytesPerElement: [1, 2, 2, 4],        type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT], };
+  t[gl.LUMINANCE_ALPHA]    = { textureFormat: gl.LUMINANCE_ALPHA, samplerType: FLOAT,    depth: false, bytesPerElement: [2, 4, 4, 8],        type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT], };
+  t[gl.RGB]                = { textureFormat: gl.RGB,             samplerType: FLOAT,    depth: false, bytesPerElement: [3, 6, 6, 12, 2],    type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT, gl.UNSIGNED_SHORT_5_6_5], };
+  t[gl.RGBA]               = { textureFormat: gl.RGBA,            samplerType: FLOAT,    depth: false, bytesPerElement: [4, 8, 8, 16, 2, 2], type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT, gl.UNSIGNED_SHORT_4_4_4_4, gl.UNSIGNED_SHORT_5_5_5_1], };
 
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/uniforms/incompatible-texture-type-for-sampler.html
```diff
@@ -69,69 +69,71 @@
...
-  t[gl.LUMINANCE_ALPHA]    = { textureFormat: gl.LUMINANCE_ALPHA, colorRenderable: true,  textureFilterable: true,  bytesPerElement: [2, 4, 4, 8],        type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT], };
-  t[gl.RGB]                = { textureFormat: gl.RGB,             colorRenderable: true,  textureFilterable: true,  bytesPerElement: [3, 6, 6, 12, 2],    type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT, gl.UNSIGNED_SHORT_5_6_5], };
-  t[gl.RGBA]               = { textureFormat: gl.RGBA,            colorRenderable: true,  textureFilterable: true,  bytesPerElement: [4, 8, 8, 16, 2, 2], type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT, gl.UNSIGNED_SHORT_4_4_4_4, gl.UNSIGNED_SHORT_5_5_5_1], };
+  t[gl.ALPHA]              = { textureFormat: gl.ALPHA,            signed: false, int: true,  normalized: true,  depth: false, bytesPerElement: [1, 2, 2, 4],        type: [gl.UNSIGNED_BYTE, gl.HALF_FLOAT, gl.HALF_FLOAT_OES, gl.FLOAT], };
```
jdashg: So all the signed/int/normalized logic /works/, but fwiw this is what Gecko uses:
https://searchfox.org/mozilla-central/rev/adec563403271e78d1a057259b3e17fe557dfd91/dom/canvas/WebGLFormats.h#217
There's really only three categories here, so we might as well just label these FLOAT/INT/UNSIGNED_INT directly.

sdk/tests/conformance2/uniforms/incompatible-texture-type-for-sampler.html
```diff
@@ -69,69 +69,71 @@
...
 
   // Sized Internal
-  t[gl.DEPTH_COMPONENT16]  = { textureFormat: gl.DEPTH_COMPONENT, colorRenderable: true,  textureFilterable: false, bytesPerElement: [2, 4],     type: [gl.UNSIGNED_SHORT, gl.UNSIGNED_INT], };
-  t[gl.DEPTH_COMPONENT24]  = { textureFormat: gl.DEPTH_COMPONENT, colorRenderable: true,  textureFilterable: false, bytesPerElement: [4],        type: [gl.UNSIGNED_INT], };
-  t[gl.DEPTH_COMPONENT32F] = { textureFormat: gl.DEPTH_COMPONENT, colorRenderable: true,  textureFilterable: false, bytesPerElement: [4],        type: [gl.FLOAT], };
-  t[gl.DEPTH24_STENCIL8]   = { textureFormat: gl.DEPTH_STENCIL,   colorRenderable: true,  textureFilterable: false, bytesPerElement: [4],        type: [gl.UNSIGNED_INT_24_8], };
-  t[gl.DEPTH32F_STENCIL8]  = { textureFormat: gl.DEPTH_STENCIL,   colorRenderable: true,  textureFilterable: false, bytesPerElement: [4],        type: [gl.FLOAT_32_UNSIGNED_INT_24_8_REV], };
+  t[gl.DEPTH_COMPONENT16]  = { textureFormat: gl.DEPTH_COMPONENT,  signed: false, int: true,  normalized: true,  depth: true, bytesPerElement: [2, 4],     type: [gl.UNSIGNED_SHORT, gl.UNSIGNED_INT], };
```
jdashg: Also consider removing `foo.depth` in favor of `IsDepth(foo.textureFormat)`.

sdk/tests/conformance2/uniforms/incompatible-texture-type-for-sampler.html
jdarpinian: Sorry, I was confused about normalized textures. I was thinking that they could be sampled as integer too, but they can't. This is much simpler now.

sdk/tests/conformance2/uniforms/incompatible-texture-type-for-sampler.html
jdarpinian: I think I prefer having this information inline instead of in a separate array.

jdarpinian:
Hmm, I just discovered that the Nvidia driver prints a warning when you sample a non-depth texture with TEXTURE_COMPARE_MODE set, claiming this is undefined behavior. I can't find anything in the spec that would make it undefined. Should we forbid it anyway to be safe?

The spec says: If the currently bound texture’s base internal format is DEPTH_COMPONENT or
DEPTH_STENCIL, then DEPTH_TEXTURE_MODE, TEXTURE_COMPARE_MODE and
TEXTURE_COMPARE_FUNC control the output of the texture unit as described below. Otherwise, the texture unit operates in the normal manner and texture comparison is bypassed.

jdashg:
Non-NONE mode with non-shadow/depth is invalid_op in Firefox. I can find
the spec text for you.

jdashg:
Or it might be just incomplete texture.

jdarpinian:
It's not INVALID_OPERATION right now. But I think it should be for consistency, if it is undefined behavior. I would like to see the spec text, but I guess it doesn't matter that much because if at least one driver thinks it's undefined behavior we should probably forbid it anyway. I don't think anybody will mind.

jdashg:
GLES3.0.5 p168:
>  These requests must be consistent; the results of a texture lookup areundefined if:
> •The  sampler  used  in  a  texture  lookup  function  is  not  one  of  the  shadowsampler  types,  the  texture  object’s  internal  format  isDEPTH_COMPONENTorDEPTH_STENCIL, and theTEXTURE_COMPARE_MODEis notNONE.
> •The sampler used in a texture lookup function is one of the shadow sam-pler  types,  the  texture  object’s  internal  format  isDEPTH_COMPONENTorDEPTH_STENCIL, and theTEXTURE_COMPARE_MODEisNONE.
> •The sampler used in a texture lookup function is one of the shadow samplertypes, and the texture object’s internal format is notDEPTH_COMPONENTorDEPTH_STENCIL

I don't love how this logic is laid out, so I prefer to simplify as:
```
is_depth_compare = format.depth_bits && tex.compare_mode
if shadow_sampler:
  is_undef = (format.depth_bits && !tex.compare_mode) || !format.depth_bits
  is_undef = !( !(format.depth_bits && !tex.compare_mode) && format.depth_bits )
  is_undef = !( (!format.depth_bits || tex.compare_mode) && format.depth_bits )
  is_undef = !( (!format.depth_bits && format.depth_bits) || (tex.compare_mode && format.depth_bits) )
  is_undef = !( false || (tex.compare_mode && format.depth_bits) )
  is_undef = !(tex.compare_mode && format.depth_bits)
  is_undef = !is_depth_compare
else:
  is_undef = format.depth_bits && tex.compare_mode
  is_undef = is_depth_compare

is_undef = shadow_sampler != is_depth_compare
```
Firefox generates INVALID_OP here:
https://searchfox.org/mozilla-central/rev/ff46b36ac2ebb243ec95fdab01340391963d62e5/dom/canvas/WebGLContextDraw.cpp#123

The WebGL 2 spec doesn't have text about this, though it falls under the general category of undefined behavior -> INVALID_OP.

jdarpinian:
I'm talking about the case where TEXTURE_COMPARE_MODE is COMPARE_REF_TO_TEXTURE but the texture is not a depth texture and the sampler is not a shadow sampler. None of those three undefined behavior conditions from the spec apply. Current Firefox Nightly does not generate INVALID_OPERATION in this case.

jdashg:
Seems clear to me that's allowed in ES3:
> (p163) If  the  currently  bound  texture’s  base  internal  format  isDEPTH_COMPONENTorDEPTH_STENCIL,  thenTEXTURE_COMPARE_MODEandTEXTURE_COMPARE_-FUNCcontrol  the  output  of  the  texture  unit  as  described  below.   Otherwise,  thetexture unit operates in the normal manner and texture comparison is bypassed.

jdashg:
GL4.5core has the same text on p274.

jdashg:
My read is that the NV driver is mistaken if it thinks that this situation is undefined.

jdarpinian:
Yes, I agree. But if drivers think it's undefined anyway, we might want to forbid it. I don't think anyone would mind the restriction. WDYT?

jdashg:
I think we should handle that in another issue/pr. This here is good to go in, if you want to leave `depth:false` inline instead of IsDepthFormat().

jdashg:
I can just merge this. It's just a test, so code style isn't a big sticking point.

PR #2751, opened 2018-12-07
Color mask must be preserved during implicit clears.

kenrussell:
Catches a bug that manifested on Chrome on macOS, but could have
affected any other platform. This new test already passes on Firefox.

Regression test for http://crbug.com/911918 .

### DIFF
--- a/sdk/tests/conformance/rendering/color-mask-preserved-during-implicit-clears.html
+++ b/sdk/tests/conformance/rendering/color-mask-preserved-during-implicit-clears.html
@@ -80,7 +80,8 @@ function runTest() {
   gl.bindFramebuffer(gl.FRAMEBUFFER, null);
   // Set clear color to transparent green.
   gl.clearColor(0, 1, 0, 0);
-  // Clear. Result should be opaque green.
+  // Clear. Result should be opaque green. Was transparent green when
+  // bug was encountered.
   gl.clear(gl.COLOR_BUFFER_BIT);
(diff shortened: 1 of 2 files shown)


kenrussell:
@kainino0x @jdarpinian in particular please review.

Any other comments welcome.

Note that this test case is as reduced as it can be; eliminating any of the calls causes the bug to stop reproducing.

sdk/tests/conformance/rendering/color-mask-preserved-during-implicit-clears.html
```diff
@@ -0,0 +1,108 @@
...
+  // latched into the internal caching code.
+  gl.clear(gl.COLOR_BUFFER_BIT);
+
+  // Switch back to default framebuffer.
+  gl.bindFramebuffer(gl.FRAMEBUFFER, null);
+  // Set clear color to transparent green.
+  gl.clearColor(0, 1, 0, 0);
+  // Clear. Result should be opaque green.
+  gl.clear(gl.COLOR_BUFFER_BIT);
+  wtu.checkCanvasRect(gl, 0, 0, sz, sz,
```
jdarpinian: Maybe add a comment saying what the values were when the bug was hit?

kenrussell: Thanks, good suggestion. Done.

kainino0x:
It's the same one.

kenrussell:
+1, same bug. No driver bug here, just a bug in Chrome.

PR #2752, opened 2018-12-09
Update multiview transform feedback validation.

null77:
This updates the CTS to be in line with the recent extension wording
change. It allows multiview rendering when Transform Feedback is bound
and paused.

See http://anglebug.com/3012

@kenrussell @jdashg @kainino0x  PTAL

Note anglebug.com links are temporarily broken but should be fixed at some point. Full issue here:

https://bugs.chromium.org/p/angleproject/issues/detail?id=3012

Also was having trouble testing this with Canary. It seems to ignore when ANGLE exposes the multiview extension. Not sure how to test this locally.


jdashg:
Thanks!

kenrussell:
LGTM after the fact.

PR #2755, opened 2018-12-12
Test uploading texture from animating transparent WebM

kainino0x:
Regression test for https://crbug.com/912842 , where we improperly blended when reusing a cached buffer, causing successive frames of a transparent video to be blended together instead of replaced.

Passes in Firefox.


sdk/tests/conformance/textures/misc/texture-video-transparent.html
```diff
@@ -0,0 +1,166 @@
...
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
+
+    gl.enable(gl.BLEND);
+    gl.blendFunc(gl.SRC_ALPHA, gl.ONE_MINUS_SRC_ALPHA);
+    gl.clearColor(1, 0, 1, 1);
+    gl.clear(gl.COLOR_BUFFER_BIT);
+    // Point the uniform sampler to texture unit 0
+    gl.uniform1i(textureLoc, 0);
```
jdashg: Since you never change this, you can do this immediately after linkProgram+useProgram, and you'll no longer need this to be a global var.

sdk/tests/conformance/textures/misc/texture-video-transparent.html
```diff
@@ -0,0 +1,166 @@
...
+    const texture = gl.createTexture();
+    // Bind the texture to texture unit 0
+    gl.bindTexture(gl.TEXTURE_2D, texture);
+    // Upload the videoElement into the texture
+    if (useTexSubImage2D) {
+        // Initialize the texture to black first
+        gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA,
+                      videoElement.videoWidth, videoElement.videoHeight, 0,
+                      gl.RGBA, gl.UNSIGNED_BYTE, null);
+        gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGBA, gl.UNSIGNED_BYTE, videoElement);
```
jdashg: I wonder if videoWidth and videoHeight are always going to be what you want here. You might just call texImage2D(Video)+texSubImage(Video) just to be sure we're getting that right.

jdashg: Also, will this video always have data after body.onload? There's no chance it's still waiting on data for the first frame? Can we assert that we haveData? (or whatever that attribute is? I think Ken knows)

kainino0x: This runs after wtu.startPlayingAndWaitForVideo above.

kenrussell: Right, startPlayingAndWaitForVideo should handle this.

kainino0x: Copied this from texture-npot-video.html, but I don't particularly care about the texSubImage version here, so: Done.

kainino0x: Whoops, wait, it's the texSubImage version that I DO care about - texImage doesn't seem to repro (at least on Linux). I'd like to leave both here. (This behavior is the same as texture-npot-video.html anyway.)

kainino0x: Oh, I see what you were suggesting. That DOES repro the issue. Done. (sorry for churn)

sdk/tests/conformance/textures/misc/texture-video-transparent.html
```diff
@@ -0,0 +1,160 @@
...
+            wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be no errors");
+            finishTest();
+        }
+    });
+
+}
+
+function runOneIteration(videoElement, useTexSubImage2D)
+{
+    const texture = gl.createTexture();
```
kenrussell: Just noticed this is leaking a texture per iteration. Could you fix that?

kainino0x: Oops, my bad, I noticed this at some point but it slipped my mind. Just moved this up into init.

PR #2756, opened 2018-12-14
Add more variants of implicit-clear test.

kenrussell:
Test against all variants of antialias and premultipliedAlpha.

Follow-on to #2751. Addresses additional scenarios from
http://crbug.com/913033 .


kenrussell:
On Chrome without the bug fix, all four variants fail; with the bug fix, all four variants pass.

kenrussell:
Merging now to make progress - will gladly address any further comments in follow-ons.

PR #2699, opened 2018-08-24
Add references to OffscreenCanvas.

kenrussell:
OffscreenCanvas is documented in the WHATWG's current specifications,
and browsers are beginning to ship support for fetching WebGL contexts
against them, as well as use them as texture sources. Document this
properly in the WebGL specification.

In Chromium, this work was tracked under issue http://crbug.com/563816
and many dependent bugs.

PR #2753, opened 2018-12-11
Add "desynchronized" flag to WebGLContextAttributes.

kenrussell:
Developers of stylus-based drawing applications (e.g. Google Keep)
have found this hint to be critical in order to be competitive with
native applications. It has been prototyped in Chromium and yields
significant latency improvements there.

This is the WebGL analogue of https://github.com/whatwg/html/pull/4360
and partially addresses https://github.com/whatwg/html/issues/4087 .

Principally authored by @yellowdoge, Miguel Casas-Sanchez <mcasas@chromium.org>

PR #2757, opened 2018-12-15
Re-run IDL generator after OffscreenCanvas changes.

kenrussell:
Neglected to do this in #2699.


kenrussell:
@jdashg please review. This addresses one concern in #2753.

jdashg:
Thanks!

PR #2758, opened 2018-12-15
COMPARE_MODE is required to use shadow samplers

jdarpinian:
This test uses shadow samplers, so it is required to set TEXTURE_COMPARE_MODE to avoid INVALID_OPERATION. We can set it unconditionally because it is ignored for non-depth textures.


jdashg:
Yup, thanks!

PR #2759, opened 2018-12-18
Change WebGLContextAttributes to boolean from GLboolean.

kenrussell:
Based on review feedback from #2753.


kenrussell:
@jdashg in particular please review. Any other comments welcome.

PR #2764, opened 2019-01-05
Add deleted object test; strengthen transform feedback test.

kenrussell:
Framebuffers and programs are the primary container objects in the
WebGL 1.0 specification. Test deleting attachments of both bound and
unbound framebuffers. Deleting programs and shaders is well tested
elsewhere.

Strengthen the WebGL 2.0-specific transform feedback test to cover
deleting output buffers, as well as active transform feedback objects.

Vertex Array Objects (VAOs), the last container type, are already well
tested by conformance/extensions/oes-vertex-array-object.html and
conformance2/vertex_arrays/vertex-array-object.html .

Regression test for http://crbug.com/739604 .

### DIFF
--- a/sdk/tests/conformance2/transform_feedback/transform_feedback.html
+++ b/sdk/tests/conformance2/transform_feedback/transform_feedback.html
@@ -67,6 +67,7 @@ var gl = wtu.create3DContext(canvas, null, 2);
 var tf = null;
 var tf1 = null;
 var buf = null;
+let out_add_buffer = null;
 var program = null;
 var activeInfo = null;
 var query = null;
@@ -81,8 +82,11 @@ if (!gl) {
     runTFBufferBindingTest();
     runObjectTest();
     runGetBufferSubDataTest();
-    runOneOutFeedbackTest();
-    runTwoOutFeedbackTest();
+    runUnboundDeleteTest();
+    runBoundDeleteTest();
+    runOneOutputFeedbackTest();
+    // Must be the last test, since it's asynchronous and calls finishTest().
+    runTwoOutputFeedbackTest();
 }
 
 function runBindingTest() {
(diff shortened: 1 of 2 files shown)


kenrussell:
@kainino0x @jdashg in particular please review.

These new and strengthened tests catch several bugs in Chromium which are resolved by the forthcoming fix for http://crbug.com/739604 . All of the tests already pass in Firefox Nightly.

Any other comments welcome. @jdarpinian @grorg @RafaelCintron

sdk/tests/conformance2/transform_feedback/transform_feedback.html
```diff
@@ -324,6 +346,80 @@
...
+    debug("isBuffer should report false after deletion");
+    shouldBe("gl.isBuffer(out_add_buffer)", "false");
+
+    gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, tf);
+    debug("Transform feedback object should keep output buffer alive");
+    shouldBe("gl.getIndexedParameter(gl.TRANSFORM_FEEDBACK_BUFFER_BINDING, 0)", "out_add_buffer");
+    // Deleting the buffer again while the transform feedback is bound shouldn't unbind it.
+    gl.deleteBuffer(out_add_buffer);
+    debug("Deleting output buffer again should be a no-op");
+    shouldBe("gl.getIndexedParameter(gl.TRANSFORM_FEEDBACK_BUFFER_BINDING, 0)", "out_add_buffer");
```
kainino0x: Maybe test unbinding and rebinding the transform feedback object as well? Just to make sure that unbinding the transform feedback object doesn't somehow propagate to the buffer being unbound.

kenrussell: Thank you, good suggestion. Added that test here.

kenrussell:
Thanks for your review @kainino0x . Merging now.

PR #2766, opened 2019-01-16
Add framebuffer tests to oes-texture-float

natashalee:
Add framebuffer tests to oes-texture-float, copied test format from oes-texture-half-float.
The oes-texture-half-float pr: https://github.com/KhronosGroup/WebGL/pull/1622.


jdashg:
It would be nice to share (most of) this code between the tests, but this is ok.

jdashg:
Thanks!

PR #2768, opened 2019-01-18
Fix webgl2 support for oes-texture-float-linear test.

jdashg:


kenrussell:
Nice work clearing this up; looks good.

PR #2774, opened 2019-01-22
Skip texture_functions.texturelodoffset.sampler3d_float_vertex

jdarpinian:
It is skipped in native dEQP. See:
https://android.googlesource.com/platform/external/deqp/+/40ff528%5E%21/
https://bugs.chromium.org/p/angleproject/issues/detail?id=3094


kenrussell:
Since this seems to be failing on lots of devices, it is appropriate to skip it in the WebGL 2.0 conformance suite.

PR #2778, opened 2019-01-24
Test for A8-on-R8 swizzle being applied to the wrong tex unit.

jdashg:
Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1503813

### DIFF
--- a/sdk/tests/conformance/textures/misc/texture-active-bind.html
+++ b/sdk/tests/conformance/textures/misc/texture-active-bind.html
@@ -50,6 +50,20 @@ void main()
   texCoord = texCoord0;
 }
 </script>
+<script id="trivial-vs" type="text/plain">
+void main()
+{
+  gl_Position = vec4(0,0,0,1);
+}
+</script>
+<script id="tex-point-fs" type="text/plain">
+precision mediump float;
+uniform sampler2D uSampler;
+void main()
+{
+  gl_FragColor = texture2D(uSampler, vec2(0));
+}
+</script>
 <script>
 "use strict";
 var gl;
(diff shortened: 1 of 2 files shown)


lexaknyazev:
Is it expected that Chrome 72 doesn't pass this test? Namely:
```
Tests that glActiveTexture and glBindTexture work as expectedSpecifically texture targets are per active texture unit.

On success, you will see a series of "PASS" messages, followed by "TEST COMPLETE".

PASS getError was expected value: NO_ERROR : 
PASS getError was expected value: NO_ERROR : 
PASS getError was expected value: NO_ERROR : 
PASS getError was expected value: NO_ERROR : 
PASS getError was expected value: NO_ERROR : 
PASS should be 0,192,128,255
PASS should be 128,64,255,255
PASS should be 192,255,64,255
PASS should be 200,0,255,255

Swizzle applied to correct ActiveTexture index
PASS getError was expected value: NO_ERROR : 
FAIL window.control should be 10,20,30,40. (was 0,192,128,255)
FAIL window.after_swizzle_applied should be 10,20,30,40. (was 0,192,128,255)
PASS successfullyParsed is true

TEST COMPLETE
```

It seems that `drawArrays` here has no effect on the subsequent `readPixels` call.
https://github.com/KhronosGroup/WebGL/blob/c2419a295f3e4299eb9ade4c845344f8636b0edd/sdk/tests/conformance/textures/misc/texture-active-bind.html#L168-L174

jdarpinian:
We have a bug filed for this already: http://crbug.com/931006

lexaknyazev:
@jdarpinian Thanks for the link!
Further investigation shows that Chrome doesn't render points unless vertex shader sets `gl_PointSize`.
So, this test should be fixed here (#2819), and the spec should probably be more strict (#2818).

PR #2779, opened 2019-01-24
Add basic test for EXT_float_blend.

jdashg:
Also test that float32 blending is forbidden without EXT_float_blend.


kenrussell:
Before we merge this can we backport EXT_float_blend to WebGL 1.0 as you suggested? We should at least have an escape hatch in case this regresses WebGL 1.0 content. Thanks.

PR #2780, opened 2019-01-25
Require each draw buffer with an attachment has a matching frag outpu…

jdashg:
…t type.

### DIFF
--- a/conformance-suites/1.0.3/conformance/extensions/webgl-draw-buffers.html
+++ b/conformance-suites/1.0.3/conformance/extensions/webgl-draw-buffers.html
@@ -561,6 +561,12 @@ function runDrawTests() {
   gl.bindFramebuffer(gl.FRAMEBUFFER, fb);
   ext.drawBuffersWEBGL(bufs);
   gl.useProgram(redProgram);
+
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no errors");
+  wtu.drawUnitQuad(gl);
+  gl.getError(); // Consume possible error.
+
+  ext.drawBuffersWEBGL([gl.COLOR_ATTACHMENT0]);
   wtu.drawUnitQuad(gl);
 
   checkAttachmentsForColorFn(attachments, function(attachment, index) {
--- a/conformance-suites/2.0.0/conformance2/rendering/draw-buffers.html
+++ b/conformance-suites/2.0.0/conformance2/rendering/draw-buffers.html
@@ -380,6 +380,12 @@ function runDrawTests() {
   gl.bindFramebuffer(gl.FRAMEBUFFER, fb);
   gl.drawBuffers(bufs);
   gl.useProgram(blueProgramESSL1);
+
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no errors");
+  wtu.drawUnitQuad(gl);
+  gl.getError(); // Consume possible error.
+
+  gl.drawBuffers([gl.COLOR_ATTACHMENT0]);
   wtu.drawUnitQuad(gl);
 
   checkAttachmentsForColorFn(attachments, function(attachment, index) {
@@ -392,6 +398,12 @@ function runDrawTests() {
   gl.bindFramebuffer(gl.FRAMEBUFFER, fb);
   gl.drawBuffers(bufs);
   gl.useProgram(redProgram);
+
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no errors");
+  wtu.drawUnitQuad(gl);
+  gl.getError(); // Consume possible error.
+
+  gl.drawBuffers([gl.COLOR_ATTACHMENT0]);
   wtu.drawUnitQuad(gl);
 
   checkAttachmentsForColorFn(attachments, function(attachment, index) {
--- a/extensions/WEBGL_draw_buffers/extension.xml
+++ b/extensions/WEBGL_draw_buffers/extension.xml
@@ -61,7 +61,7 @@
         </ul>
       </addendum>
       <addendum>
-        Although the extension name is prefixed with WEBGL the extension must be enabled with the 
+        Although the extension name is prefixed with WEBGL the extension must be enabled with the
         <code>#extension GL_EXT_draw_buffers</code> directive, as shown in the sample code, to use
         the extension in a shader.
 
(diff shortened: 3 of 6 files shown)


jdashg:
cc @null77

jdashg:
This would supplant #2462.

jdashg:
This change adds a new error case, and so is not strictly backwards-compatible with the current spec. It's possible that this change will break content, and I will be trialing this change in Firefox Nightly and checking for bug reports.

kenrussell:
@jdashg thanks for putting this together. Would it be okay if we wait to merge this until we hear the results of your trial in Firefox? Or would you prefer to merge it now?

@jdarpinian opinions on this? This will change the WebGL 2.0 conformance target a bit more. We should file a Chrome implementation bug about tracking this change.

jdashg:
We can delay this for a few weeks as we wait for any web-compat issues.

jdashg:
FWIW, this is blocking Firefox on 1.0.3 conformance.

jdashg:
This has landed in Nightly:
https://bugzilla.mozilla.org/show_bug.cgi?id=1290774

kenrussell:
Great. Please tell us what you find / hear back from users.

jdashg:
@kenrussell Please approve/merge when ready!

kenrussell:
Per discussion in the WebGL working group, we believe that this change will not affect a significant amount of content - if any content at all - and providing the guarantee that unwritten fragment shader outputs are untouched turns out to be both fragile and expensive on some platforms. For this reason WebGL will surface this area of undefined behavior in the OpenGL specification as a draw-time error.

Thanks @jdashg for driving this investigation.

Heads up @vonture @null77 @jdarpinian @kainino0x @austinEng @shrekshao in particular that we'll need to suppress the failure of this test and make the associated validation changes as quickly as possible in ANGLE and Chromium to stay conformant.

null77:
Awesome. Should simplify ANGLE's life.

shrekshao:
Hi. 
Background: I'm working on adapting this change for ANGLE.
I'd like to bring this up as it seems to have some ambiguity in how this should be expected to work with WebGL 1: 
```glsl
#extension GL_EXT_draw_buffers : require
precision highp float;
void main()
{
    gl_FragData[1] = vec4(0.0, 1.0, 0.0, 1.0);
    gl_FragData[3] = vec4(0.0, 1.0, 0.0, 1.0);
}
```
Given this shader file with GL_EXT_draw_buffers required, cur ANGLE implementation turns on all 8 fragment output variable at once. Even only channel [1] and [3] has value gets written to.

While on WebGL 2
```glsl
#version 300 es
precision highp float;
layout(location = 1) out vec4 output1;
layout(location = 3) out vec4 output2;
void main()
{
    output1 = vec4(0.0, 1.0, 0.0, 1.0);
    output2 = vec4(0.0, 1.0, 0.0, 1.0);
}
```
Only channel 1 and 3 is turned on for `mActiveOutputVariables` mask for the shader (as specified in the shader)

Not sure if this inconsistent behavior for WebGL 1 and 2 is expected. Though seems like this defined here: https://github.com/jdashg/WebGL/blob/b93ed5f35d58fe85b4314c8cca1287b79c498a60/extensions/WEBGL_draw_buffers/extension.xml#L87
Some clarification would still be great.

jdashg:
It sounds like you understand it correctly. The perceived ambiguity is not
between WebGL 1 and 2, but rather between arrays and non-arrays: `
layout(location
= 2) out vec4 colors[3];` is valid ESSL3. We expect the restrictions to
match for output arrays for ESSL1+exts and ESSL3: The whole array is
defined and given output types, so these output types must match any
attached draw-buffer attachment, regardless of static use of individual
elements in the arrays.

shrekshao:
@jdashg 
Another case would like you to help clarify.
Would a fragment shader with no output consider valid when there's active render target?
e.g. 
```glsl
void main(){}
```
It seems firefox has updated to fit the change of this spec, and consider an empty shader with active draw buffer to be a valid case. (https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/conformance2/buffers/get-buffer-sub-data-validity.html#L49)

CC: @null77

jdashg:
An empty essl1 shader effectively implicitly declares `(location = 0) out
lowp vec4 gl_FragColor;`, and is validated as such.

null77:
@jdashg can you point to where this implicit declaration happens in ANGLE or Firefox? And any spec related reasons why this should happen? I looked around but wasn't able to find it in the ANGLE translator or WebGL spec. But I could have missed it.

jdashg:
ESSL1 p61:
The built-in variables that are accessible from a fragment shader are
intrinsically given types as follows:
mediump vec4 gl_FragCoord;
bool gl_FrontFacing;
mediump vec4 gl_FragColor;
mediump vec2 gl_PointCoord;

Based on static use, gl_FragColor xor gl_FragData is implicitly
declared, is how I think about it.

In the draw_buffers extension, we basically define gl_FragColor to be
an alias of gl_FragData[0], since we forbid broadcast.

null77:
I'm not sure I agree with that reading. You could say gl_FragColor xor gl_FragData are implicitly declared based on static use. But if there is no static use I don't see why it follows that gl_FragColor is implicitly declared as an alias of gl_FragData[0]. At least I don't see where it the spec it would imply that. Rather if a shader doesn't write to either it seems like a compiler could reasonably optimize out both. This is a valid use case for programs that want to write to the depth buffer. Thus a color attachment in this case should be an error.

Also, have we discussed requiring an attachment for all shader outputs? Or does the current spec just validate that there are no dangling unwritten attachments.

shrekshao:
https://www.khronos.org/files/opengles_shading_language.pdf section 7.2 p60
> It is not a requirement for the fragment shader to write to either gl_FragColor or gl_FragData. There are
many algorithms, such as shadow volumes, that include rendering passes where a color value is not
written.

I think this statement supports @null77's argument. @jdashg

jdashg:
We don't require an attachment for all shader outputs, just that there are no dangling unwritten attachments.

I suppose if neither FragColor or FragData are written, we can have a shader with no outputs. (we can definitely have an essl3 shader with no outputs, so we have to handle that anyway!)
In this case, there are never any dangling unwritten attachments, so a no-output-program would always pass that phase of validation.

null77:
I think we should enforce there are no dangling attachments in this case. It make the spec consistent and slightly optimizes validation. Do you agree?

jdashg:
Dangling attachments are attachments that:
* Are declared in the fragment shader
* FB attachment is non-null
* glDrawBuffers has not masked out the attachment with NONE

So there can't be dangling attachments in the no-outputs case, at least according to current WebGL spec.

I don't want to decide anything else here, hidden in this issue. Please file a new issue.

null77:
I'm sorry but this affects ANGLE's implementation so it is germane to this issue.

> Dangling attachments are attachments that:
> 
> Are declared in the fragment shader
> FB attachment is non-null
> glDrawBuffers has not masked out the attachment with NONE

Maybe you made a typo somewhere. If the FB attachment is non-null and the output is declared in the FS there doesn't seem to be anything dangling.

jdashg:
Yeah, you're totally right. 

Dangling attachments are attachments that:
* Attached to a fragment output location with no static use in the shader
* FB attachment is non-null
* glDrawBuffers has not masked out the attachment with NONE

This would include COLOR_ATTACHMENT0 with a frag shader that doesn't use any outputs. Presently, the WebGL 1.0 spec requires that this leaves COLOR_ATTACHMENT0's data unaffected. Changing that is more likely to break content, but I would like to try it.

null77:
Jeff, one more question. When you say 

> Presently, the WebGL 1.0 spec requires that this leaves COLOR_ATTACHMENT0's data unaffected. Changing that is more likely to break content, but I would like to try it.

Were you going to try making this an error in a Firefox nightly build?

jdashg:
Yes.

shrekshao:
Thanks for the clarification. (Though I'm expecting quite some conformance tests need updating)

shrekshao:
There's a recent discussion by the skia team. Seems that we didn't mention whether it's still an invalid operation when glColorMask is set to all false here.
I didn't find any discussion or sentences from spec declaring this active drawbufer and frag output mismatch is regardless of mask settings yet. (Let me know if there is).
However I do remember we do have tests indicating feedback loop detection is regardless of mask settings.
So intuitively I'd say we should not take into account the mask setting here as well. Maybe there's more solid reasons for not considering mask setting in driver level as it's a fine buffer control and there are evidence showing that even if it's set to fully false there's still samples combined (OpenGL ES 3 spec section 4.2.2.1)
If we all agree on this I can update the tests with some glColorMask call following the pattern of those in feedback loop tests, showing they should not affect anything here.

jdashg:
I would like us to not have to check ColorMask, but I could go either way on it. Clearly if ColorMask is disabled, no color outputs will be written. Can you spawn another issue for this?

PR #2781, opened 2019-01-25
Stencil mismatch is only an error if there are active stencil bits.

jdashg:
The 1.0.3 version of webgl-specific.html should request stencil:true, and enable(STENCIL_TEST).


jdashg:
@kenrussell Can we merge this?

PR #2783, opened 2019-01-28
Fix XML error in EXT_draw_buffers_indexed

lexaknyazev:
#2777 had an error in XML. This PR fixes that.
/cc @jdashg


jdashg:
Thanks!

Issue #2772, opened 2019-01-21
[parallel_shader_compile] Number of threads should not be specified by the Web author


litherum:
https://www.khronos.org/registry/webgl/extensions/KHR_parallel_shader_compile/

```idl
void maxShaderCompilerThreadsKHR(GLuint count);
```
This function should be removed from the extension.

- I can’t imagine any application that would choose to limit itself; Applications often don’t like opting in to slower performance
- There is no way for a Web application to know what a good value to supply is
- Implementations often handle this kind of resource allocation and scheduling deep within the system, depending on things like system load, etc.
- Existing implementations so far seem to ignore this value.

litherum:
Similarly, this call in the example code in the extension should be removed:

```JavaScript
gl.getParameter(ext.MAX_SHADER_COMPILER_THREADS_KHR);
```

kenrussell:
Fixed by #2787 .

PR #2787, opened 2019-01-30
Remove maxShaderCompilerThreadsKHR (#2772)

jchen10:


jchen10:
@kenrussell PTAL

jdashg:
This is what I remember our decision being. Thanks for making the change!

kenrussell:
Yes, this is what was agreed on. Thanks @jchen10 for taking care of this!

jchen10:
Thank you all reviewing!

PR #2792, opened 2019-02-03
Add &quick=1 option for deqp/functional/gles3/fbocolorbuffer.

jdashg:


jdashg:
I'm looking at increasing how often I run the DEQP tests, but they take so long!

Thoroughness is appreciated for conformance purposes, but I would love to get 95% of the coverage with <30% of the running time.

Before this change it took 109s, but after this change reducing the canvas and texture sizes:
deqp/functional/gles3/fbocolorbuffer (Passed: 213/213 in 19.32 seconds)

@kenrussell What do you think?

PR #2796, opened 2019-02-05
Remove skipCorner logic from texImage *_integer-unsigned_byte tests

kainino0x:
Seems to no longer be necessary to skip these cases/checks!
Tested on 10.13.6 (Chrome, Firefox) and 10.12.6 (Chrome).

crbug.com/665197


kenrussell:
Great work Kai! LGTM after the fact.

Note that this is related to #1819 which I'll close now.

PR #2797, opened 2019-02-05
Disable incorrect builtinprecision tests

jdarpinian:
The builtinprecision tests are not correct. Errors were introduced when porting the tests from C++ to JavaScript. Tests that fail here pass in the C++ version of dEQP ported to WASM. This change disables the tests until they can be fixed.


kenrussell:
Thanks @jdarpinian for confirming this. Looks good to disable these.

PR #2799, opened 2019-02-07
glBindBufferRange with size param less than 4 generates INVALID_VALUE

austinEng:
According to section 2.12.6.3 Uniform Buffer Object Storage of
the OpenGL ES 3.0 spec, the smallest element of a uniform block,
a bool, is still read as a typed uint (4 bytes).

Binding a buffer range of size less than 4, since it is not large enough
to even fill the smallest possible uniform block, would generate
INVALID_OPERATION at draw time. It should be reasonable to add
validation that the size param is not less than 4.

Note that it is impossible to have a uniform block size that is not
a multiple of 4. Regardless, the spec allows calling
glBindBufferRange with a size that is not a multiple of 4 (You
can have an 11-byte buffer range for a 8-byte uniform block).

We're requesting this change because non-4-byte-aligned sizes
are permitted by the spec but produce INVALID_VALUE on some
drivers. To workaround this, size can be rounded down to the
nearest multiple of 4 so the driver does not generate an error.

This does not work in the few cases where rounding down a value
less than 4 to the nearest multiple of 4 produces 0. It is an error
to pass 0 to glBindBufferRange. Specing that size must be >=4
makes it safe to always round `size` down.


jdashg:
Why can't you just not call into the driver in that case? You'll never issue draw-calls that are valid with this binding.

austinEng:
Looked at this again and realized you are correct that we do not have to call into the driver. We have all the necessary state tracking to do validation and make queries correct as-if we did call the driver.

PR #2802, opened 2019-02-09
Add default transform feedback object test

jdashg:
Taking over stalled #2348.


jdashg:
@greggman Let me know if you'd rather fix up your PR instead!

PR #2804, opened 2019-02-10
Add quick mode support to more tests.

jdashg:
Also spew time elapsed when finishTest() is called from deqp if not in
quiet mode.


jdashg:
If these look reasonable, I'll add them to ToT as well.

PR #2805, opened 2019-02-11
Specify DOM image conversion

### LINKED ISSUE #2789: Tighten DOM uploads from PNG
PNG images can contain various combinations of channels and bit depths. The WebGL 2.0 spec should disallow certain ambiguous cases or clearly describe the behavior. Namely:

### Components
Texels in PNG image can be:
- Grayscale (one component)
- Grayscale with alpha (two components)
- RGB (three components)
- RGBA (four components)

#### Expectations:
- Grayscale
  - fills R as R001, RG as RR01, RGB as RRR1, and LUMINANCE as RRR1;
  - fills RGBA and LUMINANCE_ALPHA as RRR1;
  - fills ALPHA as 0001.
- Grayscale with alpha
  - fills R as R001, RG as RR01, RGB as RRR1, and LUMINANCE as RRR1; source alpha values are lost;
  - fills RGBA and LUMINANCE_ALPHA as RRRA;
  - fills ALPHA as 000A.
- RGB
  - fills R as R001, RG as RG01, RGB as RGB1, and RGBA as RGBA;
  - fills LUMINANCE and LUMINANCE_ALPHA as RRR1;
  - fills ALPHA as 0001.
- RGBA
  - fills R as R001, RG as RG01, RGB as RGB1, and RGBA as RGBA;
  - fills LUMINANCE as RRR1 and LUMINANCE_ALPHA as RRRA;
  - fills ALPHA as 000A.

### Bit Depth
PNG files can have up to 16 bits per channel (in all four modes above). In addition to that, PNG files can define how many [_significant_ bits](https://www.w3.org/TR/2003/REC-PNG-20031110/#11sBIT) each channel has. Here are some implications of this for GPU uploads.

#### Floating point
```
R32F/RG32F/RGB32F/RGBA32F
R16F/RG16F/RGB16F/RGBA16F
R11F_G11F_B10F
RGB9_E5
```
These should have no ambiguities with taking normalized floating point (`0.0 .. 1.0`) values from PNG decoder. Conformance tests should ensure expected target precision.

#### Unsigned integers
```
R8UI/RG8UI/RGB8UI/RGBA8UI
RGB10_A2UI
R16UI/RG16UI/RGB16UI/RGBA16UI
R32UI/RG32UI/RGB32UI/RGBA32UI
```
A WebGL implementation could fill GPU texels with source integer values (multiplied by bit-mask based on the number of significant bits per channel), if PNG decoder provides such information. Otherwise, all uploads to unsigned integers (including 8 bit) should be disallowed.

#### Normalized 8-bit
```
LUMINANCE/ALPHA/LUMINANCE_ALPHA
R8/RG8/RGB8/RGBA8
SRGB8/SRGB8_ALPHA8
```
There is a clear fast path for source data with 8 bits in all channels. Other cases may require renormalization.

#### Normalized 10-bit
```
RGB10_A2
```
There could be a fast path for PNG files with the same per-channel bit depth (10-10-10-2). Otherwise, texels need to be carefully converted to maintain precision.

### Color space
PNG files can define:
- [Primary chromaticities and white point](https://www.w3.org/TR/2003/REC-PNG-20031110/#11cHRM)
- [Image gamma](https://www.w3.org/TR/2003/REC-PNG-20031110/#11gAMA)
- [Embedded ICC profile](https://www.w3.org/TR/2003/REC-PNG-20031110/#11iCCP)
- [sRGB Rendering Intent](https://www.w3.org/TR/2003/REC-PNG-20031110/#11sRGB)

When `UNPACK_COLORSPACE_CONVERSION_WEBGL` is set to `NONE`, these four PNG chunks should be ignored.

---

Which of statements above (assuming they are accurate) require spec or conformance suite updates?

lexaknyazev:
This PR describes how DOM images are converted on texture uploads. It seems that implementations already follow this.

Partially resolves issue #2789.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -1706,9 +1706,14 @@ WebGL2RenderingContext includes WebGL2RenderingContextBase;
             <th>RG</th>
           </tr>
           <tr>
-            <td>Grayscale (1 channel)<br>Grayscale + Alpha (2 channels)</td>
+            <td>Grayscale (1 channel)</td>
             <td>R = <em>sourceGray</em><br></td>
-            <td>R = <em>sourceGray</em><br>G = <em>sourceGray</em></td>
+            <td>R = <em>sourceGray</em><br>G = 0</td>
+          </tr>
+          <tr>
+            <td>Grayscale + Alpha (2 channels)</td>
+            <td>R = <em>sourceGray</em><br></td>
+            <td>R = <em>sourceGray</em><br>G = <em>sourceAlpha</em></td>
           </tr>
           <tr>
             <td>Color (3 channels)<br>Color + Alpha (4 channels)</td>
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -1695,6 +1695,27 @@ WebGL2RenderingContext includes WebGL2RenderingContextBase;
       </dt>
       <dd>
         <p>Only differences from <a href="../1.0/index.html#TEXIMAGE2D_HTML">texImage2D</a> in WebGL 1.0 are described here. </p>
+        <p>Conversion to new formats introduced in WebGL 2.0 is performed according to the following table.</p>
+        <table class="foo">
+          <tr>
+            <th rowspan="2">Source DOM Image Format</th>
+            <th colspan="2">Target WebGL Format</th>
+          </tr>
+          <tr>
+            <th>RED</th>
+            <th>RG</th>
+          </tr>
+          <tr>
+            <td>Grayscale (1 channel)<br>Grayscale + Alpha (2 channels)</td>
+            <td>R = <em>sourceGray</em><br></td>
+            <td>R = <em>sourceGray</em><br>G = <em>sourceGray</em></td>
+          </tr>
+          <tr>
+            <td>Color (3 channels)<br>Color + Alpha (4 channels)</td>
+            <td>R = <em>sourceRed</em><br></td>
+            <td>R = <em>sourceRed</em><br>G = <em>sourceGreen</em><br></td>
+          </tr>
+        </table><br>
         <p>Uploading subregions of elements is detailed in <a href="#DOM_UPLOAD_UNPACK_PARAMS">Pixel
            store parameters for uploads from <code>TexImageSource</code></a>.
         <p>If a WebGLBuffer is bound to the <code>PIXEL_UNPACK_BUFFER</code> target, generates an <code>INVALID_OPERATION</code> error.</p>
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -2602,8 +2602,55 @@ WebGLRenderingContext includes WebGLRenderingContextBase;
 
             The source image data is conceptually first converted to the data type and format
             specified by the <em>format</em> and <em>type</em> arguments, and then transferred to
-            the WebGL implementation. If a packed pixel format is specified which would imply loss
-            of bits of precision from the image data, this loss of precision must occur. <br><br>
+            the WebGL implementation. Format conversion is performed according to the following table.
+            If a packed pixel format is specified which would imply loss of bits of precision from the
+            image data, this loss of precision must occur. <br><br>
+
+            <table class="foo">
+              <tr>
+                <th rowspan="2">Source DOM Image Format</th>
+                <th colspan="5">Target WebGL Format</th>
+              </tr>
+              <tr>
+                <th>ALPHA</th>
+                <th>RGB</th>
+                <th>RGBA</th>
+                <th>LUMINANCE</th>
+                <th>LUMINANCE_ALPHA</th>
+              </tr>
+              <tr>
+                <td>Grayscale (1 channel)</td>
+                <td>A = 255 (1.0)</td>
+                <td>R = <em>sourceGray</em><br>G = <em>sourceGray</em><br>B = <em>sourceGray</em><br></td>
+                <td>R = <em>sourceGray</em><br>G = <em>sourceGray</em><br>B = <em>sourceGray</em><br>A = 255 (1.0)</td>
+                <td>L = <em>sourceGray</em></td>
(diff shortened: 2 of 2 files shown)


kenrussell:
Thanks @lexaknyazev for putting this together.

It looks fine to me and I'm quite sure these conversions are already well tested by the conformance suite. Any other feedback from implementers or people interested in the test suite?

@jdashg @grorg @RafaelCintron @kainino0x @jdarpinian @austinEng @greggman

specs/latest/2.0/index.html
```diff
@@ -1695,6 +1695,27 @@ <h4>Texture objects</h4>
...
+            <th colspan="2">Target WebGL Format</th>
+          </tr>
+          <tr>
+            <th>RED</th>
+            <th>RG</th>
+          </tr>
+          <tr>
+            <td>Grayscale (1 channel)<br>Grayscale + Alpha (2 channels)</td>
+            <td>R = <em>sourceGray</em><br></td>
+            <td>R = <em>sourceGray</em><br>G = <em>sourceGray</em></td>
```
jdashg: G should be sourceAlpha. Don't be too literal about "RG", it's just a 2-channel format. :)

lexaknyazev: If that's the intended behavior, then browsers do it wrong.
Looks like they expand grayscale source to color and then sample R and G from RGB image.

jdashg: I think we should do the thing I said, and I don't want to make the spec match behavior that we might not want.
Hopefully we'll decide on a direction on the conference call tomorrow.

lexaknyazev: Just for reference, here's [Chromium DOM-to-RG8 conversion](https://github.com/chromium/chromium/blob/db29a96e938ab285182c7da5c8b4bd4a9b3c3934/third_party/blink/renderer/platform/graphics/gpu/webgl_image_conversion.cc#L1678). 
At that point, `source[]` has 4 components regardless of original image mode.

Firefox behaves the same way. A simple repro:
```html
<!DOCTYPE html>
<html>
<body onload="init()">
<img src="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAADklEQVR4AQEDAPz/AP8AAgEBAIlSUwAAAAAASUVORK5CYII=" style="display: none", id='img'>
<canvas id="cnv" width="64" height="64" style="border: 1px solid black"></canvas>
<script>
    function init() {
      var cnv = document.getElementById('cnv');
      var img = document.getElementById('img');
      var gl = cnv.getContext('webgl2', {antialias: false});

      var tex = gl.createTexture();
      gl.texImage2D(gl.TEXTURE_2D, 0, gl.RG8, 1, 1, 0, gl.RG, gl.UNSIGNED_BYTE, img);

      var fb = gl.createFramebuffer();
      gl.bindFramebuffer(gl.READ_FRAMEBUFFER, fb);
      gl.framebufferTexture2D(gl.READ_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, tex, 0);
      gl.readBuffer(gl.COLOR_ATTACHMENT0);

      gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, null);
      gl.drawBuffers([gl.BACK]);

      gl.blitFramebuffer(0, 0, 1, 1, 0, 0, 64, 64, gl.COLOR_BUFFER_BIT, gl.NEAREST);
    }
</script>
</body>
</html>
```
Supplied image contains just one pixel with Gray=255 and Alpha=0. Both Chrome and Firefox render it as yellow.

lexaknyazev: @jdashg Has the WG been able to decide on a direction here?

Looking at this issue again, I don't think that treating Grayscale+Alpha images just as 2-channel format is consistent with RGB and RGBA mappings in WebGL 1.0 implementations.

Compare:
- to-RGB: R=G=B=_Gray_, _Alpha_ is lost
- to-RGBA: R=G=B=_Gray_, A=_Alpha_

lexaknyazev: > Is there such a thing as e.g. a red-green PNG image?

No. PNG modes are RGB, RGBA, Grayscale, GrayscaleAlpha. JPEG modes are RGB and Grayscale.

> The rationale is that otherwise the two-channel image formats are fairly useless.

For red-green textures one could use RGB images with zeros in blue channel. They can be uploaded to gl.RG targets just fine.

> Do you have any feedback on this proposal?

I cannot think of a content pipeline that would put the second color channel into alpha. Also, it's not clear how premultiplication flag should affect such uploads.

jdashg:
Sometimes images aren't pictures, they're just data. Using a two-channel
PNG is primarily LA, but since WebGL 2 doesn't have LA formats, RG best
preserves the two channels.

Since the premultiplied unpack param is for unpack, it applies to the data
in LA form from the image, and may premult/depremult before upload.

lexaknyazev:
OK. So the WG decision is to make unpacking LA PNG to gl.RG target a special case, right? If so, I'll update the PR.

jdashg:
Please do!

lexaknyazev:
What should green channel (in gl.RG) be when the source is grayscale-only: gray or zero?

jdashg:
I'd love to forbid channel count changes, tbh, in part because of these ambiguities. I vote RG=[R,0].

lexaknyazev:
PTAL

jdashg:
Thanks!

kenrussell:
Thanks for this update. Ideally conformance tests verifying the new behavior would have been submitted at the same time as this specification change.

@lexaknyazev will you be able to adjust the WebGL 2.0 conformance tests verifying the new behavior, referencing #2789? It would be a great help. Of course browsers won't be passing the new tests until their implementations are changed, but we need to have a target to reach.

Thanks.

lexaknyazev:
@kenrussell 
I've opened #2814 with a few questions re conformance tests. Will gladly propose testing suite update once WebGL WG clarifies those.

PR #2806, opened 2019-02-12
integer-cubemap-specification-order-bug.html: add more test cases

kainino0x:
Makes this test more likely to catch the intended driver bug in the presence of attempted workarounds. (This caught that a few of my attempts at workarounds were not good enough. The eventual fix is (probably) to just preallocate POSITIVE_X, level 0 before everything else.)

These tests were generated by shuffling the initialization order
thousands of times and picking specifically the cases that failed
on the affected driver.

https://crbug.com/905003


jdashg:
Crazy, thanks!

kainino0x:
Considered adding more test cases here, to wriggle around more workaround attempts. But after thinking about this more, it's just too hard to even work around. I'm going to merge these just to make the test a little more robust, but I've sent this bug to NVIDIA (again) so hopefully they can take care of it in their driver instead.

Issue #2770, opened 2019-01-19
gl_VertexID appears to be wrong in ANGLE if offset parameter to gl.drawArrays is non-zero


greggman:
Run this sample http://jsfiddle.net/greggman/fLabtv81/7/

What is the expected output? 

<img width="312" alt="screen shot 2019-01-20 at 5 14 14" src="https://user-images.githubusercontent.com/234804/51431885-65b94180-1c72-11e9-9c89-46e010ceeb08.png">

two sets of 5 red dots. The first set starting from the center of the canvas. The second set starting from 75% across the canvas.

What do you see instead?

<img width="317" alt="screen shot 2019-01-20 at 5 14 38" src="https://user-images.githubusercontent.com/234804/51431889-6c47b900-1c72-11e9-8e7b-4dc378a87d1c.png">

In ANGLE/DirectX both sets of 5 dots start in the center.

Note: Running on Windows 10 with --use-angle=gl works correctly. it also works correctly on MacOS nvidia or intel and on a Pixel2 Android.

Filed a bug on ANGLE here

https://bugs.chromium.org/p/angleproject/issues/detail?id=3090

Probably needs a conformance test.

jdashg:
Huh, I didn't realize that was the behavior! Thanks for finding this.

jdashg:
Test added in #2807.

PR #2807, opened 2019-02-12
Add test for gl_VertexID.

jdashg:
In response to #2770, tests drawArrays, drawElements, and drawArrays+transform-feedback.

### DIFF
--- a/sdk/tests/conformance2/rendering/vertex-id.html
+++ b/sdk/tests/conformance2/rendering/vertex-id.html
@@ -1,0 +1,239 @@
+﻿<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL 2 gl_VertexID Tests</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/desktop-gl-constants.js"></script>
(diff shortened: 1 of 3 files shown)


jdashg:
FWIW, this fails in Firefox+Windows+ANGLE, but mostly works for Firefox+Windows+GL. The drawElements tests fail even with GL, but I think that's a Firefox bug.

sdk/tests/conformance2/rendering/vertex-id.html
```diff
@@ -0,0 +1,239 @@
...
+</script>
+<script src="../../js/js-test-post.js"></script>
+
+</body>
+</html>
+
+<!--
+
+/*
+** Copyright (c) 2013 The Khronos Group Inc.
```
kenrussell: Please move closer to the top of the file and update the copyright date.

jdashg: Let's talk briefly about this on the concall tomorrow.

jdashg:
Decision from the concall was to tentatively allow this, but try to move towards a smaller copyright block that does not incur charset reparse so easily. (Point people at a COPYRIGHT.txt?)

PR #2808, opened 2019-02-14
Add serve_localhost.py for more robust local test runs.

jdashg:
This invokes a modified basic http.server that adds anti-caching
headers. Otherwise, browsers cache test files, making testing and
debugging less reliable.

### DIFF
--- a/serve_localhost.py
+++ b/serve_localhost.py
@@ -1,0 +1,38 @@
+#! /usr/bin/env python3
+
+# Invoke http.server to host a basic webserver on localhost /without/ caching.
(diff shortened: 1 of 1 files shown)


jdashg:
Also beneficial is that this defaults to running on localhost, instead of allowing remote hosts to make requests.

kainino0x:
+1 to having a script for this. However, we've seen a lot of unreliability with SimpleHTTPServer/http.server, where it will hang in some cases (when it receives too many requests at the same time?). These might have been fixed, but I seem to recall that even pretty recent versions of Python had these bugs. I think this is the bug, and it's marked as resolved: https://bugs.python.org/issue31639

PR #2815, opened 2019-03-01
add more ext-float-blend tests

shrekshao:
Added more ext-float-blend tests, similar to those new tests added to angle.
- testExtFloatBlendMRT
- testExtFloatBlendNonFloat32Type

Follow up work for this one: https://github.com/khronosgroup/webgl/pull/2779

### DIFF
--- a/sdk/tests/js/tests/ext-float-blend.js
+++ b/sdk/tests/js/tests/ext-float-blend.js
@@ -19,7 +19,7 @@ void main()
 }
 `;
 const trivialFsMrtSrc = `#version 300 es
-precision lowp float;
+precision mediump float;
 layout(location = 0) out vec4 o_color0;
 layout(location = 1) out vec4 o_color1;
 void main()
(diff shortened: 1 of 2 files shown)


sdk/tests/js/tests/ext-float-blend.js
```diff
@@ -12,6 +12,22 @@ void main()
...
 }
 `;
+const trivialVsMrtSrc = `#version 300 es
+void main()
+{
+    gl_Position = vec4(0,0,0,1);
+}
+`;
+const trivialFsMrtSrc = `#version 300 es
+precision lowp float;
```
jdashg: mediump unless you have a reason not to. (You probably don't even need it, since trivialFsSrc doesn't)

shrekshao: Changed to `mediump` (Removing doesn't work, I think it's because the explicit out vec4?)

jdashg:
Cool, works for me. Please sign the CLA and I'll merge this.

jdashg:
Oh! Just CI we're waiting for, my bad!

Issue #2473, opened 2017-08-10
Consider exposing EXT_pvrtc_sRGB to WebGL


kainino0x:
Consider exposing [EXT_pvrtc_sRGB](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_pvrtc_sRGB.txt) to WebGL.

Discussion here:
https://www.khronos.org/webgl/public-mailing-list/public_webgl/1707/msg00027.php
> iOS 7.0 devices, including PowerVR SGX 543 and 554 supports sRGB PVRTC throught EXT_pvrtc_sRGB  extensions.

msc-:
> iOS 7.0 devices, including PowerVR SGX 543 and 554 supports sRGB PVRTC throught EXT_pvrtc_sRGB extensions.

How many of those also support ASTC sRGB?

jdashg:
@grorg @kainino0x @kenrussell 
Is there any will to support these?

kainino0x:
If there is customer demand, sure. However this would really only benefit iOS and the tiny number of PowerVR Android devices out there: http://opengles.gpuinfo.org/listreports.php?extension=GL_EXT_pvrtc_sRGB

And given that iOS has deprecated OpenGL ES, it's hard to say how long that last (or how reliably it works).

MarkCallow:
PVRTC is supported in Metal, see, e.g., MTLPixelFormatPVRTC_RGBA_2BPP, so there is no reason to expect it to disappear from iOS devices or become unreliable just because they've deprecated OpenGL.

My question about ASTC remains unanswered. Recent iOS devices all support it so there is really no need to use PVRTC.

kainino0x:
According to [this impossible-to-read PDF](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf), the devices with PVRTC but not ASTC are `iOS_GPUFamily1_v[1-5]`.

According to [this page](https://developer.apple.com/documentation/metal/mtlfeatureset?language=objc) that means Apple A7 only (and according to [this page](https://developer.apple.com/library/archive/documentation/DeviceInformation/Reference/iOSDeviceCompatibility/HardwareGPUInformation/HardwareGPUInformation.html#//apple_ref/doc/uid/TP40013599-CH106-SW13), that's iPhone 5s, iPad Air, iPad mini 3, iPad mini 2). Everything newer has both.

jdashg:
[...]

PR #2816, opened 2019-03-04
Reject PVRTC extension in favor of ASTC, ETC, and S3TC.

kdashg:


jdashg:
We recently passed on creating an extension for PVRTC+sRGB in #2473, so rejecting PVRTC as a whole might make sense.

MarkCallow:
I'm not sure any iOS devices support S3TC so we need to be cautious about suggesting that as an alternative. My not too ancient iPad Pro does not. I suppose expiration of the S3 patent might have changed things though.

jdashg:
Generally iOS devices should have at least ETC, though sufficiently old devices might not have even that, I suppose.

jdashg:
It looks like iOS Webkit has not yet implemented either ETC extension:
https://webglstats.com/webgl/extension/WEBGL_compressed_texture_etc?platforms=000000840000000000

lexaknyazev:
Compressed texture support on a hardware side:
- pre-A7 (before iPhone 5S): PVRTC only
- A7: PVRTC & ETC 1/2
- A8 and newer: PVRTC, ETC 1/2, ASTC

For some reasons, iOS WebKit exposes only PVRTC extension.

kenrussell:
Until we hear approval from Apple and until they implement WebGL's ETC compressed texture extension, this should be left as community approved to avoid invalidating their implementation.

jdashg:
Since the jury's still out, this should stay open until we have more information.

kenrussell:
After discussion with folks on Slack (@kkinnunen-apple and @lexaknyazev ), the preference was to call out this extension as deprecated, rather than rejecting it. This was done in #3398 .

Issue #2818, opened 2019-03-05
Require gl_PointSize for points rendering or define a default value


lexaknyazev:
When a vertex shader doesn't set `gl_PointSize` value (while drawing using `gl.POINTS` mode), the behavior is undefined by OpenGL ES:

GLSL ES 1.0 (section 7.1):
> `mediump float gl_PointSize; // may be written to`
> If any of these variables are not written to, their values are undefined. 

GLSL ES 3.0 (section 7.1):
> If `gl_PointSize` is not written to, its value is undefined in subsequent pipe stages.

Chrome and Firefox show different behavior: Chrome doesn't render anything while Firefox seems to default to `1.0`. In spirit of #2780, the WebGL spec should be adjusted.

jdashg:
The easiest thing to do is set it to a sane default like 1.0, like gl_Position.

However, it's useless to set this if not rasterizing points, such as with TRIANGLES, or even POINTS with RASTERIZER_DISCARD. There's probably not much of a perf overhead here, and we can elide this write as an optimization in implementations if this does become a problem.

We could also do the more complicated thing, where we forbid draw calls that use gl_PointSize but never write to it. (draw(POINTS) without RASTERIZER_DISCARD)

kenrussell:
I recall that the Chrome team ran into this issue a long time ago, and attempted to solve it by injecting code into the vertex shader which always initialized gl_PointSize to 1.0. I remember that this was not as easy as it sounded, and that on some platforms, this broke rendering of non-point geometry. Again, it's been a long time and I don't remember the details. We might be able to dig them up in the Chrome issue tracker.

The most reliable solution is to dynamically recompile the current program if POINTS are drawn and it doesn't statically initialize gl_PointSize. In Chrome (and Firefox to the best of my knowledge) the only platform on which shaders are dynamically recompiled is on Windows, in ANGLE's D3D backend. We don't want to introduce such complexity on the other platforms.

I'm concerned that as we are attempting to take another snapshot of the conformance suite, and finally have all browsers reach conformance again, that we keep moving the bar farther and farther out.

This is one area where I would strongly prefer to leave the current undefined behavior and change the conformance test to not hit this undefined behavior.

kenrussell:
If ANGLE on Windows is really the only problematic platform then I support fixing it and adding this test. Filed https://bugs.chromium.org/p/angleproject/issues/detail?id=3236 for comment by the ANGLE team. @vonture @null77

PR #2819, opened 2019-03-05
Fix A8-on-R8 swizzle test on Chrome

lexaknyazev:
Fixes this test on Chrome due to #2818.

### DIFF
--- a/sdk/tests/conformance/textures/misc/texture-active-bind.html
+++ b/sdk/tests/conformance/textures/misc/texture-active-bind.html
@@ -54,6 +54,7 @@ void main()
 void main()
 {
   gl_Position = vec4(0,0,0,1);
+  gl_PointSize = 1.0;
 }
 </script>
 <script id="tex-point-fs" type="text/plain">


lexaknyazev:
I think that this particular test (active vs bound texture) shouldn't depend on an unrelated implementation issue that could be resolved separately.

lexaknyazev:
> `gl_PointSize` should always be being clamped to within the point size range, which does not allow zero.

GLES doesn't suggest that (from 3.0 section 3.4):
> If the value written to `gl_PointSize` is less than or equal to zero, or if no value is written, the point size is undefined.

jdashg:
To ensure safe behavior, browsers must ensure that gl_PointSize is clamped to within its valid values, even if otherwise undefined.


The ES3 spec says two things here:
```

jdashg:
I filed #2822 to add an explicit test for this.
Curiously, with transform-feedback enabled, chrome passes this test.
This sounds a lot like Chrome is being over-eager to skip draw calls.

jdashg:
Let's talk about this in that PR, but I generally don't like working around implementation bugs in unrelated tests.

jdashg:
To keep coverage, I would like to resolve #2822 before merging this.

kenrussell:
This test was not intended to test the undefined behavior of failing to write gl_PointSize in a vertex shader used to draw points. Let's continue to discuss that in #2818 and #2822 . Merging this now.

Issue #2813, opened 2019-02-28
Various tests and demos checked into the conformance tests are not themselves conformant


greggman:
These tests seem to have issues

https://www.khronos.org/registry/webgl/sdk/tests/conformance/more/

For example this test

https://www.khronos.org/registry/webgl/sdk/tests/conformance/more/performance/CPUvsGPU.html

gets this error

> Error: Failed to compile shader hblur-frag, Shader info log: ERROR: 0:11: '[]' : array constructor supported in GLSL ES 3.00 and above only

this test

https://www.khronos.org/registry/webgl/sdk/tests/conformance/more/performance/bandwidth.html

> TypeError: Failed to execute 'texImage2D' on 'WebGLRenderingContext': parameter 9 is not of type 'ArrayBufferView'.

this test

https://www.khronos.org/registry/webgl/sdk/tests/conformance/more/glsl/longLoops.html

lots of shader errors

Same with this demo get shader errors

https://www.khronos.org/registry/webgl/sdk/tests/conformance/more/demos/opengl_web.html

This demo doesn't run, gets JavaScript errors

https://www.khronos.org/registry/webgl/sdk/tests/conformance/more/demos/video.html

kenrussell:
These were contributed a long time ago and since they're unmaintained should just be deleted.

kenrussell:
@greggman any help you could offer in deleting these broken tests/demos out of sdk/tests/conformance/more/ would be appreciated. Thanks.

PR #2820, opened 2019-03-05
remove unused parts of more tests

greggman:
See https://github.com/KhronosGroup/WebGL/issues/2813

I didn't remove everything, only the parts that were not being used


kenrussell:
Thanks for cleaning these up Gregg.

PR #2824, opened 2019-03-11
Test webgl canvases as well in projlod test

mysteryDate:
glsl-function-texture2dprojlod.html tests only use a 2D canvas as a source. Expand the test to use webgl backed canvases as well.


kenrussell:
Just waiting for the Travis CI build to complete.

Issue #2827, opened 2019-03-14
Lack of an easy way to get the specification of WebGL 2 compute


Richard-Yunchao:
Thanks to your review and the discussions about the specification of WebGL 2 compute. Now it was merged at https://github.com/KhronosGroup/WebGL/pull/2769. The specification is based on bikeshed. But currently we can't generate the html file automatically on github. As a result, if web developers want to read the specification, there is no simple way. They need to 
1) download (git clone) the project https://github.com/KhronosGroup/WebGL to their local machine 
2) enter to file folder specs/latest/2.0-compute in their local copy
3) run "make"
4) then get the html file and read the specification

If web developers can visit the specification like WebGL and WebGL 2 specification via a link ( https://www.khronos.org/registry/webgl/specs/latest/1.0 or https://www.khronos.org/registry/webgl/specs/latest/2.0/) directly. It's much more convenient for them to try WebGL 2 compute (compute shader on web).  

Considering that it will take some time to make this project automatically generate html file via bikeshed. So I propose that we upload a html version of the specification in this project. If we can auto-gen  the html file for the specification via bikeshed in future, we can remove the html file at that time. 

The disadvantage is that we can't revise the html file if we want to make a change for WebGL 2 compute specification. Instead, we should revise .bs file and run "make", and let it generate the html file. Then upload both .bs file and .html file. But I think it's not a big problem. Because only a few people will change the specification, and we can take care of this rule. And the reviewers like @kenrussell can also keep this in mind before such a change is merged. 

WDYT? 

FYI, @kenrussell , @gyagp , @qjia7

kainino0x:
extensions/ has a Makefile that I'm pretty sure is automatically run when updates to this git repository are being rolled out to khronos.org. It seems like it should be easy to do the same thing for this spec.

Richard-Yunchao:
@kainino0x , Yes. files under extensions/ are based on a template written particularly for extensions in Khronos WebGL. It has its own rule to write extensions in xml format, Then you can run "make" to generate html file in you local machine, and it can also auto-gen html file online. But I don't know how to enable this function. Kai, do you know how to do this? can you help on this?

MarkCallow:
It is possible to have the CI update the repos's GitHub pages. See  the `deploy:` section in https://github.com/KhronosGroup/KTX-Specification/blob/master/.travis.yml for an example.

Unfortunately you can only easily have one version in GitHub pages. Perhaps ToT version is appropriate there.

[...]

PR #2828, opened 2019-03-14
Add the html version for WebGL 2 compute specification

Richard-Yunchao:
Please refer to the discussions at https://github.com/KhronosGroup/WebGL/issues/2827.


kenrussell:
Let's hold off on this until we hear back from Khronos' webmaster about just rebuilding this from the Bikeshed source on Khronos' servers, as is done for the extension repository.

Richard-Yunchao:
Close this pr because we can auto-gen the html version for WebGL 2 compute specification.

PR #2829, opened 2019-03-16
 get-buffer-sub-data-validity.html: asyncify + regression test for crbug.com/941930

kainino0x:
This adds a minimal regression test for crbug.com/941930, where several things were being incorrectly managed in the nonblocking readback path in Chrome's passthrough command decoder.

It also `async`ifies this test to make it more readable.

I recommend reviewing the two commits separately - the first should make no functional changes.

### DIFF
--- a/sdk/tests/conformance2/buffers/get-buffer-sub-data-validity.html
+++ b/sdk/tests/conformance2/buffers/get-buffer-sub-data-validity.html
@@ -89,23 +89,22 @@ function resolvable() {
     return promise;
 }
 
-function fence() {
-    const promise = resolvable();
+function wait() {
+    return new Promise(res => {
+        setTimeout(res, 0);
+    });
+}
 
+async function fence() {
     const sync = gl.fenceSync(gl.SYNC_GPU_COMMANDS_COMPLETE, 0);
     gl.flush();
-    function check() {
-        const status = gl.clientWaitSync(sync, 0, 0);
-        if (status == gl.ALREADY_SIGNALED || status == gl.CONDITION_SATISFIED) {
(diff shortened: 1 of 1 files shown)


kainino0x:
cc @vonture

sdk/tests/conformance2/buffers/get-buffer-sub-data-validity.html
```diff
@@ -239,6 +239,18 @@
...
+    {
+        debug("");
+        debug("write-delete-recreate-fence-wait-read");
+        gl.copyBufferSubData(gl.COPY_READ_BUFFER, gl.COPY_WRITE_BUFFER, 0, 0, 8);
+        deleteReadbackBuffer();
+        recreateReadbackBuffer();
+        await fence();
+        checkGetBufferSubData(gl.NO_ERROR, "noData");
+        deleteReadbackBuffer();
+    }
```
kenrussell: Any chance we want to add a stress test which does this multiple times? Does this reliably reproduce the failure the customer found?

kainino0x: It's reliable for me. Although it could pass erroneously if the client/service ids line up in the right way.

I added one more test that should (I think) catch it a little more reliably (or at least differently).

kenrussell: Looks great.

kenrussell:
Merging without squashing in order to retain the async-ifying of the test separately.

vonture:
Awesome, thanks for adding the tests.

PR #2830, opened 2019-03-16
Implicitly enable EXT_float_blend and promote it to Community Approved.

jdashg:
Update tests to require implicit activation accordingly.

### DIFF
--- a/extensions/EXT_float_blend/extension.xml
+++ b/extensions/EXT_float_blend/extension.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<draft href="EXT_float_blend/">
+<extension href="EXT_float_blend/">
   <name>EXT_float_blend</name>
 
   <contact> <a href="https://www.khronos.org/webgl/public-mailing-list">WebGL
@@ -34,6 +34,20 @@
         WEBGL_color_buffer_float</a> extensions must be enabled.</p>
       </feature>
     </features>
+    <features>
+      On supported systems, this extension is implicitly enabled when any of the
+      following extensions are enabled:
+      <ul>
+        <li><a href="http://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float">
+        EXT_color_buffer_float</a>
+        <li><a href="http://www.khronos.org/registry/webgl/extensions/OES_texture_float">
+        OES_texture_float</a>
+        <li><a href="http://www.khronos.org/registry/webgl/extensions/WEBGL_color_buffer_float">
+        WEBGL_color_buffer_float</a>
+      </ul>
+      <p>
+      This preserves functionality for content written before this extension was exposed.
+    </features>
   </overview>
 
   <idl xml:space="preserve">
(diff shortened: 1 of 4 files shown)


MarkCallow:
This is going to get us into the same situation we got into with float rendering where apps will rely on the implicit enabling and then fail on platforms where this the feature is no available. In this case, that is every mobile platforms. 

Why are we doing this? Since it is easy to update web applications why are we so keen not to break apps. They'll break anyway when run on mobile. We should force the authors to fix them by not implicitly enabling.

jdashg:
We can't update apps that are no longer maintained, so we do this dance where we enable it implicitly, but should warn in the console when it's used implicitly, and use *that* to convince authors to improve the portability of their content.

jdashg:
@kenrussell PTAL
I want to take this change in the tail end of Firefox 67 (which is merging to Beta soon) since we started doing the strict validation here in 67.

sdk/tests/conformance2/extensions/ext-float-blend.html
```diff
@@ -28,10 +28,10 @@
     }
 
     debug("");
-    debug("Testing that float32 blending is forbidden without EXT_float_blend");
-    testExtFloatBlend(false, gl.RGBA32F);
-    testExtFloatBlendMRT(false);
-    testExtFloatBlendNonFloat32Type(false);
+    debug("Testing that float32 blending without EXT_float_blend");
```
kainino0x: "that float32 blending" what?

sdk/tests/js/tests/ext-float-blend.js
```diff
@@ -29,7 +29,9 @@ void main()
 }
 `;
 
-function testExtFloatBlend(shouldBlend, internalFormat) {
+function testExtFloatBlend(internalFormat) {
+    const shouldBlend = gl.getSupportedExtensions().indexOf('EXT_float_blend') != -1;
```
kainino0x: These mean the browser MUST expose EXT_float_blend if it supports float blend, right?

extensions/EXT_float_blend/extension.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<draft href="EXT_float_blend/">
+<extension href="EXT_float_blend/">
```
kainino0x: Shouldn't there be a public_webgl post about this first?

jdashg: I can post about it, but I don't think there's an ordering requirement.

sdk/tests/js/tests/ext-float-blend.js
jdashg: That's right.

sdk/tests/conformance2/extensions/ext-float-blend.html
jdashg: oops, s/that //

sdk/tests/js/tests/ext-float-blend.js
kainino0x: That's very strict, but it is nice. Seems ok with me.

kenrussell: SGTM too.

jdashg:
Thanks!

PR #2832, opened 2019-03-18
Renaming WEBGL_multiview to OVR_multiview2 spec and test changes

mingyuhu:
Removed all references to opaque frame buffer. Removed support of WebGL, only WebGL2 2DTextureArrays are supported.

Function rename: framebufferTextureMultiviewWEBGL -> framebufferTextureMultiviewOVR
extension directive rename: GL_OVR_multiview -> GL_OVR_multiview2

Renamed all test files from webgl_multiview... to ovr_multiview2...

### DIFF
--- a/extensions/OVR_multiview2/extension.xml
+++ b/extensions/OVR_multiview2/extension.xml
@@ -1,22 +1,23 @@
 <?xml version="1.0"?>
 
-<draft href="WEBGL_multiview/">
-  <name>WEBGL_multiview</name>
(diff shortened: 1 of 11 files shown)


jdashg:
Thanks for doing this!

kenrussell:
Thanks for this contribution.

A heads up that this broke the test suite. It was also necessary to update 00_test_list.txt in the same directory. This was fixed in #2836 .

PR #2833, opened 2019-03-18
Fix XML for EXT_float_blend extension.

kenrussell:
Prevented building of the extension registry. Follow-up to #2830.


kenrussell:
@jdashg FYI. Merging myself.

jdashg:
Oops, thanks!

PR #2835, opened 2019-03-22
update test to sync up with spec update #2780

shrekshao:
https://github.com/KhronosGroup/WebGL/pull/2780

Active draw buffers should have its corresponding outputs from fragment shader.


kenrussell:
LGTM after the fact.

Are there positive tests for these scenarios? Looks like conformance2/rendering/draw-buffers.html likely contains them?

PR #2180, opened 2016-11-30
Add links to Github repo and issue tracker.

kenrussell:
Addresses #2144.

Issue #2810, opened 2019-02-21
Expose extension for EXT_texture_norm16


jdashg:
Even with incomplete support on mobile, this is still probably nice to expose.

lexaknyazev:
FWIW, ANGLE already supports it.

lexaknyazev:
Rationale:
- Integer formats do not support filtering.
- Floating point formats with comparable memory consumption provide less precision.
  - `float16` can represent only about 15K distinct values in `0.0 .. 1.0` range.
  - `float11` and `float10` - fewer than 1K and 500 respectively.
- DOM image sources (e.g. PNG) with 16 bits per channel could be copied to GPU without intermediate floating point processing.

kenrussell:
SGTM. Could someone please contribute an extension proposal?

astojilj:
+1
Thanks. Important for 16-bit (depth, IR) camera capture ([W3C mediacapture-depth](https://w3c.github.io/mediacapture-depth/#upload-video-frame-to-webgl-texture)) that is now [using gl.FLOAT](https://w3c.github.io/mediacapture-depth/#upload-video-frame-to-webgl-texture) texture as the only way to access the data, with minor data loss.

riju:
> SGTM. Could someone please contribute an extension proposal?

I will put up an extension proposal soon.

jdashg:
Moved to PR #2837.

PR #2837, opened 2019-03-28
Expose extension for EXT_texture_norm16.

riju:
As suggested by @jdashg, that even with incomplete support on mobile,
this is still probably nice to expose.

@lexaknyazev succinctly summarized the rationale
in https://github.com/KhronosGroup/WebGL/issues/2810#issuecomment-466387837

1. Integer formats do not support filtering.
2. Floating point formats with comparable memory consumption provide less precision.
   float16 can represent only about 15K distinct values in 0.0 .. 1.0 range.
   float11 and float10 - fewer than 1K and 500 respectively.
3. DOM image sources (e.g. PNG) with 16 bits per channel could be copied to GPU
   without intermediate floating point processing

Fixes #2180

### DIFF
--- a/extensions/proposals/EXT_texture_norm16/extension.xml
+++ b/extensions/proposals/EXT_texture_norm16/extension.xml
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<proposal href="proposals/EXT_texture_norm16/">
+
+  <name>EXT_texture_norm16</name>
+
+  <contact> <a href="https://www.khronos.org/webgl/public-mailing-list/">WebGL
+  working group</a> (public_webgl 'at' khronos.org) </contact>
+
+  <contributors>
+  <contributor>Rijubrata Bhaumik (rijubrata.bhaumik 'at' intel.com)</contributor>
+    <contributor>Members of the WebGL working group</contributor>
+  </contributors>
+
+  <number>NN</number>
+
+  <depends>
+    <api version="2.0"/>
+  </depends>
+
+  <overview>
+
+    <mirrors href="https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_norm16.txt"
+             name="EXT_texture_norm16">
+        <addendum>Optional support for <code>UNSIGNED_SHORT</code> textures as FBO
+        attachments.</addendum>
+    </mirrors>
+
+    <features>
+      <feature> This extension provides a set of new 16 bit signed normalized and unsigned normalized fixed point texture, renderbuffer and texture buffer formats.
+      </feature>
+
+      <feature> The <code>texImage2D</code> and <code>texSubImage2D</code>
+      entry points taking <code>ArrayBufferView</code> are extended to accept
+      <code>Uint16Array</code> with the pixel type <code>UNSIGNED_SHORT</code> and
+      <code>Int16Array</code> with the pixel type <code>SHORT</code>.
+      </feature>
+
+      <feature> The <code>texImage2D</code> and <code>texSubImage2D</code>
+      entry points taking <code>ImageData</code>,
+      <code>HTMLImageElement</code>, <code>HTMLCanvasElement</code> and
+      <code>HTMLVideoElement</code> are extended to accept the pixel type
+      <code>UNSIGNED_SHORT</code>. </feature>
+
+    </features>
+  </overview>
+
+  <idl xml:space="preserve">
+
+[NoInterfaceObject]
+interface EXT_texture_norm16 { };
+  </idl>
+
+  <history>
+    <revision date="2019/03/27">
+      <change>Initial Draft.</change>
+    </revision>
+  </history>
+</proposal>
--- a/extensions/proposals/EXT_texture_norm16/extension.xml
+++ b/extensions/proposals/EXT_texture_norm16/extension.xml
@@ -14,19 +14,19 @@
   <number>NN</number>
 
   <depends>
-    <api version="2.0"/>
+    <api version="1.0"/>
   </depends>
 
   <overview>
 
     <mirrors href="https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_norm16.txt"
              name="EXT_texture_norm16">
-        <addendum>Optional support for <code>UNSIGNED_SHORT</code> textures as FBO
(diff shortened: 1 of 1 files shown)


riju:
PTAL @kenrussell  
+cc @Richard-Yunchao

lexaknyazev:
What are rules for converting data from 8-bit DOM sources?
- One option is to "renormalize", i.e., linearly map 8-bit `255` to 16-bit `65535`.
- Other is to "cast" ubyte texels to ushort, effectively making them invisible.

jdashg:
Conversion would have to requantize: 0xff -> 0xffff, 0x42 -> 0x4242 (generally)

extensions/proposals/EXT_texture_norm16/extension.xml
```diff
@@ -0,0 +1,58 @@
...
+
+  <contributors>
+  <contributor>Rijubrata Bhaumik (rijubrata.bhaumik 'at' intel.com)</contributor>
+    <contributor>Members of the WebGL working group</contributor>
+  </contributors>
+
+  <number>NN</number>
+
+  <depends>
+    <api version="2.0"/>
```
jdashg: 1.0, unless there's a reason this can't be exposed for WebGL1.

extensions/proposals/EXT_texture_norm16/extension.xml
```diff
@@ -0,0 +1,58 @@
...
+      <code>HTMLVideoElement</code> are extended to accept the pixel type
+      <code>UNSIGNED_SHORT</code>. </feature>
+
+    </features>
+  </overview>
+
+  <idl xml:space="preserve">
+
+[NoInterfaceObject]
+interface EXT_texture_norm16 { };
```
jdashg: We need to add the enums:
        R16_EXT                                     0x822A
        RG16_EXT                                    0x822C
        RGB16_EXT                                   0x8054
        RGBA16_EXT                                  0x805B
        R16_SNORM_EXT                               0x8F98
        RG16_SNORM_EXT                              0x8F99
        RGB16_SNORM_EXT                             0x8F9A
        RGBA16_SNORM_EXT                            0x8F9B

extensions/proposals/EXT_texture_norm16/extension.xml
```diff
@@ -0,0 +1,58 @@
...
+      <code>Uint16Array</code> with the pixel type <code>UNSIGNED_SHORT</code> and
+      <code>Int16Array</code> with the pixel type <code>SHORT</code>.
+      </feature>
+
+      <feature> The <code>texImage2D</code> and <code>texSubImage2D</code>
+      entry points taking <code>ImageData</code>,
+      <code>HTMLImageElement</code>, <code>HTMLCanvasElement</code> and
+      <code>HTMLVideoElement</code> are extended to accept the pixel type
+      <code>UNSIGNED_SHORT</code>. </feature>
+
```
jdashg: Also add [R,RG,RGBA]16_EXT as Renderbuffer internalformats.

extensions/proposals/EXT_texture_norm16/extension.xml
```diff
@@ -0,0 +1,58 @@
...
+  <depends>
+    <api version="2.0"/>
+  </depends>
+
+  <overview>
+
+    <mirrors href="https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_norm16.txt"
+             name="EXT_texture_norm16">
+        <addendum>Optional support for <code>UNSIGNED_SHORT</code> textures as FBO
```
jdashg: Why is this called out as optional?

extensions/proposals/EXT_texture_norm16/extension.xml
riju: ext_texture_norm16 is written against the OpenGL ES 3.1, feature-wise corresponding to is WebGL 2. There is no technical reason to avoid WebGL1.

extensions/proposals/EXT_texture_norm16/extension.xml
riju: Done.

extensions/proposals/EXT_texture_norm16/extension.xml
riju: Done.

riju:
Thanks @jdashg for the review comments. PTAL.

jdashg:
@kenrussell Do we need anything else here before merging?

extensions/proposals/EXT_texture_norm16/extension.xml
```diff
@@ -0,0 +1,68 @@
...
+  <overview>
+
+    <mirrors href="https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_norm16.txt"
+             name="EXT_texture_norm16">
+        <addendum>Support for <code>UNSIGNED_SHORT</code> textures as FBO
+        attachments.</addendum>
+    </mirrors>
+
+    <features>
+      <feature>This extension provides a set of new 16-bit signed normalized and unsigned normalized fixed point texture, renderbuffer and texture buffer formats. The 16-bit normalized fixed point types <code>R16_EXT</code>, <code>RG16_EXT</code> and <code>RGBA16_EXT</code> become available as color-renderable formats. Renderbuffers can be created in these formats. These and textures created with <code>type = UNSIGNED_SHORT</code>, which will have one of these internal formats, can be attached to framebuffer object color attachments for rendering.
```
kenrussell: Please wrap this line to 80 or 100 columns.

extensions/proposals/EXT_texture_norm16/extension.xml
kenrussell: Did you push an update modifying the accepted values for renderbufferStorage? Might be necessary to "git push origin [branch-name] --force" if you've rebased.

Also, does renderbufferStorageMultisample accept these internal formats?

extensions/proposals/EXT_texture_norm16/extension.xml
riju: Done.

riju:
<code>R16_EXT</code>, <code>RG16_EXT</code> and <code>RGBA16_EXT</code> are also accepted by renderbufferStorageMultisample.

kenrussell:
Thanks for updating. After re-reading the first paragraph, this looks like it covers both the texture and renderbuffer cases well.

PR #2840, opened 2019-03-29
Replace dispatchTask with awaitable dispatchPromise.

jdashg:
Makes it really easy to write code like:
```
async function Test() {
  for (const x of TEST_ARGS) {
    await wtu.dispatchPromise(() => {
      TestOnce(x);
    });
  }
  finishTest();
}
Test();
```


jdashg:
For example: https://github.com/jdashg/WebGL/commit/b7ae0667c67c0378864d6df32b29ce47362de8ae

kenrussell:
Looks nice!

PR #2841, opened 2019-03-29
User highp for VertexID in vertex-id.html test

shrekshao:
https://buganizer.corp.google.com/issues/129495560#comment3

As stated here, default precision for `int` is `mediump` which is often implemented as 16 bits on mobile parts. This would leads to the failure of this test on pixel 2 and 3 (Qualcomm)

### DIFF
--- a/sdk/tests/conformance2/rendering/vertex-id.html
+++ b/sdk/tests/conformance2/rendering/vertex-id.html
@@ -14,7 +14,7 @@
 <!-- Shaders for testing instanced draws -->
 <script id="vs" type="text/plain">
 #version 300 es
-flat out int vVertexID;
+flat out highp int vVertexID;
 
 void main() {
     vVertexID = gl_VertexID;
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/rendering/vertex-id.html
```diff
@@ -14,7 +14,7 @@
 <!-- Shaders for testing instanced draws -->
 <script id="vs" type="text/plain">
 #version 300 es
-flat out int vVertexID;
+flat out highp int vVertexID;
```
jdashg: Vertex shaders default to highp int.

jdashg:
Thanks!

sdk/tests/conformance2/rendering/vertex-id.html
kenrussell: Then the problem must have been that the fragment shader didn't declare highp?

PR #2844, opened 2019-04-01
Add Fluid example

haoyunfeix:
This example is ported from
https://github.com/walbourn/directx-sdk-samples/tree/master/FluidCS11.

Try this example using latest chrome canary with below commands:
Chrome.exe --use-cmd-decoder=passthrough --enable-webgl2-compute-context --use-gl=angle

### DIFF
--- a/sdk/demos/intel/fluid/LICENSE
+++ b/sdk/demos/intel/fluid/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+Copyright © 2019 Intel Corporation
+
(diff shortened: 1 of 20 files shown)


haoyunfeix:
@kenrussell  PTAL. Thanks.

kenrussell:
Very nice! Looks great!

Would be super if you could add mouse interaction so the fluid could be swirled around. Also a way to adjust the number of particles. On my mid-range NVIDIA GPU all three options reach 60 FPS.

PR #2846, opened 2019-04-03
Add conformance/rendering/blending.html.

jdashg:
Also:
* Add wtu.destroyContext and wtu.destroyAllContexts to facilitate
cleanup.
* Add maxArrayDiff and expectArray.


kenrussell:
Thanks, looks good. Tested in Chrome on macOS; 1.0 and 2.0 versions of the test both pass.

PR #2847, opened 2019-04-04
WebKit-specific rule to stop the slide-over button from looking horrible on iOS

grorg:
- also tweaked the position of the test list to reduce
  the gap between it and the header


jdashg:
Cool, merge after CLA.

kenrussell:
Dean, to be sure, we've all needed to click through the CLA. Can you please do that?

jdashg:
@grorg ping!

PR #2850, opened 2019-04-05
Fix rendering-sampling-feedback-loop test for old drivers

shrekshao:
This test also works for old driver not supporting `#extension GL_EXT_draw_buffers : require` for shaders. (d3d9 and some old android devices)

### DIFF
--- a/sdk/tests/conformance/rendering/rendering-sampling-feedback-loop.html
+++ b/sdk/tests/conformance/rendering/rendering-sampling-feedback-loop.html
@@ -181,12 +181,12 @@ function rendering_sampling_feedback_loop(draw_buffers) {
 
     gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
     wtu.clearAndDrawUnitQuad(gl);
-    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "Feedback loop detection ignores drawBuffers settings.");
+    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "Feedback loop detection should ignore drawBuffers settings.");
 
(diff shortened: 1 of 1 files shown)


shrekshao:
It should be the former (SUPPOSED to ignore drawBuffers settings). I update the glErrorShouldBe message ("should ignore")

sdk/tests/conformance/rendering/rendering-sampling-feedback-loop.html
```diff
@@ -125,10 +134,12 @@
     }
 }
 
-function init() {
+function init(drawBuffers) {
```
kainino0x: Oops I forgot to submit my other comment. drawBuffers can be captured by scope without passing it as an argument. Since rendering_sampling_feedback_loop captures by scope, let's be consistent and do it here too. (remove the drawBuffers arg)

shrekshao:
I think according to https://github.com/shrekshao/WebGL/blob/2984a85e124af4a668e06f473a43d8031d67cf31/sdk/tests/conformance/rendering/rendering-sampling-feedback-loop.html#L115

If it's available, it is always requested. Thus drawBuffers is assigned to a function, as a result it would use multiple render target. (Does this answer the question?)

kainino0x:
I'm talking about the browser's behavior, and whether we test it completely. This test always requests it if it's available. But do we test what happens if it's not available? Unrequested and unavailable should behave identically.

shrekshao:
Oh so it's about if the browser would fail to compile the shader when `WEBGL_draw_buffers` is supported but not requested.

Yep I think this is tested here: https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/conformance/extensions/webgl-draw-buffers.html#L262

kainino0x:
Great, looks good.

PR #2852, opened 2019-04-09
Fix depth and stencil texture completeness in depth-stencil-feedback-loop.html

shrekshao:
To form a feedback loop, texture needs to be complete. As stated in OpenGL ES 3.0.5 Section 3.8.13 (P161)

> Using the preceding definitions, a texture is complete unless any of the following conditions hold true:
...
• The effective internal format specified for the texture arrays is a sized
internal depth or depth and stencil format (see table 3.14), the value of
TEXTURE_COMPARE_MODE is NONE, and either the magnification filter is
not NEAREST or the minification filter is neither NEAREST nor NEAREST_-
MIPMAP_NEAREST.

Added some `gl.texParameteri` to make depth and depth_stencil texture "texture-complete"

### DIFF
--- a/sdk/tests/conformance2/rendering/depth-stencil-feedback-loop.html
+++ b/sdk/tests/conformance2/rendering/depth-stencil-feedback-loop.html
@@ -109,8 +109,15 @@ function init() {
     tex1 = gl.createTexture();
     tex2 = gl.createTexture();
     wtu.fillTexture(gl, tex0, width, height, [0x0, 0xff, 0x0, 0xff], 0, gl.RGBA, gl.UNSIGNED_BYTE, gl.RGBA);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
     wtu.fillTexture(gl, tex1, width, height, [0x80], 0, gl.DEPTH_COMPONENT, gl.UNSIGNED_INT, gl.DEPTH_COMPONENT16);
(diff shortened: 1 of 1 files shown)


shrekshao:
Supplement material: Yes, the texture bound to framebuffer and gets sampled from needs to be possible to be sampled from to form a feedback loop. As stated in OpenGL ES 3.0.5 Section 4.4.3.1 Rendering Feedback Loops: 
> Specifically, the values of rendered fragments are undefined if all of the following conditions are true:
• an image from texture object T is attached to the currently bound draw framebuffer at attachment point A
• the texture object T is currently bound to a texture unit U, and
• the current programmable vertex and/or fragment processing state makes it
possible (see below) to sample from the texture object T bound to texture
unit U
while either of the following conditions are true:
• the value of TEXTURE_MIN_FILTER for texture object T is NEAREST or
LINEAR, and the value of FRAMEBUFFER_ATTACHMENT_TEXTURE_LEVEL
for attachment point A is equal to the value of TEXTURE_BASE_LEVEL for
the texture object T
• the value of TEXTURE_MIN_FILTER for texture object T is one
of NEAREST_MIPMAP_NEAREST, NEAREST_MIPMAP_LINEAR, LINEAR_-
MIPMAP_NEAREST, or LINEAR_MIPMAP_LINEAR, and the value of
FRAMEBUFFER_ATTACHMENT_TEXTURE_LEVEL for attachment point A is
within the range specified by the current values of TEXTURE_BASE_LEVEL
to q, inclusive, for the texture object T. (q is defined in the Mipmapping
discussion of section 3.8.10).

PR #2853, opened 2019-04-09
Remove Program Pipeline from WebGL 2.0 compute

Richard-Yunchao:
Program Pipeline is not a core feature for compute shader, it is
not very important. Furthermore, program pipeline is not very
useful. We didn't find that some demos or applications rely on it
and state that it is a must-to-have feature. In addition, we didn't
see a simple way to implement this feature on Direct3D 11. So we tend
to remove this feature from WebGL 2.0 compute.


Richard-Yunchao:
PTAL. @kenrussell . 
FYI, @qjia7 and @gyagp .

kenrussell:
Seems fine to remove this concept from this specification. LGTM

PR #2854, opened 2019-04-10
Fix context-attribute-preserve-drawing-buffer.html

shrekshao:
There's no way to guarantee the compositor finished preparing the frame.

So we are gonna change the test to the following behavior: 

Do a for loop for `await nextFrame()` and keep doing readpixels and checking within the loop until we reach some time limit (say 0.5s). For testing `preserveDrawingBuffer: false` we could early terminate if the pixel data matches. And when testing `preserveDrawingBuffer: true` we could assert failure once the pixel doesn't match the expected ones.

### DIFF
--- a/sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
+++ b/sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
@@ -40,8 +40,8 @@
 <script>
 "use strict";
 description("This test checks whether OffscreenCanvas webgl context honors the preserveDrawingBuffer flag.");
-var wtu = WebGLTestUtils;
-var pixels = new Uint8Array(4);
+const wtu = WebGLTestUtils;
+const pixels = new Uint8Array(4);
 
 function checkPixels(color) {
   return (color[0] === pixels[0]) &&
@@ -53,15 +53,16 @@ function checkPixels(color) {
 const nextFrame = () => new Promise(r => requestAnimationFrame(r));
 
 async function getPixelsFromOffscreenWebgl(preserveFlag, color, msg) {
-  var canvas = document.createElement("canvas");
-  var offscreenCanvas = transferredOffscreenCanvasCreation(canvas, 10, 10);
-  var gl = offscreenCanvas.getContext("webgl", {preserveDrawingBuffer: preserveFlag});
+  const canvas = document.createElement("canvas");
+  const offscreenCanvas = transferredOffscreenCanvasCreation(canvas, 10, 10);
+  const gl = offscreenCanvas.getContext("webgl", {preserveDrawingBuffer: preserveFlag});
 
   // Draw some color on gl
   gl.clearColor(1, 0, 1, 1);
   gl.clear(gl.COLOR_BUFFER_BIT);
 
-  var t, t0 = await nextFrame();
+  let t;
+  const t0 = await nextFrame();
   for (;;) {
     t = await nextFrame();
     
--- a/sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
+++ b/sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
@@ -52,10 +52,38 @@ function checkPixels(color) {
 
 const nextFrame = () => new Promise(r => requestAnimationFrame(r));
 
-async function getPixelsFromOffscreenWebgl(preserveFlag, color, msg) {
+async function getPixelsFromOffscreenWebglPreserveFlagTrue(color, msg) {
   var canvas = document.createElement("canvas");
   var offscreenCanvas = transferredOffscreenCanvasCreation(canvas, 10, 10);
-  var gl = offscreenCanvas.getContext("webgl", {preserveDrawingBuffer: preserveFlag});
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
```diff
@@ -53,19 +61,45 @@
   gl.clearColor(1, 0, 1, 1);
   gl.clear(gl.COLOR_BUFFER_BIT);
 
-  await nextFrame();
-  var pixels = new Uint8Array(4);
-  wtu.checkCanvas(gl, color, msg);
+  var t, t0 = await nextFrame();
+  for (var i = 0, len = 5; i < len; i++) {
```
kenrussell: This only performs at most 5 iterations. I thought the idea was to keep iterating until the first failure, or the timeout occurred (as the code intends to below). Was this leftover from an earlier version of the code?

sdk/tests/conformance2/rendering/depth-stencil-feedback-loop.html
```diff
@@ -109,8 +109,15 @@
     tex1 = gl.createTexture();
     tex2 = gl.createTexture();
     wtu.fillTexture(gl, tex0, width, height, [0x0, 0xff, 0x0, 0xff], 0, gl.RGBA, gl.UNSIGNED_BYTE, gl.RGBA);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
```
kenrussell: These changes were already merged under an earlier pull request. Could you clean this up? Thanks.

sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
shrekshao: Oh I thought I misunderstood this at first. Will fix.

sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
```diff
@@ -41,33 +41,83 @@
...
+  var canvas = document.createElement("canvas");
+  var offscreenCanvas = transferredOffscreenCanvasCreation(canvas, 10, 10);
+  var gl = offscreenCanvas.getContext("webgl", {preserveDrawingBuffer: false});
+
+  // Draw some color on gl
+  gl.clearColor(1, 0, 1, 1);
+  gl.clear(gl.COLOR_BUFFER_BIT);
+
+  var t, t0 = await nextFrame();
+  for (var i = 0, len = 5; i < len; i++) {
```
jdarpinian: Should be an infinite loop, same as above

sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
```diff
@@ -41,33 +41,83 @@
...
+  var gl = offscreenCanvas.getContext("webgl", {preserveDrawingBuffer: true});
 
   // Draw some color on gl
   gl.clearColor(1, 0, 1, 1);
   gl.clear(gl.COLOR_BUFFER_BIT);
 
-  await nextFrame();
-  var pixels = new Uint8Array(4);
-  wtu.checkCanvas(gl, color, msg);
+  var t, t0 = await nextFrame();
```
jdarpinian: I prefer to avoid declaring multiple variables on the same line when one has an initializer, as it can be confusing. Maybe put t0 on its own line and also declare it const instead of var. While we're rewriting this code, maybe also change var to let (or const as appropriate) throughout the file.

jdarpinian:
Thanks for fixing this! Just a few comments.

sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
```diff
@@ -40,34 +40,68 @@
...
+  }
 }
 
 (async () => {
   if (!window.OffscreenCanvas) {
       testPassed("No OffscreenCanvas support");
   } else {
-      // Test if OffscreenCanvas.webgl retains context if preserveDrawingBuffer is true.
-      await getPixelsFromOffscreenWebgl(true, [255,0,255,255], "should be preserved");
+    // Test if OffscreenCanvas.webgl retains contents if preserveDrawingBuffer is true.
```
kenrussell: Here I would write "Test that". "Test if" sounds ambiguous.

sdk/tests/conformance/offscreencanvas/context-attribute-preserve-drawing-buffer.html
```diff
@@ -40,34 +40,68 @@
...
+    await getPixelsFromOffscreenWebgl(true, [255,0,255,255], "should be preserved");
 
-      // Test if OffscreenCanvas.webgl loses context if presereDrawingbuffer is false.
-      await getPixelsFromOffscreenWebgl(false, [0, 0, 0, 0], "should not be preserved");
-      finishTest();
+    // Test if OffscreenCanvas.webgl loses contents if preserveDrawingBuffer is false.
```
kenrussell: Similarly here.

PR #2855, opened 2019-04-10
Promote KHR_parallel_shader_compile to community approved.

kenrussell:
This extension significantly improves startup time and smoothness of
WebGL applications loading many shader programs.

### DIFF
--- a/extensions/KHR_parallel_shader_compile/extension.xml
+++ b/extensions/KHR_parallel_shader_compile/extension.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- vi:set sw=2 ts=4: -->
 <?xml-stylesheet href="../../extension.xsl" type="text/xsl"?>
-<draft href="KHR_parallel_shader_compile/">
+<extension href="KHR_parallel_shader_compile/">
   <name>KHR_parallel_shader_compile</name>
 
   <contact> <a href="https://www.khronos.org/webgl/public-mailing-list/">WebGL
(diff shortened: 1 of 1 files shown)


kenrussell:
@jdashg @RafaelCintron @grorg please review - as discussed at the last WebGL working group meeting, and as emailed to public_webgl.

jdashg:
(copied from the email thread)

         
           
      
I expect this to be best-practice with or without this extension enabled:
```
for (const x of shaders)
   gl.compileShader(x);
for (const x of programs)
   gl.linkProgram(x);
```
With the extension, apps would be able to poll whether programs have
linked without janking, but these are likely to take the same total
wall time to link:
```
function* linkingProgress(programs) {
   const ext = gl.getExtension('KHR_parallel_shader_compile');
   let todo = programs.slice();
   while (todo.length) {
      if (ext) {
         todo = todo.filter(x => !gl.getProgramParameter(x, ext.COMPLETION_STATUS_KHR));
      } else {
         const x = todo.pop();
         gl.getProgramParameter(x, gl.LINK_STATUS);
      }
      yield 1.0 - (todo.length / programs.length);
   }
}
```

kenrussell:
Since no objections and since some users have confirmed on public_webgl that with the latest fixes to Chrome, querying the completion status of shaders is fast and avoids blocking, this should be easily implementable on all browsers by updating ANGLE. Merging now.

jdashg:
Just to be clear: Firefox does not integrate with ANGLE in such a way that this is trivially implementable for non-windows platforms.

kenrussell:
Chrome also has more work to do in order to expose this extension on non-Windows platforms. However, the Windows platform is the one with the longest compilation times.

Issue #2845, opened 2019-04-01
Specify behavior of no-output fragment shader with active draw buffer attachments


shrekshao:
Some of the early discussion happened under this thread: https://github.com/KhronosGroup/WebGL/pull/2780#issuecomment-476877226

shrekshao:
CC @jdashg @null77

kenrussell:
Updated the synopsis slightly. Need to resolve this in order to finish the conformance tests in this area.

shrekshao:
@jdashg @null77 A reminder for finishing this discussion after Jeff finishing with some experiments.
What we have so far based on latest comment (correct me if there's anything wrong): 
For a no-output fragment shader:
* A version 100 es shader would be considered to have an implicit output gl_FragColor/gl_FragData[0].  Thus making the case with a `gl.COLOR_ATTACHMENT0` or `gl.BACK` unaffected.
* A version 300 es shader would be considered to have exactly no output, under which circumstances, any active `gl.COLOR_ATTACHMENTi` or `gl.BACK` would lead to a `gl.INVALID_OPERATION` error.

jdashg:
We should try this, though I have some fear of breaking content here.

kenrussell:
Also concerned about the potential compatibility impact of this change. I think it may be better to change the rule to not make this an INVALID_OPERATION.

jdashg:
Investigation by @shrekshao in #2856 seems to point towards this being ok to restrict.

kenrussell:
Yes. Real-world content does not seem affected by this restriction.

PR #2856, opened 2019-04-11
Empty glsl fragment shader fixes

shrekshao:
Based on the discussion here: https://github.com/KhronosGroup/WebGL/pull/2780, and finalization here: https://github.com/KhronosGroup/WebGL/issues/2845

Fix a test case to avoid `gl.INVALID_OPERATION` error generated;

FYI: By doing a quick search for `void main() {}`, there's another appearance in https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/conformance2/context/incorrect-context-object-behaviour.html
However it doesn't cause any error since there's no gl.draw(Arrays*/Elements*) call. Leave it unmodified.

### DIFF
--- a/sdk/tests/conformance/attribs/gl-vertex-attrib-unconsumed-out-of-bounds.html
+++ b/sdk/tests/conformance/attribs/gl-vertex-attrib-unconsumed-out-of-bounds.html
@@ -51,7 +51,9 @@
 </script>
 
 <script id="fshader" type="x-shader/x-fragment">
-    void main() { }
+    void main() {
+      gl_FragColor = vec4(1);
+    }
 </script>
 
 <script>
--- a/sdk/tests/conformance/extensions/webgl-draw-buffers.html
+++ b/sdk/tests/conformance/extensions/webgl-draw-buffers.html
@@ -531,7 +531,7 @@ function runDrawTests() {
   } else {
     gl.useProgram(noWriteProgram);
     wtu.drawUnitQuad(gl);
-
+    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "Active draw buffers with missing frag outputs.");
     drawBuffersUtils.checkAttachmentsForColor(attachments, [0, 255, 0, 255]);
     gl.deleteProgram(noWriteProgram);
   }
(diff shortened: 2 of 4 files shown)


shrekshao:
Sorry, there's at least one other tests with empty fragment shader
https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/conformance/attribs/gl-vertex-attrib-unconsumed-out-of-bounds.html

Let me do some further clean up before review

shrekshao:
Searched with regex `void main\(\)(\n|\s)*\{(\n|\s)*\}` for empty fragment shader (empty vertex shader is okay). (Not sure if it could cover all appearances)

* es100, add trivial output for gl_FragColor in shader
    - conformance/attribs/gl-vertex-attrib-unconsumed-out-of-bounds.html
* es100, add wtu.glErrorShouldBe to clear up errors
   - conformance/extensions/webgl-draw-buffers.html
* es300, add gl.drawBuffers calls
    - conformance2/buffers/get-buffer-sub-data-validity.html
* left unmodified since there's no gl.drawArrays/Elements call
    - conformance/essl3-shaders-with-webgl1.html
    - conformance/ogles/GL/build/build*.html
    - conformance2/context/incorrect-context-object-behaviour.html

Now it's ready for review

shrekshao:
Tested on locally built chrome with Patch on Win10 (angle) and Linux (cmd-decoder) with Sketchfab, Three.js, PlayCanvas, ShaderToy, xeogl, babylonjs, web-scene-viewer, webgl 2 particle. Seems to work fine. With developer console opened, didn't see INVALID_OPERATION error.

kenrussell:
Thanks @shrekshao for testing this change against real-world content. Could you please run a few more tests with your change?

A few of the recent HTML5 exports from the Unreal Engine:
https://s3.amazonaws.com/unrealengine/HTML5/TestBuilds/index.html

Unity's Angry Bots demo:
https://beta.unity3d.com/jonas/AngryBots/

and Unity's BoatAttack demo:
https://files.unity3d.com/marcot/boatattack-mt/

Just want to be as sure as possible that we won't regress high-end content which might be rendering to the depth buffer without a color buffer.

I was a little surprised to see how many shaders in the conformance suite had to be updated for this stricter rule, but looking back at the spec (the most recent OpenGL ES version, 3.2, section 15.2.1 "Selecting Buffers for Writing"), it's clear that this is undefined behavior even for the default back buffer:

"If a fragment color is undefined after fragment shader execution, or if the value written for a fragment color does not match the format of a draw buffer it is to be written to, the resulting value in the draw buffer is undefined."

So I think resolving this undefined behavior is the best thing to do, and tightening the spec and making it an error is the clearest signal to developers.

shrekshao:
Seems to work fine with these unity and unreal demos.

shrekshao:
@jdashg Just a reminder

jdashg:
Thanks!

PR #2861, opened 2019-04-16
Fix a typo in Copyright for WebGL 2 compute

Richard-Yunchao:
Current <Copyright> section has two "specification": https://www.khronos.org/registry/webgl/specs/latest/2.0-compute/. It's incorrect. This small change fix the tiny bug. 

PTAL. @kenrussell


kenrussell:
LGTM after the fact.

PR #2863, opened 2019-04-18
Change WEBGL_video_texture proposal extension

shaoboyan:
This commit change some names in the proposed extension and add limitations
@kainino0x @kenrussell  PTAL.

### DIFF
--- a/extensions/proposals/WEBGL_video_texture/extension.xml
+++ b/extensions/proposals/WEBGL_video_texture/extension.xml
@@ -20,7 +20,7 @@
   <overview>
     <mirrors href="https://www.khronos.org/registry/gles/extensions/OES/OES_EGL_image_external.txt"
              name="OES_EGL_image_external">
-      <addendum>Defines a new texture target <code>TEXTURE_VIDEO_IMAGE</code>.</addendum>
+      <addendum>Defines a new texture target <code>TEXTURE_VIDEO_IMAGE_WEBGL</code> and a new sampler type <code>SAMPLER_VIDEO_IMAGE_WEBGL</code></addendum>
       <addendum>Provides a mechanism for binding <code>HTMLVideoElement</code> stream to video texture targets.</addendum>
       <addendum>Provides time of frame, texture width and height of <code>HTMLVideoElement</code>'s texture binding.</addendum>
     </mirrors>
@@ -28,6 +28,17 @@
     <features>
       <feature>Add support for <code>WEBGL_video_texture</code>
       binding of HTMLVideoElement.</feature>
+      <feature><code>TEXTURE_VIDEO_IMAGE_WEBGL</code> and <code>TEXTURE_2D</code> are distinct.</feature>
+      <feature>Textures bound to <code>TEXTURE_VIDEO_IMAGE_WEBGL</code> only accept <code>LINEAR</code> and <code>NEAREST</code>
+        for <code>TEXTURE_MIN_FILTER</code> and <code>TEXTURE_MAG_FILTER</code>.</feature>
+      <feature>Textures bound to <code>TEXTURE_VIDEO_IMAGE_WEBGL</code> only accept <code>CLAMP_TO_EDGE</code> for
+        <code>TEXTURE_WRAP_S</code> and <code>TEXTURE_WRAP_T</code>.</feature>
+      <feature>The maximum mipmap level of textures that bind to <code>TEXTURE_VIDEO_IMAGE_WEBGL</code> is 1.</feature>
+      <feature><code>shareVideoImageWEBGL</code> will lock video frame and reject another video source until call
+        <code>releaseVideoImageWEBGL</code>.</feature>
+      <feature>If the application does not call releaseVideoImageWEBGL, then all of the WebGLTextures with which video
+        frames were shared will be implicitly released when the current JavaScript code returns control to the user agent.</feature>
+
 
       <glsl extname="WEBGL_video_texture">
         <stage type="fragment"/>
@@ -53,11 +64,12 @@ interface WebGLVideoFrameInfo {
 
 [NoInterfaceObject]
 interface WEBGL_video_texture {
-    const GLenum TEXTURE_VIDEO_IMAGE             = 0x851D;
-    const GLenum SAMPLER_VIDEO_IMAGE             = 0x8B61;
+    const GLenum TEXTURE_VIDEO_IMAGE_WEBGL             = 0x851D;
+    const GLenum SAMPLER_VIDEO_IMAGE_WEBGL             = 0x8B61;
 
-    [RaisesException] WebGLVideoFrameInfo VideoElementTargetVideoTexture(
-        GLenum target, HTMLVideoElement video);
+    [RaisesException] WebGLVideoFrameInfo shareVideoImageWEBGL(
+      GLenum target, HTMLVideoElement video);
+    [RaiseException] bool releaseVideoImageWEBGL(GLenum target);
 };
   </idl>
 
@@ -84,9 +96,10 @@ interface WEBGL_video_texture {
     function update() {
         var ext = gl.getExtension('WEBGL_video_texture');
         if(ext !=== null){
-            gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE, videoTexture);
-            ext.VideoElementTargetVideoTexture(ext.TEXTURE_VIDEO_IMAGE, videoElement);
-            gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE, null);
+            gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE_WEBGL, videoTexture);
+            ext.shareVideoImageWEBGL(ext.TEXTURE_VIDEO_IMAGE_WEBGL, videoElement);
+            ext.releaseVideoImageWEBGL(ext.TEXTURE_VIDEO_IMAGE_WEBGL);
+            gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE_WEBGL, null);
         }
     }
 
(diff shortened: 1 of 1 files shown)


extensions/proposals/WEBGL_video_texture/extension.xml
```diff
@@ -20,14 +20,21 @@
   <overview>
     <mirrors href="https://www.khronos.org/registry/gles/extensions/OES/OES_EGL_image_external.txt"
              name="OES_EGL_image_external">
-      <addendum>Defines a new texture target <code>TEXTURE_VIDEO_IMAGE</code>.</addendum>
+      <addendum>Defines a new texture target <code>TEXTURE_VIDEO_IMAGE_WEBGL</code> and a new sampler type <code>GL_SAMPLER_VIDOE_IMAGE_WEBGL</code></addendum>
```
kainino0x: GL_SAMPLER_VIDOE... -> SAMPLER_VIDEO...

(remove the GL_ and fix VIDOE typo)

extensions/proposals/WEBGL_video_texture/extension.xml
```diff
@@ -20,14 +20,21 @@
...
     </mirrors>
 
     <features>
       <feature>Add support for <code>WEBGL_video_texture</code>
       binding of HTMLVideoElement.</feature>
+      <feature>Bind textures to <code>TEXTURE_VIDEO_IMAGE_WEBGL</code> may unbind textures binded to <code>TEXTURE_2D</code>.</feature>
+      <feature>Textures bind to <code>TEXTURE_VIDEO_IMAGE_WEBGL</code> only accept <code>GL_LINEAR</code> and <code>GL_NEAREST</code>
+        for <code>TEXTURE_MIN_FILTER</code> and <code>TEXTURE_MAG_FILTER</code>.</feature>
+      <feature>Textures binds to <code>TEXTURE_VIDEO_IMAGE_WEBGL</code> only accept <code>GL_CLAMP_TO_EDGE</code> for
+        <code>GL_TEXTURE_WRAP_S</code> and <code>GL_TEXTURE_WRAP_T</code>.</feature>
```
kainino0x: Same in this new text, the GL_ prefix should be removed.

extensions/proposals/WEBGL_video_texture/extension.xml
```diff
@@ -20,14 +20,21 @@
...
-      <addendum>Defines a new texture target <code>TEXTURE_VIDEO_IMAGE</code>.</addendum>
+      <addendum>Defines a new texture target <code>TEXTURE_VIDEO_IMAGE_WEBGL</code> and a new sampler type <code>GL_SAMPLER_VIDOE_IMAGE_WEBGL</code></addendum>
       <addendum>Provides a mechanism for binding <code>HTMLVideoElement</code> stream to video texture targets.</addendum>
       <addendum>Provides time of frame, texture width and height of <code>HTMLVideoElement</code>'s texture binding.</addendum>
     </mirrors>
 
     <features>
       <feature>Add support for <code>WEBGL_video_texture</code>
       binding of HTMLVideoElement.</feature>
+      <feature>Bind textures to <code>TEXTURE_VIDEO_IMAGE_WEBGL</code> may unbind textures binded to <code>TEXTURE_2D</code>.</feature>
```
kainino0x: Bind textures -> Binding textures

binded -> bound

kainino0x: Can it be more specific than "may"? I think the implementation should guarantee that these binding points are distinct, even if it has to emulate it.

extensions/proposals/WEBGL_video_texture/extension.xml
```diff
@@ -20,14 +20,21 @@
...
+      <feature>Textures bind to <code>TEXTURE_VIDEO_IMAGE_WEBGL</code> only accept <code>GL_LINEAR</code> and <code>GL_NEAREST</code>
```
kainino0x: bind -> bound

extensions/proposals/WEBGL_video_texture/extension.xml
```diff
@@ -20,14 +20,21 @@
...
+        for <code>TEXTURE_MIN_FILTER</code> and <code>TEXTURE_MAG_FILTER</code>.</feature>
+      <feature>Textures binds to <code>TEXTURE_VIDEO_IMAGE_WEBGL</code> only accept <code>GL_CLAMP_TO_EDGE</code> for
```
kainino0x: binds -> bound

extensions/proposals/WEBGL_video_texture/extension.xml
shaoboyan: "may" here refers two different situations of current implementation. On Windows, it is emulated by a texture 2D, but on Android, the implementation should be a real distinct target.
To remove "may", there are two ways:
1. bind texture to TEXTURE_VIDEO_IMAGE_WEBGL unbind textures bound to TEXTURE_2D
This is straight for Windows platform but not necessary for Android.
2. Bind texture to TEXTURE_VIDEO_IMAGE_WEBGL doesn't unbind textures bound to TEXTURE_2D
For Android, it is true, but for Windows platform, this will have complex mechanism to find a empty texture unit (and maybe not exisit)

Because I tried method 2 first and find lots of potential issues, so I tend to use method 1. WDYT?

kainino0x: I definitely think we need to spec it one way or the other.

I can see why it might be difficult to do option 2. I wasn't thinking of trying to use an empty texture unit - just use the user-provided glActiveTexture unit. The tricky part is making sure that the correct texture is bound to the texture unit's underlying binding points, each time the texture unit is used. I don't think it's ever possible to observe both binding points (TEXTURE_2D and TEXTURE_VIDEO_IMAGE_WEBGL) at the same time, but I could be wrong about that.

I would be ok specifying that it MUST replace the TEXTURE_2D binding point. However, at that point, I would prefer to say that it *IS* the TEXTURE_2D binding point. Is there any way to get rid of the extra TEXTURE_VIDEO_IMAGE_WEBGL binding point (and enum)? (Sorry, haven't had a chance to review whether this would work or not.)

jdashg: We're going to need to do some patching magic for tex unit selection (uniform1i) at draw time anyway, so that means we should be able to de-alias TEXTURE_2D[0] and TEXTURE_VIDEO_IMAGE_WEBGL[0]. I think we should treat these as always distinct, and binding to TEXTURE_VIDEO_IMAGE_WEBGL will not change bindinds to TEXTURE_2D and vice versa.

shaoboyan: Thanks for your reviewing!. I think I was persuaded by the uniform1i magic patching. I think the progress should be:
1. Bind TEXTURE_VIDEO_IMAGE_WEBGL won't do real binding.
2. Call sharing video frame API will unbind current TEXTURE_2D and do real binding for sharing.
3. Check sampler type in uniform1i (if possible) if the active texture unit has both TEXTURE_2D and TEXTURE_VIDEO_IMAGE_WEBGL bound and then choose the correct one bind to the unit and do real uniform1i.

WDYT?

kainino0x: Off-hand, seems good to me.

extensions/proposals/WEBGL_video_texture/extension.xml
```diff
@@ -53,10 +60,10 @@ interface WebGLVideoFrameInfo {
 
 [NoInterfaceObject]
 interface WEBGL_video_texture {
-    const GLenum TEXTURE_VIDEO_IMAGE             = 0x851D;
-    const GLenum SAMPLER_VIDEO_IMAGE             = 0x8B61;
+    const GLenum TEXTURE_VIDEO_IMAGE_WEBGL             = 0x851D;
+    const GLenum SAMPLER_VIDEO_IMAGE_WEBGL             = 0x8B61;
 
-    [RaisesException] WebGLVideoFrameInfo VideoElementTargetVideoTexture(
+    [RaisesException] WebGLVideoFrameInfo shareVideoImageWEBGL(
```
kenrussell: Our group's tech lead and I had a conversation about this extension last week. Sharing the video texture to WebGL conceptually "locks" the video frame, at least in Chrome's media stack, and it's important to limit the duration that the frame might be locked.

Could you please add a new operation like "releaseVideoImageWEBGL" which takes the GLenum target, and releases the video frame from the bound texture?

Could you please also add some specification text like:

```
If the application does not call releaseVideoImageWEBGL, then all of the WebGLTextures
with which video frames were shared will be implicitly released when the current
JavaScript code returns control to the user agent.
```

It's important to make it clear to application developers that they are supposed to do all of their rendering work in the same callback where they acquire those video frames.

Thanks for considering these changes.

shaoboyan: Sure, it makes sense. Thanks for your reviewing !

extensions/proposals/WEBGL_video_texture/extension.xml
shaoboyan: Done.

extensions/proposals/WEBGL_video_texture/extension.xml
shaoboyan: Done.

shaoboyan: I try to update the description, PTAL if you have time.

extensions/proposals/WEBGL_video_texture/extension.xml
shaoboyan: Done.

extensions/proposals/WEBGL_video_texture/extension.xml
shaoboyan: Done.

extensions/proposals/WEBGL_video_texture/extension.xml
shaoboyan: I try to add the new API and update the example. PTAL if you have time!

PR #2864, opened 2019-04-20
In WEBGL_depth_texture test, check filtering results.

jdashg:
This helps guarantee that WebGL1-on-ES3 works properly, since there was a question about the filterability of depth components in ES3.


kenrussell:
For the record, the question about the filterability of depth textures came up on the webgl-dev-list here:
https://groups.google.com/d/topic/webgl-dev-list/T4_bKNzEhqk/discussion

Tested this change on Windows and Mac; looks good.

@jdashg should we file a separate issue about relaxing the restriction that WebGL 2.0 can't use LINEAR filtering for depth textures?

jdashg:
Yep, I'm making that PR.

PR #2865, opened 2019-04-21
Fix Python 3 errors

saschanaz:
As Python 2 will die in 2020.

### DIFF
--- a/specs/latest/1.0/extract-idl.py
+++ b/specs/latest/1.0/extract-idl.py
@@ -61,4 +61,4 @@ def elementTextContent(el):
 preList = doc.getElementsByTagName("pre")
 idlList = [elementTextContent(p) for p in preList if elementHasClass(p, "idl") ]
 licenseTemplate = Template(LICENSE)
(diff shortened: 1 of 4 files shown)


kainino0x:
If this is python3-only, the `#!` needs to point at python3 explicitly, as Jeff pointed out.

Even if not, I don't think we want to maintain a polyglot, so it should point at one or the other explicitly.

saschanaz:
@kainino0x The CI only have Python 2.7, and I'm not sure whether Travis can install both Py2/3. Can it?

saschanaz:
Hmm, maybe we could just replace 2 with 3. I'll try it.

specs/latest/1.0/extract-idl.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
```
jdashg: `#! /usr/bin/env python3` not `#! /usr/bin/python3`.

jdashg:
@kenrussell Please double-check that we're good to upgrade to py3 on Travis. Otherwise LGTM!

PR #2866, opened 2019-04-22
Adding tests for #GL_OVR_multiview ANGLE directive within OVR_multiview2 webgl extension

mingyuhu:
GL_OVR_multiview should function the same as GL_OVR_multiview2. Refactor tests to test both extensions.


mingyuhu:
Sorry, this is not adding a webgl extension. But on the angle side, we are supporting both GL_OVR_multiview and GL_OVR_multiview2 extensions. So whether someone uses the directive #GL_OVR_multiview2 or #GL_OVR_multiview in the shader code, OVR_multiview2 (the webgl extension) should be able to work with both. 

I have a PR in angle/angle that is waiting on review that adds GL_OVR_multiview as an extension. https://chromium-review.googlesource.com/c/angle/angle/+/1575904/2

jdashg:
I don't see this as supported by OVR_multiview2 in WebGL, at least as the specs stand today.

mingyuhu:
oh yes, I was supposed to add that to the spec. Sorry, let me make that change!

extensions/Makefile
```diff
@@ -1,55 +1,55 @@
-# See README for more information about the WebGL extension registry automation.
```
kenrussell: Are the changes in this file affecting the line endings? If so please revert them.

extensions/OVR_multiview2/extension.xml
```diff
@@ -27,10 +27,10 @@
       </addendum>
       <addendum>
         The extension must be enabled with the
-        <code>#extension GL_OVR_multiview2</code> directive, as shown in the sample code, to use
+        <code>#extension GL_OVR_multiview2</code> or <code>#extension GL_OVR_multiview</code> directive, as shown in the sample code, to use
```
kenrussell: I haven't been following the development of these extensions closely but am confused why we want to support both the GL_OVR_multiview *and* GL_OVR_multiview2 extension directives. They aren't equivalent conceptually; GL_OVR_multiview2 relaxes restrictions from GL_OVR_multiview.

What is the rationale for supporting both extension directives? It seems to me this will confuse WebGL developers.

mingyuhu: From what I have been told, initially OVR_multiview2 was only exposing GL_OVR_multiview2 extension directive. 
I also made the angle change for what GL_OVR_multiview2 covers here: https://chromium-review.googlesource.com/c/angle/angle/+/1552023
as per Geoff Lang's comment in the PR, this functionality was also to be exposed in GL_OVR_multiview and that the WebGL extension can use either directive and the functionality would be the same. And these are the conformance test changes for that.

As to why this is the case, I'm not too clear. I vaguely remember something along the lines of "we shouldn't have GL_OVR_multiview2 without GL_OVR_multiview, even if they function exactly the same". Perhaps @RafaelCintron can give a better explanation? (although Rafael is OOO until 4/25)

jdashg: They aren't the same though: multiview1 is more restrictive, and has (we've been led to believe) been superseded by multiview2. (such that we don't expect to have multiview1 but not multiview2 in any large numbers)

As such, we decided not to support multiview1. If this analysis had flawed premises, we should talk about it, but so far I don't see any reason to support multiview1 validation.

kenrussell: Looking again at the [GL_OVR_multiview](https://www.khronos.org/registry/OpenGL/extensions/OVR/OVR_multiview.txt) extension spec:

```
    [[ If OVR_multiview2 is not supported ]]

    It is a compile- or link-time error if any output variable other
    than gl_Position is statically dependent on gl_ViewID_OVR. If an
    output variable other than gl_Position is dynamically dependent on
    gl_ViewID_OVR, the values are undefined.

    NOTE: Implementations that also support OVR_multiview2 may not
    generate an error if these conditions are violated, even if the
    OVR_multiview2 extension is not enabled.
```

This is pretty gross - in the native OpenGL ES extension it is legal to use "#extension GL_OVR_multiview : enable", and yes, if GL_OVR_multiview2 is supported, then the restrictions defined by GL_OVR_multiview won't be enforced.

I don't think we should carry forward this design mistake in the WebGL exposure of this extension. The WebGL implementation controls the validation of the shaders via ANGLE's shader translator, and we have full ability to change the behavior in this WebGL wrapper for the native extension.

For these reasons, I think WebGL should only expose GL_OVR_multiview2 as an extension concept, and not GL_OVR_multiview.

jdashg: I agree.

PR #2868, opened 2019-04-23
Fix markup in OVR_multiview2.

kenrussell:
Follow-on to #2867.


kenrussell:
@jdashg FYI. Merging myself.

PR #2869, opened 2019-04-23
Handle Bad Implementations

greggman:
Some browsers pass less than 20% of the WebGL2 conformance tests
yet turning on support would make it appear WebGL2 is supported
leading to improper reporting all around the net.

Here's a minor test to try to weed out such implementations
from reporting support.

### DIFF
--- a/other/get.webgl.org/webgl2/index.html
+++ b/other/get.webgl.org/webgl2/index.html
@@ -1,6 +1,6 @@
 <!--
 NOTE: If you need/want this page changed to include
-other browsers see https://github.com/KhronosGroup/WebGL/tree/master/other/get.webgl.org
+other browsers see https://github.com/KhronosGroup/WebGL/tree/master/other/get.webgl.org/webgl2/
 -->
 <!doctype html>
 <html>
--- a/other/get.webgl.org/webgl2/index.html
+++ b/other/get.webgl.org/webgl2/index.html
@@ -128,10 +128,10 @@ other browsers see https://github.com/KhronosGroup/WebGL/tree/master/other/get.w
       }
 
     </style>
-<script type="text/javascript" src="troubleshooting/DoNotCopyOrLinkThisFileElseYouWillNotGetAutoUpdatedHelpForYourUsers.js"></script>
-<script type="text/javascript" src="twgl-full.min.js"></script>
-<script type="text/javascript" src="cube.js"></script>
-<script type="text/javascript">
+<script src="troubleshooting/DoNotCopyOrLinkThisFileElseYouWillNotGetAutoUpdatedHelpForYourUsers.js"></script>
+<script src="twgl-full.min.js"></script>
+<script src="cube.js"></script>
+<script>
 /**/
 
 function $$(x) {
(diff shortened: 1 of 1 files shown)


kainino0x:
lgtm, though I feel like not many people use this site. Does webglreport.com detect and surface these bad implementations?

greggman:
>  Does webglreport.com detect and surface these bad implementations?

No it does not. Neither does webglstats.com leading the bad data. See for example this report on `MAX_SAMPLES`.

The minimum allowed value is 4 yet the graph is showing results of bad implementations that report 0

kainino0x:
Is it likely to be affecting webglstats? Are there incomplete implementations which are exposed by default? I don't think that is allowed by the WebGL conformance rules.

kainino0x:
It would be wonderful to have an update to webglreport which detects these cases. I think this is the site that developers will see.

other/get.webgl.org/webgl2/index.html
```diff
@@ -182,18 +182,44 @@
...
+            { pname: 'MAX_VERTEX_UNIFORM_BLOCKS', min:12, },
+            { pname: 'MAX_VERTEX_TEXTURE_IMAGE_UNITS', min:16, },
+            { pname: 'MAX_FRAGMENT_INPUT_COMPONENTS', min:60, },
+            { pname: 'MAX_UNIFORM_BUFFER_BINDINGS', min:24, },
+            { pname: 'MAX_COMBINED_UNIFORM_BLOCKS', min:24, },
+        ];
+        for (var i = 0; i < params.length; ++i) {
+            var param = params[i];
+            var value = gl.getParameter(gl[param.pname]);
+            if (typeof value !== 'number' || Number.isNaN(value) || value < params.min) {
```
jdashg: `Number.isNaN(value)` is technically redundant to `value < params.min`, isn't it?

other/get.webgl.org/webgl2/index.html
```diff
@@ -182,18 +182,44 @@
...
+            { pname: 'MAX_DRAW_BUFFERS', min:4, },
+            { pname: 'MAX_COLOR_ATTACHMENTS', min:4, },
+            { pname: 'MAX_VERTEX_UNIFORM_BLOCKS', min:12, },
+            { pname: 'MAX_VERTEX_TEXTURE_IMAGE_UNITS', min:16, },
+            { pname: 'MAX_FRAGMENT_INPUT_COMPONENTS', min:60, },
+            { pname: 'MAX_UNIFORM_BUFFER_BINDINGS', min:24, },
+            { pname: 'MAX_COMBINED_UNIFORM_BLOCKS', min:24, },
+        ];
+        for (var i = 0; i < params.length; ++i) {
+            var param = params[i];
```
jdashg: Generally I'd like `for (const param of params)` these days, if you're not using `i`.

greggman: I assumed since the rest of the code is using `var` that you wanted old compatible code. for of is pretty new and this site is supposed to run on old browsers to tell you to upgrade your browser.

other/get.webgl.org/webgl2/index.html
greggman: Screen Shot 2019-04-24 at 16 09 27

jdashg: Ah, yeah, ok.

other/get.webgl.org/webgl2/index.html
jdashg: Fair! If this were only executed after checking for webgl2 support, we could rely on ES6.

jdashg:
Thanks!

PR #2875, opened 2019-04-24
Not requiring compile error report at compiling stage.

shrekshao:
Change the behavior of a series of shaders compiling and linking test to the following:
* Test won't fail if compile status is success but is expected to be failure.
* GLSL es 100 spec section 10.27: 
> Should compilers be required to report any errors at compile time or can errors be deferred until link
time?
RESOLUTION: If a program cannot be compiled, on-target compilers are only required to report that an error has occurred. This error may be reported at compile time or link time or both. Development systems must generate grammar errors at compile time.

* GLSL es 300 spec section 12.32:

> Should compilers be required to report any errors at compile time or can errors be deferred until link
time?
RESOLUTION: If a program cannot be compiled, on-target compilers are only required to report that an error has occurred. This error may be reported at compile time or link time or both. Development systems must generate grammar errors at compile time.

* `linkstatAdjusted = test.linkstat && test.compstat`

### DIFF
--- a/sdk/tests/conformance/ogles/ogles-utils.js
+++ b/sdk/tests/conformance/ogles/ogles-utils.js
@@ -722,7 +722,7 @@ function runBuildTest(test, callback) {
       compileSuccess = (success[0] && success[1]);
       if (!test.compstat) {
         if (compileSuccess) {
-          testFailed("expected compile failure but was successful");
+          testPassed("expected compile failure but was successful (spec doesn't require reporting errors at compile time)");
         } else {
           testPassed("expected compile failure and it failed");
         }
@@ -736,8 +736,9 @@ function runBuildTest(test, callback) {
         var program = wtu.createProgram(gl, shaders[0], shaders[1], function() {
           linkSuccess = false;
         });
-        if (linkSuccess !== test.linkstat) {
-          testFailed("expected link to " + (test.linkstat ? "succeed" : "fail"));
+        var linkstatAdjusted = test.linkstat && test.compstat;
+        if (linkSuccess !== linkstatAdjusted) {
+          testFailed("expected link to " + (linkstatAdjusted ? "succeed" : "fail"));
         } else {
           testPassed("shaders compiled and linked as expected.");
         }
--- a/sdk/tests/conformance/ogles/ogles-utils.js
+++ b/sdk/tests/conformance/ogles/ogles-utils.js
@@ -732,17 +732,19 @@ function runBuildTest(test, callback) {
         } else {
           testFailed("expected compile success but it failed");
         }
-        var linkSuccess = true;
-        var program = wtu.createProgram(gl, shaders[0], shaders[1], function() {
-          linkSuccess = false;
-        });
-        var linkstatAdjusted = test.linkstat && test.compstat;
-        if (linkSuccess !== linkstatAdjusted) {
-          testFailed("expected link to " + (linkstatAdjusted ? "succeed" : "fail"));
-        } else {
-          testPassed("shaders compiled and linked as expected.");
-        }
       }
+
+      var linkSuccess = true;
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/ogles/ogles-utils.js
```diff
@@ -722,7 +722,7 @@ function runBuildTest(test, callback) {
       compileSuccess = (success[0] && success[1]);
       if (!test.compstat) {
         if (compileSuccess) {
-          testFailed("expected compile failure but was successful");
+          testPassed("expected compile failure but was successful (spec doesn't require reporting errors at compile time)");
```
kenrussell: I haven't looked at these tests in detail but this looks like we'll lose coverage. If compstat is false then we'll never check the link status, and we want to at least make sure that for shaders with compstat=false, we confirm that they'd never link.

For the tests which set compstat=false, do they have matching shaders which would allow a link to be attempted?

jdashg:
It sounds like implementations are allowed to say "compiled successfully"
for all shaders, deferring all errors to link time. If so, we should not
bother to check any shader compilation results, even in conformance tests.

kainino0x:
We can still have tests that valid shaders compile successfully.

I'm iffy on the "development systems" thing above. Do we want to ignore that? I think it would be ok.

kenrussell:
It seems fine to ignore the spec's statement about development systems. Also fine to have assertions that shaders which should compile, do compile.

Concretely, for existing conformance tests which assert that shader compilation fails, it's important that we try to link that shader against a known good vertex or fragment shader counterpart, and assert that the link fails. Some of the conformance tests are already written this way and hopefully we won't have to rewrite too many of these conformance/ogles tests to do the same.

sdk/tests/conformance/ogles/ogles-utils.js
shrekshao: My bad for being dizzy here.

shrekshao:
I have a question for 
> link that shader against a known good vertex or fragment shader counterpart
How are we going to decide the varying output/input of the good shader? I didn't come up with a quick solution to this. I'm gonna take a look at the existing framework to see if there's some tools for this at the same time.

kenrussell:
@shrekshao do all of the tests under conformance/ogles pass with your changes above? In other words, do all of the tests that previously expected compile failures now observe link failures?

If so, I don't think any other changes to the tests are needed. Your latest code looks good to me.

shrekshao:
@kenrussell Tested locally with `build_xx_to_xx` tests and tests including `"pattern": "build"`. And they all passed.
So yes.

jdashg:
It does look like similar text is in the ESSL3 spec.

It comes down to whether we consider our implementations "development" or "on-target". It's possible to improve Chrome's program cache to efficiently report errors at compile time, so holding ourselves to the stricter "development" bar is probably viable.

shrekshao:
FYI, testing against some common webgl apps on locally built chromium seem to be okay. (Three.js, Babylon.js, Unreal, Unity, Onshape)
Improving Chrome's program cache sounds totally reasonable. But AFIK it could be painful :sweat_smile:

kenrussell:
Let's take some more time and investigate ANGLE's program cache implementation, and how COMPILE_STATUS is computed.

If computing COMPILE_STATUS can be done entirely by ANGLE's shader translator, and doesn't involve running e.g. the HLSL compiler on its translated output, then we should continue to treat WebGL implementations as "development" machines. Users have said over the years that WebGL is a great development environment for shaders, and it's important to maintain that high standard. It's a small matter of programming to improve the shader cache, and every WebGL developer will benefit.

If it turns out that providing a correct COMPILE_STATUS result *always* requires redundant compilation work, then we can move forward with this change.

PR #2876, opened 2019-04-26
FRAMEBUFFER_ATTACHMENT_TEXTURE_BASE_VIEW_INDEX_OVR on default FB is I…

jdashg:
…NVALID_ENUM.

This matches the behavior of FRAMEBUFFER_ATTACHMENT_TEXTURE_LAYER.

### DIFF
--- a/sdk/tests/conformance2/extensions/ovr_multiview2.html
+++ b/sdk/tests/conformance2/extensions/ovr_multiview2.html
@@ -91,19 +91,11 @@ function runDefaultFramebufferQueryTest()
     debug("");
     debug("Testing querying base view index and num views on the default framebuffer");
     gl.bindFramebuffer(gl.FRAMEBUFFER, null);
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/extensions/ovr_multiview2.html
```diff
@@ -91,19 +91,11 @@
     debug("");
     debug("Testing querying base view index and num views on the default framebuffer");
     gl.bindFramebuffer(gl.FRAMEBUFFER, null);
+    // Same behavior as FRAMEBUFFER_ATTACHMENT_TEXTURE_LAYER:
     let baseViewIndex = gl.getFramebufferAttachmentParameter(gl.FRAMEBUFFER, gl.BACK, ext.FRAMEBUFFER_ATTACHMENT_TEXTURE_BASE_VIEW_INDEX_OVR);
```
kainino0x: Could get rid of the variable declarations since they're not used.

PR #2878, opened 2019-04-26
Change invariance rule for #pragma STDGL invariant(all).

kenrussell:
The ESSL 1.0 specification states that the global invariance pragma
applies only to output variables. During the most recent work on this
test in #1924, the assumption was that this pragma applied essentially
to all variables, because ESSL 1.0 also mandates that the invariant
qualifier must match between vertex shader outputs and fragment shader
inputs.

ANGLE's shader translator is relaxing the validation in this area in
order to both support a Vulkan backend and pass the ES 2.0
drawElements test suite. The best understanding of the intent of the
ES 2.0 and ESSL 1.0 specifications has been applied in coming to this
decision.

The associated WebGL conformance test in shaders-with-invariance.html,
which asserts a link failure with a variant output variable in the
vertex shader and an "invariant" input variable in the fragment shader
via the pragma, is being changed to conform to the ES 2.0 spec. The
test is being removed from earlier snapshots.

For more background please see:

https://chromium-review.googlesource.com/1558678
https://bugs.chromium.org/p/angleproject/issues/detail?id=1293
https://bugs.chromium.org/p/angleproject/issues/detail?id=3285

### DIFF
--- a/conformance-suites/1.0.3/conformance/glsl/misc/shaders-with-invariance.html
+++ b/conformance-suites/1.0.3/conformance/glsl/misc/shaders-with-invariance.html
@@ -216,6 +216,9 @@ GLSLConformanceTester.runTests([
     linkSuccess: false,
     passMsg: "vertex shader with invariant varying and fragment shader with variant varying must fail",
   },
+  // The following test has been removed as it was enforcing a rule contradictory to the ESSL 1.0
+  // specification.
+  /*
   {
     vShaderId: "vertexShaderVariant",
     vShaderSuccess: true,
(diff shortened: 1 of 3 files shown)


kenrussell:
@jdashg @RafaelCintron @grorg in particular please review.

Let's please not merge this until the authors of https://chromium-review.googlesource.com/1558678 have confirmed that all of these tests pass with their patch.

Thanks.

kenrussell:
@jdashg @grorg @RafaelCintron this is now ready for review. https://chromium-review.googlesource.com/1558678 is passing all tests except this one sub-test, where the nature of the test is being changed to better match the ESSL specification.

jdashg:
Cool, thanks!

kenrussell:
Thanks Jeff. We want to remove the suppression for this test ASAP so merging now.

Issue #2842, opened 2019-03-29
Program handling is different across browsers


greggman:
My understanding of the spec is that a `WebGLProgram` (or GL program) has internal state like this

```
class WebGLProgram {
   Shader vertexShader;
   Shader fragmentShader;

   // status from last time you called gl.linkProgram
   int linkStatus;

   // message from last time you called gl.linkProgram  
   string infoLog;

   // internal program created by last SUCCESSFUL linkProgram
   Program validProgram; 
}

class Program {
  map<string, int> attribLocations;
  map<string, WebGLUniformLocation> uniformLocations;
  GPUProgramCode compiledProgram
}

class Shader {
  string src;

  // status from last time you called gl.compileShader
  int compileStatus;

  // message from last time you called gl.compileShader  
  string infoLog;

  // internal shader created from last successful compile
  GPUShaderCode compiledShader;
}
```

And that the only thing that really matters at render time is that `someProgram.validProgram` points to valid code. `someProgram.validProgram` is only set when `gl.linkProgram` is successful.

That means after a successful link if you detach or delete shaders there is no problem. 

Similarly if you change the source of a shader and compile that shader but don't link the program has not yet been affected

Further, if you call `gl.linkProgram` and the link fails the previously valid program is still set on `someProgram.validProgram` and is still usable and locations are still valid.

It is only after a **successful** link that `someProgram.validProgram` is updated and that locations become invalid

So, believing this I wrote a test

https://jsfiddle.net/greggman/qkn3mju9/

---

Firefox shows the behavior I expect

<img width="658" alt="Screen Shot 2019-03-29 at 12 51 02" src="https://user-images.githubusercontent.com/234804/55258866-54b42000-5221-11e9-9f94-3ce51fbc5b43.png">

---

Chrome, after calling `gl.linkProgram` with a failed link appears to invalidate the locations but keep the `validProgram` so I can no longer set uniforms but I can still render

<img width="603" alt="Screen Shot 2019-03-29 at 12 52 30" src="https://user-images.githubusercontent.com/234804/55258958-9349da80-5221-11e9-8701-07789388eb56.png">

---

Safari appears to make the failed link make the program unusable

<img width="618" alt="Screen Shot 2019-03-29 at 12 54 01" src="https://user-images.githubusercontent.com/234804/55259015-beccc500-5221-11e9-8411-57002f42cbd8.png">

[...]

PR #2879, opened 2019-04-30
Port greggman new program handling test

shrekshao:
Mostly a port from @greggman 's program handling test https://github.com/KhronosGroup/WebGL/issues/2842

There is a section doing similar tests in `program-test.html`. But since it only detects gl errors without actually drawing something, it couldn't really expose problems. I removed that one test case from there. The new `program-handling.html` tests covers better for the topic of program handling for various linking situation.

### DIFF
--- a/sdk/tests/conformance/programs/00_test_list.txt
+++ b/sdk/tests/conformance/programs/00_test_list.txt
@@ -6,7 +6,7 @@ gl-get-active-uniform.html
 gl-getshadersource.html
 gl-shader-test.html
 invalid-UTF-16.html
-program-handling.html
+--min-version 1.0.4 program-handling.html
 --min-version 1.0.4 program-infolog.html
 program-test.html
 --min-version 1.0.2 use-program-crash-with-discard-in-fragment-shader.html
--- a/sdk/tests/conformance/programs/00_test_list.txt
+++ b/sdk/tests/conformance/programs/00_test_list.txt
@@ -6,6 +6,7 @@ gl-get-active-uniform.html
 gl-getshadersource.html
 gl-shader-test.html
 invalid-UTF-16.html
+program-handling.html
 --min-version 1.0.4 program-infolog.html
 program-test.html
 --min-version 1.0.2 use-program-crash-with-discard-in-fragment-shader.html
--- a/sdk/tests/conformance/programs/program-handling.html
+++ b/sdk/tests/conformance/programs/program-handling.html
@@ -1,28 +1,7 @@
 <!--
-
-/*
-** Copyright (c) 2019 The Khronos Group Inc.
-**
-** Permission is hereby granted, free of charge, to any person obtaining a
-** copy of this software and/or associated documentation files (the
-** "Materials"), to deal in the Materials without restriction, including
-** without limitation the rights to use, copy, modify, merge, publish,
-** distribute, sublicense, and/or sell copies of the Materials, and to
-** permit persons to whom the Materials are furnished to do so, subject to
-** the following conditions:
-**
-** The above copyright notice and this permission notice shall be included
-** in all copies or substantial portions of the Materials.
-**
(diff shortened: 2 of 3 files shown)


sdk/tests/conformance/programs/program-handling.html
```diff
@@ -0,0 +1,143 @@
+<!--
+
+/*
+** Copyright (c) 2019 The Khronos Group Inc.
```
jdashg: There's a new shorter copyright block:
https://github.com/KhronosGroup/WebGL/pull/2846/files

sdk/tests/conformance/programs/program-test.html
```diff
@@ -331,10 +331,6 @@
     assertMsg(gl.getProgramParameter(progGood2, gl.LINK_STATUS) == false,
               "linking should fail with in-use formerly good program, with new bad shader attached");
 
-    // Invalid link leaves previous valid program intact.
```
jdashg: I suppose we can remove this in favor of the more detailed tests, but I generally don't like removing tests like this.

shrekshao: So you prefer to leave the old test unchanged (restore the deleted section)? (Just to double check I understand correctly)

jdashg: I prefer leaving the old test, but I don't mind strongly.

sdk/tests/conformance/programs/00_test_list.txt
```diff
@@ -6,6 +6,7 @@ gl-get-active-uniform.html
 gl-getshadersource.html
 gl-shader-test.html
 invalid-UTF-16.html
+program-handling.html
```
jdashg: `--min-version 1.0.4` for new tests, sorry!

sdk/tests/conformance/programs/program-test.html
shrekshao: I have one more reason to delete this as current chromium implementation didn't handle this properly. Separate this part of testing into the new program-handling.html would make suppression a bit easier for us. So I think I will keep it removed.

sdk/tests/conformance/programs/00_test_list.txt
shrekshao: Thanks for the instruction

sdk/tests/conformance/programs/program-test.html
jdashg: I don't like this sort of gamesmanship. Please focus on fixing the failing subtests, and don't focus on overall pass rate.

shrekshao: FYI I'm working on fixing the handling issue at the same time.

shrekshao: Restored.

jdashg:
Thanks!

kenrussell:
Nice work Shrek. LGTM after the fact.

kainino0x:
Sorry if I'm misunderstanding, I'm not following this closely. If I understood correctly, the point of having the failure in a separate test was not to improve pass rate, but to allow us to suppress the failure without losing test coverage of the whole file. Were things landed in a way that doesn't lose us test coverage?

shrekshao:
Thanks Kai for clarifying this. Jeff would this change your mind? I could
open a separate pull request

jdashg:
As you can see in this PR, it only lands the new test. Pre-existing tests are not modified, so suppression hasn't changed here.
Google is free to modify the conformance test checkout they use for CI. (Firefox has definitely done this in the past)
That, or fixing this incorrect behavior, should be the primary approaches.

kainino0x:
Ok, it doesn't affect the existing coverage. We don't normally modify our conformance test checkout - our workflow would make this hard and it would make pulling CTS updates more complex.

I'm just talking about taking exactly the existing tests and splitting them into two files so the suppressions can be more targeted. If I understood correctly, in this case that would improve our coverage.

Anyway, I'm not dead set on it, but I think that being able to have more targeted suppressions is always better.

PR #2884, opened 2019-05-07
Add compliance test for UBOs sized over 65536

JasperNV:
https://bugs.chromium.org/p/angleproject/issues/detail?id=3388

### DIFF
--- a/sdk/tests/conformance2/uniforms/large-uniform-buffers.html
+++ b/sdk/tests/conformance2/uniforms/large-uniform-buffers.html
@@ -1,0 +1,138 @@
+<!--
+Copyright (c) 2019 The Khronos Group Inc.
+Use of this source code is governed by an MIT-style license that can be
+found in the LICENSE.txt file.
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL large uniform buffers conformance tests</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+<script id="vshader" type="x-shader/x-vertex">#version 300 es
+in vec4 a_vertex;
+void main(void) {
+  gl_Position = a_vertex;
+}
+</script>
+<script id="fshader" type="x-shader/x-fragment">#version 300 es
+precision mediump float;
+layout (std140) uniform color_ubo {
+  vec4 color;
+};
+out vec4 fragColor;
+void main(void) {
+  fragColor = color;
+}
+</script>
+</head>
+<body>
+<canvas id="example" width="100", height="100"></canvas>
+<div id="description"></div>
+<div id="console"></div>
+<script>
+"use strict";
+debug("");
+description("Test for UBOs sized over 65536 - https://bugs.chromium.org/p/angleproject/issues/detail?id=3388");
+
+var wtu = WebGLTestUtils;
+var gl = wtu.create3DContext("example", undefined, 2);
+
+function runTest() {
+    var program = wtu.setupProgram(gl, ['vshader', 'fshader'], ['a_vertex']);
+    var uboIndex = gl.INVALID_INDEX;
+    if (program)
+        uboIndex = gl.getUniformBlockIndex(program, "color_ubo");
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/uniforms/large-uniform-buffers.html
```diff
@@ -0,0 +1,155 @@
...
+    var uboData = new Float32Array(0x20000);
+
+    var offs0 = 0x00000;
+    // Red
+    uboData[offs0 + 0] = 1;
+    uboData[offs0 + 1] = 0;
+    uboData[offs0 + 2] = 0;
+    uboData[offs0 + 3] = 1;
+
+    var offs1 = 0x10000;
```
kenrussell: offs1 has to be computed as a multiple of gl.getParameter(UNIFORM_BUFFER_OFFSET_ALIGNMENT), or the test might be incorrect on some platforms.

sdk/tests/conformance2/uniforms/large-uniform-buffers.html
```diff
@@ -0,0 +1,155 @@
...
+
+    var uboDataSize = gl.getActiveUniformBlockParameter(
+        program, uboIndex, gl.UNIFORM_BLOCK_DATA_SIZE);
+    if (uboDataSize == 0) {
+        testFailed("uniform block data size invalid");
+        return;
+    }
+
+    var uboBuf = gl.createBuffer();
+    var uboData = new Float32Array(0x20000);
```
kenrussell: There aren't any queries of gl.getParameter(MAX_UNIFORM_BLOCK_SIZE) in this test. Should there be? I realize that the program is only reading 4 floats out of the UBO. We have seen this 65K UBO size limit on other platforms too (macOS) and it's accounted for in the UBO rendering path in this project for example:
https://github.com/kenrussell/webgl-animometer

sdk/tests/conformance2/uniforms/large-uniform-buffers.html
JasperNV: The maximum alignment ever seen in the wild is 256 (by NVIDIA GPUs): http://opengl.gpuinfo.org/displaycapability.php?name=GL_UNIFORM_BUFFER_OFFSET_ALIGNMENT , and there is also GLES spec language capping it to 256. The `65536` size isn't an OpenGL limit, it's a Direct11 cbuffer limit, so there's no easy way I can compute this offset from the details given to me from GLES alone.

sdk/tests/conformance2/uniforms/large-uniform-buffers.html
JasperNV: Based on my reading of the GLES spec, GL_MAX_UNIFORM_BLOCK_SIZE describes the maximum size of any individual GL block used by the shader. It does *not* describe the maximum size of the overall buffer. We are only binding 16 bytes from the buffer to the block here.

sdk/tests/conformance2/uniforms/large-uniform-buffers.html
kenrussell: At least could you please add an assertion that offs1 is evenly divisible by UNIFORM_BUFFER_OFFSET_ALIGNMENT? Thanks.

sdk/tests/conformance2/uniforms/large-uniform-buffers.html
kenrussell: Thanks for confirming. That's my understanding too.

jdashg:
Don't worry too much about it. I figure I'll circle around down the road and fix-up some stragglers.

kenrussell:
OK, thanks @jdashg .

sdk/tests/conformance2/uniforms/large-uniform-buffers.html
JasperNV: Sounds good. Any recommended ways to test an assertion? I'll tackle this tomorrow.

kenrussell: const alignment = gl.getParameter(gl.UNIFORM_BUFFER_OFFSET_ALIGNMENT);
if (offs1 % alignment != 0) {
  testFailed("Platform has a strange uniform buffer offset alignment: " + alignment);

or similar.

JasperNV:
Adjusted for feedback; apologies for the delay.

PR #2885, opened 2019-05-07
Use smaller copyright block in sdk/tests/*.

jdashg:


jdashg:
Here's the script I used, since sed seems to be tough with multiline matching:
```
#! /usr/bin/env python3
assert __name__ == '__main__'

import pathlib
import re
import sys

(path_str,) = sys.argv[1:]

path = pathlib.Path(path_str)
data = path.read_bytes()
#print(data[:1200])

RE_COMMENT_STR = b'/[*]\n[*][*] Copyright [(]c[)].*IN THE MATERIALS[.]\n[*]/'
RE_TAG_OR_COMMENT = re.compile(b'(<!--\\n*)?' + RE_COMMENT_STR + b'(\\n*-->)?', re.DOTALL)

SMALL_BLOCK = b'''
Copyright (c) 2019 The Khronos Group Inc.
Use of this source code is governed by an MIT-style license that can be
found in the LICENSE.txt file.
'''.strip()

m = RE_TAG_OR_COMMENT.search(data)
#print('m', m, file=sys.stderr)
if m:
  found = m.group()
  replacement = SMALL_BLOCK
  if found.startswith(b'<!--'):
    replacement = b'<!--\n' + replacement + b'\n-->'
  else:
    replacement = b'/*\n' + replacement + b'\n*/'

  data = data[:m.start()] + replacement + data[m.end():]
  #sys.stdout.buffer.write(data[:1200])
  path.write_bytes(data)
```

jdashg:
Also, this isn't all of them.

jdashg:
@kenrussell We should probably be preserving the year, shouldn't we?

jdashg:
Ok, that's all of them.

kenrussell:
LGTM

The Travis failures are fragility in the lint script which are being fixed in #2886 .

PR #2886, opened 2019-05-08
Whitelist files containing Unicode characters.

kenrussell:
Two files in the repository deliberately contain non-UTF-8 Unicode
characters. Explicitly whitelist them so if other files later accidentally
add such characters, they're discovered.

Related to #2885.

### DIFF
--- a/sdk/tests/py/lint/lint.py
+++ b/sdk/tests/py/lint/lint.py
@@ -146,10 +146,13 @@ def check_regexp_line(path, f):
 
     applicable_regexps = [regexp for regexp in regexps if regexp.applies(path)]
 
-    for i, line in enumerate(f):
-        for regexp in applicable_regexps:
-            if regexp.search(line):
-                errors.append((regexp.error, "%s line %i" % (path, i+1), i+1))
+    try:
+        for i, line in enumerate(f):
+            for regexp in applicable_regexps:
+                if regexp.search(line):
+                    errors.append((regexp.error, "%s line %i" % (path, i+1), i+1))
+    except UnicodeDecodeError as e:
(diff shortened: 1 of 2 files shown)


kenrussell:
@jdashg PTAL. Tested with your PR #2885.

sdk/tests/py/lint/lint.py
```diff
@@ -146,10 +146,13 @@ def check_regexp_line(path, f):
...
-        for regexp in applicable_regexps:
-            if regexp.search(line):
-                errors.append((regexp.error, "%s line %i" % (path, i+1), i+1))
+    try:
+        for i, line in enumerate(f):
+            for regexp in applicable_regexps:
+                if regexp.search(line):
+                    errors.append((regexp.error, "%s line %i" % (path, i+1), i+1))
+    except UnicodeDecodeError as e:
+        return [("CONTAINS UNICODE", "File %s contains Unicode characters" % path, None)]
```
jdashg: Wouldn't UnicodeDecodeError mean it contains /invalid/ unicode, not just unicode?

kenrussell: It seems these files are encoded with something called "extended ASCII" and not UTF-8, which Python doesn't recognize. I'm hesitant to touch their contents less I accidentally negate the tests.

jdashg: Alright! Can we call this CONTAINS NON-UNICODE then?

jdashg: or INVALID UNICODE would also work for me.

kenrussell: Changed to INVALID UNICODE. Thanks.

kenrussell:
Please take another look. Thanks.

PR #2887, opened 2019-05-10
Allow Uint8ClampedArray in readPixels and tex{Sub}Image.

kenrussell:
Where Uint8Array is allowed per spec, also allow Uint8ClampedArray.
This makes interaction with ImageData easier.

Firefox already allows Uint8ClampedArray in texImage2D where
Uint8Array is allowed, so this is a difference in behavior between
implementations that's now being clarified.

Reported in http://crbug.com/961658 .

### DIFF
--- a/sdk/tests/conformance/reading/read-pixels-test.html
+++ b/sdk/tests/conformance/reading/read-pixels-test.html
@@ -307,9 +307,36 @@ function runTest(canvas, antialias) {
       shouldBe("actual", "expected");
     }
     wtu.glErrorShouldBe(gl, gl.NO_ERROR, "there should be no GL errors");
+
+    debug("");
+    debug("check readback into Uint8ClampedArray");
+    continueTestFunc = continueTestPart4;
+    const kSize = 32;
+    canvas.width = kSize;
+    canvas.height = kSize;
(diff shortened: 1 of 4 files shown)


kenrussell:
@jdashg please review.

Firefox already passes the revised texture-formats-test, so this is an area that needs to be clarified between implementations.

sdk/tests/conformance/reading/read-pixels-test.html
```diff
@@ -307,9 +307,36 @@
...
+    canvas.width = kSize;
+    canvas.height = kSize;
+    if (gl.getError() != gl.CONTEXT_LOST_WEBGL) {
+      continueTestPart4();
+    }
+  }
+
+  function continueTestPart4() {
+    const kSize = 32;
+    gl.viewport(0, 0, kSize, kSize);
```
jdashg: clear ignores viewport, fyi!

jdashg:
Thanks!

sdk/tests/conformance/reading/read-pixels-test.html
kenrussell: Yeah, was just resetting it since the earlier test set it to a large value.

PR #2890, opened 2019-05-14
Add draw-buffer test against RASTERIZER_DISCARD state

shrekshao:


kenrussell:
Note that this change had to be made for compatibility with existing apps per http://crbug.com/962555 and http://crbug.com/962636 .

kainino0x:
Is this already covered by the spec or do we need a spec update?

jdashg:
I don't think this violates the spec, but I don't think it was previously required. We can formalize this interpretation, but it's a pretty reasonable interpretation to extrapolate from the rules.

PR #2891, opened 2019-05-14
Fix conformance/ogles after breakage from #2885.

jdashg:
These are generated files, and I've update the generation script.
Unfortunately, it looks like the ES conformance suite (now VK-GL-CTS)
has refactored the old mustpass.run files out of existance, so I'm not
sure we can trivially regenerate these tests. (though that's probably
ok)


kainino0x:
Fine with me to not regen. If we wanted, I'm sure we could find the old snapshot that has the same structure as when they were first generated.

PR #2894, opened 2019-05-17
Fix typo in depth renderbuffer initialization tests

Jiawei-Shao:
This patch fixes a typo in the depth renderbuffer initialization
tests that the depth renderbuffer should be initialized to 1.0
according to WebGL SPEC (WebGL 1.0, Chapter 4.1) instead of 0.


Jiawei-Shao:
@Richard-Yunchao @gyagp  PTAL, thanks!

PR #2895, opened 2019-05-17
Reject WEBGL_security_sensitive_resources.

dyna-dot:
Olli Etuaho, formerly of NVIDIA, demonstrated that it's possible to
infer textures' contents despite defenses such as these. No current
champion of this spec. Per discussion in WebGL WG, reject it.


kenrussell:
This is a spam pull request. The user has been reported.

PR #2898, opened 2019-05-30
Use three colors in tex-image tests

fritzk:
Using three colors makes it apparent if
the selected rectangle is upside down.

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
@@ -11,6 +11,7 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
     var successfullyParsed = false;
     var redColor = [255, 0, 0];
     var greenColor = [0, 255, 0];
+    var blueColor = [0, 0, 255];
     var repeatCount;
 
     function shouldRepeatTestForTextureFormat(internalFormat, pixelFormat, pixelType)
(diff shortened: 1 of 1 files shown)


jdashg:
Red and green are already stacked vertically. Can you be more precise about why you want this?

fritzk:
I would like to make sure that the selected rectangle is oriented correctly.  Currently there is nothing to say that the red or green rectangle is right side up or upside down after using texImage2D.  By adding a third color it the selected rectangle is now either red/blue or blue/green and the orientation can be derived.

fritzk:
Isn't that what tex-image-and-sub-image-2d-with-canvas-sub-rectangle.js is doing now?  Using UNPACK_SKIP_PIXELS and UNPACK_SKIP_ROWS?

I was working on this today already.  If you still want it back in tex-image-and-sub-image-2d-with-webgl-canvas.js I can start work on figuring out how to add a optional sub-rectangle to the test case.

kenrussell:
Looking again at tex-image-and-sub-image-2d-with-canvas-sub-rectangle.js, I see that it's in fact testing both 2D canvas and WebGL sources. Was concerned that it might be missing some cases.

If you're convinced that the existing tests catch all possible errors in the implementation then it's ok to not make any changes to the tests.

kenrussell:
@fritzk could you please confirm whether this CL is really needed, or whether tex-image-and-sub-image-2d-with-canvas-sub-rectangle already covered these cases (and caught the associated regressions)? Thanks.

kenrussell:
@fritzk confirmed offline that this change isn't strictly needed - the scenario is covered by the other tests - so I'm closing this. Sorry for creating unnecessary work for you Fritz.

PR #2899, opened 2019-06-07
Ext float blend test for WebGL 1

shrekshao:
Some test results:
* :heavy_check_mark: Linux Chrome 75 (OpenGL Backend)
* :heavy_check_mark: Linux Firefox Nightly
* :heavy_check_mark: Win Chromium local build with ext-float-blend enabled for D3D9 (Angle, validating cmd decoder, D3D9 Backend)
* :heavy_check_mark: Win Chrome 75 (Angle, validating cmd decoder, D3D11, OpenGL Backend)

Failing ones: RGBA float32 fbo color attachment is said to be unsupported, but it should be supported if WEBGL_color_buffer_float is available https://www.khronos.org/registry/webgl/extensions/WEBGL_color_buffer_float/ (RGB float32 is optional)

* :heavy_multiplication_x: Win Chrome 75 (Angle, passthrough cmd decoder, D3D11, OpenGL Backend)
* :heavy_multiplication_x: Win Firefox Nightly

### DIFF
--- a/sdk/tests/conformance/extensions/ext-float-blend.html
+++ b/sdk/tests/conformance/extensions/ext-float-blend.html
@@ -1,0 +1,86 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL EXT_float_blend Conformance Tests</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+<script src="../../js/tests/ext-float-blend.js"></script>
+</head>
+<body>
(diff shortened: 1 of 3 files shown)


sdk/tests/conformance/extensions/ext-float-blend.html
```diff
@@ -0,0 +1,86 @@
+<!DOCTYPE html>
```
jdashg: You didn't add this file to the test list file.
Also, I would really prefer merging the webgl1 and 2 versions of this test, if it's not toooo much work.

shrekshao: :man_facepalming: My bad. A new pull request is here: https://github.com/KhronosGroup/WebGL/pull/2900

PR #2900, opened 2019-06-08
Merge WebGL 1 and 2 Ext-float-blend tests

shrekshao:
Also fixed 00_test_list at the same time


kenrussell:
LGTM assuming this has been tested locally on both WebGL 1.0 and 2.0. Nice work unifying the tests.

PR #2901, opened 2019-06-11
Add a test for #extension after non-preprocessor code in ESSL 1.00

kainino0x:
Regression test for https://crbug.com/971660 .

See comment for details.


kenrussell:
LGTM. Please merge yourself as soon as the CI has passed.

CC @grorg @jdashg - this is needed ASAP as it's the only regression test for this important bug.

PR #2905, opened 2019-06-17
Fix MRT EXT_float_blend test for WebGL 1

shrekshao:
ES 2.0 requires that all color buffers attached to application-created framebuffer objects must have the same number of bitplanes.
Thus modify the MRT part of EXT_float_blend test for WebGL 1.


kenrussell:
Nice work tracking down this limitation Shrek. LGTM

PR #2910, opened 2019-07-12
Add test forbidding sampler-array-indexing via constant-index-express…

jdashg:
…ion in essl300.


jdashg:
ESSL3.00 p143-144:
> Indexing of arrays of samplers by constant-index-expressions is supported in GLSL ES 1.00. A constant-index-expression is an expression formed from constant-expressions and certain loop indices, defined for a subset of loop constructs.  Should this functionality be included in GLSL ES 3.00?
> 
> RESOLUTION:  No.  Arrays of samplers may only be indexed by constant-integral-expressions.

kenrussell:
LGTM. Thanks Jeff for putting together this test case.

PR #2911, opened 2019-07-16
Set preload='auto' flag on videos uploaded to WebGL textures.

kenrussell:
This is an attempt to see whether this eliminates flakiness in
video-related WebGL tests, per suggestion from Chrome's media team in
https://crbug.com/979444#c20 .


kenrussell:
@jdashg @kainino0x please take a look.

Not attempting to work around browser bugs; just trying to further diagnose intermittent failures seen in these tests. Will follow up with the results of this experiment, and revert if this didn't have any positive effect.

kenrussell:
Thanks. Let me move forward with this PR so we can test its results.

PR #2913, opened 2019-07-18
Increase timeout on long-loop-compile.html.

kenrussell:
This test assumes the HLSL compiler will finish within 5000 ms for a
long loop which it apparently has to unroll. On some slower Windows
bots this timeout is occasionally being hit, so increase the timeout
50%, to 7500 ms.

Related to http://crbug.com/691951.

### DIFF
--- a/sdk/tests/conformance/glsl/misc/large-loop-compile.html
+++ b/sdk/tests/conformance/glsl/misc/large-loop-compile.html
@@ -156,10 +156,11 @@ if (!gl) {
   setTimeout(function() {
     if (!receivedContextLost) {
       testPassed("Large loop compiled and linked without terminating the WebGL context");
-      if (endTime - startTime < 5000) {
-        testPassed("Shader compilation completed in a reasonable amount of time");
+      let timeString = ": " + (endTime - startTime) + " ms";
+      if (endTime - startTime < 7500) {
+        testPassed("Shader compilation completed in a reasonable amount of time" + timeString);
       } else {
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/glsl/misc/large-loop-compile.html
```diff
@@ -156,10 +156,11 @@
   setTimeout(function() {
     if (!receivedContextLost) {
       testPassed("Large loop compiled and linked without terminating the WebGL context");
-      if (endTime - startTime < 5000) {
-        testPassed("Shader compilation completed in a reasonable amount of time");
+      let timeString = ": " + (endTime - startTime) + " ms";
```
kainino0x: nit: const. Also, you could choose to use
```
`Shader compilation ...: ${deltaTime}ms`
```
if you want.

sdk/tests/conformance/glsl/misc/large-loop-compile.html
```diff
@@ -156,10 +156,11 @@
...
+      if (endTime - startTime < 7500) {
+        testPassed("Shader compilation completed in a reasonable amount of time" + timeString);
       } else {
-        testFailed("Shader compilation took an unreasonably long time");
+        testFailed("Shader compilation took an unreasonably long time" + timeString);
```
jdashg: I would make both of these ` time: " + timeString`.

kenrussell:
Thanks for suggestions; incorporated.

Issue #2914, opened 2019-07-20
BC1 RGB on ANGLE


lexaknyazev:
`WEBGL_compressed_texture_s3tc` and `WEBGL_compressed_texture_s3tc_srgb` extensions expose enums for BC1 RGB compressed format (`COMPRESSED_RGB_S3TC_DXT1_EXT` and `COMPRESSED_SRGB_S3TC_DXT1_EXT` respectively).

These flavors of BC1 are not supported on D3D so ANGLE maps them to BC1 RGBA. This leads to inconsistency with OpenGL. Specifically, pixels that BC1 RGB decodes to `(0, 0, 0, 1)` are decoded to `(0, 0, 0, 0)`.

Here's a repro case (the left square should be solid black):
```html
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8">
</head>
<body onload="init()">
  <canvas id="cnv1" width="64" height="64"
    style="border: 1px solid black; background-color:#CCC"></canvas>
  <canvas id="cnv2" width="64" height="64"
    style="border: 1px solid black; background-color:#CCC"></canvas>
  <script>
    function init() {
      draw('cnv1', false);
      draw('cnv2', true);
    }

    function draw(cnvId, rgba) {
      var cnv = document.getElementById(cnvId);
      var gl = cnv.getContext('webgl');
      var ext = gl.getExtension('WEBGL_compressed_texture_s3tc');

      // Setup program
      var vss =
        'attribute vec2 a_pos; varying vec2 v_tex; void main() {gl_Position = vec4(a_pos.xy, 0.0, 1.0); v_tex = a_pos * 0.5 + 0.5;}';
      var vs = gl.createShader(gl.VERTEX_SHADER);
      gl.shaderSource(vs, vss);
      gl.compileShader(vs);

      var fss =
        'precision highp float; varying vec2 v_tex; uniform sampler2D u_tex; void main() {gl_FragColor = texture2D(u_tex, v_tex);}';
      var fs = gl.createShader(gl.FRAGMENT_SHADER);
      gl.shaderSource(fs, fss);
      gl.compileShader(fs);

      var prog = gl.createProgram();
      gl.attachShader(prog, vs);
      gl.attachShader(prog, fs);
      gl.linkProgram(prog);
      gl.bindAttribLocation(prog, 0, 'a_pos');
      gl.enableVertexAttribArray(0);
      gl.useProgram(prog);
[...]

PR #2915, opened 2019-07-23
Update tests for BC1 RGB

lexaknyazev:
This PR modifies tests for BC1 RGB (both linear and sRGB) so that the test payload covers the case where BC1 RGB and BC1 RGBA compressed texture formats produce different output for the same data (see #2914).

Also, the test visualization is updated to always show produced alpha channel value.

Results are expected since ANGLE on D3D doesn't properly support BC1 RGB ([bug](https://bugs.chromium.org/p/angleproject/issues/detail?id=3729)):

- Firefox 68 (Win10)
  - ANGLE: FAIL
  - OpenGL: PASS
- Chrome 77 (Win10)
  - ANGLE on D3D11: FAIL
  - ANGLE on OpenGL: PASS

As a side note, Chrome 77 doesn't expose `WEBGL_compressed_texture_s3tc_srgb` extension when ANGLE is running on OpenGL.

@kenrussell, PTAL.


kenrussell:
Sorry for the long delay reviewing. We'll discuss this on the WebGL conference call this Thursday.

kenrussell:
Let's discuss on tomorrow's working group conference call.

lexaknyazev:
@jdashg @kenrussell 
Here're all mentioned alternatives for your convenience:

| Mitigation | Pros | Cons |
|---------------------------------------------------------|---------------------------------|------------------------------|
| Unpack to RGBA8 | Correct sampling | x8 memory<br>Heavy CPU usage |
| Extend each block with 8 zeros, upload as BC2 or BC3 | Correct sampling | x2 memory |
| Track usage and patch shaders (emulate `rgb1`swizzling) | Correct sampling | Hard to implement |
| Relax the WebGL extensions | No ANGLE changes | Incorrect sampling on D3D |
| Disallow BC1 RGB on ANGLE with D3D | Correct sampling (when allowed) | Breaking change |

lexaknyazev:
Given that this issue is relevant only to D3D, it may be reasonable to choose a mitigation that is aligned with previous work on this platform. GL-to-D3D inevitable conversions so far:

**Packed**
- `GL_RGBA4`
  - `R8G8B8A8_UNORM` (unpacking, x2) or
  - `B4G4R4A4_UNORM` (reordering, requires DXGI 1.2);
- `GL_RGB5_A1`
  - `R8G8B8A8_UNORM` (unpacking, x2) or
  - `B5G5R5A1_UNORM` (reordering, requires DXGI 1.2).

**Three-component 8-bit**
- `GL_RGB8` -> `R8G8B8A8_UNORM` (extending, x1.333);
- `GL_RGB8_SNORM` -> `R8G8B8A8_SNORM` (extending, x1.333);
- `GL_RGB8I` -> `R8G8B8A8_SINT` (extending, x1.333);
- `GL_RGB8UI` -> `R8G8B8A8_UINT` (extending, x1.333);
- `GL_SRGB8` -> `R8G8B8A8_UNORM_SRGB` (extending, x1.333).

**Three-component 16-bit**
- `GL_RGB16` -> `R16G16B16A16_UNORM` (extending, x1.333);
- `GL_RGB16_SNORM` -> `R16G16B16A16_SNORM` (extending, x1.333);
- `GL_RGB16I` -> `R16G16B16A16_SINT` (extending, x1.333);
- `GL_RGB16UI` -> `R16G16B16A16_UINT` (extending, x1.333);
- `GL_RGB16F` -> `R16G16B16A16_FLOAT` (extending, x1.333).

**Depth and Stencil**
- `GL_DEPTH_COMPONENT24` -> `D24_UNORM_S8_UINT` (extending, x1.333);
- `GL_STENCIL_INDEX8` -> `D24_UNORM_S8_UINT` (extending, x4).

Given that these conversions are already in place, adding one more (BC1 -> BC2/BC3) to achieve conformant behavior doesn't look like a big deal.

Nevertheless, I'd add a note to the spec advising against using all these formats in desktop WebGL applications.

RafaelCintron:
The description of how BC1 is decompressed in D3D is described in [section 19.5.6](https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#19.5.6%20BC1{U|G}:%202(+2%20Derived)%20Opaque%20Colors%20or%202(+1%20Derived)%20Opaque%20Colors%20+%20Transparent%20Black) of the functional spec.  

Here is the relevant text. I've omitted the definition of `promoteToUNORM8`, which can be found in the spec.
> color_0_p = promoteToUNORM8(color_0)
color_1_p = promoteToUNORM8(color_1)
if (color_0 > color_1) // unsigned compare
{
    // Four-color block:
    color_2 = (2 * color_0_p + color_1_p) / 3;
    color_3 = (color_0_p + 2 * color_1_p) / 3;
    alpha_3 = 1.0f
} else {
    // Three-color block:
    color_2 = (color_0_p + color_1_p) / 2;
    color_3 = (0.0f,0.0f,0.0f); alpha_3 = 0.0f
}
> // color_*: Actually 3 independent calculations for R,G,B.

> // The following 2-bit codes select
// a UNORM8 color for each texel:
// (MSB)00(LSB) = color_0_p, alpha=1.0f
//      01      = color_1_p, alpha=1.0f
//      10      = color_2,   alpha=1.0f
//      11      = color_3,   alpha=alpha_3

The alpha channel ends up being either 0 or 1 depending on the values of the colors and the bits set in the block.  

OpenGL and D3D differ in regards to what alpha_3 ends up being if you specify COMPRESSED_RGB_S3TC_DXT1_EXT.  OpenGL requires the alpha channel always be 0 whereas D3D mandates 1.  
 
Though D3D12 allows you to map individual components of a view by specifying a [shader component mapping](https://github.com/KhronosGroup/WebGL/pull/2915), D3D11 does not have this ability.  

I would not be in favor of solutions that involve texture decompression.  Space and bandwidth are at a premium on low end machines, especially Hololens devices.  The less memory we use for compressed textures, the better. 

If OpenGL allows us to select individual components like D3D12 does, I would be in favor requiring that WebGL spec the alpha channel be 1.0 in the `color_0 > color_1` case I wrote above.

kenrussell:
@RafaelCintron thanks for the input. Per the WebGL working group's discussion this past Thursday, the direction being pursued is to modify the incoming BC1 bitstream to promote it to BC2 or BC3. This will increase the size of the compressed texture by 2x but will provide consistent results everywhere. See http://crbug.com/angleproject/3729 .

lexaknyazev:
Apparently, extending BC1 RGB to BC2/BC3 with opaque alpha does not work. According to the specs, BC2 and BC3 change the semantics of the RGB data block in a way that the troublesome case will evaluate to completely different values.

Here's the worst-case example.

**BC1 block**
```
0x00, 0x00, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF
```
- `color_0 = (0, 0, 0)`
- `color_1 = (255, 255, 255)`
- All selectors are set to 3.

**Decoded as BC1 RGB**
- `(color_0 <= color_1)` so output color is `[0, 0, 0, 255]` - opaque black.

**Decoded as BC1 RGBA**
- `(color_0 <= color_1)` so output color is `[0, 0, 0, 0]` - transparent black.

**Decoded as BC2/BC3 RGBA (with prepended opaque alpha block)**
- No special case here, so output color is `[170, 170, 170, 255]` - opaque grey.

---

This unfortunately leaves only two options:
- relax the WebGL spec and strongly discourage use of BC1 RGB, or
- completely remove support for BC1 RGB from the extension.

kenrussell:
Thanks for investigating that Alexey. Can we try to collectively double-check that there's no way to do that? https://docs.microsoft.com/en-us/windows/win32/direct3d10/d3d10-graphics-programming-guide-resources-block-compression doesn't seem to indicate there's a difference in the evaluation of the color channels between BC1/BC3. There does seem to be a special case for BC3's alpha values on that page which will set one of the alphas to 0 and the other to 255, interpolating the rest, if alpha_1 >= alpha_0, and I wonder if that's the case being encountered here. What if alpha_0 is set to 254 and alpha_1 to 255? Would that produce "close enough" answers?

lexaknyazev:
Alpha interpolation in BC3 is independent of color channels.

The difference is in color block interpretation between BC1 and BC2/3 (as can be seen from the D3D functional spec linked above).

lexaknyazev:
Also the text in D3D10 spec on BC1-3 formats is not precise enough. The complete spec can be found in the D3D11 functional spec (linked by @RafaelCintron) or in the S3TC OpenGL extension.

RafaelCintron:
For reference, here is the spec for the BC2 format, per the D3D functional spec.  

> // Four-color block: derive the other two colors
color_0_p = promoteToUNORM8(color_0)
color_1_p = promoteToUNORM8(color_1)
color_2 = (2 * color_0 + color_1) / 3;
color_3 = (color_0 + 2 * color_1) / 3;
// color_*: Actually 3 independent calculations for R,G,B.

> // The following 2-bit codes select
// a UNORM8 color for each texel:
// (MSB)00(LSB) = color_0_p,
//      01      = color_1_p,
//      10      = color_2,
//      11      = color_3

> // Derive alpha value for texel t:
alpha = alpha[t]/15.0f

I would be in favor of either relaxing the WebGL spec  for changing the spec such that BC1 RGB == BC1 RGBA for all platforms.

kenrussell:
Resolved during WebGL F2F 2019-09-25: change spec to allow either alpha 0.0 or 1.0 when sampling these BC1 textures, to accommodate differences between D3D11 and OpenGL specifications.

Alexey, could you please update the extension spec and the test to tightly cover this behavior? Thank you very much for your help.

lexaknyazev:
Sure.

What's the WG opinion of warning against some uncompressed formats as listed in https://github.com/KhronosGroup/WebGL/pull/2915#issuecomment-526829900?
For instance, loading three-component data on D3D11 usually invokes `LoadToNative3To4` function that loops over every pixel.

jdashg:
That's a great idea. I'll file an issue.

kenrussell:
@lexaknyazev this test still needs to be updated to assert that the alpha channel is either 0.0 or 1.0, but nothing else. Could you please make that change? Thanks.

kenrussell:
Actually, the extension spec needs to be updated as well. Could you please modify this pull request to include the extension update? Thanks very much.

lexaknyazev:
> tests still need to be updated

Of course they do. I just pushed specs update first. Tests update is WIP, as I want them to cover every possible decoder branch. The same code will be reused for RGTC (technically, BC4 is a slightly enhanced version of BC3 alpha block).

lexaknyazev:
@kenrussell 
Please review the updated tests for the S3TC extension. If everything's fine, I'll apply the same changes to S3TC sRGB tests.

extensions/WEBGL_compressed_texture_s3tc/extension.xml
```diff
@@ -24,6 +24,14 @@
         <code>COMPRESSED_RGBA_S3TC_DXT1_EXT</code>, <code>COMPRESSED_RGBA_S3TC_DXT3_EXT</code>, and
         <code>COMPRESSED_RGBA_S3TC_DXT5_EXT</code> may be passed to
         the <code>compressedTexImage2D</code> and <code>compressedTexSubImage2D</code> entry points.
+
+        <div class="note">
+          Some platforms do not have native support for <code>COMPRESSED_RGB_S3TC_DXT1_EXT</code> format,
+          so WebGL implementations may map it to <code>COMPRESSED_RGBA_S3TC_DXT1_EXT</code>.
```
kenrussell: Could you please describe the consequences here, like you did in the comment in the test case below?

"This means the alpha channel is not guaranteed to be opaque (0xFF); it may be transparent (0x0) in some implementations."

extensions/WEBGL_compressed_texture_s3tc_srgb/extension.xml
```diff
@@ -24,6 +24,14 @@
         <code>COMPRESSED_SRGB_ALPHA_S3TC_DXT1_EXT</code>, <code>COMPRESSED_SRGB_ALPHA_S3TC_DXT3_EXT</code>, and
         <code>COMPRESSED_SRGB_ALPHA_S3TC_DXT5_EXT</code> may be passed to
         the <code>compressedTexImage2D</code> and <code>compressedTexSubImage2D</code> entry points.
+
+        <div class="note">
+          Some platforms do not have native support for <code>COMPRESSED_SRGB_S3TC_DXT1_EXT</code> format,
+          so WebGL implementations may map it to <code>COMPRESSED_SRGB_ALPHA_S3TC_DXT1_EXT</code>.
+          Applications expecting maximum portability should avoid authoring and uploading textures of
```
kenrussell: Same sentence here.

lexaknyazev:
@kenrussell 
Have you had a chance to review the updated text?

kenrussell:
> @kenrussell
> Have you had a chance to review the updated text?

The new text looks good. (When you push a new commit please comment indicating that you've done so, so that email notifications get sent out; thanks.)

Is this ready to be merged at this point, or are any more changes to the tests incoming?

lexaknyazev:
Pushed more updates. This should be ready to be merged.
- Fully aligned linear and sRGB tests - now they operate alike (added PBO tests to sRGB).
- Using the same payloads for linear and sRGB tests. The latter must look darker. Previously, both the looks and the data were different.
- PBO test now clears the texture before the second run. As a result, a new Firefox issue was discovered (#2994).
- Sub-region tests now use correct uncompressed data for zero-filled textures.

### Results
#### Linear transfer

| | DXT1 RGB | DXT1 RGBA | DXT3 | DXT5 |
|--------------------------|:----------------:|:-----------:|:------:|:------:|
| Chrome / D3D11 / NVIDIA | ➡️ RGBA | ✔️ | ✔️ | ✔️ |
| Chrome / D3D11 / Intel | ➡️RGBA | ✔️ | ✔️ | ✔️ |
| Chrome / GL / NVIDIA | ✔️ | ✔️ | ✔️ | ✔️ |
| Chrome / SwiftShader | ➡️ RGBA | ✔️ | ✔️ | ✔️ |
| Firefox / D3D11 / NVIDIA | ➡️RGBA | ✔️ | ✔️ | ✔️ |
| Firefox / D3D11 / Intel | ➡️RGBA | ✔️ | ✔️ | ✔️ |
| Firefox / GL / NVIDIA | ✔️ | ✔️ | ✔️ | ✔️ |
| Firefox / GL / Intel | ✔️ | ✔️ | ✔️ | ✔️ |

#### sRGB transfer

| | DXT1 RGB | DXT1 RGBA | DXT3 | DXT5 |
|--------------------------|:----------------:|:-----------:|:------:|:------:|
| Chrome / D3D11 / NVIDIA | ➡️ RGBA | ✔️ | ✔️ | ✔️ |
| Chrome / D3D11 / Intel | ➡️RGBA | ✔️ | ✔️ | ✔️ |
| Chrome / GL / NVIDIA | ❌ | ❌ | ❌ | ❌ |
| Chrome / SwiftShader | ❌ | ❌ | ❌ | ❌ |
| Firefox / D3D11 / NVIDIA | ➡️RGBA | ✔️ | ✔️ | ✔️ |
| Firefox / D3D11 / Intel | ➡️RGBA | ✔️ | ✔️ | ✔️ |
| Firefox / GL / NVIDIA | ✔️ | ✔️ | ✔️ | ✔️ |
| Firefox / GL / Intel | ❗️ | ✔️ | ✔️ | ✔️ |

#### Notes
- Chrome (both stable and canary) does not expose `WEBGL_compressed_texture_s3tc_srgb` on GL / Windows and on SwiftShader.
- It should be possible to align SwiftShader's BC1 RGB implementation with the spec thus improving conformance for native apps.
- Intel HD Graphics 4000 (driver 10.18.10.5069) produces incorrect color values for BC1 SRGB when using Firefox / OpenGL. Chrome refuses to use this GPU with OpenGL altogether.

@kenrussell @jdashg PTAL

lexaknyazev:
@kenrussell 
FYI, issues opened following findings above:
- [SwiftShader BC1 RGB](https://issuetracker.google.com/issues/147143690)
- [SwiftShader S3TC sRGB](https://issuetracker.google.com/issues/147143691)
- [ANGLE S3TC sRGB on OpenGL](https://bugs.chromium.org/p/angleproject/issues/detail?id=4266)

kenrussell:
Thank you Alexey for enhancing the tests and reporting all of these issues!

Looks like Jeff may be swamped right now, so let me proceed with merging this now.

PR #2916, opened 2019-07-24
Promote OVR_multiview2 to Community Approved status.

jdashg:


kenrussell:
Approved during working group conference call of 8/15/2019.

Issue #2006, opened 2016-09-02
Interaction with sampler objects and EXT_texture_filter_anisotropic


juj:
In WebGL 2, sampler objects move texture sampling specific state from textures to sampler objects.

The extension https://www.khronos.org/registry/webgl/extensions/EXT_texture_filter_anisotropic/ adds a new texture sampling state variable to textures:

```
const GLenum TEXTURE_MAX_ANISOTROPY_EXT       = 0x84FE;
```

When sampler objects are used, should this state variable be available in the state of the sampler object, i.e. should it be possible to set the following?

```
webGL2context.samplerParameterf(samplerObject, anisoExt.TEXTURE_MAX_ANISOTROPY_EXT, 16.0);
```

Correspondingly, get the following?

```
webGL2context.getSamplerParameter(samplerObject, anisoExt.TEXTURE_MAX_ANISOTROPY_EXT, 'float');
```

Neither on native GLES3 or desktop OpenGL side specifications, I was not able to find a spec wording about this, although testing locally on native desktop OpenGL 4.4.0 (Windows+NVidia), this works. Curiously desktop OpenGL 4.5 core spec pdf does not mention the string "anisotropy" at all.

The wording text for the native side extension at https://www.opengl.org/registry/specs/EXT/texture_filter_anisotropic.txt predates sampler objects by a long time, and it is possible that the spec text was never updated nor a new spec was created to highlight this, but native drivers just went ahead to support it on their own.

kenrussell:
Thanks for reporting this. Anisotropic texture filtering has always been an OpenGL extension to the best of my knowledge, and it's a good point that its interactions with sampler objects may not be well defined even in the native specs.

WebGL's wrapper for EXT_texture_filter_anisotropic should define the interactions with WebGL 2.0, and include a conformance test. (This will require WebGL 2.0 implementations to be updated, of course.)

If you have the time to contribute a test under sdk/tests/conformance2/extensions/ which asserts the behavior (and I think it should be assumed that anisotropy is one of the properties contained in the sampler object), that would be greatly appreciated. Presumably this can be emulated reasonably easily in WebGL 2.0 implementations if it turns out that some drivers don't implement it properly.

Grief:
Here is the code:

        let gl = canvas.getContext("webgl2");
        let extAnisotropic = gl.getExtension('EXT_texture_filter_anisotropic');
        let sampler = gl.createSampler();
[...]

PR #2925, opened 2019-09-07
Add WebGL 2.0 EXT_texture_filter_anisotropic test.

kenrussell:
Refactors and reuses the previous WebGL 1.0 test, and adds tests for
setting and getting this parameter against sampler objects, which is
supposed to work per Issue 1 in the OpenGL extension specification:
https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_filter_anisotropic.txt

Both Chrome's and Firefox's WebGL 2.0 implementations currently fail
the positive sampler-related portions of this test.

Addresses #2006.

### DIFF
--- a/sdk/tests/conformance2/extensions/00_test_list.txt
+++ b/sdk/tests/conformance2/extensions/00_test_list.txt
@@ -1,5 +1,6 @@
 ext-color-buffer-float.html
 ext-disjoint-timer-query-webgl2.html
+--min-version 2.0.1 ext-texture-filter-anisotropic.html
 promoted-extensions.html
 promoted-extensions-in-shaders.html
 --min-version 2.0.1 ovr_multiview2.html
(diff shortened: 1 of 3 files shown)


kenrussell:
BTW, Chrome also fails the query of the MAX_TEXTURE_MAX_ANISOTROPY_EXT context parameter against WebGL 2.0 contexts. Fixes underway in http://crbug.com/1001763 .

(Fingers crossed that most OpenGL and OpenGL ES drivers already support setting the TEXTURE_MAX_ANISOTROPY_EXT parameter against sampler objects.)

kenrussell:
Fix for Chrome is up for review in the linked bug. This newly revised test passes in both WebGL 1.0 and 2.0 configurations. Could it please be approved and merged? Thanks.

PR #2926, opened 2019-09-11
Create webgl2 context with antialias=false for OVR_multiview

fernandojsg:
The OVR_multiview2 extension only supports non-multisampled contexts, so it could be nice to reflect that in the example as `antialias` is true by default when creating a webgl context


jdashg:
The example code draws into a user FBO, so the backbuffer antialiasing doesn't matter here.

jdashg:
Also for clarity, multiview is definitely supported in antialias:true contexts. There's just no way to render using multiview with multisampling in the same framebuffer at the moment.

fernandojsg:
Ah ok, that makes sense, thanks for clarifying!

PR #2927, opened 2019-09-18
Loosen pixel check in context-size-change.html

jchen10:
The pixel in the case may be considered as 'edge' to be blurred by
screen space anti-alias argorithms like CMAA. So it's possible the
value read back is not exactly same as originally written.

https://crbug.com/964010#c17

### DIFF
--- a/sdk/tests/conformance/context/context-size-change.html
+++ b/sdk/tests/conformance/context/context-size-change.html
@@ -70,10 +70,10 @@ function runTest()
     gl.clearColor(1.0, 1.0, 1.0, 1.0);
     gl.clear(gl.COLOR_BUFFER_BIT);
 
-    // Check that the top-left pixel has R channel 255.
+    // Check that the top-left pixel has R channel non-zero.
     gl.readPixels(0, 0, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, buf);
     pixel[0] = buf[0];
-    shouldBeTrue("pixel[0] == 255");
+    shouldBeTrue("pixel[0] != 0");
 
     // Check that the bottom-right pixel has R channel 0.
     gl.readPixels(2, 2, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, buf);


jchen10:
@kenrussell PTAL

@gyagp

jdashg:
I'm inclined to require MSAA for browser-provided antialiasing. Can CMAA be
implemented from inside WebGL, like FXAA?

jchen10:
Yes, I also prefer to use MSAA. But for performance consideration, we had to use CMAA as an alternative for the IA chromebooks.

sdk/tests/conformance/context/context-size-change.html
```diff
@@ -70,10 +70,10 @@
     gl.clearColor(1.0, 1.0, 1.0, 1.0);
     gl.clear(gl.COLOR_BUFFER_BIT);
 
-    // Check that the top-left pixel has R channel 255.
+    // Check that the top-left pixel has R channel non-zero.
```
kenrussell: Can this check be strengthened and still pass with CMAA? The idea that this test could pass while allowing that pixel to have a value of 1 out of 255 (still looking quite black) is undesirable.

jchen10: The actual pixel value after CMAA applied is 183. I am uncomfortable to use this magic value here, as the value may change if CMAA evolves or we use other algorithms like FXAA. Maybe it fair to be more than half of 255.
Another idea in my mind is that we should apply CMAA only if there was a draw to the framebuffer.  After all MSAA doesn't interfere with clear calls as well. Do you think we should seek to refine the current WebGL implementation regarding the timing of applying screen-space anti-alias?

kenrussell:
@jchen10 if those modifications wouldn't be too intrusive then yes, I think we should make them and leave this test alone. Please note that in crbug.com/1003860 and after measurement on many Intel GPUs and discussion with senior engineers at Intel, the current plan is to remove CMAA.

jchen10:
Close this PR as we are going to remove CMAA.

PR #2929, opened 2019-09-24
Fix typo in WebGL2 sync test

imiklos:
There was a wrong check on the `gl.getSyncParameter` return value, which results that the test always pass.

### DIFF
--- a/sdk/tests/conformance2/sync/sync-webgl-specific.html
+++ b/sdk/tests/conformance2/sync/sync-webgl-specific.html
@@ -79,7 +79,7 @@ function runGetSyncParameterTest() {
   var iter = numRetries;
   while (--iter > 0) {
     syncWithGLServer();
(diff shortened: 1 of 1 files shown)


jdashg:
Thanks!

sdk/tests/conformance2/sync/sync-webgl-specific.html
```diff
@@ -79,7 +79,7 @@
   var iter = numRetries;
   while (--iter > 0) {
     syncWithGLServer();
-    if (gl.getSyncParameter(sync, gl.SYNC_STATUS) == gl.SYNC_SIGNALED) {
+    if (gl.getSyncParameter(sync, gl.SYNC_STATUS) == gl.SIGNALED) {
```
kainino0x: Should this ALWAYS return UNSIGNALED? `!== gl.UNSIGNALED` would be more strict, if it would work.

PR #2931, opened 2019-09-25
Promote EXT_clip_cull_distance to Draft.

jdashg:


extensions/EXT_clip_cull_distance/extension.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- vi:set sw=2 ts=4: -->
 <?xml-stylesheet href="../../extension.xsl" type="text/xsl"?>
-<proposal href="proposals/EXT_clip_cull_distance/">
+<draft href="proposals/EXT_clip_cull_distance/">
```
kenrussell: The href is wrong - needs to be moved out of the proposals/ folder or this won't build.

PR #2932, opened 2019-09-25
Promote EXT_texture_norm16 to Draft.

jdashg:


extensions/EXT_texture_norm16/extension.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<proposal href="proposals/EXT_texture_norm16/">
+<draft href="proposals/EXT_texture_norm16/">
```
kenrussell: The href needs to be updated here too.

PR #2943, opened 2019-10-07
Add test cases to non-ascii-comments test

jonahryandavis:
As noted in
https://bugs.chromium.org/p/chromium/issues/detail?id=940865, comments
were not stripped by WebGL in some cases. Added a test to prevent this
from happening again.


kenrussell:
LGTM. Thanks Jonah. (Anyone watching, Jonah is on the ANGLE team so has implicitly signed the CLA.)

kenrussell:
Note that the linter complained that this change broke the build:
https://travis-ci.org/KhronosGroup/WebGL/builds/594823710

We've had trouble with this test case in the past where the whitelist in sdk/tests/py/lint/lint.whitelist doesn't work for suppressing the invalid unicode errors. If the build continues to be broken for subsequent pull requests then we'll investigate more deeply.

PR #2950, opened 2019-10-24
Fix vertex-array-object-and-disabled-attributes test.

### LINKED ISSUE #2949: 3th draw black arrays may fail?
https://github.com/KhronosGroup/WebGL/blob/2701c130839edbeb226735b0775966b6423d9e83/sdk/tests/conformance2/vertex_arrays/vertex-array-object-and-disabled-attributes.html#L120

in my tests, both 2th and 3th draw is green, but in this test case,  3th draw expect to be black, so case fail;
case fail may due to the following reason,  please confirm? thanks
1. in case line 87, program vertex shader is singlevshader,  API gl.vertexAttribPointer, set vertexArray's postion attribute, first parameter (index) is singlePositionLocation;  but in line 120, program vertex shader is dualvshader, position attribute location is positionLocation;  
2. in singlevshader and dualvshader,  input variable position with no location assigned in the shader text, so when a program is linked, GL API automatically assigned location; these two position location in different shaders may not equal;  
3. if var positionLocation is not equal to singlePositionLocation,  then 3th draw black array failed; 
4. in my tests, i find singlePositionLocation is 0, but positionLocation is 1, colorLocation is 0, so 3th draw black fail;

kenrussell:
conformance2/vertex_arrays/
vertex-array-object-and-disabled-attributes.html was assuming that the
"position" attribute location would match between two separately
compiled and linked programs. This must be enforced via
bindAttribLocation.

Associated with ANGLE test bug: http://crbug.com/angleproject/4049

Fixes #2949 .

### DIFF
--- a/sdk/tests/conformance2/vertex_arrays/vertex-array-object-and-disabled-attributes.html
+++ b/sdk/tests/conformance2/vertex_arrays/vertex-array-object-and-disabled-attributes.html
@@ -57,8 +57,9 @@ description();
 
 var gl = wtu.create3DContext("example", undefined, 2);
 
-var singleProgram = wtu.setupProgram(gl, ['singlevshader', 'singlefshader']);
-var dualProgram = wtu.setupProgram(gl, ['dualvshader', 'dualfshader']);
+// Location of "position" attribute must match between shaders.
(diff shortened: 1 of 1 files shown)


jdashg:
Backports?

kenrussell:
> Backports?

Test didn't exist before 2.0.1, so no backports needed.

PR #2951, opened 2019-10-25
Update WEBGL_video_texture extension to define textureVideoWEBGL.

kenrussell:
Instead of using texture2D with a textureVideoWEBGL argument.

Related to bug and pull request up for discussion:
  http://crbug.com/776222
  https://chromium-review.googlesource.com/1866274

### DIFF
--- a/extensions/proposals/WEBGL_video_texture/extension.xml
+++ b/extensions/proposals/WEBGL_video_texture/extension.xml
@@ -44,7 +44,7 @@
         <stage type="fragment"/>
         <type name="samplerVideoWEBGL"/>
 
-        <function name="texture2D" type="vec4">
+        <function name="textureVideoWEBGL" type="vec4">
           <param name="sampler" type="samplerVideoWEBGL"/>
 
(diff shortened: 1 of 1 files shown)


kenrussell:
@shaoboyan could you please review this change?

shaoboyan:
@kenrussell  CL looks good to me. But I have a silly question. Since the extension [OES_EGL_image_external](https://www.khronos.org/registry/OpenGL/extensions/OES/OES_EGL_image_external.txt) and [OES_EGL_image_external_essl3](https://www.khronos.org/registry/OpenGL/extensions/OES/OES_EGL_image_external_essl3.txt) reused texture2D* and texture* functions. why we have a different sampler function? Is it any case that texture2D* cannot cover? Thanks

kenrussell:
The EGL extensions were designed to work with 2D textures. WEBGL_video_texture has to abstract over ARB_texture_rectangle textures on some desktop platforms (macOS in particular). Therefore the texture sampling function must be made abstract as well as the sampler type.

PR #2952, opened 2019-10-25
Add oes-fbo-render-mipmap test

shrekshao:
Add test for `OES_fbo_render_mipmap` extension

### DIFF
--- a/sdk/tests/conformance/extensions/00_test_list.txt
+++ b/sdk/tests/conformance/extensions/00_test_list.txt
@@ -26,6 +26,7 @@
 --min-version 1.0.3 --max-version 1.9.9 oes-texture-half-float-with-image.html
 --min-version 1.0.3 --max-version 1.9.9 oes-texture-half-float-with-video.html
 --min-version 1.0.2 --max-version 1.9.9 oes-element-index-uint.html
+--min-version 1.0.4 --max-version 1.9.9 oes-fbo-mipmap-renderer.html
 webgl-debug-renderer-info.html
 webgl-debug-shaders.html
 --min-version 1.0.4 webgl-compressed-texture-astc.html
--- a/sdk/tests/conformance/extensions/oes-fbo-render-mipmap.html
+++ b/sdk/tests/conformance/extensions/oes-fbo-render-mipmap.html
@@ -1,0 +1,92 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL OES_fbo_render_mipmap Conformance Tests</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<canvas id="canvas" width="128" height="128"></canvas>
+<div id="description"></div>
+<div id="console"></div>
+<script>
+"use strict";
+description("This test verifies the functionality of the OES_fbo_render_mipmap extension, if it is available.");
+
+var wtu = WebGLTestUtils;
+var canvas = document.getElementById("canvas");
+var gl = wtu.create3DContext(canvas);
+var ext = null;
+
+(function(){
+
+    const oesFboRenderMipmap = gl.getExtension("OES_fbo_render_mipmap");
+    if (!oesFboRenderMipmap) {
+        testPassed("OES_fbo_render_mipmap is allowed to be missing.");
+        return;
+    }
+
+    let texture = gl.createTexture();
+    gl.bindTexture(gl.TEXTURE_2D, texture);
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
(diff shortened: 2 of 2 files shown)


sdk/tests/conformance/extensions/oes-fbo-render-mipmap.html
```diff
@@ -0,0 +1,92 @@
...
+        gl.clear(gl.COLOR_BUFFER_BIT);
+    }
+
+    gl.bindFramebuffer(gl.FRAMEBUFFER, null);
+
+    let program = wtu.setupTexturedQuad(gl);
+
+    for (let level = 0; level <= maxLevel; level++) {
+        let size = Math.pow(2, maxLevel - level);
+        gl.viewport(0, 0, size, size);
```
jdashg: Oh this is cute, but leave a comment that viewport size is how you select which miplevel is fetched from the texture.

sdk/tests/conformance/extensions/oes-fbo-render-mipmap.html
```diff
@@ -0,0 +1,92 @@
...
+        gl.texImage2D(gl.TEXTURE_2D, level, gl.RGBA, size, size, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+        let fbo = gl.createFramebuffer();
+        gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+        gl.framebufferTexture2D(gl.FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texture, level);
+        fbos.push(fbo);
+
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "Any level of a texture can be attached to a framebuffer object.");
+    }
+
+    shouldBe("gl.checkFramebufferStatus(gl.FRAMEBUFFER)", "gl.FRAMEBUFFER_COMPLETE");
```
jdashg: Do this per FB, or not at all.

sdk/tests/conformance/extensions/oes-fbo-render-mipmap.html
```diff
@@ -0,0 +1,92 @@
...
+
+    for (let level = 0; level <= maxLevel; level++) {
+        gl.bindFramebuffer(gl.FRAMEBUFFER, fbos[level]);
```
jdashg: You can merge this with the previous loop, and not have to keep track of multiple FBOs, unless there's a reason you split the loops like this.

sdk/tests/conformance/extensions/oes-fbo-render-mipmap.html
shrekshao: Thanks for the review. I write this test against chromium with roughly hooked up angle impl of this functionality. Turns out checking framebuffer completeness before all mip levels of the texture is created will ends in FRAMEBUFFER_INCOMPLETE_ATTACHMENT.

I just checked the OpenGL ES 2 Spec https://www.khronos.org/registry/OpenGL/specs/es/2.0/es_cm_spec_2.0.pdf
And seems like it makes sense
> Performing any of the following actions may change whether the framebuffer is considered complete or
incomplete.
> • Changing the width, height, or internal format of a texture image that is attached to the framebuffer
by calling TexImage2D, CopyTexImage2D and CompressedTexImage2D.
> This means that rendering commands such as DrawArrays, DrawElements, any command that reads the framebuffer such as ReadPixels and CopyTexSubImage will generate the
error INVALID FRAMEBUFFER OPERATION if called while the framebuffer is not framebuffer complete

But it should be checked for every fbo still. So I put into the second loop (together with the clear command)

sdk/tests/conformance/extensions/oes-fbo-render-mipmap.html
shrekshao: Same reasons stated above. Calling `gl.clear` before all mip levels of texture are created seem to lead to INVALID_OPERATION

sdk/tests/conformance/extensions/oes-fbo-render-mipmap.html
shrekshao: Just test with Firefox Nightly 72 on Linux. Proved the assumption. It wouldn't work if checking framebuffer completeness and calling gl.clear before all mip levels are created.

jdashg:
Thanks!

sdk/tests/conformance/extensions/00_test_list.txt
```diff
@@ -26,6 +26,7 @@
 --min-version 1.0.3 --max-version 1.9.9 oes-texture-half-float-with-image.html
 --min-version 1.0.3 --max-version 1.9.9 oes-texture-half-float-with-video.html
 --min-version 1.0.2 --max-version 1.9.9 oes-element-index-uint.html
+--min-version 1.0.4 --max-version 1.9.9 oes-fbo-mipmap-renderer.html
```
kenrussell: Typo here: should be the name of the HTML file, e.g. oes-fbo-render-mipmap.html . Could you please put up a follow-on pull request? Thanks.

shrekshao: Thanks for the catch Ken. This actually lead to test bot failure. Pull request is here: 2954

PR #2954, opened 2019-10-28
Fix incorrect test html name in test list

shrekshao:
Tests added in `00_test_list.txt` should be `oes-fbo-render-mipmap.html`.
Thanks @kenrussell for pointing this out.


jdashg:
Thanks!

kenrussell:
LGTM after the fact.

PR #2955, opened 2019-10-29
Fix BaseVertexBaseInstance extensions number

shrekshao:
Didn't assign an extension number when promoted to draft.
Thanks for Ken for pointing this out.


kenrussell:
LGTM

PR #2956, opened 2019-10-29
Allocate enums for WEBGL_video_texture extension out of WebGL's block.

kenrussell:
The previously chosen constants collided with other OpenGL constants.


kenrussell:
@kainino0x could you please review?

@shaoboyan FYI.

Khronos' internal WebGL wiki page is the place where these enums are allocated.

kenrussell:
@kainino0x yes, on Khronos' internal WebGL wiki page.

Thanks for your review.

shaoboyan:
Thank you for working on this!

PR #2959, opened 2019-10-30
Revert antialias cases

xhcao:
Previously modified canvas size for CMAA algorithm,
but CMAA had been removed from chromium, revert them.

PR #2960, opened 2019-10-30
Revert antialias cases

xhcao:
Previously modified canvas size to 3\*3 for CMAA
algorithm, but CMAA had been removed from chromium,
revert canvas size to 2\*2.

### DIFF
--- a/conformance-suites/1.0.0/conformance/context-attributes-alpha-depth-stencil-antialias.html
+++ b/conformance-suites/1.0.0/conformance/context-attributes-alpha-depth-stencil-antialias.html
@@ -200,7 +200,6 @@ function testAntialias(antialias)
 
     shouldBeTrue("webGL.canvas.width == webGL.canvas.height");
     // Both the width and height of canvas are N.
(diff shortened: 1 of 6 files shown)


xhcao:
@gyagp Please take a look, thank you.

gyagp:
LGTM

xhcao:
@kenrussell Please take a look, thank you.

jdashg:
We can revert these, but I don't think there's a strong reason to. We could consider leaving them. I have no preference.

kenrussell:
Thanks for cleaning this up. I'd prefer to move forward with removing changes made just for CMAA. However: does #2959 subsume this one? That one looks more thorough, so if they're mutually exclusive, I'd rather merge #2959 and close this one.

gyagp:
#2959 reverts every change we introduced for CMAA. However, the previous code is not flexible to extend, so we just made it flexible first, then did necessary changes for CMAA. So I prefer this PR (#2960), which keeps the changes for flexibility (maybe we will have requirement in future to change this case, and this will make the changes much easier).

kenrussell:
Thanks @gyagp for the explanation. Sounds fine then to merge this one and close #2959. Thanks for cleaning this up!

PR #2962, opened 2019-10-30
Accept all NaNs when sampling invalid LDR ASTC non-sRGB

jdarpinian:
When sampling invalid LDR ASTC textures which are not sRGB, hardware is allowed to return either the normal error color (magenta) or all NaNs. This modifies the ASTC test to accept an all NaN result in this case.

This fixes the test on some Qualcomm Android devices including Pixel phones.

### DIFF
--- a/sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
+++ b/sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
@@ -1592,29 +1592,21 @@ var program = wtu.setupTexturedQuad(gl);
 // For LDR and non-sRGB formats, the error color is magenta *or* all NaNs. To
 // make sure we accept this we create a program that transforms an all NaN
 // sampling result to magenta.
-var nanToMagentaProgram = wtu.setupTexturedQuad(gl);
-var fragmentShader = null;
-for (let shader of gl.getAttachedShaders(nanToMagentaProgram)) {
-    if (gl.getShaderParameter(shader, gl.SHADER_TYPE) == gl.FRAGMENT_SHADER) {
-        fragmentShader = shader;
-    }
-}
-gl.detachShader(nanToMagentaProgram, fragmentShader);
-fragmentShader = wtu.loadShader(gl,
-   `precision mediump float;
-    uniform sampler2D tex;
-    varying vec2 texCoord;
-    void main() {
-        vec4 c = texture2D(tex, texCoord);
-        if (all(notEqual(c, c))) {
-            gl_FragData[0] = vec4(1, 0, 1, 1);
-        } else {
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
```diff
@@ -1588,6 +1588,35 @@
 var canvas = document.getElementById("canvas");
 var gl = wtu.create3DContext(canvas, {antialias: false});
 var program = wtu.setupTexturedQuad(gl);
+
+// For LDR and non-sRGB formats, the error color is magenta *or* all NaNs. To
+// make sure we accept this we create a program that transforms an all NaN
+// sampling result to magenta.
+var nanToMagentaProgram = wtu.setupTexturedQuad(gl);
```
jdashg: Prefer const to var.

sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
```diff
@@ -1588,6 +1588,35 @@
...
+var fragmentShader = null;
+for (let shader of gl.getAttachedShaders(nanToMagentaProgram)) {
+    if (gl.getShaderParameter(shader, gl.SHADER_TYPE) == gl.FRAGMENT_SHADER) {
+        fragmentShader = shader;
```
jdashg: I don't really like patching the program with a different shader like this, but ok.

sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
```diff
@@ -1588,6 +1588,35 @@
...
+        vec4 c = texture2D(tex, texCoord);
+        if (all(notEqual(c, c))) {
+            gl_FragData[0] = vec4(1, 0, 1, 1);
+        } else {
+            gl_FragData[0] = texture2D(tex, texCoord);
+        }
+    }`, gl.FRAGMENT_SHADER);
+gl.attachShader(nanToMagentaProgram, fragmentShader);
+gl.linkProgram(nanToMagentaProgram);
+wtu.glErrorShouldBe(gl, gl.NO_ERROR, "linking nan program");
```
jdashg: LinkProgram generally doesn't generate errors, so don't bother checking. If you want to check status, you can check the error after useProgram, though.

sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
kenrussell: I also think it'd be nicer to explicitly set up the program the intended way the first time.

kenrussell:
Very nice. Thanks for the improvement James!

PR #2968, opened 2019-11-06
Remove unnecessary executable bit on WEBGL_compress_texture proposal

kenrussell:
Noticed in #2966.


kenrussell:
Merging myself to unblock the other contributor.

PR #2970, opened 2019-11-08
Probe for max non-OOM 3D texture size

hujiajie:
Large image requests can trigger an out-of-memory error (#1925). In #2579, we try to probe the max non-OOM level for 2D and cube map textures. This is applied to 3D textures now.

### DIFF
--- a/sdk/tests/conformance2/textures/misc/tex-3d-size-limit.html
+++ b/sdk/tests/conformance2/textures/misc/tex-3d-size-limit.html
@@ -62,22 +62,48 @@ function test3DTextureDimensions() {
         "texImage3D should fail for dimension out of range.");
 
     for (var i = 0; i < maxLevels; ++i) {
-        var size = 1 << i;
         var level = maxLevels - i - 1;
-        var badSize = size * 2;
+        var badSize = 1 << (i + 1);
 
-        gl.texImage3D(gl.TEXTURE_3D, level, gl.RGBA, size, 1, 1, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
-        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texImage3D should succeed.");
-        gl.texImage3D(gl.TEXTURE_3D, level, gl.RGBA, 1, size, 1, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
-        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texImage3D should succeed.");
-        gl.texImage3D(gl.TEXTURE_3D, level, gl.RGBA, 1, 1, size, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
-        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texImage3D should succeed.");
         gl.texImage3D(gl.TEXTURE_3D, level, gl.RGBA, badSize, 2, 2, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/textures/misc/tex-3d-size-limit.html
```diff
@@ -80,6 +73,39 @@
...
+    while (true) {
+        gl.texImage3D(gl.TEXTURE_3D, maxTestLevels-1, gl.RGBA, 1, 1, 1, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+        var err = gl.getError();
+        if (err == gl.OUT_OF_MEMORY) {
+            debug("Probe failed for level=" + (maxTestLevels-1) + ", reducing...");
+            maxTestLevels -= 1;
+            if (!maxTestLevels) {
+                testFailed("Failed to allocate any levels");
+                return;
+            }
```
jaibeer72: Taking that the test fails here don't you think it is better to just break it out of the loop here itself and make it go to clean up? 
if I am missing something and it is testing multiple things should it not be either in different tests or enumerate test cases?

sdk/tests/conformance2/textures/misc/tex-3d-size-limit.html
```diff
@@ -80,6 +73,39 @@
...
+    for (var i = 0; i < maxTestLevels; ++i) {
+        var level = maxTestLevels - i - 1;
+        var size = 1 << i;
+
+        gl.texImage3D(gl.TEXTURE_3D, level, gl.RGBA, size, 1, 1, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texImage3D should succeed.");
+        gl.texImage3D(gl.TEXTURE_3D, level, gl.RGBA, 1, size, 1, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texImage3D should succeed.");
+        gl.texImage3D(gl.TEXTURE_3D, level, gl.RGBA, 1, 1, size, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texImage3D should succeed.");
```
jaibeer72: being nit-pikey : again is there no way to enumerate these test conditions in JS? rather than having so many asserts that imply the same message and are not very explicit on where and what failed?  it seems that it just returns 3 pass messages rather than what component passed exactly.

sdk/tests/conformance2/textures/misc/tex-3d-size-limit.html
kenrussell: @jaibeer72 the test has to find the maximum mipmap level that can be allocated in order for the test loop below to work. It's appropriate for the test to fail if discovery of the maximum supported mipmap level fails.

sdk/tests/conformance2/textures/misc/tex-3d-size-limit.html
kenrussell: It's a fair point that these error messages could read something like "texImage3D with maximum [x/y/z] dimension should succeed".

jaibeer72: Got it :D Thanks.

hujiajie:
@kenrussell @jaibeer72 Thank you for reviewing this!

PR #2974, opened 2019-11-19
Webdriver tools update

grorg:
A collection of commits that update the tools/conformance.py script to:

- be compatible with Python3
- support Safari/WebKit
- various minor clean-ups

The documentation is also updated to show how to set up a virtual
python environment.

### DIFF
--- a/sdk/tests/webgl-conformance-tests.html
+++ b/sdk/tests/webgl-conformance-tests.html
@@ -287,6 +287,7 @@ function start() {
     div.appendChild(check);
     var button = reporter.localDoc.createElement('input');
     button.type = 'button';
+    button.id = this.elementId + "-button";
     button.value = 'run';
     button.onclick = function() {
       autoScroll = false;
--- a/tools/conformance.py
+++ b/tools/conformance.py
@@ -566,10 +566,6 @@ class Browser(object):
                 self.path = '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome'
             elif self.name == 'chrome_canary':
                 self.path = '/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary'
-            elif self.name == 'safari':
-                self.path = '/usr/bin/safaridriver'
-            elif self.name == 'safari_technology_preview':
-                self.path = '/Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver'
         elif self.os.is_win():
             if self.name == 'chrome' or self.name == 'chrome_stable':
                 self.path = '%s/Google/Chrome/Application/chrome.exe' % self.os.programfilesx86
@@ -658,6 +654,10 @@ class Webdriver(object):
         # path
         if target_os.is_cros():
             self.path = '/usr/local/chromedriver/chromedriver'
+        elif browser.name == 'safari':
+            self.path = '/usr/bin/safaridriver'
+        elif browser.name == 'safari_technology_preview':
+            self.path = '/Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver'
 
         executable_suffix = Util.get_executable_suffix(host_os)
         if not self.path and browser.is_chrome() and host_os == target_os:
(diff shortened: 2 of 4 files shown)


jdarpinian:
@gyagp Mind taking a look at this? GitHub won't let me add you as a reviewer, but I think you wrote this code originally.

tools/conformance.py
```diff
@@ -557,12 +564,16 @@ def __init__(self, name, path, options, os):
         elif self.os.is_mac():
             if self.name == 'chrome' or self.name == 'chrome_stable':
                 self.path = '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome'
-            if self.name == 'chrome_canary':
+            elif self.name == 'chrome_canary':
                 self.path = '/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary'
+            elif self.name == 'safari':
+                self.path = '/usr/bin/safaridriver'
```
gyagp: This should be browser binary path, instead of webdriver path.

tools/conformance.py
```diff
@@ -557,12 +564,16 @@ def __init__(self, name, path, options, os):
...
+            elif self.name == 'safari_technology_preview':
+                self.path = '/Applications/Safari Technology Preview.app/Contents/MacOS/safaridriver'
```
gyagp: same as above

tools/conformance.py
```diff
@@ -636,7 +650,7 @@ class Webdriver(object):
         'chrome_public': 'org.chromium.chrome',
     }
 
-    def __init__(self, path, browser, host_os, target_os, android_device=None, debug=False):
+    def __init__(self, path, browser, host_os, target_os, android_device=None, debug=False, tools=False):
         self._logger = Util.get_logger()
         self.path = path
         self.target_os = target_os
```
gyagp: I couldn't add comment in below code, but Safari webdriver path can be initialized the similar way with ChromeOS as below. Safari is a bit complicated as you have 2 webdrivers.

tools/conformance.py
```diff
@@ -728,12 +742,15 @@ def __init__(self, path, browser, host_os, target_os, android_device=None, debug
                     service_args = []
                 self.driver = selenium.webdriver.Chrome(executable_path=self.path, chrome_options=chrome_options, service_args=service_args)
             elif browser.is_safari():
-                Util.not_implemented()
+                capabilities = DesiredCapabilities.SAFARI.copy()
+                if tools:
+                    capabilities['automaticInspection'] = True
+                self.path = browser.path
+                self.driver = selenium.webdriver.safari.webdriver.WebDriver(desired_capabilities=capabilities, executable_path=self.path)
```
gyagp: browser.path should be assigned to capabilities['binary'], and self.path is actually webdriver path.

gyagp:
Thanks for the update, especially for bringing in Safari and Python3 support. Most of changes LGTM, but  the browser binary path and webdriver path need to be handled correctly. If "TODO Features" in README.md can be also updated accordingly, that would be great!

tools/conformance.py
grorg: Not for Safari. You don't worry about the browser path - instead you talk to the `safaridriver` binary and it will handle the browser for you. There is a strong mapping between a built `safaridriver` and the browser. (with the exception of iOS, where any `safaridriver` will forward requests onto the hidden driver running on the device)

tools/conformance.py
grorg: Same.

sdk/tests/webgl-conformance-tests.html
```diff
@@ -287,6 +287,7 @@
     div.appendChild(check);
     var button = reporter.localDoc.createElement('input');
     button.type = 'button';
+    button.id = this.elementId + "-button";
```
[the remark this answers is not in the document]
grorg: This probably isn't necessary. I used it to debug an issue in Safari where clicks were not getting through, but that turned out to be a Safari bug.

tools/conformance.py
grorg: In fact, this could be left empty for Safari since it doesn't need to be used.

gyagp:
LGTM!

grorg:
Thanks @gyagp!!

PR #2978, opened 2019-11-27
Require gl arg to glEnumToString and fix buggy callsites.

jdashg:


sdk/tests/js/webgl-test-utils.js
```diff
@@ -40,10 +40,11 @@ var loggingOff = function() {
...
 var glEnumToString = function(gl, value) {
-  // Avoid returning "NO_ERROR" if the arguments are totally wrong.
-  if (gl.NO_ERROR === undefined || value === undefined) {
-    return undefined;
-  }
+  if (gl.NO_ERROR === undefined)
+    throw new Error('glEnumToString: `gl` must be a WebGL2?RenderingContext');
+  if (value === undefined)
+    throw new Error('glEnumToString: `value` must not be undefined');
+
```
hujiajie: There seems to be a problem here: `value` could be an enum defined by extensions, and it's an exception here if `glEnumToString()` is called like [this](https://github.com/KhronosGroup/WebGL/blob/a782558f12df78e43d23089a70cd975e257c2381/sdk/tests/js/tests/webgl-compressed-texture-size-limit.js#L160) (note the second `glEnumToString()` of the line).

jdashg:
Ugh, yeah, that'd be a problem.

jdashg:
Fixed in #2980.

shrekshao:
Seems that this change leads to a failure at https://github.com/KhronosGroup/WebGL/blob/e0d986560015ba50517875dcdc9f5a6c01b13cff/sdk/tests/js/tests/webgl-compressed-texture-size-limit.js#L160
called in webgl-compressed-texture-size-limit.html

@jdashg any idea to get this covered (when enum exists only in ext)? Maybe you could fix it in #2980 at the same time.

shrekshao:
Oh didn't see this isse is already addressed by hujiajie. Please ignore.

PR #2980, opened 2019-11-29
glEnumToString allow glOrExt arg.

jdashg:
* Caches the value->enum unmappings, instead of linearly searching.
* Combine same-valued enum names: 0 -> NONE/ZERO/POINTS/NO_ERROR.
* glErrorShouldBeImpl still maps 0 -> NO_ERROR though.

### DIFF
--- a/sdk/tests/js/webgl-test-utils.js
+++ b/sdk/tests/js/webgl-test-utils.js
@@ -33,6 +33,10 @@ var loggingOff = function() {
   error = function() {};
 };
 
+const ENUM_NAME_REGEX = RegExp('[A-Z][A-Z0-9_]*');
+const ENUM_NAME_BY_VALUE = {};
+const ENUM_NAME_PROTOTYPES = new Map();
+
 /**
  * Converts a WebGL enum to a string.
  * @param {!WebGLRenderingContext} gl The WebGLRenderingContext to use.
@@ -43,24 +47,23 @@ var glEnumToString = function(glOrExt, value) {
   if (value === undefined)
     throw new Error('glEnumToString: `value` must not be undefined');
 
-  if (glOrExt.keyByValue === undefined) {
-    const keyByValue = glOrExt.keyByValue = {};
+  const proto = glOrExt.__proto__;
+  if (!ENUM_NAME_PROTOTYPES.has(proto)) {
+    ENUM_NAME_PROTOTYPES.set(proto, true);
 
-    for (const k in glOrExt) {
-      if (k == 'drawingBufferWidth' ||
-          k == 'drawingBufferHeight') {
-        continue;
-      }
+    for (const k in proto) {
+      if (!ENUM_NAME_REGEX.test(k)) continue;
+
       const v = glOrExt[k];
-      if (keyByValue[v] === undefined) {
-        keyByValue[v] = k;
+      if (ENUM_NAME_BY_VALUE[v] === undefined) {
+        ENUM_NAME_BY_VALUE[v] = k;
       } else {
-        keyByValue[v] += '/' + k;
+        ENUM_NAME_BY_VALUE[v] += '/' + k;
       }
     }
   }
 
-  const key = glOrExt.keyByValue[value];
+  const key = ENUM_NAME_BY_VALUE[value];
   if (key !== undefined) return key;
 
   return "0x" + Number(value).toString(16);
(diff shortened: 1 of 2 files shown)


sdk/tests/js/webgl-test-utils.js
```diff
@@ -39,24 +39,30 @@ var loggingOff = function() {
...
-  if (value === gl.NO_ERROR) {
-    return "NO_ERROR";
-  }
-  for (var p in gl) {
-    if (gl[p] == value) {
-      if (p == 'drawingBufferWidth' || p == 'drawingBufferHeight') {
+  if (glOrExt.keyByValue === undefined) {
+    const keyByValue = glOrExt.keyByValue = {};
+
+    for (const k in glOrExt) {
```
kainino0x: old code, but it would be tighter to iterate `glOrExt.__proto__` to avoid seeing anything that's hanging off the object.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -39,24 +39,30 @@ var loggingOff = function() {
...
 
-  // Optimization for the most common enum:
-  if (value === gl.NO_ERROR) {
-    return "NO_ERROR";
-  }
-  for (var p in gl) {
-    if (gl[p] == value) {
-      if (p == 'drawingBufferWidth' || p == 'drawingBufferHeight') {
+  if (glOrExt.keyByValue === undefined) {
+    const keyByValue = glOrExt.keyByValue = {};
```
kainino0x: keyByValue could be associated with the prototype instead of the object, I think

kainino0x: instead of the invasive expando, how about using a global `Map` from prototype object -> keyByValue dict?

sdk/tests/js/webgl-test-utils.js
```diff
@@ -39,24 +39,30 @@ var loggingOff = function() {
...
+
+    for (const k in glOrExt) {
+      if (k == 'drawingBufferWidth' ||
+          k == 'drawingBufferHeight') {
         continue;
       }
-      return p;
+      const v = glOrExt[k];
+      if (keyByValue[v] === undefined) {
+        keyByValue[v] = k;
```
kainino0x: I think this is stringifying all of the functions and setting those strings as keys in keyByValue?

sdk/tests/js/webgl-test-utils.js
```diff
@@ -39,24 +39,30 @@ var loggingOff = function() {
...
-  }
-  for (var p in gl) {
-    if (gl[p] == value) {
-      if (p == 'drawingBufferWidth' || p == 'drawingBufferHeight') {
+  if (glOrExt.keyByValue === undefined) {
+    const keyByValue = glOrExt.keyByValue = {};
+
+    for (const k in glOrExt) {
+      if (k == 'drawingBufferWidth' ||
+          k == 'drawingBufferHeight') {
```
kainino0x: To generalize this and catch the possible problem below, maybe test this against `/[A-Z][A-Z_]*[A-Z]/` or something.

shrekshao:
Jeff could you take a look at this pull request? I wish this could be addressed and merged as it blocks the rolling of webgl conformance now.

jdashg:
Sorry about that! I'll get it merged later today.

sdk/tests/js/webgl-test-utils.js
jdashg: Oh, that makes sense: FOO_BAR only.

sdk/tests/js/webgl-test-utils.js
jdashg: Poooossibly. The new regex should fix this.

sdk/tests/js/webgl-test-utils.js
jdashg: Map should be efficient.

jdashg:
PTAL

sdk/tests/js/webgl-test-utils.js
```diff
@@ -33,6 +33,10 @@ var loggingOff = function() {
   error = function() {};
 };
 
+const ENUM_NAME_REGEX = RegExp('[A-Z][A-Z0-9_]*');
+const ENUM_NAME_BY_VALUE = {};
```
kainino0x: Hm, should this also be a Map? (Looks correct to me either way, though.)

jdashg: It can be, but it doesn't need to be.
I used a Map for the prototype so I didn't need to rely on its stringification.

kainino0x: 👍

PR #2982, opened 2019-12-05
Use an empty string for the default bind address

grorg:
The comment says the default is all interfaces but then explicitly specifies localhost. By using an empty string we can bind everywhere.

### DIFF
--- a/serve_localhost.py
+++ b/serve_localhost.py
@@ -22,7 +22,7 @@ if __name__ == '__main__':
     import argparse
 
     parser = argparse.ArgumentParser()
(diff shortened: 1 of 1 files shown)


jdashg:
Oops, I didn't fix that comment. I deliberately wanted to bind localhost instead of turning your dev machine into a server. (Though you should be able to do that if you like!)

grorg:
Sure! I'll update the comment. I'll also make a note for how to bind everywhere. It tripped me up when running the web driver tests on a device, because it couldn't see the remote host.

jdashg:
Oh, you just pushed directly! LGTM!

PR #2983, opened 2019-12-05
Add support for iOS to WebDriver script

grorg:
This patch adds support for iOS devices.

- Renamed generic "device"-like things from mentioning Android
- Took the existing Android objects and made base classes with
  Android and iOS subclasses if necessary

### DIFF
--- a/tools/conformance.py
+++ b/tools/conformance.py
@@ -73,8 +73,13 @@ class Util(object):
         exit(1)
 
     @staticmethod
+    def warn(msg):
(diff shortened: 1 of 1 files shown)


grorg:
@gyagp you might want to look at this one too

gyagp:
LGTM, though I don't have any mobile devices at hand for test :) Could you please also update the README to add iOS (It's also OK in another PR)? Thanks!

PR #2984, opened 2019-12-06
Add test that context creation reclaims vram to prevent creation fail…

jdashg:
…ure.

This would be a nice invariant to have, but we can talk about how much we want to guarantee here.

### DIFF
--- a/sdk/tests/conformance/context/context-creation-reclaims-vram.html
+++ b/sdk/tests/conformance/context/context-creation-reclaims-vram.html
@@ -1,0 +1,85 @@
+<!--
+Copyright (c) 2019 The Khronos Group Inc.
+Use of this source code is governed by an MIT-style license that can be
+found in the LICENSE.txt file.
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Test that contexts are freed and garbage collected reasonably</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/context/context-creation-reclaims-vram.html
```diff
@@ -0,0 +1,85 @@
...
+  });
+  assertMsg(gl, `Creating context ${canvas.id} should succeed.`);
+  if (gl) {
+    const width = gl.drawingBufferWidth;
+    const height = gl.drawingBufferHeight;
+    debug(`   allocated size: ${width*height*4/1024/1024|0} MB`);
+    assertMsg(width >= minSize,
+              `   drawingBufferWidth (${width}) should be >= ${minSize}`);
+    assertMsg(height >= minSize,
+              `   drawingBufferHeight (${height}) should be >= ${minSize}`);
```
kenrussell: Chromium's WebGL implementation handles the scenario of users allocating many WebGL contexts with large back buffers by scaling down the size of newly created back buffers, rather than losing older contexts. We've tried in the past to lift the global limits on the amount of back buffer VRAM that can be allocated, but have never succeeded; crashes have been seen inside OpenGL drivers on various platforms, instead of GL_OUT_OF_MEMORY being reported. I'm not sure that requiring that newly-allocated contexts have the same back buffer size as the originally-allocated context is an invariant we should enforce; let's discuss in real time on tomorrow's working group conference call.

PR #2990, opened 2019-12-24
Add conformance test for dynamic indexing of row_major matrix arrays

themikelester:
This is failing on OSX Chrome. A bug will also be filed with Chrome.

I'm seeing this issue on a personal project. In my environment, with `layout(row_major)` the test fails, `layout(column_major)` test succeeds.

Edit: ANGLE bug filed at https://bugs.chromium.org/p/angleproject/issues/detail?id=4242

### DIFF
--- a/sdk/tests/conformance2/glsl3/matrix-row-major-dynamic-indexing.html
+++ b/sdk/tests/conformance2/glsl3/matrix-row-major-dynamic-indexing.html
@@ -1,0 +1,78 @@
+<!--
+
+/*
+** Copyright (c) 2019 The Khronos Group Inc.
+**
+** Permission is hereby granted, free of charge, to any person obtaining a
+** copy of this software and/or associated documentation files (the
+** "Materials"), to deal in the Materials without restriction, including
+** without limitation the rights to use, copy, modify, merge, publish,
+** distribute, sublicense, and/or sell copies of the Materials, and to
+** permit persons to whom the Materials are furnished to do so, subject to
+** the following conditions:
+**
+** The above copyright notice and this permission notice shall be included
+** in all copies or substantial portions of the Materials.
+**
+** THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+** EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+** MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/glsl3/matrix-row-major-dynamic-indexing.html
```diff
@@ -0,0 +1,78 @@
...
+{
+  fShaderId: 'fshaderIndexMatrixArrayInUniformBlock',
+  fShaderSuccess: true,
+  linkSuccess: true,
+  passMsg: '',
+  uniformBlocks: [{name: "a", value: new Float32Array([
+    0, 0, 0, 0,
+    0, 1, 0, 0,
+    0, 0, 0, 0,
+    0, 0, 0, 0,
```
kenrussell: This matrix is symmetric - how does it uncover a bug in row_major vs. column_major indexing?

To be more robust, could the fragment shader read two different values from this matrix - ideally not mirrored across the diagonal of the matrix?

sdk/tests/conformance2/glsl3/matrix-row-major-dynamic-indexing.html
```diff
@@ -0,0 +1,78 @@
...
+};
+
+void main() {
+    float f = u_mats[u_zero + 0][1][1];
+    my_FragColor = vec4(1.0 - f, f, 0.0, 1.0);
+}
+</script>
+<script type="application/javascript">
+"use strict";
+description("Indexing matrices within a uniform block should work");
```
kenrussell: Indexing row-major matrices...

sdk/tests/conformance2/glsl3/matrix-row-major-dynamic-indexing.html
magcius: MacGL appears to have a bug where, when row_major is specified, dynamically indexing into an array of matrices always returns 0.

kenrussell: OK, but would it not make more sense to at least put the non-zero value off the diagonal, so that if another bug comes up where row-major is treated as column-major, the test fails?

kenrussell: While hooking up the workaround for this driver bug in http://anglebug.com/2273 I made a mistake and the fact that the matrix is symmetric would have let a bug slip through. Since there hasn't been any reply since my initial review, I'm going to merge this and apply my suggested changes on top of it.

PR #2972, opened 2019-11-13
EXT_texture_norm16 test copyTex/ReadPixels RGBA only

shrekshao:
According to https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_norm16.txt

"ReadPixels format and type used during CopyTex*." is limited to RGBA.

PR #2995, opened 2020-01-03
Fix alpha channel set to 1 in ext-texture-norm16 test

shrekshao:
Part of the fix after figuring out the root cause for the failure hasn't landed.
https://github.com/KhronosGroup/WebGL/pull/2972/files


jdashg:
LGTM

PR #2998, opened 2020-01-07
Fix samplers.html: enumToString might return multiple enums separated by slash

shrekshao:
Now glEnumToString might return multiple enums sharing one values #2980.

Quick fix for a test where https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/conformance2/samplers/samplers.html#L177
```
eval("gl['POINTS/ZERO/NO_ERROR/NONE']");
```
will return undefined leading to a test failure..

This is a quick fix but might not be optimal.

### DIFF
--- a/sdk/tests/conformance2/samplers/samplers.html
+++ b/sdk/tests/conformance2/samplers/samplers.html
@@ -174,7 +174,7 @@ function runParameterTest() {
         var pname = testCases[ii].pname;
         var param = testCases[ii].param;
         wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.samplerParameteri(s, " + pname + ", " + param + ")");
-        shouldBe("gl.getSamplerParameter(s, " + pname + ")", "gl['" + (enumToString(param).split('/'))[0] + "']");
+        shouldBe("gl.getSamplerParameter(s, " + pname + ")", param);
         wtu.glErrorShouldBe(gl, gl.NO_ERROR);
     }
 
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance2/samplers/samplers.html
```diff
@@ -174,7 +174,7 @@
         var pname = testCases[ii].pname;
         var param = testCases[ii].param;
         wtu.shouldGenerateGLError(gl, gl.NO_ERROR, "gl.samplerParameteri(s, " + pname + ", " + param + ")");
-        shouldBe("gl.getSamplerParameter(s, " + pname + ")", "gl['" + enumToString(param) + "']");
+        shouldBe("gl.getSamplerParameter(s, " + pname + ")", "gl['" + (enumToString(param).split('/'))[0] + "']");
```
jdashg: This is a weird use of this function. Let's just do:
`shouldBe("gl.getSamplerParameter(s, " + pname + ")", param);`

shrekshao: I changed it. Though https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/js/js-test-pre.js#L372 
WARN: shouldBe() expects string arguments?

jdashg:
@shrekshao Looks good.
Generally "request changes" means I want to review it again, whereas "accepted" means merge when ready.

shrekshao:
Thanks for the explanation.

PR #2999, opened 2020-01-08
Use ESSL300 Texture Quad shader for ext_texture_norm16.html

shrekshao:
It seems on some mobile device (Pixel 2/3XL, Qualcomm) ESSL 100 doesn't provide us enough precision for 16 bit normalized value even with `precision highp float`
* 27189 is the value we used in the test, output discrete values: 27184, 27200
* `gl.getShaderPrecisionFormat(gl.FRAGMENT_SHADER, gl.HIGH_FLOAT)` returns WebGLShaderPrecisionFormat {rangeMin: 127, rangeMax: 127, precision: 23}

Not sure if this is a driver precision bug for normalized integer or it's the nature of ESSL 100 shader. (If it is the limitation of the ESSL 100 shader then it might be an obstacle for us to support this extension in WebGL 1 context?)

### DIFF
--- a/sdk/tests/conformance2/extensions/ext-texture-norm16.html
+++ b/sdk/tests/conformance2/extensions/ext-texture-norm16.html
@@ -153,7 +153,8 @@ function runTestExtension() {
 
   wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texture and framebuffer setup succeed");
 
-  wtu.setupTexturedQuad(gl);
+  wtu.setupSimpleTextureProgramESSL300(gl);
+  wtu.setupUnitQuad(gl);
 
   testNorm16Texture(ext.R16_EXT, gl.RED, gl.UNSIGNED_SHORT);
   testNorm16Texture(ext.RG16_EXT, gl.RG, gl.UNSIGNED_SHORT);
--- a/sdk/tests/js/webgl-test-utils.js
+++ b/sdk/tests/js/webgl-test-utils.js
@@ -113,6 +113,20 @@ var simpleTextureVertexShader = [
   '}'].join('\n');
 
 /**
+ * A vertex shader for a single texture.
+ * @type {string}
+ */
+var simpleTextureVertexShaderESSL300 = [
+  '#version 300 es',
+  'layout(location=0) in vec4 vPosition;',
+  'layout(location=1) in vec2 texCoord0;',
+  'out vec2 texCoord;',
+  'void main() {',
+  '    gl_Position = vPosition;',
+  '    texCoord = texCoord0;',
+  '}'].join('\n');
+
+/**
  * A fragment shader for a single texture.
  * @type {string}
  */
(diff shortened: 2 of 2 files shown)


jdashg:
I suspect this is due to the default precision of samplers, which is lowp in both essl100 and essl300:
`precision lowp sampler2D;`

Texturing operations behave like the template:
`vec4<precisionT> texture2D(sampler2D<precisionT>, vec2<precisionU>)`

To get a higher precision output from texturing commands, declare the sampler as:
`mediump sampler2D foo;`

shrekshao:
Thanks for the explanation on precision on sampler2D. I mixed it up with the floating point precision.

Yet declaring the sampler as `uniform mediump sampler2D tex;` in a ESSL 100 shader still produce the same incorrect result.

I check the table in section 4.5.2 from [OpenGL ES Shading Language 1.0](https://www.khronos.org/files/opengles_shading_language.pdf) and compare it with that from section 4.7.1 [OpenGL ES Shading Language 3.0](https://www.khronos.org/registry/OpenGL/specs/es/3.0/GLSL_ES_Specification_3.00.pdf)
It seems that in 1.00 it lacks the range definition for unsigned integer. Would that be possible cause?

jdashg:
For 27189, you probably need highp. Is that what you tried?

Rule of thumb is lowp exactly represents +/-256, and mediump is +/-MAX_TEXTURE_SIZE (technically only +/-1024 is required), and highp for needs beyond that.

shrekshao:
Yep. I also tried highp. But I think highp is not required for the implementation to support if I understand the spec correctly.
```
precision highp float;
precision highp int;
uniform highp sampler2D tex;
```
this setting still produce the same failure result

jdashg:
highp is optional in fragment shaders, yes. Basically the tests here should use the highest precision available.
This sounds like a bug to me. `precision highp float; uniform highp sampler2D tex;` should work fine.

shrekshao:
Well after all this is being tested on WIP chromium implementation with ext_texture_norm16 support so it might still be bugs in browser implementation.

I think what we need here is Floating point precision of 2<sup>-16</sup>. And in ESSL 100 highp should provide such precision in theory (it it's supported). So that this test should just work with ESSL 100 but there's some bug (maybe in the driver).

Using ESSL 300 could guarantee it works on all hardware (seen at this moment). Do you want to merge in this change, or have a new pull request still using ESSL 100 but with an opt_fragmentShaderOverride passed in with `precision highp float; uniform highp sampler2D tex;` declared?

jdashg:
Since this looks like a bug, we want to test this, so the essl100 test needs to be written.
This should probably be generalized, since this is effectively needed for all tests with >8 bits per channel, including all float/half-float tests.
IIRC, iOS in particular is picky about precision qualifiers for float texture sampling.

shrekshao:
Thanks for the input. I will close this and open a new one still using ESSL 100.
Tip for generalization: it seems that using some values other than 1 (MAX) could help reveal this kind of precision problem. This test is ported from an ANGLE texture test that use `0x6a35` (27189) as the pixel value and now I kind of get the reason for choosing this.

PR #3000, opened 2020-01-08
Use high precision float and sampler2D for ext-texture-norm16.html

shrekshao:
By default the fragment shader uses `lowp` for `sampler2D`. It is not enough for RGBA16 UNSIGNED_SHORT normalized value which has 16 bit for each channel. Use an override shader with `precision highp float; uniform highp sampler2D tex;` declared instead.

This should probably be generalized, since this is effectively needed for all tests with >8 bits per channel, including all float/half-float tests.

Previous discussion could be found in #2999

### DIFF
--- a/sdk/tests/conformance2/extensions/ext-texture-norm16.html
+++ b/sdk/tests/conformance2/extensions/ext-texture-norm16.html
@@ -153,7 +153,7 @@ function runTestExtension() {
 
   wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texture and framebuffer setup succeed");
 
-  wtu.setupTexturedQuad(gl);
+  wtu.setupTexturedQuad(gl, 0, 1, wtu.simpleHighPrecisionTextureFragmentShader);
 
   testNorm16Texture(ext.R16_EXT, gl.RED, gl.UNSIGNED_SHORT);
   testNorm16Texture(ext.RG16_EXT, gl.RG, gl.UNSIGNED_SHORT);
--- a/sdk/tests/js/webgl-test-utils.js
+++ b/sdk/tests/js/webgl-test-utils.js
@@ -125,6 +125,18 @@ var simpleTextureFragmentShader = [
   '}'].join('\n');
 
 /**
+ * A fragment shader for a single texture with high precision.
+ * @type {string}
+ */
+var simpleHighPrecisionTextureFragmentShader = [
+  'precision highp float;',
+  'uniform highp sampler2D tex;',
+  'varying vec2 texCoord;',
+  'void main() {',
+  '    gl_FragData[0] = texture2D(tex, texCoord);',
+  '}'].join('\n');
+
+/**
  * A fragment shader for a single cube map texture.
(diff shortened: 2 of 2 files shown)


jdashg:
Oh, the issue on essl100 might be the precision of the *output* variable, actually! Consider splitting it into vec4(x / 256, mod(x, 256), 0, 1) for testing, or doing the testing in the shader directly.

shrekshao:
You are right, tested with
```
vec4 x = texture2D(tex, texCoord);
gl_FragData[0] = vec4(abs(x.r * 65535.0 - 27189.0) < 0.1 ? 1.0 : 0.0, abs(x.g * 65535.0 - 27184.0) < 0.1 ? 1.0 : 0.0, x.b, 1);
```
The output.r = 1 and output.g = 0 meaning the sampled value is 27189 (correct).

Is the output precision changeable? (I haven't found the def in spec yet)

jdashg:
It's not changeable in ESSL100. (you can search for "gl_FragData" to find the implicit definition)

shrekshao:
Definition found in ESSL spec 100 section 7.2, declared as `mediump`
```
mediump vec4 gl_FragCoord;
 bool gl_FrontFacing;
mediump vec4 gl_FragColor;
mediump vec2 gl_PointCoord;
```
Required floating point precision is 2<sup>-10</sup>. That's insufficient for 16 bit normalized integer. 

But that is the same in ESSL 100 and 300. I tried setting `out mediump/lowp vec4 out_color;` in the ESSL 300 shader but that still works (might be because the implementation could provide higher precision)

So it seems to me that we still need ESSL 300 for this test, with highp declared explicitly to request highest possible precision.

We can also wait and discuss it on WebGL WG meeting.

jdashg:
Just test for it in the shader, or break it into uint8 parts for testing.

jdashg:
Another option is to make sure it's within mediump tolerance, but stricter than lowp tolerance. (fuzzy ranges)

shrekshao:
I could do the workaround value checking in the shader, but that wouldn't actually test the norm16 texture rendering ability. Either the in-shader test or breaking into uint8 is only testing norm16 texture sampling ability, not rendering.

I still think at least for `testNorm16Render` we should use ESSL 300 shader.

shrekshao:
That might affect how we are designing the extension. We might need to have two different extension versions, one for WebGL 2 with full features, the other for WebGL 1 with only sampling?

kenrussell:
https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_texture_norm16.txt states that OpenGL ES 3.1 is required - so implicitly the extension requires ESSL 3.00 support.

Perhaps it should only be advertised for WebGL 2.0 contexts. That might be a reasonable start.

jdashg:
I don't see it as requiring essl300. This issue is just unusual fallout from interactions between specs. We should add a disclaimer to the extension text about color-renderability, but the core sampling behavior is working perfectly. The RTT precision truncation is within spec, even if superficially unexpected.

PR #3001, opened 2020-01-09
Use ESSL300 Texture Quad shader for ext_texture_norm16

shrekshao:
In OpenGL ES Shading language spec 100 section 7.2
```
mediump vec4 gl_FragCoord;
 bool gl_FrontFacing;
mediump vec4 gl_FragColor;
mediump vec2 gl_PointCoord;
```
These declarations show the output precision would be mediump which is at least 2<sup>-10</sup> for floating point precision and is unchangeable. Thus it is not sufficient for render ability required by EXT_texture_norm16 for 16 bit normalized unsigned short.

This might also affect how we design this extension. If we want it to be exposed to WebGL 1 in the future, we might need a separate version of it lacking the rendering ability. I think we could discuss this part of the topic in WebGL WG meeting.

Previous discussion in #3000

### DIFF
--- a/sdk/tests/conformance2/extensions/ext-texture-norm16.html
+++ b/sdk/tests/conformance2/extensions/ext-texture-norm16.html
@@ -153,7 +153,8 @@ function runTestExtension() {
 
   wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texture and framebuffer setup succeed");
 
-  wtu.setupTexturedQuad(gl, 0, 1, wtu.simpleHighPrecisionTextureFragmentShader);
+  wtu.setupSimpleTextureProgramESSL300(gl);
+  wtu.setupUnitQuad(gl);
 
   testNorm16Texture(ext.R16_EXT, gl.RED, gl.UNSIGNED_SHORT);
   testNorm16Texture(ext.RG16_EXT, gl.RG, gl.UNSIGNED_SHORT);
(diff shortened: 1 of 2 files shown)


shrekshao:
Thanks for the review. This is mostly identical to #2999 where Jeff has viewed. Let me merge and roll to move forward with the extension impl in chromium.

sdk/tests/conformance2/extensions/ext-texture-norm16.html
```diff
@@ -153,7 +153,8 @@
 
   wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texture and framebuffer setup succeed");
 
-  wtu.setupTexturedQuad(gl, 0, 1, wtu.simpleHighPrecisionTextureFragmentShader);
```
jdashg: I would like to see the ESSL100 test retained, but testing for the result within a tolerance. It's important to test that ESSL100 shaders sample from this correctly, and ideally that RTT works within the restrictions of mediump.

jdashg:
To be clear, EXT_texture_norm16 doesn't guarantee that fragment output below highp isn't truncated. It's unfortunate fallout from ESSL1's choice of output precisions, but it's all compatible spec language as far as I can see. We should add a disclaimer about this interaction in the extension spec, but otherwise I'm happy to have this extension in both.

Note also: ESSL1 shaders in WebGL2/ES3 will have the same limitation as in WebGL1.

PR #3004, opened 2020-01-14
Add tests for query, sync and timer queries on workers.

kenrussell:
Test the following scenarios with OffscreenCanvas on workers:

 - Query objects with WebGL 2.0 contexts
 - Sync objects with WebGL 2.0 contexts
 - Timer queries with WebGL 1.0 and 2.0 contexts

Regression test for http://crbug.com/1010877 .

### DIFF
--- a/sdk/tests/conformance/offscreencanvas/offscreencanvas-timer-query.html
+++ b/sdk/tests/conformance/offscreencanvas/offscreencanvas-timer-query.html
@@ -1,0 +1,84 @@
+<!--
+Copyright (c) 2019 The Khronos Group Inc.
+Use of this source code is governed by an MIT-style license that can be
+found in the LICENSE.txt file.
+-->
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Test for Timer Query objects with OffscreenCanvas</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+  <div id="description"></div>
+  <div id="console"></div>
+  <script id='myWorker' type='text/worker'>
+  function tick(callback) {
+      function tickImpl() {
+          const res = callback();
+          if (res) {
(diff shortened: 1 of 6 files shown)


kenrussell:
Tested in Chrome with fix for http://crbug.com/1010877 . Tested with Safari and Firefox Nightly; confirmed test fails gracefully ("No OffscreenCanvas support").

@jdarpinian please review.

sdk/tests/conformance/offscreencanvas/offscreencanvas-timer-query.html
```diff
@@ -0,0 +1,84 @@
...
+      if (!ext) {
+          self.postMessage("PASSED - no EXT_disjoint_timer_query extension - this is legal");
+          return false;
+      }
+      let query = ext.createQueryEXT();
+      ext.beginQueryEXT(ext.TIME_ELAPSED_EXT, query);
+      ext.endQueryEXT(ext.TIME_ELAPSED_EXT);
+      gl.clearColor(0.0, 1.0, 0.0, 1.0);
+      gl.clear(gl.COLOR_BUFFER_BIT);
+      tick(function() {
```
jdashg: FWIW, consider this async pattern:
```
requestAnimationFrame = requestAnimationFrame || function(task) {
  setTimeout(task, 1000/60);
}
async function awaitRaf() {
  await new Promise(res => {
    requestAnimationFrame(() => {
      res();
    });
  });
}
```
```
while (true) {
  await awaitRaf();
  
  const status = ext.getQueryObjectEXT(query, ext.QUERY_RESULT_AVAILABLE_EXT);
  if (status) {
    self.postMessage("PASSED - timer query object completed successfully on worker");
    break;
  }
  const err = gl.getError();
  if (err != 0) {
    self.postMessage("FAILED - GL error " + err);
    break;
  }
  continue;
}
```

kenrussell: Thanks for the suggestion. I'm guessing this would require the onmessage handler to be an async function? Going to leave this as is, but open to a refactoring in the future.

PR #3005, opened 2020-01-15
Spec zero-sized canvas behavior and test.

jdashg:
0x0 canvas yields 1x1 WebGL context.


jdashg:
From this Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1609006

Issue #3003, opened 2020-01-13
WEBGL_draw_arrays not working as per spec?


greggman:
AFAIK it should be an error to try to draw to the canvas like this

```
ext = gl.getExtensions('WEBGL_draw_buffers');
ext.drawBuffersWEBGL([
     ext.COLOR_ATTACHMENT0_WEBGL,
]);
gl.drawArrays(...);
```

And it is but, if you first draw to a framebuffer instead of the canvas, then draw to the canvas Firefox and Chrome don't generate an error. Safari does

In other words. These are the steps

1.  enable extension
2. create framebuffer and attach texture,
3. bindFramebuffer(framebuffer)
4. drawBuffersEXT([ ext.COLOR_ATTACHMENT0_WEBGL ]);
5. draw (to framebuffer)  // should be no error
6.  bindFramebuffer(null)
7. draw (to canvas)  // should be INVALID_OPERATION

https://jsfiddle.net/greggman/9euqaL13/

From [the spec](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_draw_buffers.txt)

> If the GL is bound to the default framebuffer, then &lt;n&gt; must be 1
    and the constant must be BACK or NONE. When draw buffer zero is
    BACK, color values are written into the sole buffer for single-
    buffered contexts, or into the back buffer for double-buffered
    contexts. If DrawBuffersEXT is supplied with a constant other than
    BACK and NONE, or with a value of <n> other than 1, the error
    INVALID_OPERATION is generated.

greggman:
Ok, well, I guess Safari is wrong in this case? It sounds like `drawBuffers` state is per framebuffer state not global state and in that case Safari is wrong

greggman:
Ok well,  testing that it is drawBuffers state is per framebuffer

https://jsfiddle.net/greggman/4Lcmzp3w/

Chrome: Pass
Safari: Pass
Firefox: Fail

Also WebGL2  https://jsfiddle.net/greggman/k04fj7cv/

Chrome: Pass
Safari: N/A
Firefox: Fail

jdashg:
Firefox looks to have a bug there in getParameter, though rendering should work properly.

kenrussell:
Thanks Gregg for reporting and Jeff for the regression test.

PR #3006, opened 2020-01-17
Add test for bindFramebuffer+getParameter(DRAW_BUFFERn) without drawing.

jdashg:
Regression test for a bug in Firefox.


jdashg:
Addresses #3003.

PR #3007, opened 2020-01-20
Update test for dynamically-indexed row-major matrices.

kenrussell:
Follow-on to #2990.


kenrussell:
Minor changes. Merging myself.

kenrussell:
Note: the workaround for http://anglebug.com/2273 makes this test pass on macOS.

PR #3018, opened 2020-02-06
Clarify behaviors and timelines in WEBGL_lose_context.

jdashg:
This resolves ambiguities discovered in
https://bugzilla.mozilla.org/show_bug.cgi?id=1144115

### DIFF
--- a/extensions/WEBGL_lose_context/extension.xml
+++ b/extensions/WEBGL_lose_context/extension.xml
@@ -52,14 +52,30 @@ interface WEBGL_lose_context {
 
     </function>
     <function name="restoreContext" type="void">
+      <p>
       When this function is called while the context is lost, and the conditions
       defined by the WebGL specification for restoring the context are
       met, simulate the context being restored so as to trigger the steps
       described in the WebGL spec for handling the context being restored.
+      Because this only enqueues the 'webglcontextrestored' event, the context will not be
+      immediately restored after <code>restoreContext()</code>, but rather is first usably restored
+      within the 'webglcontextrestored' event callback.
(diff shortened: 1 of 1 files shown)


extensions/WEBGL_lose_context/extension.xml
```diff
@@ -52,14 +52,30 @@ interface WEBGL_lose_context {
...
       the conditions in the WebGL specification for restoring the
       context are not satisfied, or if the context was not lost via <code>loseContext()</code>,
       generate an <code>INVALID_OPERATION</code> error.
+      </p>
+
+      <p>
+      For example, if <code>e.preventDefault()</code> is not called on the 'webglcontextlost'
+      event, the context will not be marked restorable, and <code>restoreContext()</code> will fail.
+      Additionally, the context is not restorable before the 'webglcontextlost' event is complete:
+      Calling <code>restoreContext()</code> before 'webglcontextlost' has completed will fail.
```
kenrussell: Calling -> calling

extensions/WEBGL_lose_context/extension.xml
```diff
@@ -52,14 +52,30 @@ interface WEBGL_lose_context {
...
+      </p>
     </function>
   </newfun>
   <history>
```
kenrussell: Would you please add a revision entry below?

extensions/WEBGL_lose_context/extension.xml
jdashg: Wow, but I think you're right! :)

jdashg:
I don't know if we test all these things directly.

PR #3019, opened 2020-02-07
Check sampling from allocated compressed texture array

lexaknyazev:
In Chrome (80-82), sampling from freshly-allocated compressed 2D texture array results in `INVALID_ENUM` error on `drawArrays` call.

https://crbug.com/1051217

### DIFF
--- a/sdk/tests/conformance2/textures/misc/tex-storage-compressed-formats.html
+++ b/sdk/tests/conformance2/textures/misc/tex-storage-compressed-formats.html
@@ -98,7 +98,7 @@ function runTexCompressedFormatsTest(texCompressedFormats)
         // Test a 2D texture.
         var tex = gl.createTexture();
         gl.bindTexture(gl.TEXTURE_2D, tex);
-        gl.texStorage2D(gl.TEXTURE_2D, 1, internalformat, 4, 4);
+        gl.texStorage2D(gl.TEXTURE_2D, 1, internalformat, 1, 1);
         wtu.glErrorShouldBe(gl, gl.NO_ERROR,
                             "texStorage2D should succeed for " + enumToString(internalformat));
         gl.deleteTexture(tex);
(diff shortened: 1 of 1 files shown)


lexaknyazev:
I was able to reproduce this issue on GL Desktop with `--use-cmd-decoder=validating`.

lexaknyazev:
With the latest commit, the issue is consistently reproduced with validating command decoder on mobile and desktop, on D3D and GL.

sdk/tests/conformance2/textures/misc/tex-storage-compressed-formats.html
```diff
@@ -29,17 +29,48 @@
 var wtu = WebGLTestUtils;
 var canvas = document.getElementById("canvas");
 var gl = wtu.create3DContext(canvas, null, 2);
+
+// Either of these extensions should be available on all WebGL 2.0 contexts
 var WEBGL_compressed_texture_etc = gl.getExtension("WEBGL_compressed_texture_etc");
-var vao = null;
+var WEBGL_compressed_texture_s3tc = gl.getExtension("WEBGL_compressed_texture_s3tc");
```
jdashg: const

lexaknyazev: Done. Also changed dimensions to 4x4 because Firefox doesn't like 1x1 for compressed level 0.

jdashg: Oh! We should test that though. Let me check the specs.

jdashg: I think that's a bug in Firefox. The spec doesn't say anything explicitly, so I think it's implicitly allowed. (We have to support e.g. 1x1 mip:2 levels anyways)

jdashg: Bug filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1615838

lexaknyazev:
> Let's just test 1x1, and I'll fix Firefox. :)

Done.

jdashg:
Thanks!

Issue #2996, opened 2020-01-05
Clarify compressed texture entrypoints


lexaknyazev:
All WebGL extensions for compressed texture formats explicitly define new allowed enums only for `compressedTex[Sub]Image2D ` entrypoints (with ETC1 not supporting partial updates).

Strictly speaking, this is not fully accurate because in WebGL 2.0 context, compressed formats can also be passed to `texStorage2d` entrypoint.

This leads to a few additional questions:
- Which formats can be used for sliced 3D and 2D arrays in WebGL 2.0? Some GLES extensions may support those.
- OpenGL ES extensions for S3TC, S3TC sRGB, RGTC, and BPTC formats extend not only `compressedTex[Sub]Image2D` entrypoints but also regular `texImage2D`. If it's not the case with WebGL, extension specs should mention this.

Whatever the answers are, we probably need more tests to cover remaining cases.

kenrussell:
Please feel free to enhance the respective compressed texture extension specifications with notes about WebGL 2.0 support. Some of the compressed texture conformance tests have already been enhanced to call texStorage2D for testing purposes; your help would be greatly appreciated getting all of the commonly supported formats up to the same level of testing.

lexaknyazev:
Before I can update the specs and the tests, what is the WebGL WG position on those questions (2D arrays, sliced 3D, regular `texImage2D` entrypoints)?

jdashg:
The definition of TexStorage is in terms of TexImage and CompressedTexImage:
> If internalformat is a compressed texture format, then references to TexImage* should be replaced by CompressedTexImage*, with format, type and data replaced by any valid imageSize and data.

Generally I don't think there's any divergence between acceptable formats for 2D and 3D, so expect if 2D is allowed, for 3D to be supported too. It would be nice to clarify some spec text in this regard.

lexaknyazev:
> The definition of TexStorage is in terms of TexImage and CompressedTexImage

That's great - we don't need to be too explicit in the specs. Tests should still cover that, though.

---

> I don't think there's any divergence between acceptable formats for 2D and 3D

For example, GLES RGTC (BC4/BC5) extension says:
[...]

PR #3020, opened 2020-02-07
Expand ASTC tests

### LINKED ISSUE #3010: ASTC tests do not cover LDR/HDR profiles support
The extension declares `sequence<DOMString> getSupportedProfiles()` function that should provide accurate information about supported profiles. Currently, tests do not even call it.

### LINKED ISSUE #3016: ASTC NaN-to-magenta conversion does not work on Intel OpenGL
According to the ASTC spec, a block filled with zeros is considered to have "Reserved" mode. For such a case, the spec says:
> Decoders are required to recognize all illegal blocks and emit the standard error color value upon encountering an illegal block. Here is a comprehensive list of situations that represent illegal block encodings:
>- The block mode specified is one of the modes explicitly listed as Reserved.
>- ...

> For linear RGB decode, the error color may be either opaque fully-saturated magenta `(1.0, 0.0, 1.0, 1.0)` or a vector of four NaNs.

Tests convert NaNs to magenta in GLSL (added in #2962):
```glsl
vec4 c = texture2D(tex, texCoord);
if (all(notEqual(c, c))) {
    gl_FragData[0] = vec4(1, 0, 1, 1);
} else {
    gl_FragData[0] = texture2D(tex, texCoord);
}
```

This snippet does not work as expected on Intel OpenGL and produces (0, 0, 0, 0) for NaNs.

For this reason, WebGL 2.0 CTS fails with
```
FAIL texture should be initialized to error color at (0, 0) expected: 255,0,255,255 was 0,0,0,0
```

/cc @jdarpinian

lexaknyazev:
- Added `getSupportedProfiles` validation (fixes #3010)
  - ~~A simple polyfill should stay there (line 1935) until https://crbug.com/1049806 is fixed~~
- Added sRGB test vectors and refactored error color validation
- For every ASTC format (14 * 3), many code paths are validated:
  - non-zero border
  - mipmap generation attempt 
  - `NEAREST` and `LINEAR` filtering
  - `TEXTURE_2D`, `TEXTURE_2D_ARRAY`  (WebGL 2.0), and `TEXTURE_3D` (WebGL 2.0) targets
    - 3D slices are expected to work only if HDR profile is supported (see #2996)
  - initializing mutable textures from data
    - `compressedTexImage*D`
  - initializing mutable textures with zeros and filling actual afterwards
    - `compressedTexImage*D` followed by `compressedTexSubImage*D`
  - immutable textures (WebGL 2.0)
    - `texStorage*D` followed by `compressedTexSubImage*D`
  - error color generation from zero-initialized textures; formats that may result in NaNs are checked only on WebGL 2.0 (fixes #3016)

Results:
- Firefox (stable)
  - [x] Windows / OpenGL / Intel UHD 600 
  - [x] Android / OpenGL ES / Adreno 540 
- Chrome (canary)
  - [x] ~~immutable compressed 3D textures cannot be created due to [ANGLE validation bug](https://bugs.chromium.org/p/angleproject/issues/detail?id=4385)~~
  - [ ] immutable compressed 3D textures cannot be created due to [Chromium's command decoder bug](https://crbug.com/1051908)
  - [ ] `getSupportedProfiles` does not report HDR profle, https://crbug.com/1049880
  - [ ] sampling from freshly-allocated compressed texture array leads to unexpected error on validating command decoder (see https://crbug.com/1051217 and #3019)

/cc @kenrussell @shrekshao @jdarpinian


sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
```diff
@@ -1678,9 +1932,104 @@
     // Test that texture upload buffer size is validated correctly.
     ctu.testFormatRestrictionsOnBufferSize(gl, validFormats, expectedByteLength, getBlockDimensions);
 
-    // Tests ASTC texture with every format
-    testLDRTextures();
-    testHDRTextures();
+    // TODO: remove after http://crbug.com/1049806 is fixed
+    ext.getSupportedProfiles = ext.getSupportedProfiles || (() => ["ldr", "hdr"]);
```
shrekshao: getSupportedProfiles is added to chrome (simply returns ["ldr"] for now). I think it is safe to remove it.

lexaknyazev: Without HDR profile support, `TEXTURE_3D` target must generate `INVALID_OPERATION`. This is already implemented in ANGLE and covered in WebGL tests by this PR. To pass this test, Chromium has to accurately report supported profiles.

sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
```diff
@@ -1646,6 +1893,13 @@
 };
 var name;
 
+var errorColor = [255, 0, 255, 255];
```
jdashg: const

sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
```diff
@@ -1646,6 +1893,13 @@
...
+var hasHdr = false;
```
jdashg: let

sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
```diff
@@ -1678,9 +1932,101 @@
...
+    if (profiles.length > 0) {
+        // Check for HDR support first, since it affects sliced 3D support for LDR
+        if (profiles.indexOf("hdr") != -1) {
+            hasHdr = true;
+            testHDRTextures();
+        } else {
+            testPassed("HDR profile is not supported.");
+        }
+
+        if (profiles.indexOf("ldr") != -1) {
```
jdashg: Instead of checking for an empty array, and also requiring ldr, you can simply require ldr to be present, and that guarantees we don't have an empty array.

sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
```diff
@@ -1772,62 +2130,234 @@
...
+        return tex; // Keep reference to the texture so it could be deleted
+    }
+
+    function uploadSubData(target) {
+        if (target == gl.TEXTURE_2D) {
+            gl.compressedTexSubImage2D(target, 0, 0, 0, width, height, format, data);
+            wtu.glErrorShouldBe(gl, gl.NO_ERROR, "uploading compressed texture data via compressedTexSubImage2D");
+        } else if (target == gl.TEXTURE_2D_ARRAY || target == gl.TEXTURE_3D) {
+            gl.compressedTexSubImage3D(target, 0, 0, 0, 0, width, height, 1, format, data);
+            wtu.glErrorShouldBe(gl, gl.NO_ERROR, "uploading compressed texture data via compressedTexSubImage3D");
```
jdashg: Consider running both 2D and 3D, and checking for errors based on the target.

sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
lexaknyazev: Done

lexaknyazev:
FWIW, Firefox runs this test much faster than Chromium.

jdashg:
Thanks!

sdk/tests/conformance/extensions/webgl-compressed-texture-astc.html
```diff
@@ -1772,62 +2126,248 @@
...
+        // 2D texture
+        if (check2DTarget) {
+            switchProgram(gl.TEXTURE_2D);
+
+            // immutable
+            tex = setupEmptyTexture(gl.TEXTURE_2D, true);
+            wtu.glErrorShouldBe(gl, gl.NO_ERROR, "allocating compressed texture via texStorage2D");
+            checkErrorColor();
+            uploadSubData(gl.TEXTURE_2D);
+            checkSampling(gl.TEXTURE_2D)
```
shrekshao: chekSampling and the followings are missing semicolons `;`

PR #3025, opened 2020-02-19
Use 127/255 instead of 0.5 when we expect 127 later.

jdashg:


jdashg:
This fixes a failure in Firefox.

PR #3026, opened 2020-02-20
Add compressedTexSubImage validation test for bptc and rgtc

shrekshao:
CTS used to miss the validation test for bptc and rgtc (There are some compressedTexSubImage* tests for pvrtc, s3tc and astc). Now latest Firefox and Chromium on Linux Nvidia are tested passing these newly added tests.

### DIFF
--- a/sdk/tests/js/tests/compressed-texture-utils.js
+++ b/sdk/tests/js/tests/compressed-texture-utils.js
@@ -156,6 +156,42 @@ var testFormatRestrictionsOnBufferSize = function(gl, validFormats, expectedByte
     }
 };
 
+/**
+ * @param {WebGLRenderingContextBase} gl
+ * @param {Object} validFormats Mapping from format names to format enum values.
+ * @param expectedByteLength A function that takes in width, height and format and returns the expected buffer size in bytes.
+ * @param getBlockDimensions A function that takes in a format and returns block size in pixels.
+ * @param {number} width Width of the image in pixels.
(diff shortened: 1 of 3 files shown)


sdk/tests/js/tests/compressed-texture-utils.js
```diff
@@ -156,14 +156,51 @@ var testFormatRestrictionsOnBufferSize = function(gl, validFormats, expectedByte
...
 return {
     formatToString: formatToString,
     insertCaptionedImg: insertCaptionedImg,
     makeScaledImage: makeScaledImage,
     testCompressedFormatsListed: testCompressedFormatsListed,
     testCompressedFormatsUnavailableWhenExtensionDisabled: testCompressedFormatsUnavailableWhenExtensionDisabled,
     testCorrectEnumValuesInExt: testCorrectEnumValuesInExt,
-    testFormatRestrictionsOnBufferSize: testFormatRestrictionsOnBufferSize
+    testFormatRestrictionsOnBufferSize: testFormatRestrictionsOnBufferSize,
+    testTexSubImageDimensions: testTexSubImageDimensions
```
jdashg: Can you include the trailing comma, so we don't get these spurious line changes in csets?

shrekshao: Sure. Added. (Thanks for the explanation, now I get the validity of these trailing comma)

PR #3027, opened 2020-02-22
Backport changes to shaders-with-invariance to 2.0.0 and 1.0.3.

jdashg:


jdashg:
Firefox passes this on master, but not in the 2.0.0 snapshot.

PR #3028, opened 2020-02-25
Add GLSL preprocessor comment behavior tests.

jdashg:
Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1617512


jdashg:
There's a (SFW :)) standalone testcase here: https://bug1617512.bmoattachments.org/attachment.cgi?id=9128672

jdashg:
This matches Chrome's behavior, because that seems like the best choice both technically and pragmatically, to avoid breaking changes.

PR #3029, opened 2020-02-27
Duplicate and centralize glsl-comment testcases into this file.

jdashg:
It's nicer to have a one-stop-shop for validating tests for this
behavior.


jdashg:
The ones without attribution are new tests.

kenrussell:
Sounds good. Nice work Jeff on the new tests.

Issue #2902, opened 2019-06-12
Discrepancy in unsized DEPTH_STENCIL handling between Firefox and Chrome


kenrussell:
In https://github.com/mrdoob/three.js/issues/16756 it was pointed out that Firefox's WebGL 2.0 implementation seems to be allowing the unsized DEPTH_STENCIL format as an argument to renderbufferStorageMultisample, while Chrome's implementation rejects it.

WebGL 2.0 and OpenGL ES 3.0 tightened the rules around texture and renderbuffer format specification so a test should be written to make all implementations behave identically. Ideally it would be an error to specify the unsized DEPTH_STENCIL format in WebGL 2.0.

CC @jdashg

jdashg:
We process renderbufferStorage as renderbufferStorageMultisample(samples=0), so I'd prefer to Chrome to expand what it accepts for renderbufferStorageMultisample, and maintain this layered structure of these entrypoints.

jdashg:
This is easy enough to fix I guess.

PR #3031, opened 2020-02-28
 Make renderbufferStorageMultisample with DEPTH_STENCIL and samples>0 generate INVALID_OPERATION.

jdashg:
Firefox allows DEPTH_STENCIL with samples >= 0.
Chrome allows samples==0, but INVALID_ENUM for samples >= 1.

### DIFF
--- a/sdk/tests/conformance2/renderbuffers/multisampled-depth-renderbuffer-initialization.html
+++ b/sdk/tests/conformance2/renderbuffers/multisampled-depth-renderbuffer-initialization.html
@@ -29,6 +29,19 @@ if (!gl) {
     // Set the clear color to green. It should never show up.
     gl.clearColor(0, 1, 0, 1);
 
+    debug("Test renderbufferStorageMultisample with webgl1's DEPTH_STENCIL.");
+    {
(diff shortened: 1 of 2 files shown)


jdashg:
The other option is to make `format==DEPTH_STENCIL && samples > 0` generate INVALID_OPERATION, since the entrypoint accepts the enum (INVALID_ENUM), just not the way it is used. (INVALID_OPERATION)

jdashg:
This is a resolution for #2902.

kenrussell:
My preference would be to reject the unsized DEPTH_STENCIL enum with samples > 0 with INVALID_OPERATION. Would you please consider updating the test to assert that instead? We'll be happy to change Chrome's implementation to generate INVALID_OPERATION instead of INVALID_ENUM to follow.

CC @jdarpinian @shrekshao for any thoughts.

jdashg:
Sure!

jdashg:
PTAL!

kenrussell:
@jdarpinian @shrekshao heads up that we need to file a bug about changing this validation in Chrome because this test will start failing during the next conformance roll.

PR #3033, opened 2020-03-05
Add WebGL-specific limitations to EXT_draw_buffers_indexed

### LINKED ISSUE #2938: [EXT_draw_buffers_indexed] Constant Alpha Blend Factors
D3D11 supports only "constant color" blending factors, so ANGLE emulates "constant alpha" factors by rewriting the constant color to `aaaa`. See [`StateManager11::syncBlendState`](https://github.com/google/angle/blob/087f1384f233792921d34c0ca5b5a90fac75b070/src/libANGLE/renderer/d3d/d3d11/StateManager11.cpp#L1246-L1264).

This is reflected in the WebGL spec, [Blending With Constant Color](https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.15):
> In the WebGL API, constant color and constant alpha cannot be used together as source and destination factors in the blend function. A call to `blendFunc` will generate an `INVALID_OPERATION` error if one of the two factors is set to `CONSTANT_COLOR` or `ONE_MINUS_CONSTANT_COLOR` and the other to `CONSTANT_ALPHA` or `ONE_MINUS_CONSTANT_ALPHA`. A call to `blendFuncSeparate` will generate an `INVALID_OPERATION` error if `srcRGB` is set to `CONSTANT_COLOR` or `ONE_MINUS_CONSTANT_COLOR` and `dstRGB` is set to `CONSTANT_ALPHA` or `ONE_MINUS_CONSTANT_ALPHA` or vice versa.

The OpenGL ES `EXT_draw_buffers_indexed` extension (and the corresponding OpenGL 4.0 feature) enables separate blend factors per each color output, but the constant color remains the same for all of them. This means that the WebGL `EXT_draw_buffers_indexed` extension has to have an additional restriction to work on D3D11.

There are two options.
- Disallow use of `CONSTANT_ALPHA` and `ONE_MINUS_CONSTANT_ALPHA` factors with indexed blend state completely (easier to implement).
- Extend existing WebGL restriction to indexed blend states. That would require more runtime validation on D3D11.

lexaknyazev:
Fixes #2938.

There is also an identical `OES_draw_buffers_indexed` extension. Should we expose it instead?

### DIFF
--- a/extensions/EXT_draw_buffers_indexed/extension.xml
+++ b/extensions/EXT_draw_buffers_indexed/extension.xml
@@ -19,6 +19,14 @@
   <overview>
     <mirrors href="https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_draw_buffers_indexed.txt"
              name="EXT_draw_buffers_indexed">
+             <addendum>Core WebGL <a href="https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.15">blend factors limitation</a>
+              is also applied to the new entrypoints.
+             </addendum>
+             <addendum>
+              When <code>CONSTANT_COLOR</code> or <code>ONE_MINUS_CONSTANT_COLOR</code> are used as blend color factors of an active draw buffer and
+              <code>CONSTANT_ALPHA</code> or <code>ONE_MINUS_CONSTANT_ALPHA</code> are used as blend color factors of another active draw buffer,
+              any draw call will generate an <code>INVALID_OPERATION</code> error.
+             </addendum>
     </mirrors>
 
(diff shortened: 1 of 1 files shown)


jdashg:
We should actually rename ours to OES_, since that's the ES extension,
generally.

lexaknyazev:
OK. Could we merge this PR?
I'll open a new one with renaming.

jdashg:
This looks like what we agreed on.

lexaknyazev:
FWIW, it's already implemented in ANGLE for D3D11, including WebGL validation.

PR #3034, opened 2020-03-05
Rename EXT_draw_buffers_indexed to OES_draw_buffers_indexed

lexaknyazev:


lexaknyazev:
Both of them are defined only for ES.

jdashg:
Yep, my bad. OES_ is ES's ARB, so OES_ is the ratified one, which still sounds like the right choice.

PR #3038, opened 2020-03-13
Add Render();NoOpResize();Render(); tests.

jdashg:
Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1621523

### DIFF
--- a/sdk/tests/conformance/canvas/render-after-resize-test.html
+++ b/sdk/tests/conformance/canvas/render-after-resize-test.html
@@ -1,5 +1,5 @@
 <!--
-Copyright (c) 2019 The Khronos Group Inc.
+Copyright (c) 2020 The Khronos Group Inc.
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/canvas/render-after-resize-test.html
```diff
@@ -14,58 +14,116 @@
...
-const smallWidth = 300;
-const smallHeight = 150;
+const SMALL = 2;
 // Changing this size to something smaller produces
 // different results. Sometimes wrong, sometimes correct.
-const largeWidth = 1200;
-const largeHeight = 672;
+const LARGE = 1200;
 
 function render() {
```
kenrussell: Could you remove this empty function?

jdashg:
Thanks!

PR #3039, opened 2020-03-13
Non ascii via codepoints

jdashg:
Use arrays of code-points instead of having utf-8 strings.

This makes it easier to handle these files, and I've confirmed the test data is the same.                                                                                                Some of Mozilla's integration infra doesn't handle utf-8 gracefully,
making it harder than it needs to be to update.


jdashg:
The tooling bug on our side: https://github.com/mystor/phlay/issues/66

sdk/tests/conformance/glsl/misc/non-ascii.vert.html
```diff
@@ -18,18 +18,35 @@
...
+
+vertexShader.textContent = `
 // Non ascii data in source should fail
 // See GLSL ES Spec 1.0.17 section 3.1 and 3.2
-// ����͂`�r�b�h�h�ł͂Ȃ��ł��B
-// �s�������@�h���@�m�����@�`�r�b�h�h
-uniform mat4 �m�����`�r�b�h�h;
+// ${a}
+// ${b}
+uniform mat4 ${b};
```
kenrussell: This identifier's changed from the original test. The comment has spaces but the uniform's name didn't. Would you consider restoring this to the original value?

jdashg: You know, this is the first time I've seen it render as ~"NotASCII". I've only ever seen garbage elsewhere. :S

PR #3040, opened 2020-03-18
Add test for compressedTex(Sub)Image2D(DataView).

jdashg:
Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1622536


jdashg:
I prefer to keep it this way, but if you want to refactor it, that is probably slightly better.
I find it a little annoying because then I have to hunt open multiple files. We have this this depends-on-version in a number of tests, though, so I always load the conformance suite runner first, and run tests through that, or at least use that to generate the url with the right flags.

PR #3041, opened 2020-03-18
Re-add BufferSource to idl.

jdashg:
It had disappeared at some point.


kenrussell:
I'm pretty sure this was removed because it was promoted to the core Web IDL spec:
https://heycam.github.io/webidl/#BufferSource

Does that seem sufficient or would you like this re-added here?

jdashg:
I should have looked this up better!

PR #3043, opened 2020-03-19
Do is-webgl2 checks by duck-typing, not instanceof.

jdashg:
Doing instanceof makes it hard for webgl-1on2 testing, since the rendering context there is webgl1 behavior, but instanceof WebGL2RenderingContext.


jdashg:
Yes that would be better.

kenrussell:
I'll take care of merging this and following up with the desired refactor.

kenrussell:
Merging now. Have follow-on ready.

PR #3044, opened 2020-03-19
Refactor compressed-tex-image into WebGL 1.0 and 2.0 tests.

kenrussell:
Follow-on to #3040.

### DIFF
--- a/sdk/tests/conformance2/textures/misc/compressed-tex-image.html
+++ b/sdk/tests/conformance2/textures/misc/compressed-tex-image.html
@@ -16,7 +16,7 @@ found in the LICENSE.txt file.
 <div id="description"></div>
 <div id="console"></div>
 <script>
-var contextVersion = 2;
+const contextVersion = 2;
 </script>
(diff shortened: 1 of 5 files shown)


kenrussell:
Tested both versions both inside the test harness as well as standalone - both have the desired behavior.

sdk/tests/conformance2/textures/misc/compressed-tex-image.html
```diff
@@ -0,0 +1,24 @@
...
+<meta charset="utf-8">
+<link rel="stylesheet" href="../../../resources/js-test-style.css"/>
+<script src="../../../js/js-test-pre.js"></script>
+<script src="../../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<script>
+var contextVersion = 2;
```
jdashg: `const` please!

kenrussell: Thanks, done.

PR #3045, opened 2020-03-23
Upstream some copyTexSubImage2D test cases from WebKit.

kenrussell:
These tests uncover bugs in some WebGL implementations that will be
filed separately.


kenrussell:
@austinEng @kainino0x @grorg FYI.

kenrussell:
Thanks @jdashg for your quick review. Merging now.

PR #3047, opened 2020-03-27
Add tests for SharedArrayBuffers to entrypoints that should take them.

jdashg:
Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1624944

### DIFF
--- a/sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
+++ b/sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
@@ -94,6 +94,11 @@ function bufferDataSizesTest() {
 
         { parameter: new ArrayBuffer(0), expectedBufferSize: 0 },
         { parameter: new ArrayBuffer(4), expectedBufferSize: 4 },
+        { parameter: new SharedArrayBuffer(3), expectedBufferSize: 3 },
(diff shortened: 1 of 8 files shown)


jdashg:
Oh, I figured that we'd be able to unconditionally create SABs, just not transfer them. Sure, will fix.

sdk/tests/conformance/buffers/buffer-data-and-buffer-sub-data.html
```diff
@@ -94,15 +94,24 @@
...
         { parameter: new ArrayBuffer(0), expectedBufferSize: 0 },
         { parameter: new ArrayBuffer(4), expectedBufferSize: 4 },
+        { parameter: new Uint8Array(new ArrayBuffer(5)), expectedBufferSize: 5 },
+        { parameter: new DataView(new ArrayBuffer(7)), expectedBufferSize: 7 },
 
         { parameter: "WebGL Rocks!", expectedBufferSize: 0 },
         { parameter: { mystring: "WebGL Rocks!" }, expectedBufferSize: 0 },
     ];
 
+    if (SharedArrayBuffer) {
```
kenrussell: This needs to be:
`  if (window.SharedArrayBuffer)
`
to avoid throwing an exception in Safari. Could you update here and throughout?

PR #3049, opened 2020-04-08
Update for Web IDL extended attribute changes

domenic:
[NoInterfaceObject] → [LegacyNoInterfaceObject]. This follows https://github.com/heycam/webidl/pull/870.

This also replaces a couple instances of [NoInterfaceObject] with the interface mixin concept, as was previously done in 2df151be79b1a44d41464b149d6044fbf6bbf2f7 and a6f99093cbc1e22112b0d36ae368438d268f2376.

---

I imagine some of these changes might not be desired; it seems there are old copies of the spec in the master branch of the repo and maybe those need to stay for archival purposes. Let me know which, if any, changes to revert.

### DIFF
--- a/extensions/ANGLE_instanced_arrays/extension.xml
+++ b/extensions/ANGLE_instanced_arrays/extension.xml
@@ -27,7 +27,7 @@
     </p>
   </overview>
   <idl xml:space="preserve">
-[NoInterfaceObject]
+[LegacyNoInterfaceObject]
 interface ANGLE_instanced_arrays {
     const GLenum VERTEX_ATTRIB_ARRAY_DIVISOR_ANGLE = 0x88FE;
     void drawArraysInstancedANGLE(GLenum mode, GLint first, GLsizei count, GLsizei primcount);
--- a/extensions/extension.xsl
+++ b/extensions/extension.xsl
@@ -411,7 +411,7 @@
 <xsl:template match="interface" mode="newfun">
   <dt class="idl-code">
 	<xsl:if test="@noobject = 'true'">
-	  <xsl:text>[LegacyLegacyNoInterfaceObject]</xsl:text><br/>
+	  <xsl:text>[LegacyNoInterfaceObject]</xsl:text><br/>
 	</xsl:if>
     <xsl:text>interface </xsl:text>
     <em><xsl:value-of select="@name" /></em>
--- a/extensions/extension.xsl
+++ b/extensions/extension.xsl
@@ -411,7 +411,7 @@
 <xsl:template match="interface" mode="newfun">
   <dt class="idl-code">
 	<xsl:if test="@noobject = 'true'">
-	  <xsl:text>[NoInterfaceObject]</xsl:text><br/>
+	  <xsl:text>[LegacyLegacyNoInterfaceObject]</xsl:text><br/>
 	</xsl:if>
     <xsl:text>interface </xsl:text>
     <em><xsl:value-of select="@name" /></em>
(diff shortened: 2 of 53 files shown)


extensions/extension.xsl
```diff
@@ -411,7 +411,7 @@
 <xsl:template match="interface" mode="newfun">
   <dt class="idl-code">
 	<xsl:if test="@noobject = 'true'">
-	  <xsl:text>[NoInterfaceObject]</xsl:text><br/>
+	  <xsl:text>[LegacyLegacyNoInterfaceObject]</xsl:text><br/>
```
kenrussell, line 414: Typo.

extensions/ANGLE_instanced_arrays/extension.xml
```diff
@@ -27,7 +27,7 @@
     </p>
   </overview>
   <idl xml:space="preserve">
-[NoInterfaceObject]
+[LegacyNoInterfaceObject]
```
kenrussell, line 30: For this and other WebGL extensions, do you have any guidance about how to change the spec text to avoid the use of [LegacyNoInterfaceObject], but achieve the same goal - allow just type "object" to be returned from WebGLRenderingContextBase.getExtension(), not have the extensions' interface names appear in the global namespace, but have those objects include those interfaces? Would interface mixins work?

domenic: I'm not quite sure what the intent is here. `[LegacyNoInterfaceObject]` doesn't change the type returned; this is still a proper class, named `ANGLE_instanced_arrays`. You can verify that using e.g. `ctx.getExtension("...").constructor.name === "ANGLE_instanced_arrays"`. (Not sure I got the syntax right for obtaining an instance, but you get the idea.)

The only thing `[LegacyNoInterfaceObject]` does is prevent `window.ANGLE_instanced_arrays` from existing. This, in turn, prevents feature detection (both of the overall feature, e.g. `if (window.ANGLE_instanced_arrays)`, and of individual methods, e.g. `if (window.ANGLE_instanced_arrays.drawArraysInstancedANGLE)`.

So my advice would be to just remove the `[LegacyNoInterfaceObject]` and allow the property to exist on the global. Hiding it doesn't seem to give much benefit. Maaaybe it'd be worth renaming them at the same time to something more idiomatic, e.g. `interface ANGLEInstancedArrays { ... }`, but I can see how having to define two names for things (one for use in the `getExtension()` argument and one for the global property) is not appealing, so having non-idiomatic naming for Web GL extension interfaces would be reasonable.

kainino0x: It's arguably beneficial to prevent feature detection on `window.ANGLE_instanced_arrays`. Checking whether that class exists doesn't tell you anything about whether you can use it; it only tells you whether the browser implements the interface, so allowing it could be misleading. Feature detection must be done using getExtension() because support is hardware dependent.

jdashg: I disagree, because it makes up-front prototype patching impossible. We should preserve our existing behavior in this PR, and make a new issue for whether to change behavior.

jdashg:
@domenic Can you sign the CLA?

domenic:
I'm not sure; I'm looking into whether that's allowed for Google employees.

jdashg:
@kenrussell from Google can probably help! :)

domenic:
Indeed, @kenrussell confirmed to me it's fine, so I've signed now :).

BTW this should not land until the https://github.com/heycam/webidl/pull/870 pull request does; I'm happy to ping this thread after that happens.

jdashg:
> BTW this should not land until the [heycam/webidl#870](https://github.com/heycam/webidl/pull/870) pull request does; I'm happy to ping this thread after that happens.

Please do!

domenic:
Web IDL PR merged, so this can be merged to!

kenrussell:
Thanks for following up, sounds good! Merging now.

PR #3032, opened 2020-02-28
Remove now-broken `while (!img.complete) {}`.

jdashg:
`HTMLImageElement.complete` no longer changes except on microtask boundaries.
Backported minimal fixes to 1.0.3 and 2.0.0.
More invasive fix to make things nicer moving forward for top-of-tree.

PR #3052, opened 2020-04-10
Fix crossorigin tests to default to remote-web-url, but fallback to l…

jdashg:
…ocal on localhost.

Fixes regressions from #3032:
> all/conformance/more/functions (Passed: 90/93 Failed: 2/93 Timeout: 1/93 in 26.81 seconds)
> 
> readPixelsBadArgs.html(*timeout*)
> 
> texImage2DHTML.html (Passed: 1/2 Failed: 1/2 in 170.0 ms)
> 
>     failed: testTexImage2DNonSOP
> 
> texSubImage2DHTML.html (Passed: 1/2 Failed: 1/2 in 251.0 ms)
> 
>     failed: testTexImage2DNonSOP

### DIFF
--- a/sdk/tests/conformance/textures/misc/origin-clean-conformance-offscreencanvas.html
+++ b/sdk/tests/conformance/textures/misc/origin-clean-conformance-offscreencanvas.html
@@ -12,6 +12,12 @@ found in the LICENSE.txt file.
 <link rel="stylesheet" href="../../../resources/js-test-style.css"/>
 <script src="../../../js/js-test-pre.js"></script>
 <script src="../../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<canvas id="canvas"></canvas>
+<img id="img" style="display:none;">
 <script>
 "use strict";
 var wtu = WebGLTestUtils;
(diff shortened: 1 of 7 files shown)


sdk/tests/conformance/textures/misc/origin-clean-conformance-offscreencanvas.html
```diff
@@ -105,13 +109,15 @@
   finishTest();
 }
 
-wtu.setupImageForCrossOriginTest("#img", defaultImgUrl, localImgUrl, imageLoaded);
+(async function() {
+    const img = document.getElementById('img');
+    try {
+        await wtu.awaitOrTimeout(wtu.loadCrossOriginImage(img, defaultImgUrl, localImgUrl), 1000);
+    } catch (e) {
+        console.log(`Image setup failed (${e}), running tests anyway.`);
```
kenrussell, line 117: The image setup is not supposed to fail given the localhost fallback. Silently running the tests in this case will cause silent flakiness on automated test systems. Could you please change this and the other tests to fail if an exception is thrown at this point?

jdashg: This PR matches the existing behavior, but I'm happy to change the behavior to fail-on-error in a follow-up.

kenrussell: OK, sounds fine to leave this as a pure refactoring.

PR #3054, opened 2020-04-10
Fix flakiness of context-attribute-preserve-drawing-buffer.html.

jdashg:
Previously, at smaller window sizes, the preserveDrawingBuffer:false
test canvas was off the visible region of the page, and thus not cleared
since it was not Painted.

The fix:
* Increase the min-width of the test-runner test iframe.
* Decrease the size of canvases in context-attribute-preserve-drawing-buffer.html.


kainino0x:
Whoops, I was hoping the sidebar thing wouldn't cause issues like this. But I guess it's not strictly any more error prone than the old iframe (which was wider but shorter).

Issue #3058, opened 2020-04-24
Switch statement tests


samsinsane:
I encountered this bug the other day in both Chrome and Firefox, and I'm not entirely sure what to do about it. The `default` case is not executed in this scenario:
```glsl
switch(0)
{
  default:
    ...
}
```
Bit of backstory: At work we use Shader Conductor to convert our DirectX shaders to Metal, OpenGL, and OpenGL ES/WebGL shaders. This pattern seems to be generated due to functions being inlined into `main`, and the functions with multiple returns end up generating this `switch` statement. For what it's worth, this pattern worked fine for OpenGL on Windows, and Metal on macOS.

I was able to workaround this issue by replacing `default` with `case 0`, but figured I should at least report this somewhere and I stumbled across this repo while trying to find out if this was "by design" in WebGL. Since I couldn't find anything saying that `default` statements can be ignored under certain conditions, and since it seems to be browser independent, I figure that it's a conformance test problem and should be reported here? Unfortunately, I don't really have that much time to dedicate to this, so chasing individual browser vendors around on this isn't feasible, but I can add more tests to `switch-case.html` if that gets the ball rolling?

kenrussell:
This sounds like a bug. Yes, please create a pull request adding more tests to switch-case.html.

Please also provide the contents of about:gpu from Chrome or about:support from Firefox on the machines where this isn't working. That will help narrow down whether this is a browser, shader translator, OS-specific, or graphics driver problem. Thanks.

samsinsane:
> This sounds like a bug. Yes, please create a pull request adding more tests to switch-case.html.

No worries, I can do that.

> Please also provide the contents of about:gpu from Chrome or about:support from Firefox on the machines where this isn't working.

Both have a "Copy to Clipboard" button which seems to contain significantly less information but seems more inline with what you want? Is this the data that I should post, or do you need the full thing, tables and all?

kainino0x:
In chrome at least, the button should include everything (it just uses JS to select-all and copy).

samsinsane:
Strange, it seems to be copying more than 7 lines today, I must have done something wrong yesterday. I've attached the reports here:
[HomeDesktopFirefox.txt](https://github.com/KhronosGroup/WebGL/files/4534657/HomeDesktopFirefox.txt)
[HomeDesktopChrome.txt](https://github.com/KhronosGroup/WebGL/files/4534659/HomeDesktopChrome.txt)
[...]

PR #3060, opened 2020-04-25
Added switch statement tests for default and literal values

samsinsane:
Related to #3058

### DIFF
--- a/sdk/tests/conformance2/glsl3/switch-case.html
+++ b/sdk/tests/conformance2/glsl3/switch-case.html
@@ -36,6 +36,59 @@ void main()
     }
 }
 </script>
+<script id="fshaderDeclarationInsideSwitchDefault" type="x-shader/x-fragment">#version 300 es
+
+precision highp float;
+out vec4 my_FragColor;
+
+uniform int u_zero;
+
+void main()
+{
+    my_FragColor = vec4(1, 0, 0, 1);
+    switch (u_zero)
+    {
+        default:
+            ivec2 i;
+            i = ivec2(1, 0);
+            my_FragColor = vec4(0, i[0], 0, 1);
+    }
+}
+</script>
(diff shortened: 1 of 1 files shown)


jdashg:
fshaderDeclarationInsideSwitchLiteralDefault fails on Firefox Windows with ANGLE, but passes without ANGLE (native gl amd).

jdashg:
This might be a D3D bug, given the failure is with this seemingly-reasonable hlsl:
```
// FRAGMENT SHADER BEGIN
// GLSL BEGIN
#version 300 es
precision highp float;
out vec4 my_FragColor;
void main()
{
    my_FragColor = vec4(1, 0, 0, 1);
    switch (0)
    {
        default:
            ivec2 i;
            i = ivec2(1, 0);
            my_FragColor = vec4(0, i[0], 0, 1);
    }
}
// GLSL END
// INITIAL HLSL BEGIN
float4 vec4_ctor(int x0, int x1, int x2, int x3)
{
    return float4(x0, x1, x2, x3);
}
#ifdef ANGLE_ENABLE_LOOP_FLATTEN
#define LOOP [loop]
#define FLATTEN [flatten]
#else
#define LOOP
#define FLATTEN
#endif
#define ATOMIC_COUNTER_ARRAY_STRIDE 4
// Varyings
static float4 out_my_FragColor = {0, 0, 0, 0};
cbuffer DriverConstants : register(b1)
{
};
@@ PIXEL OUTPUT @@
PS_OUTPUT main(@@ PIXEL MAIN PARAMETERS @@){
@@ MAIN PROLOGUE @@
(out_my_FragColor = float4(1.0, 0.0, 0.0, 1.0));
{
int2 _i2324 = {0, 0};
switch (0) {
default:
(_i2324 = int2(1, 0));
(out_my_FragColor = vec4_ctor(0, _i2324[0], 0, 1));
break;
}
}
return generateOutput();
}
// INITIAL HLSL END
// FRAGMENT SHADER END
```

Issue #3059, opened 2020-04-24
getShaderPrecisionFormat conformance test not sufficient


greggman:
Safari is hard coded to just return desktop values for  `getShaderPrecisionFormat`. 

https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/platform/graphics/opengl/GraphicsContextGLOpenGLBase.cpp#L362

It seems like the getShaderPrecisionFormat test needs to test by actually rendering. If the implementation says for mediump float that precision is 23, rangeMin is 127 and rangeMax is 127 then basically just try to render something that's out of range for an typical meduimp value and see that you get that value back.

As it is apparently the only way for an app to test if mediump really gets mediump is to do that test themselves since asking via `getShaderPrecisionFormat` will fail to tell you on iOS.

kenrussell:
Could you please help improve the conformance test in this area? A pull request would be greatly appreciated.

As an aside, collaboration is actively ongoing to improve Safari's WebGL implementation. Please feel free to file bugs on bugs.webkit.org under the WebGL component. Still, improving the conformance test is the best way to highlight gaps in implementations.

greggman:
I filed a bug.

https://bugs.webkit.org/show_bug.cgi?id=211013

Will try to update conformance test

PR #3061, opened 2020-04-26
add rendering test to shader precisions format test

greggman:
to weed out implementions that incorrectly report precision

#3059 

I'm not entirely sure this is a valid test but it seems to work. My reading of the spec is that there is all kinds of leeway here about when an were the precision stuff takes effect. 

The 2.0 spec says overflowing an int is undefined so anything could happen. The code is uploading an int that should overflow in Safari iOS and it is failing as it should but I tried a bunch of other things before I got to this example that succeded where I expected it to fail and failed where I expected it to succeed.

All that is a way of saying I'm not confident in this code but it is running and does what I expect on the devices I have to test with.

### DIFF
--- a/sdk/tests/conformance/misc/shader-precision-format.html
+++ b/sdk/tests/conformance/misc/shader-precision-format.html
@@ -128,6 +128,209 @@ shouldBeTrue('shaderPrecisionFormat.rangeMin == shaderPrecisionFormat2.rangeMin'
 shouldBeTrue('shaderPrecisionFormat.rangeMax == shaderPrecisionFormat2.rangeMax');
 shouldBeTrue('shaderPrecisionFormat.precision == shaderPrecisionFormat2.precision');
 
+debug("");
+debug("Test that specified precision matches rendering results");
+debug("");
+
+function testRenderPrecisionSetup(gl, shaderPair) {
+  const program = wtu.setupProgram(gl, shaderPair);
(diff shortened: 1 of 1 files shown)


kenrussell:
Thanks for this contribution. I haven't had the time to confirm that this passes on WebKit on iOS with https://bugs.webkit.org/show_bug.cgi?id=211013 . Perhaps @grorg or @jdarpinian could help verify it?

kenrussell:
I've tested this on the iOS Simulator with top-of-tree WebKit, both with and without the patch for https://bugs.webkit.org/show_bug.cgi?id=211013 applied. Unfortunately, at least on that platform, this new test passes both before and after.

@grorg would it be possible for you to test on a physical iOS device to make sure that these new tests catch the bug that was previously present?

Or - if anyone has an iOS device with the iOS 14 Beta installed and can compare to results on a different device with iOS 13, as far as I understand, Safari on the beta contains this fix.

grorg:
I checked on an iOS device and confirmed this addition will fail when the WebKit commit 260789 is missing.

kenrussell:
Thanks @grorg for confirming that this test catches the bug!

kenrussell:
Note that we've found that this newly added test fails on at least one ARM GPU: http://crbug.com/1108368 . Still needs to be investigated.

PR #3063, opened 2020-04-27
Add deleted-object-behavior.html to 00_test_list.txt

RafaelCintron:
deleted-object-behavior was added to the conformance test but never added to any test_list.txt files.

Add deleted-object-behavior.html to 00_test_list.txt as part this change.

### DIFF
--- a/sdk/tests/conformance/context/00_test_list.txt
+++ b/sdk/tests/conformance/context/00_test_list.txt
@@ -12,7 +12,7 @@ context-attributes-alpha-depth-stencil-antialias.html
 context-lost-restored.html
 context-lost.html
 --max-version 1.9.9 context-type-test.html
-deleted-object-behavior.html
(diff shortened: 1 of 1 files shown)


kenrussell:
Yikes, sorry about that! I forgot this when adding this test. Thanks Rafael for catching and fixing it!

sdk/tests/conformance/context/00_test_list.txt
```diff
@@ -12,6 +12,7 @@ context-attributes-alpha-depth-stencil-antialias.html
 context-lost-restored.html
 context-lost.html
 --max-version 1.9.9 context-type-test.html
+deleted-object-behavior.html
```
kenrussell, line 15: Rafael, could you please add --min-version 1.0.4 here? This is just for bookkeeping, to know roughly when the new tests were added. Thanks.

kenrussell:
Thanks @RafaelCintron !

PR #3065, opened 2020-04-30
Update test utils to use video.rVFC

tguilbert-google:
The startPlayingAndWaitForVideo function uses various heuristics to
determine whether or not a video has started playing. These heuristics
mostly rely on currentTime progress, but don't aren't necessarily
reliable when determining wether there is actually new video data to
be consumed.

This commit adds cross browser support for the
video.requestVideoFrameCallback API, which can be used to replace the
heuristics, and potentially deflake tests.

### DIFF
--- a/sdk/tests/js/webgl-test-utils.js
+++ b/sdk/tests/js/webgl-test-utils.js
@@ -2913,6 +2913,23 @@ var requestAnimFrame = function(callback) {
   _requestAnimFrame.call(window, callback);
 };
 
+var _requestVidFrameCallback = null;
+
+
+/**
+ * Provides video.requestVideoFrameCallback in a cross browser way.
+ * Returns a property, or undefined if unsuported.
+ */
+var getRequestVidFrameCallback = function() {
+  if (_requestVidFrameCallback === null) {
+    var vid = document.createElement('video');
+    _requestVidFrameCallback =
+      getPrefixedProperty(vid, "requestVideoFrameCallback");
+  }
+
+  return _requestVidFrameCallback;
+};
+
 var _cancelAnimFrame;
 
 /**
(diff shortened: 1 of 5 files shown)


tguilbert-google:
I didn't update the test utils in 1.0.0, 1.0.1 or 1.0.2, because there wasn't already the "getPrefixedProperty" function available, and I didn't know which files should or shouldn't be updated.

I am not really a web developer, so please let me know if there is a more elegant way to do this :)

sdk/tests/js/webgl-test-utils.js
```diff
@@ -2913,6 +2913,23 @@ var requestAnimFrame = function(callback) {
...
+
+/**
+ * Provides video.requestVideoFrameCallback in a cross browser way.
+ * Returns a property, or undefined if unsuported.
+ */
+var getRequestVidFrameCallback = function() {
+  if (_requestVidFrameCallback === null) {
+    var vid = document.createElement('video');
+    _requestVidFrameCallback =
+      getPrefixedProperty(vid, "requestVideoFrameCallback");
```
kenrussell, line 2927: Is it necessary to try all of the browser-specific prefixes? Since the current web ecosystem development process discourages shipping prefixed APIs, it would be better to just look for the specified name.

In this case, can this be simpler?

`  return HTMLVideoElement.prototype["requestVideoFrameCallback"]
`
? Works in both Chrome Canary (has support) and Stable (doesn't have support).

tguilbert-google: It isn't necessary, and your proposed method is much simpler :) My experience with JS is quite limited, and I didn't know this was an option.

I've also added the support to the 1.0.1 and 1.0.2 versions, now that I'm not reliant on the "getPrefixedProperty".

I'm also happy to directly inline `HTMLVideoElement.prototype["requestVideoFrameCallback"]`, since it's only used only once per file, but maybe there will be other users in the future and the "get" function makes sense.

tguilbert-google:
Thanks! Yes, I ran tex-image-and-sub-image-2d-with-video.html on 1.0.1 and up.

PR #3067, opened 2020-05-06
Add demo for WEBGL_video_texture

shaoboyan:
This is a demo about using WEBGL_video_texture extension.
It uses 8k 60fps vp9 pure green video to reduce the video size.
It can fallback to the texImage2D path by adding disable_WEBGL_video_texture=1
in URL.


shaoboyan:
@kenrussell  PTAL if you have time.

shaoboyan:
Update the demo to remove some library dependencies.
@kenrussell  ping softly for review if you have time.

sdk/demos/intel/WEBGL_video_texture.html
```diff
@@ -0,0 +1,123 @@
...
+      if (enable_WEBGL_video_texture) {
+        textNode = document.createTextNode("WEBGL_video_texture: On");
+      } else {
+        textNode = document.createTextNode("WEBGL_video_texture: Off");
+      }
+     document.body.appendChild(textNode);
+
+      gl = webglCanvas.getContext("webgl");
+
+      if (!gl) {
```
kenrussell, line 87: Let's remove the experimental-webgl code - at this point, "webgl" is the well-established context name.

sdk/demos/intel/js/renderer.js
```diff
@@ -0,0 +1,236 @@
...
+          gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, self.videoElement);
+
+          gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR);
+          gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR);
+          gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
+          gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
+        } else {
+          if (this.ext !== null) {
+            gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE, self.texture);
+            ext.VideoElementTargetVideoTexture(ext.TEXTURE_VIDEO_IMAGE, video);
```
kenrussell, line 139: Should this be using shareVideoImageWEBGL?

sdk/demos/intel/js/renderer.js
```diff
@@ -0,0 +1,236 @@
...
+      gl.bindTexture(gl.TEXTURE_2D, this.texture);
+
+      if (this.videoElement && !this.videoElement.paused) {
+        gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, this.videoElement);
+      }
+    } else {
+      gl.bindTexture(ext.TEXTURE_VIDEO_IMAGE, this.texture);
+
+      if (this.videoElement && !this.videoElement.paused && this.ext !== null) {
+        ext.VideoElementTargetVideoTexture(ext.TEXTURE_VIDEO_IMAGE, this.videoElement);
```
kenrussell, line 225: Similarly, should this be using shareVideoImageWEBGL, with releaseVideoImageWEBGL after the call to drawElements?

sdk/demos/intel/js/renderer.js
shaoboyan: I'm sorry. In a long time ago, I think there is some disagree with changing the name to shareVideoImageWEBGL and without a proper name. So I remember the decision should be keep the VideoElementTargetVideoTexture. (I'll update the spec if this is true)

sdk/demos/intel/js/renderer.js
shaoboyan: The comment on releaseVideoImageWEBGL:
Quick try and find that I missed implementing this API in blink now.... Will do a quick update.

sdk/demos/intel/WEBGL_video_texture.html
shaoboyan: Done.

sdk/demos/intel/js/renderer.js
kenrussell: I vaguely remember there was some disagreement about the share/release semantics too, but I can't find it in the WebGL working group's meeting minutes folder. If you can search in there and you find it, please tell us which document you found it in.

At this point, https://www.khronos.org/registry/webgl/extensions/proposals/WEBGL_video_texture/ defines the share/release APIs. The share/release semantic is absolutely necessary for some implementations - like Chrome's - to minimize the duration the video frames are locked. So, could you please update Blink's IDL and implementation to follow the current extension proposal, and then update this demo to use the share/release APIs? Thanks.

PR #3071, opened 2020-05-12
Test both alpha:true and alpha:false in webgl_canvas tests.

kenrussell:
Remove explicit use of Promises in the inner loop, and change to use
async/await, to provide better control of the execution order of the
test cases.

Verified in Chrome, Firefox and Safari on macOS. ToT of all browsers
pass the new tests; Firefox has a couple of preexisting failures with
the RGBA10_A2/RGBA/UNSIGNED_INT_2_10_10_10_REV combination.

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
@@ -4,13 +4,15 @@ Use of this source code is governed by an MIT-style license that can be
 found in the LICENSE.txt file.
 */
 
+"use strict";
+
 function generateTest(internalFormat, pixelFormat, pixelType, prologue, resourcePath, defaultContextVersion) {
     var wtu = WebGLTestUtils;
     var tiu = TexImageUtils;
     var gl = null;
     var successfullyParsed = false;
-    var redColor = [255, 0, 0];
-    var greenColor = [0, 255, 0];
+    var redColor = [255, 0, 0, 255];
+    var greenColor = [0, 255, 0, 255];
     var repeatCount;
 
     function shouldRepeatTestForTextureFormat(internalFormat, pixelFormat, pixelType)
@@ -28,7 +30,7 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
                 (internalFormat == 'RGB' && pixelFormat == 'RGB' && pixelType == 'UNSIGNED_BYTE'));
     }
 
-    function init()
+    async function init()
     {
         description('Verify texImage2D and texSubImage2D code paths taking webgl canvas elements (' + internalFormat + '/' + pixelFormat + '/' + pixelType + ')');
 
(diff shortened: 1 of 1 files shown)


kenrussell:
@jdashg and perhaps @jdarpinian or @shrekshao could you please review?

This does increase the run time of this subset of the conformance2/ tests, but it's a fairly small subset, so shouldn't have a large impact.

Note that these new tests actually pass in all current WebGL 2.0 implementations. It could be argued that they don't catch any new bugs, but on the other hand, it's probably a good idea to increase testing coverage of alpha:false.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
```diff
@@ -67,7 +69,7 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
         runTest();
```
shrekshao, line 69: Do we need an `await` here?

kenrussell: Good point and good question. I just tested both with and without the await and the tests seem to behave the same. The Promise returned by init() is explicitly waited on in the anonymous function returned at the bottom of the file.

One behavioral difference with this change is that the intermediate results aren't displayed any more. I tried changing "await wtu.dispatchPromise(...)" to "await wtu.dispatchRAFPromise(...)", and doing a requestAnimationFrame inside it. Still no difference - the web page only displays the final results, not the intermediate ones.

I'm not an expert in Promises. Can you think of any improvements to make here?

kenrussell: Figured it out - needed to await wtu.dispatchPromise() after every test case. Thanks for your catch; please re-review.

kenrussell:
Update: these new alpha:false tests actually do catch a bug in ToT WebKit on iOS! Will appreciate your reviews to get these in. Thanks.

kenrussell:
@grorg FYI please see above comment! :)

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-webgl-canvas.js
```diff
@@ -191,93 +209,92 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
...
-        }
+        for (let alpha of [ true, false ]) {
+            let ctx = wtu.create3DContext(null, { alpha:alpha });
+            let canvas = ctx.canvas;
+            // Note: We use preserveDrawingBuffer:true to prevent canvas
+            // visibility from interfering with the tests.
+            let visibleCtx = wtu.create3DContext(null, { preserveDrawingBuffer:true, alpha:alpha });
+            if (!visibleCtx) {
+                testFailed("context does not exist");
+                finishTest();
```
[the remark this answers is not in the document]
kenrussell: Yes, thanks, that and another finishTest() can be removed.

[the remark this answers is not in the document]
kenrussell: Removed in the current version.

kenrussell:
New tests have caught bugs, so merging now.

PR #3072, opened 2020-05-13
Update WEBGL_video_texture

shaoboyan:
In current spec, API releaseVideoImageWEBGL returns boolean.
Since it can raise exception and in the spec example the return value
doesn't been checked, void should be a better return value.


shaoboyan:
@kenrussell  PTAL if you have time.

shaoboyan:
@kenrussell good questions. I'll do some explore to give the answer.

Issue #3076, opened 2020-05-18
Image loading tests have become flaky after #3052


kenrussell:
#3052 seems to have introduced significant flakiness into the tests which load cross-origin images. On Chromium's automated testing system, the following bugs have been found:

WebGL origin-clean-conformance-* test fail because image loading failed
http://crbug.com/1082525

conformance/textures/misc/origin-clean-conformance.html is flaky
http://crbug.com/1083100

@jdashg @shrekshao @jdarpinian how can we fix this flakiness? In hindsight, 1000 ms is too short a timeout for some slow bots.

kenrussell:
One more bug:

conformance/more/functions/readPixelsBadArgs.html is flaky
http://crbug.com/1083320

jdashg:
Make the delay bigger I guess.

PR #3077, opened 2020-05-20
Increase timeout for cross-origin image loading tests.

kenrussell:
Change WebGLTestUtils.awaitOrTimeout to offer a default timeout of
5000 ms, and all tests currently using it to use the default.

Tested both manually and in Chromium's automated WebGL test harness.
Unclear whether this will fix the problems seen on Chromium's bots,
but need to try.

Follow-on to #3052. Addresses #3076.


kenrussell:
@jdashg @shrekshao please review - thanks.

PR #3080, opened 2020-05-22
Fix nonexistent drawQuad reference in top-of-tree PVRTC conformance t…

kenrussell:
…est.

Follow-on to #3079.


kenrussell:
@grorg FYI. Merging myself.

PR #3089, opened 2020-05-31
Add tests to verify toDataURL images are the correct orientation for premultiplyAlpha=false

nkronlage:
This adds a test to verify the image returned by toDataURL with a premultipliedAlpha=false canvas is the correct orientation.

This was suggested by:
https://bugs.webkit.org/show_bug.cgi?id=156129

I ran the test on Chrome/FireFox and got:

    testing 16 by 16 attributes: {"premultipliedAlpha":false}
    PASS should be 0,255,0,255
    PASS should be 255,0,0,255
    PASS should be 0,0,255,255

And got this on Safari:

    testing 16 by 16 attributes: {"premultipliedAlpha":false}
    FAIL should be 0,255,0,255
    at (0, 0) expected: 0,255,0,255 was 255,0,0,255
    FAIL should be 255,0,0,255
    at (0, 8) expected: 255,0,0,255 was 0,255,0,255
    PASS should be 0,0,255,255

### DIFF
--- a/sdk/tests/conformance/canvas/to-data-url-test.html
+++ b/sdk/tests/conformance/canvas/to-data-url-test.html
@@ -42,7 +42,7 @@ var main = function() {
 
     canvas.width = width;
     canvas.height = height;
-    
+
     if (!gl) {
       gl = wtu.create3DContext(canvas, attributes);
     }
--- a/sdk/tests/conformance/canvas/to-data-url-test.html
+++ b/sdk/tests/conformance/canvas/to-data-url-test.html
@@ -21,7 +21,6 @@ found in the LICENSE.txt file.
 var wtu = WebGLTestUtils;
 var numTests = 10;
 var gl;
-var glAttributes;
 var ctx;
 
 var main = function() {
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/canvas/to-data-url-test.html
```diff
@@ -71,6 +97,9 @@
     });
   };
 
+  var premultipliedAlphaAttributes =  {
```
jdashg, line 100: Prefer `const`/`let` in all new code.

sdk/tests/conformance/canvas/to-data-url-test.html
```diff
@@ -21,29 +21,55 @@
...
     gl.scissor(x, y, width, height);
     gl.clear(gl.COLOR_BUFFER_BIT);
   };
 
-  var testSize = function(gl, width, height, callback) {
-    debug("testing " + width + " by " + height);
-    gl.canvas.width = width;
-    gl.canvas.height = height;
+  var testSize = function(width, height, attributes, callback) {
+    var attributesDebugMessage = (attributes ? " attributes: " + JSON.stringify(attributes) : "");
```
jdashg, line 65: Parens around compound components of a trinary:
`(attributes ? (" attributes: " + JSON.stringify(attributes)) : "")`
Or more ideally avoid trinaries.

sdk/tests/conformance/canvas/to-data-url-test.html
```diff
@@ -21,29 +21,55 @@
 var wtu = WebGLTestUtils;
 var numTests = 10;
 var gl;
+var glAttributes;
```
jdashg, line 24: Embed it in `gl.attributes`.

jdashg:
Thanks for this addition! I'm fine to merge it without the changes, but it'd be nice to do these fixups. Let me know which you'd prefer!

nkronlage:
Updated the PR with changes requested. Take another look and let me know if there's anything else. Thanks!

jdashg:
Thanks!

PR #3092, opened 2020-06-03
Temporarily disable requestVideoFrameCallback for some tests.

kenrussell:
Disable the use of this new API for the image_bitmap_from_video
conformance tests to work around crbug.com/1090960. It will be
re-enabled once the underlying bug is fixed.

Follow-on to #3065.


kenrussell:
@shrekshao please review - this should unblock the WebGL conformance roll in https://chromium-review.googlesource.com/2227304 .

PR #3097, opened 2020-06-05
Race web+localhost resource loads for cors tests.

jdashg:
This should address #3095.

### DIFF
--- a/sdk/tests/conformance/textures/misc/origin-clean-conformance.html
+++ b/sdk/tests/conformance/textures/misc/origin-clean-conformance.html
@@ -18,7 +18,6 @@ found in the LICENSE.txt file.
 <div id="console"></div>
 <canvas id="canvas1"></canvas>
 <canvas id="canvas2"></canvas>
-<img id="img" style="display:none;">
 <script>
 "use strict";
 var wtu = WebGLTestUtils;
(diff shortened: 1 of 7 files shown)


sdk/tests/conformance/textures/misc/origin-clean-conformance.html
```diff
@@ -122,9 +121,10 @@
 }
 
 (async function() {
-  const img = document.getElementById('img');
+  //const img = document.getElementById('img');
```
kenrussell, line 124: Could you delete this commented-out code?

kenrussell:
BTW this small addition made it easier to debug:

<pre>
diff --git a/sdk/tests/js/webgl-test-utils.js b/sdk/tests/js/webgl-test-utils.js
index d231ff53f..c5b1bfaa2 100644
--- a/sdk/tests/js/webgl-test-utils.js
+++ b/sdk/tests/js/webgl-test-utils.js
@@ -3256,6 +3256,7 @@ async function makeCrossOriginImage(webUrl, localUrl) {
       makeImage('web', webUrl),
       makeImage('local', localUrl),
     ]);
+    console.log('[makeCrossOriginImage] returning image with src: ' + ret.src);
     return ret;
   } catch {}
   throw 'createCrossOriginImage failed';
</pre>

kenrussell:
@jdashg : @gyagp reported on http://crbug.com/1082525 that this works, but introduces warnings in Chromium's test harness about not being able to cleanly shut down the proxy server which it sets up. This is probably because of the attempted network connections which are still timing out at test shutdown

If there are specific reasons to prefer this approach over that in #3095 , could you please describe them? It would be better in my opinion to avoid adding more non-determinism to the test suite.

kenrussell:
Per discussion on the WebGL Dev+Implementer Matrix chat room hosted by Mozilla, we've agreed to move forward with #3095 and close this PR for the time being. If #3095 destabilizes Mozilla's CI system then we'll revisit this as an alternative.

PR #3099, opened 2020-06-18
readBuffer of out of range color attachment should cause INVALID_OPERATION

jdarpinian:
Looks like the test was written to the man page, but the man page disagrees with the spec and with dEQP. The spec should win.


jdashg:
If I could do WebGL again, I would only have GL_OOM, and some generic INVALID_CALL.

kenrussell:
This was a mistake while writing the WebGL conformance tests; they disagree with the OpenGL ES conformance tests.

Thanks for your review.

jdashg:
I know, I just don't like changing our expectations in this case. Enforcing conformance here doesn't really help anyone.

PR #3100, opened 2020-06-19
Test JPEG with linear gamma in image_bitmap_from_blob tests.

kenrussell:
Restructure the tests to use the Fetch API and Promises to make this
addition easier to read.

Make tolerance configurable, and use a large value for compatibility
with Firefox Nightly on macOS.

The new JPEG resource was exported from the included Photoshop
document which was constructed from scratch for this test case, based
on the original report from Figma.

Regression test for http://crbug.com/1058160 .


kenrussell:
All CC'd - please review. Thanks.

This change has been tested locally on macOS with Safari Technology Preview, Firefox, and Chromium with the needed bug fix.

- Firefox Nightly passes all cases except the RGB10_A2 tests.
- Safari Technology Preview was failing several of the original test cases with red-green-semi-transparent.png, and also fails about the same number of the new test cases.
- Chromium with the needed bug fix passes all cases.

MarkCallow:
I don't think a JPEG file with linear gamma is valid. I hope browser vendors aren't trying to encourage this.

kenrussell:
> I don't think a JPEG file with linear gamma is valid. I hope browser vendors aren't trying to encourage this.

Can you provide more details on why this is invalid? A JPEG with an embedded color profile can have the gamma adjusted to 1.0. Figma receives files like this on an ongoing basis.

Not trying to encourage invalid behavior, but a regression test for this is absolutely needed.

MarkCallow:
Which file format are we talking about? I was thinking JFIF in which the color space is required to be BT.601 which has a non-linear transfer function. If you are talking about EXIF, although the default color space is sRGB, complete color space information can be provided including a transfer function which is specified as a table.

I got permission denied when I tried to look at the bug this PR is a regression test for.

lexaknyazev:
@MarkCallow 
ICC profile may be embedded in a JFIF file using `APP2` marker as per ICC.1:2010-12, Annex B.4.

kenrussell:
The Chromium bug was automatically marked restricted view because it's a crash report. It's not a security bug though. You can provoke the same crash by patching this PR into your checkout of the WebGL conformance suite and running any of the conformance[2]/textures/image_bitmap_from_blob/ tests in Chromium.

I don't know the details of the JPEG file which I exported from Photoshop - please feel free to examine the one in this PR. To create it, I defined a custom color profile, set the gamma to 1.0, and exported the file as a JPEG with an embedded color profile. The original PSD file is included.

MarkCallow:
I found how the profile is being embedded. Profiles are indeed not part of JFIF. However the [ICC spec.](http://color.org/specification/ICC1v43_2010-12.pdf) in Annex 4, Section B.4 describes how to use _JPEG Interchange Format_ (note the missing "F")  APP2 markers to store an ICC profile. .jpg files with profiles are valid _JPEG Interchange Format_ files and valid in some subset of the family of _JPEG Interchange Format_ files, not including the JFIF format.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-blob.js
```diff
@@ -31,17 +31,36 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
...
         gl.clearDepth(1);
 
-        var xhr = new XMLHttpRequest();
-        xhr.open("GET", resourcePath + "red-green-semi-transparent.png");
-        xhr.responseType = 'blob';
-        xhr.onload = function() {
-            var blob = xhr.response;
-            runImageBitmapTest(blob, 0.5, internalFormat, pixelFormat, pixelType, gl, tiu, wtu, false)
+        fetch(resourcePath + "red-green-semi-transparent.png")
+            .then(response => {
```
jdashg, line 35: `then` is sooo 2018. :)
We should be writing this with `await`:
```
let response, blob;
debug('*** Running tests against red-green-semi-transparent.png ***');
response = await fetch(resourcePath + "red-green-semi-transparent.png");
blob = await response.blob();
await runImageBitmapTest(blob, 0.5, internalFormat, pixelFormat, pixelType, gl, tiu, wtu, false);

debug('*** Running tests against red-green-128x128-linear-profile.jpg ***');
response = await fetch(resourcePath + "red-green-128x128-linear-profile.jpg");
blob = await response.blob();
// This test requires a huge tolerance because browsers - at least currently - vary
// widely in the colorspace conversion results for this image.
const tolerance = 120;
await runImageBitmapTest(blob, 1.0, internalFormat, pixelFormat, pixelType, gl, tiu, wtu, false, tolerance);
```

sdk/tests/js/tests/tex-image-and-sub-image-with-image-bitmap-utils.js
```diff
@@ -363,29 +360,31 @@ function runImageBitmapTestInternal(bitmaps, alphaVal, internalFormat, pixelForm
...
+                internalFormat, pixelFormat, pixelType, gl, tiu, wtu, tolerance);
         }
     }
 }
 
-function runImageBitmapTest(source, alphaVal, internalFormat, pixelFormat, pixelType, gl, tiu, wtu, is3D)
+function runImageBitmapTest(source, alphaVal, internalFormat, pixelFormat, pixelType, gl, tiu, wtu, is3D, opt_tolerance)
 {
+    if (opt_tolerance === undefined)
+        opt_tolerance = 10;
```
jdashg, line 387:
```
    if (opt_tolerance === undefined) {
        opt_tolerance = 10;
    }
```

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image-bitmap-from-blob.js
kenrussell: Thanks for the suggestions. :) I should have realized that possible transformation. Done throughout, and tested and removed the check of response.ok. (The tests will fail later if the image couldn't be fetched.)

sdk/tests/js/tests/tex-image-and-sub-image-with-image-bitmap-utils.js
kenrussell: Done.

kenrussell:
Thanks for your reviews. Merging now.

PR #3110, opened 2020-06-30
EXT_clip_cull_distance: allow for an implementation to return 0 for the number of supported cull planes

krogovin:
This will allow browsers on mobile to support HW clip planes for those GPU's that expose GL_APPLE_clip_distance only (for example many Imagination Technologies GPU's). In addition, on desktop, it enlarges the ability to implement the extension on all GL versions that are at 4.4 or before. This, for example, includes all Apple desktop hardware.

### DIFF
--- a/extensions/EXT_clip_cull_distance/extension.xml
+++ b/extensions/EXT_clip_cull_distance/extension.xml
@@ -7,6 +7,7 @@
 
   <contributors>
     <contributor>Members of the WebGL working group</contributor>
+    <contributor>Kevin Rogovin (kevinrogovin[at]invisionapp.com</contributor>
   </contributors>
 
   <number>43</number>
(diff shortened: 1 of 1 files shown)


jdashg:
Well, is this extension useful if 0 is returned? If not, we'd rather just not expose the extension on those machines.

krogovin:
The extension has two values:
- the number of clip planes
- the number of cull planes

The PR allows to return 0 for the number of cull planes, but leaves in place that the number of clip planes is atleast 8.

The main use case is to allow for HW clip-plan support for pre GL 4.5 driver/hardware and for those GLES driver/hardware that expose GL_APPLE_clip_distance. For both of these, there is HW-clip planes, but no cull planes.

krogovin:
Added to the commit an addition to the history. My apologies for missing that the first time around.

kenrussell:
No worries. There aren't any automated checks for this. Thanks for the update! Merging now.

PR #3112, opened 2020-07-07
Update bptc and rgtc tests with level and dimension test

shrekshao:
Updating test according to recent change: https://github.com/KhronosGroup/WebGL/pull/3111

addressing chrome issue: http://crbug.com/1100834

### DIFF
--- a/sdk/tests/js/tests/compressed-texture-utils.js
+++ b/sdk/tests/js/tests/compressed-texture-utils.js
@@ -192,6 +192,28 @@ var testTexSubImageDimensions = function(gl, validFormats, expectedByteLength, g
     gl.deleteTexture(tex);
 };
 
+var testTexImageLevelDimensions = function(gl, validFormats, expectedByteLength, getBlockDimensions, imageConfigs) {
+    var tex = gl.createTexture();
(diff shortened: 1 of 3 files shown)


sdk/tests/js/tests/compressed-texture-utils.js
```diff
@@ -192,6 +192,28 @@ var testTexSubImageDimensions = function(gl, validFormats, expectedByteLength, g
     gl.deleteTexture(tex);
 };
 
+var testTexImageLevelDimensions = function(gl, validFormats, expectedByteLength, getBlockDimensions, imageConfigs) {
```
kenrussell, line 195: Might be a good time to start changing "var" to "let", at least inside this function.

shrekshao:
> has this been tested with revised validation code in ANGLE and/or Chromium?
The level validation already passed on Win (using D3D angle).
Not yet on Mac (current chrome fails test). Updated validation is still compiling (once run out of disk storage). I will post updates

shrekshao:
FYI. Local test on Mac Chromium shows updated validation code passed the new test.

kenrussell:
Thanks for confirming. Merging now.

PR #3113, opened 2020-07-10
Promote WEBGL_multi_draw to community approved

shrekshao:
I'm working on adding WEBGL_multi_draw to emscripten at the same time.
What do you think about promoting?


kenrussell:
I'd love to see this promoted to community approved. While there's currently only one implementation, in Chromium, @austinEng 's early prototypes in Safari showed a big performance benefit there, too. Would love to get this in the hands of developers sooner rather than later.

shrekshao:
Trivial change to update change log to have promote to community approved to be the last one. Merging my self.

PR #3118, opened 2020-07-17
Add WebGL 2.0 test of pixelStorei with PACK_SKIP_PIXELS > 0.

kenrussell:
All of the existing tests for this pixel pack parameter with a value
greater than 0 were assumed to fail validation. Adding a positive test
reveals an apparent OpenGL driver bug on macOS.

Test case for http://crbug.com/angleproject/4849 .

This new test passes in Chromium and Firefox on macOS, and in WebKit
on top of ANGLE with the necessary driver bug workaround.


kenrussell:
Reviewers (ideally two) please review. Thanks.

kenrussell:
Thanks for your reviews. Merging now.

PR #3119, opened 2020-07-20
Fix WEBGL_compressed_texture_etc1 section of compressed-tex-image test

sugoi1:
According to the spec:
https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_etc1/

This WebGL extension does not allow using compressedTexSubImage2D:
"Remove ability to use the format with compressedTexSubImage2D,
 as per the GLES extension spec."

The compressed-tex-image test was modified to reflect the spec.

### DIFF
--- a/sdk/tests/js/tests/compressed-tex-image.js
+++ b/sdk/tests/js/tests/compressed-tex-image.js
@@ -59,7 +59,7 @@ if (!gl) {
 
   gl.bindTexture(gl.TEXTURE_2D, tex);
 
-  function validateExt(extName, enumName, blockSize, blockByteSize) {
+  function validateExt(extName, enumName, blockSize, blockByteSize, expectedSubImageError) {
     debug('\n---------------------------');
     debug('\n' + extName);
     ext = gl.getExtension(extName);
@@ -97,7 +97,7 @@ if (!gl) {
       wtu.shouldGenerateGLError(gl, gl.NO_ERROR,
           `gl.compressedTexImage2D(gl.TEXTURE_2D, 0, ext.${enumName}, ${blockSize},${blockSize}, 0, g_view)`);
 
-      wtu.shouldGenerateGLError(gl, gl.NO_ERROR,
+      wtu.shouldGenerateGLError(gl, expectedSubImageError,
           `gl.compressedTexSubImage2D(gl.TEXTURE_2D, 0, 0,0, ${blockSize},${blockSize}, ext.${enumName}, g_view)`);
     }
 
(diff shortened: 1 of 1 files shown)


sdk/tests/js/tests/compressed-tex-image.js
```diff
@@ -97,8 +97,10 @@ if (!gl) {
       wtu.shouldGenerateGLError(gl, gl.NO_ERROR,
           `gl.compressedTexImage2D(gl.TEXTURE_2D, 0, ext.${enumName}, ${blockSize},${blockSize}, 0, g_view)`);
 
-      wtu.shouldGenerateGLError(gl, gl.NO_ERROR,
-          `gl.compressedTexSubImage2D(gl.TEXTURE_2D, 0, 0,0, ${blockSize},${blockSize}, ext.${enumName}, g_view)`);
+      if (supportsSubImage) {
+        wtu.shouldGenerateGLError(gl, gl.NO_ERROR,
+            `gl.compressedTexSubImage2D(gl.TEXTURE_2D, 0, 0,0, ${blockSize},${blockSize}, ext.${enumName}, g_view)`);
+      }
```
kenrussell, line 103: Please add in the negative test - that this should generate INVALID_OPERATION per https://www.khronos.org/registry/OpenGL/extensions/OES/OES_compressed_ETC1_RGB8_texture.txt .

sdk/tests/js/tests/compressed-tex-image.js
```diff
@@ -116,23 +118,25 @@ if (!gl) {
...
+      if (supportsSubImage) {
+        wtu.shouldGenerateGLError(gl, gl.NO_ERROR,
+            `gl.compressedTexSubImage2D(gl.TEXTURE_2D, 0, 0,0, ${blockSize},${blockSize}, ext.${enumName}, ${blockByteSize}, 0)`);
+        wtu.shouldGenerateGLError(gl, gl.NO_ERROR,
+            `gl.compressedTexSubImage2D(gl.TEXTURE_2D, 0, 0,0, ${blockSize},${blockSize}, ext.${enumName}, ${blockByteSize}, 1)`);
+        wtu.shouldGenerateGLError(gl, gl.NO_ERROR,
+            `gl.compressedTexSubImage2D(gl.TEXTURE_2D, 0, 0,0, ${blockSize},${blockSize}, ext.${enumName}, ${blockByteSize}, ${blockByteSize})`);
+        wtu.shouldGenerateGLError(gl, gl.INVALID_OPERATION,
+            `gl.compressedTexSubImage2D(gl.TEXTURE_2D, 0, 0,0, ${blockSize},${blockSize}, ext.${enumName}, ${blockByteSize}, ${blockByteSize+1})`);
+      }
```
kenrussell, line 130: Please add in the negative tests here too.

kenrussell:
@jdashg could you please review this too?

Any other comments welcome. CC @jdarpinian @shrekshao @grorg @RafaelCintron @kainino0x

sdk/tests/js/tests/compressed-tex-image.js
kenrussell: Also - do you think you could update the associated tests in the conformance-suites/1.0.3 and conformance-suites/2.0.0 snapshots? Those need to stay valid before and after this validation change - so essentially this current change is the one that should be applied to those older snapshots.

Thanks in advance for your help!

sdk/tests/js/tests/compressed-tex-image.js
sugoi1: While WEBGL_compressed_texture_etc1 does not support compressedTexSubImage2D, WEBGL_compressed_texture_etc does support compressedTexSubImage2D, so negative tests would be conditional on having WEBGL_compressed_texture_etc1 and NOT WEBGL_compressed_texture_etc.

I'll add it, but it will add a bit of complexity to the test.

sugoi1: Oh, never mind, WEBGL_compressed_texture_etc is exclusively for ETC2, so the 2 are unrelated

sdk/tests/js/tests/compressed-tex-image.js
sugoi1: Unless I'm reading this wrong, the same compressed-tex-image.js file is used for both test suites.

Older versions of the test only used compressedTexImage2D, so the compressedTexSubImage2D issue wasn't exposed.

MarkCallow:
> Oh, never mind, WEBGL_compressed_texture_etc is exclusively for ETC2, so the 2 are unrelated

At the hardware level ETC2 implementations can use ETC1 textures. I sincerely hope that WEBGL_compressed_texture_etc does not forbid implementations from accepting ETC1 textures.

sugoi1:
Ok, should be clean now

kenrussell:
Thanks @jdashg for your review!

@sugoi1 still hoping for your feedback on backporting this change to the snapshots.

sdk/tests/js/tests/compressed-tex-image.js
sugoi1: > Thanks @jdashg for your review!
> 
> @sugoi1 still hoping for your feedback on backporting this change to the snapshots.

Hi Ken, as I mentioned, I don't see anything that needs to be backported to the snapshots. Only compressed-tex-image.js contains code that attempts to call compressedTexSubImage2D in this manner. If I'm wrong, point me to the file(s) and I'll do it.

kenrussell: Ah, sorry - I missed that comment above. Thanks for checking and confirming the old test suites don't need to be updated.

kenrussell:
Per above, @sugoi1 confirmed the snapshots don't include compressedTexSubImage tests so don't need to be updated.

Thanks for fixing this. Merging now.

PR #3122, opened 2020-07-24
Add null-object-behaviour-2 test for WebGL 2 objects.

jdarpinian:


sdk/tests/conformance2/misc/00_test_list.txt
```diff
@@ -1,6 +1,7 @@
 expando-loss-2.html
 getextension-while-pbo-bound-stability.html
 instanceof-test.html
+null-object-behaviour-2.html
```
jdashg, line 4: I think technically this should be `>= 2.0.1`.
Given git, I'm not sure we care though?

jdashg: @kenrussell

kenrussell: Just for bookkeeping purposes - since we've done this so far - it would be helpful to add "--min-version 2.0.1".

PR #3123, opened 2020-07-24
Use more inclusive terms in lint script.

kenrussell:
Addresses http://crbug.com/1097235 .


kenrussell:
Tested locally - lint scripts still run.

Please review, perhaps @jdashg or @jdarpinian or @shrekshao . Thanks.

kenrussell:
Thanks @jdashg for your review. Merging now.

PR #3124, opened 2020-07-25
Shader source character set and preprocess validation test

shrekshao:
Addressing issue reported at: https://bugs.chromium.org/p/chromium/issues/detail?id=1108588

The main issue is that WebGL should not validate shader string before preprocessing for cases like:
```
#if defined(GL_GOOGLE_cpp_style_line_directive)
#line 45 "foo.txt"
#endif
```
where " is not a valid character but should be cleared by preprocessing step thus not failing the validation and compilation.

(Check https://www.khronos.org/registry/OpenGL/specs/es/3.0/GLSL_ES_Specification_3.00.pdf Section 3.2)

Tested on Win 10 and fails in Chrome and Firefox. Passed for modified local Chromium build.

### DIFF
--- a/sdk/tests/conformance/glsl/bugs/00_test_list.txt
+++ b/sdk/tests/conformance/glsl/bugs/00_test_list.txt
@@ -5,6 +5,7 @@
 --min-version 1.0.3 array-of-struct-with-int-first-position.html
 --min-version 1.0.4 assign-to-swizzled-twice-in-function.html
 --min-version 1.0.4 bool-type-cast-bug-int-float.html
+--min-version 1.0.4 character-set.html
 --min-version 1.0.3 compare-loop-index-to-uniform.html
 --min-version 1.0.3 complex-glsl-does-not-crash.html
 --min-version 1.0.4 compound-assignment-type-combination.html
--- a/sdk/tests/conformance/glsl/bugs/character-set.html
+++ b/sdk/tests/conformance/glsl/bugs/character-set.html
@@ -32,6 +32,16 @@ void main() {
 }
 </script>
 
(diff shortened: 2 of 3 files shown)


sdk/tests/conformance/glsl/bugs/character-set.html
```diff
@@ -0,0 +1,118 @@
...
+// some comment: \n\\r\n
+void main() {
+  gl_FragColor = vec4(1, 0, 0, 1);
+}
+</script>
+
+<script id="fs-cleared-by-preprocessing-valid" type="x-shader/x-fragment">
+precision mediump float;
+#if defined(GL_GOOGLE_cpp_style_line_directive)
+#line 9 "foo.txt"
```
jdashg, line 49: Add a test of an uncommented 一些注释 within ifdefs that fails. (Ideally we'd check that this fails in shaderSource)

sdk/tests/conformance/glsl/bugs/00_test_list.txt
```diff
@@ -5,6 +5,7 @@
 --min-version 1.0.3 array-of-struct-with-int-first-position.html
 --min-version 1.0.4 assign-to-swizzled-twice-in-function.html
 --min-version 1.0.4 bool-type-cast-bug-int-float.html
+--min-version 1.0.4 character-set.html
```
jdashg, line 8: I'm not really sure this fits into glsl/bugs, but it works.

jdashg:
Thanks!

shrekshao:
(Still trying to add a pure shaderSource test though...
Let me open a separate pull request then)

kenrussell:
Great work Shrek. LGTM after the fact.

PR #3125, opened 2020-07-27
Fix object-deletion-behaviour-2 to create WebGL 2.0 context.

kenrussell:
The 2.0.0 snapshot of this test was incorrectly creating a WebGL 1.0
context.


kenrussell:
FYI @jdarpinian @shrekshao @jdashg @grorg @RafaelCintron . Discovered this bug while revising WebKit's snapshot of the conformance suite. It was probably introduced by other refactorings.

kenrussell:
Merging myself.

PR #3126, opened 2020-07-27
Loosen readBuffer error condition in readbuffer.html.

kenrussell:
Since the required error was changed from INVALID_ENUM to
INVALID_OPERATION in #3099, the 2.0.0 snapshot must allow both values
in order to admit both old and new implementations.


kenrussell:
FYI @jdarpinian @shrekshao @jdashg @grorg @RafaelCintron .

kenrussell:
Merging myself.

PR #3127, opened 2020-07-27
Fix integer-cubemap-specification-order-bug.html in 2.0.0 snapshot.

kenrussell:
Incorporate fix from #2327, which never made it into the 2.0.0
snapshot. This test was substantially rewritten on top-of-tree in #2806.


kenrussell:
@jdarpinian @kainino0x @shrekshao @jdashg @grorg @RafaelCintron FYI.

kenrussell:
Merging myself.

PR #3128, opened 2020-07-27
Expect `gl.INVALID_VALUE` for `gl.shaderSource` for invalid shader in character-set.html

shrekshao:
Following up https://github.com/KhronosGroup/WebGL/pull/3124

We now expect `gl.INVALID_VALUE` for `gl.shaderSource` instead of use GLSLConformanceTester for the overall test.


shrekshao:
As this is addressing comments from #3124 
I'm going to merge this for further webgl roll and chrome patch impl

kenrussell:
LGTM after the fact.

PR #2386, opened 2017-04-27
Prepare tests for landing texImageSource & GL_RGB10_A2 support.

astojilj:
Splitting part of the https://github.com/KhronosGroup/WebGL/pull/2375 to
allow landing https://crrev.com/2821363003/ before landing the
https://github.com/KhronosGroup/WebGL/pull/2375.

This is related to #2274: GL_RGB10_A2 and texImageSource variants of tex*Image*

PR #3129, opened 2020-07-27
Remove invalid RGB10_A2 tests from 2.0.0 conformance snapshot.

kenrussell:
These tests were invalidated in #2386. In order to allow both old and
new implementations to pass these tests, and to minimize complexity in
the 2.0.0 conformance snapshot, simply remove the now-invalid tests.


kenrussell:
@jdarpinian @shrekshao @jdashg @RafaelCintron @grorg FYI.

kenrussell:
Merging myself.

PR #3131, opened 2020-07-28
Fix function signatures in multi-draw extensions.

kenrussell:
Add "new functions" section to WEBGL_multi_draw.

Add needed parentheses to union types in the Web IDL in
WEBGL_multi_draw_instanced_base_vertex_base_instance.


kenrussell:
@shrekshao please review.

@enga @jdashg (mainly) FYI.

PR #3132, opened 2020-07-28
Increase tolerance to 3 in clear-srgb-color-buffer.html .

kenrussell:
Found necessary in https://bugs.webkit.org/show_bug.cgi?id=214893 .


kenrussell:
FYI @jdarpinian @jdashg @shrekshao @grorg @RafaelCintron . Merging myself.

PR #3134, opened 2020-08-04
Skip some ImageBitmap from ImageData tests w/o ImageBitmapOptions.

kenrussell:
When ImageBitmapOptions is not supported by the browser, the implicit
behavior of the premultiplyAlpha option is the default. Per HTML
specification:

https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#cropped-to-the-source-rectangle-with-formatting

this is implementation-specific (either premultiplied or not), so when
the input ImageData has alpha != 1.0, there isn't a single correct
answer. Skip these tests.

This was discovered while implementing ImageBitmap from ImageData
support in WebKit in https://bugs.webkit.org/show_bug.cgi?id=183438 ,
though WebKit supports ImageBitmapOptions. Firefox currently doesn't
support ImageBitmapOptions and happens to have been passing these
tests, but WebKit wouldn't have.


kenrussell:
@jdashg @grorg in particular please review.

PR #3135, opened 2020-08-06
Test WebGL 2's new vertex attribute types

jdarpinian:
INT, UNSIGNED_INT, HALF_FLOAT, INT_2_10_10_10_REV, and UNSIGNED_INT_2_10_10_10_REV.

Also test that FIXED is not supported in WebGL 2, as the restriction from WebGL 1 was not lifted. Firefox seems to currently support FIXED in WebGL 2.

### DIFF
--- a/sdk/tests/conformance/attribs/gl-vertexattribpointer-offsets.html
+++ b/sdk/tests/conformance/attribs/gl-vertexattribpointer-offsets.html
@@ -52,12 +52,7 @@ There is supposed to be an example drawing here, but it's not important.
                 componentSize: 4,
                 normalize: false,
               },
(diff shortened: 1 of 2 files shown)


jdarpinian:
I have no opinion on whether it should be supported. I don't know why it was excluded to begin with.

kenrussell:
Thanks for revising these tests @jdarpinian .

Regarding WebGL's removal of FIXED type vertex attributes - I recall that they were only supported on OpenGL ES when WebGL 1.0 was under active development, so to eliminate costly emulation on desktop they were removed from the spec. Since there's been essentially zero demand for them since, I agree with asserting that they're disallowed in WebGL 2.0.

Issue #3114, opened 2020-07-10
Make EXT_texture_norm16 exposed to WebGL 2 only and promote to community approved.


shrekshao:
@jdashg We agreed on https://gitlab.khronos.org/webgl/member-tools/-/wikis/26-Mar-2020#issues-and-pull-requests-issues-and-pull-requests that we would make EXT_texture_norm16 to only exposed to WebGL 2 and make it public then. I found https://github.com/emscripten-core/emscripten/pull/10227 and seems emscripten and unity would expect this change back in Jan -_-!!
I think we've done the implementation and cleared known issues in Chromium (http://crbug.com/1000354). Would you mind modifying the spec and promote this extension to community approved? Thanks!

kenrussell:
Let me take this since it seems Jeff is busy.

kenrussell:
Good work Shrek getting this over the finish line!

PR #3116, opened 2020-07-14
Specialize EXT_texture_norm16 to WebGL 2.0; promote to community appr…

kenrussell:
…oved.

This extension refers to sized internal formats that are unavailable
in the WebGL 1.0 specification. Per agreement in the WebGL WG,
restrict it to being exposed on WebGL 2.0 contexts.

Promote to community approved.

Addresses #3114.

PR #3136, opened 2020-08-10
Specialize EXT_texture_norm16 to WebGL 2.0.

kenrussell:
Revise conformance2/extensions/ext-texture-norm16.html to mandate that
16-bit signed normalized textures are not color renderable. That
functionality is covered under EXT_render_snorm, not currently exposed
to WebGL.

Partially addresses #3114.

### DIFF
--- a/sdk/tests/conformance2/extensions/ext-texture-norm16.html
+++ b/sdk/tests/conformance2/extensions/ext-texture-norm16.html
@@ -23,7 +23,7 @@ description("This test verifies the functionality of the EXT_texture_norm16 exte
 debug("");
 
 var wtu = WebGLTestUtils;
-var gl = wtu.create3DContext();
+var gl = wtu.create3DContext(null, null, 2);
 var ext;
 
 var formats = null;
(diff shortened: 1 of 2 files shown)


kenrussell:
Note that earlier change #3116 also promoted this to community approved but let's hold off on that until we have agreement on all of the details of this extension spec.

kenrussell:
Revised to remove the EXT_render_snorm dependence, and added negative tests forbidding the SNORM16 texture formats from being advertised as color-renderable.

sdk/tests/conformance2/extensions/ext-texture-norm16.html
```diff
@@ -174,6 +189,11 @@
   testNorm16Render(ext.R16_EXT, gl.RED, gl.UNSIGNED_SHORT, 0);
   testNorm16Render(ext.RG16_EXT, gl.RG, gl.UNSIGNED_SHORT, 0);
   testNorm16Render(ext.RGBA16_EXT, gl.RGBA, gl.UNSIGNED_SHORT, 0);
+
+  testSnorm16Unrenderable(ext.R16_SNORM_EXT, gl.RED, gl.SHORT);
+  testSnorm16Unrenderable(ext.RG16_SNORM_EXT, gl.RG, gl.SHORT);
+  testSnorm16Unrenderable(ext.RGB16_SNORM_EXT, gl.RGB, gl.SHORT);
```
shrekshao, line 195: https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_render_snorm.txt doesn't include RGB16_SNORM
But since it's negtive test, I think it be okay as long as this wouldn't lead to crash.

jdashg: If it crashes, that would make this an essential negative test! :)

sdk/tests/conformance2/extensions/ext-texture-norm16.html
```diff
@@ -23,7 +23,7 @@
 debug("");
 
 var wtu = WebGLTestUtils;
-var gl = wtu.create3DContext();
+var gl = wtu.create3DContext(null, null, 2);
```
jdashg, line 26: Don't need this, since this is in conformance2/*.

kenrussell: It's needed to support navigating directly to this test rather than running it inside the harness, which makes debugging simpler. Most of the other tests under conformance2/ do this; it was an oversight that this test didn't.

sdk/tests/conformance2/extensions/ext-texture-norm16.html
kenrussell: Thanks for pointing that out. It's probably an oversight in EXT_render_snorm; there is a reference to RGB16_SNORM_EXT in the section "Interactions with EXT_texture_norm16".

kenrussell:
Thanks @shrekshao @jdashg for your reviews. Merging now.

PR #3137, opened 2020-08-11
Support WebKit's MiniBrowser in webglHarnessCollectGarbage().

kenrussell:
This allows the expando-loss.html and expando-loss-2.html tests to be
run correctly in the MiniBrowser with the following two environment
variables set:
  export JSC_useDollarVM=1
  export __XPC_JSC_useDollarVM=1

Thanks to Yusuke Suzuki from Apple for this tip.


kenrussell:
@grorg please review. Tested this change in the MiniBrowser.

Any other feedback welcome.

PR #3146, opened 2020-08-31
Support EXT_color_buffer_half_float in WebGL 2.0 contexts.

### LINKED ISSUE #3093: WebGL 2.0 can't render to FP textures on devices with only EXT_color_buffer_half_float
While adding support for EXT_color_buffer_half_float to WebKit's ANGLE backend for WebGL in https://bugs.webkit.org/show_bug.cgi?id=212181 , it was discovered that the OpenGL ES 3.0 implementation on iOS advertises EXT_color_buffer_half_float, but not EXT_color_buffer_float.

WebGL's https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_half_float/ extension declares that it was subsumed by https://www.khronos.org/registry/webgl/extensions/EXT_color_buffer_float/ in WebGL 2.0. EXT_color_buffer_float is only allowed to be exposed if the device supports FP32 render targets, which iOS does not.

How can this be resolved? Options include:

- Allow EXT_color_buffer_half_float to be exposed from WebGL 2.0 implementations, with behavioral differences from WebGL 1.0, like dropping support for rendering to GL_RGB half-float textures.
- Define a new half-float renderability extension for WebGL 2.0.
- Expose EXT_color_buffer_float on WebGL 2.0, but allow implementations to say that framebuffers with 32-bit floating-point texture attachments are incomplete.

Thoughts on other options and opinions on these?

kenrussell:
Allows floating-point rendering in WebGL 2.0 on devices that only have
16-bit-per-channel floating point render targets (iOS, specifically),
so which can not support EXT_color_buffer_float.

Closes #3093.

### DIFF
--- a/extensions/EXT_color_buffer_half_float/extension.xml
+++ b/extensions/EXT_color_buffer_half_float/extension.xml
@@ -16,44 +16,55 @@
   <depends>
     <api version="1.0"/>
 
-    <ext name="OES_texture_half_float" require="true"/>
-
-    <subsumed version="2.0" by="EXT_color_buffer_float" />
+    <ext name="OES_texture_half_float" require="webgl1"/>
   </depends>
 
   <overview>
     <mirrors href="http://www.khronos.org/registry/gles/extensions/EXT/EXT_color_buffer_half_float.txt"
              name="EXT_color_buffer_half_float">
       <addendum>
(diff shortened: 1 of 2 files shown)


kenrussell:
@jdashg and/or @grorg and/or @jdarpinian : please review. Thanks.

As discussed some time ago in WebGL working group meetings.

PR #3147, opened 2020-09-01
Test EXT_color_buffer_half_float

jdarpinian:
There wasn't an extension test for EXT_color_buffer_half_float. This tests the new change in #3146. I added the test to the 2.0.0 suite as well; old browsers will pass because they don't support it.

### DIFF
--- a/conformance-suites/2.0.0/conformance2/extensions/ext-color-buffer-half-float.html
+++ b/conformance-suites/2.0.0/conformance2/extensions/ext-color-buffer-half-float.html
@@ -1,0 +1,508 @@
+<!--
+Copyright (c) 2019 The Khronos Group Inc.
+Use of this source code is governed by an MIT-style license that can be
+found in the LICENSE.txt file.
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL EXT_color_buffer_half_float Conformance Tests</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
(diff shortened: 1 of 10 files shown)


conformance-suites/2.0.0/conformance2/extensions/ext-color-buffer-half-float.html
```diff
@@ -0,0 +1,508 @@
+<!--
+Copyright (c) 2019 The Khronos Group Inc.
```
kenrussell, line 2: 2020

sdk/tests/conformance2/extensions/ext-color-buffer-half-float.html
```diff
@@ -0,0 +1,508 @@
+<!--
+Copyright (c) 2019 The Khronos Group Inc.
```
kenrussell, line 2: 2020

sdk/tests/conformance2/extensions/00_test_list.txt
```diff
@@ -1,4 +1,5 @@
 ext-color-buffer-float.html
+ext-color-buffer-half-float.html
```
kenrussell, line 2: Could you use --min-version 2.0.1 in this top-of-tree version? This is mainly for bookkeeping. Thanks.

jdarpinian:
The test required major refactoring to work in WebGL 1 and I discovered some issues in Chrome and ANGLE, since the extension was never tested before. Should be good to go now.

kenrussell:
Thanks for doing the refactor! I'm swamped until Wednesday - can others help review this like @shrekshao ?

shrekshao (approved):
The refactor LGTM. I think the most work is addressing those issues in Chrome.

Issue #2558, opened 2017-12-06
Platform and browser inconsistencies rendering to 3D textures


tsherif:
There are sharp inconsistencies in how rendering to 3D textures works across platforms and browsers, and it's unclear what the correct behaviour is since it isn't specified in much detail in the either WebGL 2 or OpenGL ES 3.0 specifications.

I've tested Chrome 62/Firefox 57 on Ubuntu 16.04/Quadro M1000 and Windows 10/GT 750m using the following scenes (also available in [rt-3Dtexture-bug.tar.gz](https://github.com/KhronosGroup/WebGL/files/1534769/rt-3Dtexture-bug.tar.gz)):
1. Render to TEXTURE_2D color attachment/TEXTURE_2D depth: https://tsherif.github.io/webgl2bugs/unknown/render-to-texture2D.html
2. Render to TEXTURE_3D color attachment/TEXTURE_2D depth: https://tsherif.github.io/webgl2bugs/unknown/render-to-texture3D.html
3. Render to TEXTURE_3D color attachment/TEXTURE_2D_ARRAY depth: https://tsherif.github.io/webgl2bugs/unknown/render-to-texture3D-array-depth.html

The first works everywhere and demonstrates what the scene should look like:
- "This is WebGL" texture used for offscreen draw
- Gray bar is the clear color for the offscreen draw
- Black bar is the clear color for the main draw

The second works in both browsers on Ubuntu, but fails in both on Windows. Chrome reports FRAMEBUFFER_UNSUPPORTED as the framebuffer status. Firefox reports no errors, and seems to do the clear (gray), but doesn't draw.

The third works in both browsers on Ubuntu and fixes things in Chrome on Windows. Firefox behaves the same on Windows as for the previous example.

I was tipped off to trying the third setup by the following language from the OpenGL 4.6 spec (section 9.4.2): "If any framebuffer attachment is layered, all populated attachments must be layered. Additionally, all populated color attachments must be from textures of the same target (three-dimensional, one- or two-dimensional array, cube map, or cube map array textures)."

However, neither the ES 3.0, nor the WebGL 2 spec mention any such restriction. @shannonwoods  brought it to my attention that it does appear in section 9.4.2 of the ES 3.2 spec.

zhenyao:
ATTN some ANGLE folks: @vonture @null77 @Kangz

jdashg:
This is crazy awesome. I'll read up on this.

jdashg:
[...]

PR #3148, opened 2020-09-01
Add test for fb with mismatched attachment rb/tex-targets.

jdashg:
Addresses #2558.

### DIFF
--- a/sdk/tests/conformance2/rendering/framebuffer-mismatched-attachment-targets.html
+++ b/sdk/tests/conformance2/rendering/framebuffer-mismatched-attachment-targets.html
@@ -34,7 +34,7 @@ void main() {
 "use strict";
 debug("");
 
-description("Test that WebGL2 can render to layers in 3D textures");
+description("Test framebuffer attachments with different targets");
 
 const wtu = WebGLTestUtils;
 const gl = wtu.create3DContext("example", undefined, 2);
--- a/sdk/tests/conformance2/rendering/framebuffer-mismatched-attachment-targets.html
+++ b/sdk/tests/conformance2/rendering/framebuffer-mismatched-attachment-targets.html
@@ -132,7 +132,8 @@ function runTest() {
         debug(`mipLevel: ${mipLevel}`);
         color.attach(gl.COLOR_ATTACHMENT0, mipLevel);
         depth.attach(gl.DEPTH_ATTACHMENT, mipLevel);
-        if (!wtu.framebufferStatusShouldBe(gl, gl.FRAMEBUFFER, [gl.FRAMEBUFFER_COMPLETE, gl.FRAMEBUFFER_UNSUPPORTED])) {
+        const maybeStatus = wtu.framebufferStatusShouldBe(gl, gl.FRAMEBUFFER, [gl.FRAMEBUFFER_COMPLETE, gl.FRAMEBUFFER_UNSUPPORTED]);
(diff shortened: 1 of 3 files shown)


kenrussell:
To be clear: the above failures are on macOS 10.15.6, using the system's OpenGL implementation. ANGLE isn't involved in Chrome's stack, though I tested both with and without it.

jdashg:
(Addresses #2558, since having it in the title didn't auto-link it)

kenrussell:
Good point @jdashg - updated the title and description so they'll look good in the eventual commit message.

Are you planning to push an update which allows FRAMEBUFFER_UNSUPPORTED for some or all of these combinations? If so, then we could go ahead and merge this, since https://bugs.chromium.org/p/angleproject/issues/detail?id=5009 tracks fixing the test on ANGLE on Windows.

jdashg:
PTAL

sdk/tests/conformance2/rendering/framebuffer-mismatched-attachment-targets.html
```diff
@@ -0,0 +1,161 @@
...
+  const fb = gl.createFramebuffer();
+  gl.bindFramebuffer(gl.FRAMEBUFFER, fb);
+  for (const color of colorResList) {
+    for (const depth of depthResList) {
+      debug(`\ncolor: ${color.target}; depth: ${depth.target}`);
+      for (let mipLevel = 0; mipLevel < MIP_LEVELS; mipLevel++) {
+        debug(`mipLevel: ${mipLevel}`);
+        color.attach(gl.COLOR_ATTACHMENT0, mipLevel);
+        depth.attach(gl.DEPTH_ATTACHMENT, mipLevel);
+        if (!wtu.framebufferStatusShouldBe(gl, gl.FRAMEBUFFER, [gl.FRAMEBUFFER_COMPLETE, gl.FRAMEBUFFER_UNSUPPORTED])) {
```
kenrussell, line 135: The cases mentioned earlier still fail on macOS, including in Firefox Nightly. This has to continue here (skip the rendering and clearing tests) if the framebuffer status is FRAMEBUFFER_UNSUPPORTED.

jdashg: It's sort of a pain to spin up a machine to fix this. If you see what's wrong it's most efficient if you can suggest/provide a fix collaboratively!

kenrussell: Got it. I'll post a diff here shortly which you can apply.

kenrussell: Never mind, your revision is more elegant than the one I was going to propose. :)

sdk/tests/conformance2/rendering/framebuffer-mismatched-attachment-targets.html
```diff
@@ -0,0 +1,162 @@
...
+</head>
+<body>
+<canvas id="example" width="1", height="1"></canvas>
+<div id="description"></div>
+<div id="console"></div>
+<script>
+"use strict";
+debug("");
+
+description("Test that WebGL2 can render to layers in 3D textures");
```
jdarpinian, line 37: Description needs updating.

jdarpinian (approved):

kenrussell:
@jdashg please merge yourself once @jdarpinian 's review feedback is addressed. Thanks!

PR #3154, opened 2020-10-01
Add R16F/RG16F GLenum to EXT_color_buffer_half_float

shrekshao:
Follow up to https://github.com/KhronosGroup/WebGL/pull/3146

`R16F` and `RG16F` might be missing in the idl.


kenrussell:
Thanks for suggesting this. R and RG textures are only supported in WebGL 2.0. Since EXT_color_buffer_half_float now applies to both WebGL 1.0 and 2.0, I think the user should pick up these enums from the WebGL2RenderingContext. What do you think? CC @jdarpinian .

PR #3158, opened 2020-10-05
Test many format/type combos for ReadPixels.

jdashg:


jdashg:
Blink currently allows ReadPixels(RGBA/UNSIGNED_SHORT) from the RGBA8-ish backbuffer. @kenrussell

jdashg:
This would be allowed, I suppose, if the backbuffer were RGBA16, but I think that would break a lot of other interactions?

sdk/tests/conformance/reading/read-pixels-test.html
```diff
@@ -206,37 +199,113 @@
...
         }
-      ]);
+      );
+    }
+
+    const ext = gl.getExtension('OES_texture_half_float');
+    if (ext) {
+      TYPES.push(
+        {
+          type: gl.HALF_FLOAT_OES,
```
kenrussell, line 286: This should be `ext.HALF_FLOAT_OES`. Please test the WebGL 1.0 version of this modification. e.g.:

http://localhost:8080/sdk/tests/conformance/reading/read-pixels-test.html

kenrussell (approved):
LGTM with one bug fix.

PR #3160, opened 2020-10-06
Add RGTC drawing tests to existing S3TC test.

jdarpinian:
The existing S3TC tests are very thorough. S3TC corresponds to the BC1, BC2, and BC3 formats in DirectX. The RGTC extension adds BC4 and BC5, which are 1 and 2 channel variants of BC3. Rather than build a new test, I just added BC4 and BC5 support to the existing S3TC test.

RGTC was not fully tested before, and unsurprisingly there are bugs in both Chrome and Firefox.

### DIFF
--- a/sdk/tests/conformance/extensions/s3tc-and-rgtc.html
+++ b/sdk/tests/conformance/extensions/s3tc-and-rgtc.html
@@ -12,7 +12,7 @@ found in the LICENSE.txt file.
 <script src="../../js/js-test-pre.js"></script>
 <script src="../../js/webgl-test-utils.js"></script>
 <script src="../../js/tests/compressed-texture-utils.js"></script>
-<title>WebGL WEBGL_compressed_texture_s3tc Conformance Tests</title>
+<title>WebGL WEBGL_compressed_texture_s3tc and EXT_texture_compression_rgtc Conformance Tests</title>
 <style>
 img {
  border: 1px solid black;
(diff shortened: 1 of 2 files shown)


lexaknyazev:
@jdarpinian 
Could you please add links to the browser bugs?

sdk/tests/conformance/extensions/s3tc-and-rgtc.html
```diff
@@ -130,6 +130,29 @@
     0x00, 0xF8, 0xE0, 0x07, 0x1B, 0x5B, 0xAB, 0xFF
 ]);
 
+// BC4 - just the alpha block from BC3 above, interpreted as the red channel.
```
kenrussell, line 133: Any additional comments you could leave here regarding how you constructed this data? Pointers to docs, etc.?

kenrussell (approved):
Fantastic work James improving these crucial tests!

Confirmed that these run correctly in Chrome Canary on macOS with ANGLE + the passthrough command decoder, and fail in Chrome Stable with the validating command decoder.

shrekshao (approved):

jdarpinian:
Chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1136231
Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1669848

PR #3161, opened 2020-10-06
Print error log when neither left is green nor right is blue

shrekshao:
Print detailed pixels value error when neither left is green nor right is blue to help better understanding of flakiness.

### DIFF
--- a/sdk/tests/conformance/textures/misc/texture-video-transparent.html
+++ b/sdk/tests/conformance/textures/misc/texture-video-transparent.html
@@ -92,21 +92,22 @@ function runOneIteration(videoElement, useTexSubImage2D)
     const tolerance = 5;
     const red = [255, 0, 0];
     const green = [0, 255, 0];
-    const blue = [0, 0, 255];
+    const blue = [0, 0, 200];
 
     // Check the left and right sides. Make sure that EITHER:
 
     // - Left is green and right is transparent-blended-with-red
     let leftIsGreen = false, leftIsRed = false, rightIsBlue = false, rightIsRed = false;
     let greenRedError = "", redBlueError = "";
+    let leftGreenError = "", rightBlueError = "";
     wtu.checkCanvasRectColor(gl, 4, 4, 8, 24, green, tolerance,
-        /* sameFn */ () => { leftIsGreen = true; }, /* differentFn */ m => {}, debug);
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/textures/misc/texture-video-transparent.html
```diff
@@ -123,7 +123,7 @@
             testFailed("right is blue, but: " + redBlueError);
         }
     } else {
-        testFailed("neither left is green nor right is blue")
+        testFailed("neither left is green nor right is blue \n" + greenRedError + "\n" + redBlueError);
```
kenrussell, line 126: It seems to me that if the only error is that the left side isn't green, that this detail log won't show any more information.

kenrussell (approved):
Have you tried injecting various failures into this test (making each of the comparisons fail individually) and seeing what the error messages are? If not, please do and revise as necessary before merging. Thanks.

kenrussell:
Note: I'd recommend printing out all of the read-back values from all of the locations very clearly in the situation where the test fails for any reason.

sdk/tests/conformance/textures/misc/texture-video-transparent.html
```diff
@@ -92,21 +92,22 @@
     const tolerance = 5;
     const red = [255, 0, 0];
     const green = [0, 255, 0];
-    const blue = [0, 0, 255];
+    const blue = [0, 0, 200];
```
kenrussell, line 95: Did you mean to make this change, or was it only for testing?

shrekshao: that was meant to be deleted. BTW I'm extending to print the whole buf.

shrekshao:
Now outputing whole buffer value when failed. Will merge and roll if that looks good. @kenrussell

sdk/tests/js/webgl-test-utils.js
```diff
@@ -1283,7 +1283,7 @@ var checkCanvasRectColor = function(gl, x, y, width, height, color, opt_errorRan
           was += "," + buf[offset + j];
         }
         differentFn('at (' + (x + (i % width)) + ', ' + (y + Math.floor(i / width)) +
-                    ') expected: ' + color + ' was ' + was);
+                    ') expected: ' + color + ' was ' + was, buf);
```
kenrussell, line 1286: Hopefully this has no effect on other tests.

kenrussell (approved):
Looks good!

Issue #3141, opened 2020-08-18
Expand EXT_texture_norm16 conformance test to test uploading DOM elements / ImageBitmaps / etc. to texture image


shrekshao:
I'm gonna add this one.

kenrussell:
Thanks Shrek - assigned this one to you.

shrekshao:
I have a stupid question (some basic concept understanding issue....)
For passing image src data/imageBitmaps to texImage2D, is the webgl expecting an 8bit input image (png and jpg only seems to support 8bit image)? (But the EXT_texture_norm16 spec doesn't expect a valid combination of UNSIGNED_BYTE as the type and RGBA16_EXT as the internalformat)
Or is the webgl expecting an 16bit input image as input? (But tif is not supported for dom img object)

lexaknyazev:
PNG supports 16-bit pixels: https://www.w3.org/TR/2003/REC-PNG-20031110/#11IHDR

shrekshao:
Ah... Thanks for the pointer. It's a shame I didn't know that

kenrussell:
In a meeting this week, @juj from Unity asked whether WebP supports 16 bits per channel - we're investigating that.

jaikrishnan:
> In a meeting this week, @juj from Unity asked whether WebP supports 16 bits per channel - we're investigating that.

WebP assumes 8-bits per channel for [lossless](https://developers.google.com/speed/webp/docs/webp_lossless_bitstream_specification?hl=en#1_introduction)

shrekshao:
Thanks for the detail.

Closing this as we now put DOM elements uploading for 16 bpcc into a new extension. Proposal is here: https://github.com/KhronosGroup/WebGL/pull/3191

PR #3179, opened 2020-11-10
Update EXT_texture_norm16 extension spec with support constrained to ArrayBufferView only

shrekshao:
Update the spec as discovered during implementing the feature in Chrome:
* Add behavior descriptions for uploading various types of image elements with texImage2d
  - Uploading image elements with signed normalized integer type is not allowed (according to https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.7.6
`[throws] void texImage2D(GLenum target, GLint level, GLint internalformat, GLsizei width, GLsizei height, GLint border, GLenum format, GLenum type, TexImageSource source) // May throw DOMException (OpenGL ES 3.0.4 §3.8.4, man page)`)
  - ~~Uploading image elements with 8-bit content with `UNSIGNED_SHORT` type should raise a `DOMException`. And the behavior is undefined.~~
* Add a getSupportedInputs function to make support for image elements types other than arrayBufferView become easier to be optional

### DIFF
--- a/extensions/EXT_texture_norm16/extension.xml
+++ b/extensions/EXT_texture_norm16/extension.xml
@@ -53,7 +53,7 @@
       can be queried from calling the <code>getSupportedInputs</code> function.
       When the data source is a DOM element (HTMLImageElement, HTMLCanvasElement, or HTMLVideoElement),
       or is an ImageBitmap, ImageData, or OffscreenCanvas object with only 8 bits unsigned integer content,
-      a <code>DOMException</code> should be rasied and the behavior is undefined.
+      the conversion to 16 bit would have to requantize (e.g. 0x42 -> 0x4242).
       </feature>
     </features>
   </overview>
--- a/extensions/EXT_texture_norm16/extension.xml
+++ b/extensions/EXT_texture_norm16/extension.xml
@@ -74,7 +74,7 @@ interface EXT_texture_norm16 {
       <change>Promoted to Community Approved.</change>
     </revision>
     <revision date="2020/12/03">
-      <change>Constraint the scope to only support ArrayBufferView</change>
+      <change>Constrain the scope to only support ArrayBufferView.</change>
     </revision>
   </history>
 </extension>
--- a/extensions/EXT_texture_norm16/extension.xml
+++ b/extensions/EXT_texture_norm16/extension.xml
@@ -53,7 +53,7 @@
       can be queried from calling the <code>getSupportedInputs</code> function.
       When the data source is a DOM element (HTMLImageElement, HTMLCanvasElement, or HTMLVideoElement),
       or is an ImageBitmap, ImageData, or OffscreenCanvas object with only 8 bits unsigned integer content,
-      the conversion to 16 bit would have to requantize (e.g. 0x42 -> 0x4242).
+      the conversion to 16 bit is requantized (e.g. 0x42 -> 0x4242).
       </feature>
     </features>
   </overview>
@@ -70,13 +70,23 @@ interface EXT_texture_norm16 {
   const GLenum RGB16_SNORM_EXT = 0x8F9A;
   const GLenum RGBA16_SNORM_EXT = 0x8F9B;
 };
+
+enum TextureNorm16SupportedInputTypes {
+  "arraybufferview",
+  "imagebitmap",
+  "imagedata",
+  "htmlimageelement",
+  "htmlcanvaselement",
+  "htmlvideoelement",
+  "offscreencanvas"
+};
   </idl>
 
   <newfun>
     <function name="getSupportedInputs" type="sequence&lt;DOMString&gt;">
       Returns the names of the input types supported by the implementation. As of this writing,
-      possibly valid return values will include "ArrayBufferView", "ImageBitmap", "ImageData",
-      "HTMLImageElement", "HTMLCanvasElement", "HTMLVideoElement", and "OffscreenCanvas".
+      possibly valid return values are "arraybufferview", "imagebitmap", "imagedata",
+      "htmlimageelement", "htmlcanvaselement", "htmlvideoelement", and "offscreencanvas" as stated in the idl <code>enum TextureNorm16SupportedInputTypes</code>.
     </function>
     <div class="note">
       The intent of the <code>getSupportedInputs</code> function is to not blocking broswers from advertising
--- a/extensions/EXT_texture_norm16/extension.xml
+++ b/extensions/EXT_texture_norm16/extension.xml
@@ -46,13 +46,13 @@
       entry points taking <code>ImageBitmap</code>, <code>ImageData</code>,
       <code>HTMLImageElement</code>, <code>HTMLCanvasElement</code>, <code>HTMLVideoElement</code>,
       and <code>OffscreenCanvas</code> are extended to accept the pixel type
-      <code>UNSIGNED_SHORT</code>, with combination internal format being one of
+      <code>UNSIGNED_SHORT</code>, with associated internal format being one of
       <code>R16_EXT</code>, <code>RG16_EXT</code>,
       <code>RGB16_EXT</code>, and <code>RGBA16_EXT</code>.
       Information on whether each type is supported
       can be queried from calling the <code>getSupportedInputs</code> function.
       When the data source is a DOM element (HTMLImageElement, HTMLCanvasElement, or HTMLVideoElement),
-      or is an ImageBitmap, ImageData, or OffscreenCanvas object with only 8 bits unsigned integer content,
+      or is an ImageBitmap, ImageData, or OffscreenCanvas object with only 8 bit per channel unsigned integer content,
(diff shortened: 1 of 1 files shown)


shrekshao:
Thanks to @lexaknyazev pointing at some early discussion on 8bit DOM image src: https://github.com/KhronosGroup/WebGL/pull/2837#issuecomment-477757794

extensions/EXT_texture_norm16/extension.xml
```diff
@@ -43,16 +43,22 @@
...
+      <code>HTMLImageElement</code>, <code>HTMLCanvasElement</code>, <code>HTMLVideoElement</code>,
+      and <code>OffscreenCanvas</code> are extended to accept the pixel type
+      <code>UNSIGNED_SHORT</code>, with combination internal format being one of
+      <code>R16_EXT</code>, <code>RG16_EXT</code>,
+      <code>RGB16_EXT</code>, and <code>RGBA16_EXT</code>.
+      Information on whether each type is supported
+      can be queried from calling the <code>getSupportedInputs</code> function.
+      When the data source is a DOM element (HTMLImageElement, HTMLCanvasElement, or HTMLVideoElement),
+      or is an ImageBitmap, ImageData, or OffscreenCanvas object with only 8 bits unsigned integer content,
+      the conversion to 16 bit would have to requantize (e.g. 0x42 -> 0x4242).
```
kenrussell, line 56: would have to requantize -> is requantized

extensions/EXT_texture_norm16/extension.xml
```diff
@@ -66,6 +72,19 @@ interface EXT_texture_norm16 {
 };
   </idl>
 
+  <newfun>
+    <function name="getSupportedInputs" type="sequence&lt;DOMString&gt;">
+      Returns the names of the input types supported by the implementation. As of this writing,
+      possibly valid return values will include "ArrayBufferView", "ImageBitmap", "ImageData",
```
kenrussell, line 78: Please rephrase: "valid return values are".

Also, could you please define an enumeration for this return value?
https://heycam.github.io/webidl/#idl-enums

Please use lowercase for the enum values. I think not having hyphens is easier to read: e.g. "arraybufferview".

kenrussell (commented):
Looks good overall Shrek. A couple of minor comments before approving.

extensions/EXT_texture_norm16/extension.xml
shrekshao: Like what we have in the latest?

extensions/EXT_texture_norm16/extension.xml
```diff
@@ -64,8 +70,31 @@ interface EXT_texture_norm16 {
...
+  "imagedata",
+  "htmlimageelement",
+  "htmlcanvaselement",
+  "htmlvideoelement",
+  "offscreencanvas"
+};
   </idl>
 
+  <newfun>
+    <function name="getSupportedInputs" type="sequence&lt;DOMString&gt;">
```
kenrussell, line 86: Can the return type be sequence&lt;TextureNorm16SupportedInputTypes&gt; ? Could you verify in Chrome that that IDL will parse?

extensions/EXT_texture_norm16/extension.xml
```diff
@@ -43,16 +43,22 @@
...
       <feature> The <code>texImage2D</code> and <code>texSubImage2D</code>
-      entry points taking <code>ImageData</code>,
-      <code>HTMLImageElement</code>, <code>HTMLCanvasElement</code> and
-      <code>HTMLVideoElement</code> are extended to accept the pixel type
-      <code>UNSIGNED_SHORT</code>. </feature>
-
+      entry points taking <code>ImageBitmap</code>, <code>ImageData</code>,
+      <code>HTMLImageElement</code>, <code>HTMLCanvasElement</code>, <code>HTMLVideoElement</code>,
+      and <code>OffscreenCanvas</code> are extended to accept the pixel type
+      <code>UNSIGNED_SHORT</code>, with combination internal format being one of
```
kenrussell, line 49: Suggest "combination" -> "associated"

extensions/EXT_texture_norm16/extension.xml
```diff
@@ -43,16 +43,22 @@
...
+      <code>R16_EXT</code>, <code>RG16_EXT</code>,
+      <code>RGB16_EXT</code>, and <code>RGBA16_EXT</code>.
+      Information on whether each type is supported
+      can be queried from calling the <code>getSupportedInputs</code> function.
+      When the data source is a DOM element (HTMLImageElement, HTMLCanvasElement, or HTMLVideoElement),
+      or is an ImageBitmap, ImageData, or OffscreenCanvas object with only 8 bits unsigned integer content,
```
kenrussell, line 55: 8 bits -> 8 bit per channel

kenrussell (approved):
Thanks for revising; looks good overall. A couple more minor requests; please squash and merge yourself after reviewing and potentially updating this again. Thanks.

extensions/EXT_texture_norm16/extension.xml
shrekshao: Oops. Verified. Also example from WebGLPowerPreference

shrekshao:
void texImage2D(GLenum target, GLint level, GLint internalformat, GLsizei width, GLsizei height, GLint border, GLenum format, GLenum type, TexImageSource source)


| internalFormat | format | type | img src | behavior |
| --------------- | ------ | ---- | -------- | -------- |
| `RGBA16` | `RGBA` | `UNSIGNED_SHORT` | 16 bit | ✅ |
| `RGBA16` | `RGBA` | `UNSIGNED_SHORT` | 8 bit | requantize (0x42 -> 0x4242) |
| `RGBA16_SNORM` | `RGBA` | `SHORT` | 16 bit | `INVALID_ENUM`/`INVALID_OPERATION` |
| `RGBA16_SNORM` | `RGBA` | `SHORT` | 8 bit | `INVALID_ENUM`/`INVALID_OPERATION` |

kenrussell (approved):
The table in your comment above looks very nice. Would you consider embedding it into this extension spec as HTML? Note that texSubImage2D's behavior should also be documented.

jdashg:
I don't know how I feel about this.
I think it would be more consistent to either allow re-quantization to unorm and snorm both, or forbid re-quantization for unorm8->unorm16 for now.

For the time being I would feel better if you couldn't use UNSIGNED_SHORT with TexImageSource, and I don't think that breaks any use-cases.

I think this gets into how our upload-from-image/media story is generally not great.
I'm pretty sure what nine out of ten users actually want is closer to TexImage2D(target, mip, TexImageSource)
I want to think more about this, but I think it's valuable to have unorm16 textures from non-TexImageSource without any other changes.

shrekshao:
+1 for "forbid re-quantization for unorm8->unorm16 for now."

kenrussell:
@jdashg : @shrekshao has been working on supporting 16-bit PNG HTMLImageElements to EXT_texture_norm16 texture formats/types per your feedback during the WebGL WG conference call on 2020-08-13.

Unity's told us that there's value in supporting image uploads (HTMLImageElement, in particular) to 16-bit-per-channel textures. Authors today create some images, like height maps for terrain, as 16-bit-per-channel PNGs. It would be nice for browsers to support this. They don't need ImageBitmap support, but that'd be a good next step, since it lets images be decoded on background threads.

Removing the requantizing of 8-bit images sounds good to me. Making uploads of 8-bit images to 16-bit-per-channel textures raise INVALID_OPERATION is my preference.

I think that adding the query of getSupportedInputs makes it easier to incrementally add functionality to browsers. Would you agree to add it to the spec?

shrekshao:
Referencing #3141

shrekshao:
@jdashg @lexaknyazev Do we have any new thoughts?
I think what we have discussed at WebGL WG could be concluded as:
* If we can decide the minimum support level we want (say arrayBufferView, HTMLImageElement and ImageBitmap) then we just chop the support list and make the statement in the extension spec.
* If we cannot safely decide the scope, then
  - We limit EXT_texture_norm16 to support only arrayBufferView. We will make new extensions like EXT_texture_norm16_img, EXT_texture_norm16_videos etc.
  - We use a query function (as stated in this pull request)

kenrussell:
@shrekshao and I discussed this. Of those options, I'd like to advocate for:
  - Supporting only uploads of HTMLImageElement and ImageBitmap to 16-bit-per-channel textures - no canvases or video; those throw DOMException
  - No "re-quantization" - attempting to upload 8-bit-per-channel data to 16-bit-per-channel textures throws DOMException
  - Either keeping everything in one extension, or adding EXT_texture_norm16_img which adds this capability (would make it easier to ship the baseline ArrayBuffer-based extension, which has the most value, on other browsers)

Other opinions on this?

shrekshao:
Simpy removed support for anything else other than ArrayBufferView. Support for other texture source types would be in a new separate extension (EXT_texture_norm16_dom_source).

kenrussell:
Great. Thanks @shrekshao for pushing this forward.

PR #3181, opened 2020-11-14
Fix serve_localhost.py to bind to localhost again.

jdashg:
It was only binding to [::1], sometimes causing pages to fail to load.
This was probably due to changes in http.server.test, which we probably
shouldn't have been abusing for this.


kenrussell (approved):
Looks good to me!

PR #3187, opened 2020-11-24
texImage2D takes in gl.RGB10_A2UI, gl.RGBA_INTEGER for DOMElement gen…

shrekshao:
…erates either gl.INVALID_VALUE or gl.INVALID_ENUM

Removing some redundant validations in Chromium fails this test (that unrelated validation accidently make it pass this test): https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc;l=7493-7499?q=webgl_rendering_context_base&ss=chromium

Usually invalid formats enum combinations generates `gl.INVALID_ENUM`. `gl.INVALID_VALUE` is used mostly for integer value error (width, level, border etc.)
(This one is a bit different as it's valid for takingin arrayBuffer but invalid various DOMElements; Although I still interprets this as invalid enum for DOMElements)

Simply extending expecting result to either `gl.INVALID_VALUE` or `gl.INVALID_ENUM`, which requires the least effort and shall create the least noise if any.

(Related spec discussion at: https://github.com/KhronosGroup/WebGL/issues/2274#issue-203074505)


kenrussell (approved):
Thanks for the detailed comment. This seems straightforward and non-controversial.

PR #3188, opened 2020-11-24
KHR_parallel_shader_compile: lost context behavior spec & test

jdarpinian:
This was proposed on the public WebGL mailing list, where it was pointed out that KHR_robustness seems to imply that this should be the behavior of all queries. https://www.khronos.org/webgl/public-mailing-list/public_webgl/1904/msg00004.php. The OpenGL working group agrees: https://github.com/KhronosGroup/OpenGL-Registry/issues/263


kenrussell (approved):
Nice test. I wish it were possible to make assertions that the completion status of programs that take a long time to compile was false and then transitions to true - let's try to strengthen the test in the future.

PR #3190, opened 2020-12-03
Add uint-int-shift-bug test

### LINKED ISSUE #2989: GLSL ES 3.0 uint on Adreno 5xx/6xx
From the spec:

**Shift operators (`<<`) and (`>>`)**
> For both operators, the operands must be signed or unsigned integers or integer vectors. One operand can be signed while the other is unsigned. In all cases, the resulting type will be the same type as the left operand. If the first operand is a scalar, the second operand has to be a scalar as well. If the first operand is a vector, the second operand must be a scalar or a vector with the same size as the first operand, and the result is computed component-wise. The result is undefined if the right operand is negative, or greater than or equal to the number of bits in the left expression's base type. The value of `E1 >> E2` is `E1` right-shifted by `E2` bit positions. If `E1` is a signed integer, the right-shift will extend the sign bit. If `E1` is an unsigned integer, the right-shift will zero-extend.

**And (`&`)**
> The operands must be of type signed or unsigned integers or integer vectors. The operands cannot be vectors of differing size. If one operand is a scalar and the other a vector, the scalar is applied component-wise to the vector, resulting in the same type as the vector. The fundamental types of the operands (signed or unsigned) must match, and will be the resulting fundamental type. For and (`&`), the result is the bitwise-and function of the operands.

**Conversion and Scalar Constructors**
> The constructor `int(uint)` preserves the bit pattern in the argument, which will change the argument's value if its sign bit is set. The constructor `uint(int)` preserves the bit pattern in the argument, which will change its value if it is negative.

Following the definitions above, these two expressions must produce the same value.
```glsl
precision highp int;
ivec3 d; // must not be compile-time constant
int v1 = (d.x >> 25) & 0x7F;
int v2 = int(uint(d.x & 0xFE000000) >> 25U);
```
On Adreno 5xx (checked 508 and 540) devices, `uint(...)` cast is apparently ignored, so the right shift always operates on a signed value.

lexaknyazev:
Fixes #2989.

Casting the value twice (`uint(uint(int))`) produces correct result. Vector types are not affected.

### DIFF
--- a/sdk/tests/conformance2/glsl3/00_test_list.txt
+++ b/sdk/tests/conformance2/glsl3/00_test_list.txt
@@ -42,6 +42,7 @@ short-circuiting-in-loop-condition.html
 texture-offset-out-of-range.html
 --min-version 2.0.1 texture-offset-uniform-texture-coordinate.html
 --min-version 2.0.1 tricky-loop-conditions.html
+--min-version 2.0.1 uint-int-shift-bug.html
 --min-version 2.0.1 unary-minus-operator-in-dynamic-loop.html
 uniform-block-layouts.html
 uniform-block-layout-match.html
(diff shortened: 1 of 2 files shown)


kenrussell (approved):
Meant to say: LGTM.

PR #3192, opened 2020-12-04
Add precision-side-effects-bug test

lexaknyazev:
When this test is run on Adreno (no repros on other vendors so far):
- the result of the expression `0 | (int(e0 == e0) * 0x8000)` somehow returns `-32768` instead of `32768` despite the variable using `highp` precision;
 - splitting the expression along `|` fixes the issue (could also be observed with other operators).


kenrussell:
Thanks @lexaknyazev for the test. Looks fine.

Please file a bug on the Chromium issue tracker about this and comment here with the bug ID. We'll try to contact Qualcomm to get them to fix it.

kenrussell (approved):
LGTM

lexaknyazev:
Filed https://crbug.com/1155942.

PR #3193, opened 2020-12-09
Extend KHR_parallel_shader_compile test

jdarpinian:
Construct a shader that takes a long time to compile and actually test that COMPLETION_STATUS_KHR 
transitions from false to true without blocking. Also add a button for manual verification that animations aren't blocked.

### DIFF
--- a/sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
+++ b/sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
@@ -65,13 +65,6 @@ ${extra || ''}
     gl_FragColor = result;
 }`;
 
-const generateExtraCode = (amount) => {
-    let code = '\n    if (true) { result += vec4(0.0000001) }';
-    for (let i = 0; i < amount; i++)
-        code += code;
-    return code;
-};
-
 let vs = gl.createShader(gl.VERTEX_SHADER);
 let fs = gl.createShader(gl.FRAGMENT_SHADER);
 let program = gl.createProgram();
@@ -119,10 +112,11 @@ let extraCode = '';
         status = gl.getProgramParameter(program, COMPLETION_STATUS_KHR);
         if (status !== true && status !== false) testFailed("status should be a boolean");
 
-        debug("Constructing shader that takes >500ms to compile");
+        const minimumShaderCompileDurationMs = 500;
+        debug(`Constructing shader that takes > ${minimumShaderCompileDurationMs} ms to compile.`);
         let measuredCompileDuration = 0;
         extraCode = '\n    if (true) { result += vec4(0.0000001); }';
-        for (let i = 0; measuredCompileDuration < 500; i++) {
+        for (let i = 0; measuredCompileDuration < minimumShaderCompileDurationMs; i++) {
             extraCode += extraCode;
             extraCode += extraCode;
             if (i < 4) continue;
@@ -153,18 +147,19 @@ let extraCode = '';
         gl.getShaderParameter(fs, COMPLETION_STATUS_KHR);
         gl.getShaderParameter(vs, COMPLETION_STATUS_KHR);
         let duration = performance.now() - start;
-        if (performance.now() - start > 100)
+        if (duration > 100)
             testFailed(`Querying shader status should not wait for compilation. Took ${duration} ms`);
 
         let frames = 0;
+        const maximumTimeToWait = measuredCompileDuration * 4;
         while (!gl.getProgramParameter(program, COMPLETION_STATUS_KHR)
-            && performance.now() - start < measuredCompileDuration * 4) {
+            && performance.now() - start < maximumTimeToWait) {
             frames++;
             await new Promise(requestAnimationFrame);
         }
         duration = performance.now() - start;
         if (!gl.getProgramParameter(program, COMPLETION_STATUS_KHR)) {
-            testFailed(`Program took longer than 4x the expected duration to compile. Expected: ${measuredCompileDuration} ms, actual: ${duration} ms`);
+            testFailed(`Program took longer than ${maximumTimeToWait} ms to compile. Expected: ${measuredCompileDuration} ms, actual: ${duration} ms`);
         } else if (!gl.getShaderParameter(vs, COMPLETION_STATUS_KHR) || !gl.getShaderParameter(fs, COMPLETION_STATUS_KHR)) {
             testFailed('Program linked before shaders finished compiling.');
         } else if (frames <= 6) {
--- a/sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
+++ b/sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
@@ -11,14 +11,39 @@ found in the LICENSE.txt file.
 <link rel="stylesheet" href="../../resources/js-test-style.css"/>
 <script src="../../js/js-test-pre.js"></script>
 <script src="../../js/webgl-test-utils.js"></script>
+<style>
+.spinner {
+    width: 100px;
+    height: 100px;
+    border: 20px solid transparent;
+    border-top: 20px solid black;
+    border-radius: 100%;
+    text-align: center;
+    padding: 10px;
+}
+@keyframes rotation {
+    from { transorm: rotate(0); }
+    to { transform: rotate(360deg); }
+}
+</style>
 </head>
 <body>
 <div id="description"></div>
+<button onclick='compileShaders()'>The spinners below should not stutter when you click this button.</button>
+<div class="spinner" style="animation: rotation 2s infinite linear;">CSS</div>
+<div class="spinner" id=spinner>JS</div>
 <div id="console"></div>
-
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
```diff
@@ -11,33 +11,70 @@
...
-    gl_FragColor = vec4(0.${counter++}, 1, 1, 1);
+    vec4 result = vec4(0.${counter++});
+${extra || ''}
+    gl_FragColor = result;
 }`;
 
-const vs = gl.createShader(gl.VERTEX_SHADER);
-const fs = gl.createShader(gl.FRAGMENT_SHADER);
-const program = gl.createProgram();
+const generateExtraCode = (amount) => {
```
kenrussell, line 68: This doesn't seem to be referenced.

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
```diff
@@ -58,36 +95,131 @@
...
+async function compileShaders() {
+    console.log('Compiling shaders');
+    const gl = canvas.getContext('webgl');
+    const vs = gl.createShader(gl.VERTEX_SHADER);
+    const fs = gl.createShader(gl.FRAGMENT_SHADER);
+    const program = gl.createProgram();
+    gl.getExtension(wtu.getExtensionWithKnownPrefixes(gl, "KHR_parallel_shader_compile"));
+    gl.attachShader(program, vs);
+    gl.attachShader(program, fs);
+    gl.shaderSource(vs, vertexSource(extraCode));
```
kenrussell, line 208: Since invoking this function doesn't change extraCode, it seems likely to me that these shader compilations and program link could hit the browser's program binary caches, and not actually do any work. Would it be possible to better randomize the shader's code? Repeatability is ideal, and a pseudorandom number generator would work for that (see http://crbug.com/1132587#c9 ), but that would also probably hit the browser's cache from run to run. Maybe seeding the PRNG with the current time and printing the seed for each run would allow debugging and preserve repeatability.

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
```diff
@@ -58,36 +95,131 @@
...
+        gl.shaderSource(fs, fragmentSource(extraCode));
+        gl.compileShader(vs);
+        gl.compileShader(fs);
+        gl.linkProgram(program);
+
+        let start = performance.now();
+        gl.getShaderParameter(fs, COMPLETION_STATUS_KHR);
+        gl.getShaderParameter(vs, COMPLETION_STATUS_KHR);
+        let duration = performance.now() - start;
+        if (performance.now() - start > 100)
```
kenrussell, line 156: Should this read "if (duration > 100)"?

Also concerned about performance on slower machines, like Chromium's testing fleet - though I agree there's no reason these calls should take any appreciable period of time.

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
```diff
@@ -58,36 +95,131 @@
...
+        gl.compileShader(vs);
+        let status = gl.getShaderParameter(vs, COMPLETION_STATUS_KHR);
+        if (status !== true && status !== false) testFailed("status should be a boolean");
+
+        gl.linkProgram(program);
         status = gl.getProgramParameter(program, COMPLETION_STATUS_KHR);
-        if (status !== true) testFailed("status should be true when context is lost");
+        if (status !== true && status !== false) testFailed("status should be a boolean");
+
+        debug("Constructing shader that takes >500ms to compile");
```
kenrussell, line 122: Factor out constant into a separate variable?

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
```diff
@@ -58,36 +95,131 @@
...
+        let start = performance.now();
+        gl.getShaderParameter(fs, COMPLETION_STATUS_KHR);
+        gl.getShaderParameter(vs, COMPLETION_STATUS_KHR);
+        let duration = performance.now() - start;
+        if (performance.now() - start > 100)
+            testFailed(`Querying shader status should not wait for compilation. Took ${duration} ms`);
+
+        let frames = 0;
+        while (!gl.getProgramParameter(program, COMPLETION_STATUS_KHR)
+            && performance.now() - start < measuredCompileDuration * 4) {
```
kenrussell, line 161: Use a separate for the factor?

Especially concerned about performance on slower machines, such as those in Chromium's testing fleet. I think this will end up being very flaky.

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
```diff
@@ -58,36 +95,131 @@
...
+            frames++;
+            await new Promise(requestAnimationFrame);
+        }
+        duration = performance.now() - start;
+        if (!gl.getProgramParameter(program, COMPLETION_STATUS_KHR)) {
+            testFailed(`Program took longer than 4x the expected duration to compile. Expected: ${measuredCompileDuration} ms, actual: ${duration} ms`);
+        } else if (!gl.getShaderParameter(vs, COMPLETION_STATUS_KHR) || !gl.getShaderParameter(fs, COMPLETION_STATUS_KHR)) {
+            testFailed('Program linked before shaders finished compiling.');
+        } else if (frames <= 6) {
+            testFailed(`Program should have taken many more than 6 frames to compile. Actual value: ${frames} frames, duration ${performance.now() - start} ms.`);
```
kenrussell, line 171: With the browser's program binary caches I'm concerned this may be hit often, and introduce flakiness in this test.

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
```diff
@@ -11,33 +11,70 @@
...
+}
+@keyframes rotation {
+    from { transorm: rotate(0); }
+    to { transform: rotate(360deg); }
+}
+</style>
 </head>
 <body>
 <div id="description"></div>
+<button onclick='compileShaders()'>The spinners below should not stutter when you click this button.</button>
```
kenrussell, line 32: Can you think of a way - like measuring requestAnimationFrame timestamps - to programmatically catch dropped frames over a period of time, rather than relying on manual testing?

kenrussell (commented):
Nice work @jdarpinian enhancing these tests. Looks like a good direction overall; a few questions and comments.

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
jdarpinian: I agree that flakiness is a danger, but I think with some tuning on the thresholds it can be made quite reliable. It shouldn't matter how slow or fast the machine is in absolute terms, because we're comparing it to itself. This would only be flaky if the shader takes more than 4 times longer to compile the second time vs the first time.

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
jdarpinian: The vertexSource and fragmentSource functions increment a counter and put the changed value into the code every time they are called, so it should bypass any cache.

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
jdarpinian: The test kind of does, by requiring several frames to elapse during the compilation time. But it can't verify that pixels on the screen actually change. That's what the manual test is for.

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
kenrussell: I see - thanks for pointing that out.

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
kenrussell: Seems this has been resolved with the counter being incremented in the source code.

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
kenrussell: Other tests - conformance/rendering/texture-switch-performance.html specifically - have attempted to compare against baseline performance but were not successful, and were ultimately too flaky to be turned on on Chromium's test bots. See the blanket suppression for that test in Chromium's webgl_conformance_expectations.txt.

We can certainly try this on the bots, though I'd suggest factoring out the factor into a separate constant.

sdk/tests/conformance/extensions/khr-parallel-shader-compile.html
kenrussell: I see. Thanks for clarifying this.

kenrussell (approved):
Looks good. Please revisit the unresolved conversations before merging - thanks.

PR #3194, opened 2020-12-09
Add more comments to precision-side-effects test

kenrussell:
Follow-on to #3192 .


kenrussell:
Adding these before reporting the bug to Qualcomm.

cc @lexaknyazev . Merging myself.

PR #3195, opened 2020-12-20
Replace GLuint64EXT with GLuint64

saschanaz:
Back in the old days [there was no GLuint64](https://github.com/KhronosGroup/WebGL/pull/2076#discussion_r81674534), but [now there is](https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.1).

The current definition causes IDL parser warning because there are two definitions, one here and another in [EXT_disjoint_timer_query](https://www.khronos.org/registry/webgl/extensions/EXT_disjoint_timer_query/).


jdashg:
Works for me. No need for us to be pedantic here.

PR #3198, opened 2020-12-22
Allow makeXRCompatible as an optional method

grorg:
The WebXR device specification mixes in a new function makeXRCompatible.
The methods.html and methods-2.html tests should allow for the possibility
of the method existing.

While here, update the code to more modern JavaScript.

### DIFF
--- a/sdk/tests/conformance/context/methods.html
+++ b/sdk/tests/conformance/context/methods.html
@@ -20,7 +20,7 @@ found in the LICENSE.txt file.
 "use strict";
 description("This test ensures that the WebGL context has all the methods in the specification.");
 
-var methods = [
+const methods = [
   "getContextAttributes",
   "activeTexture",
   "attachShader",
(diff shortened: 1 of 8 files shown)


sdk/tests/conformance/context/methods.html
```diff
@@ -161,47 +161,51 @@
...
 // spec that are backward-compatible with this version
-var ignoredMethods = [
+const ignoredMethods = [
   // There is no official spec for the commit API yet, the proposal link is:
   // https://wiki.whatwg.org/wiki/OffscreenCanvas
   "commit"
 ];
 
+const optionalMethods = [
+  "makeXRCompatible"
```
kenrussell, line 171: Should the presence of this method be tied to the existence of some other method or global interfaces?

kenrussell (approved):
This looks fine. One small question, but we can merge this as is.

Does this need to be backported in particular to the 1.0.3 and 2.0.0 snapshots?

sdk/tests/conformance/context/methods.html
grorg: I thought about that, but I decided it was the place of the WebXR tests to check for conformance there. I consider this patch to simply be a workaround for a well-known method name.

grorg:
> This looks fine. One small question, but we can merge this as is.
> 
> Does this need to be backported in particular to the 1.0.3 and 2.0.0 snapshots?

Yes, it does. Would you like that in this patch also?

kenrussell:
> > This looks fine. One small question, but we can merge this as is.
> > Does this need to be backported in particular to the 1.0.3 and 2.0.0 snapshots?
> 
> Yes, it does. Would you like that in this patch also?

Sure, would you mind? That way there are fewer commits to track down if necessary later. Thanks!

jdashg (approved):

jdashg:
Refactored and backported. PTAL @kenrussell @grorg

kenrussell (approved):
Super. Thank you for doing the backport @jdashg . Looks good.

@jdarpinian could you please also review and test these changes? @jdashg could we wait to merge this until James has had a chance to verify? Thanks.

jdarpinian (approved):

PR #3199, opened 2020-12-23
Add test for calling texImage2D on video element with "rotate" metadata tag.

nihohit:
This test verifies that the process of copying the content of an HTMLVideoElement
into a WebGL texture takes into account rotation information from the metadata,
in a similar manner to how the process handles an image's EXIF information.


sdk/tests/conformance/textures/misc/video-rotation.html
```diff
@@ -0,0 +1,114 @@
...
+                let y = loc[1];
+                output("  Checking " + place);
+                wtu.checkCanvasRect(gl, Math.floor(canvas.width * x), Math.floor(canvas.height * y), 1, 1,
+                    color.color, "shouldBe " + color.color + " +/-" + tolerance, tolerance);
+            }
+        }
+
+        function loadVideoElement(filename) {
+            return new Promise((resolve) => {
+                const video = new HTMLVideoElement();
```
kenrussell, line 60: This doesn't work in any browser - you can use `document.createElement('video')`.

sdk/tests/conformance/textures/misc/video-rotation.html
```diff
@@ -0,0 +1,114 @@
...
+                video.addEventListener('loadeddata', () => {
+                   resolve(video);
```
kenrussell, line 62: Please use WebGLTestUtils.startPlayingAndWaitForVideo. That function already contains several improvements to make videos behave more portably between browsers.

kenrussell (changes requested):
Thanks for contributing this test! Some more work on it is needed before it can be merged, but we certainly appreciate your help with it!

kenrussell (changes requested):
Also: please look at sdk/tests/js/tests/tex-image-and-sub-image-2d-with-video.js. It's necessary to use HTMLVideoElement.canPlayType to see whether the test should try .mp4 videos at all. It would be ideal if you would also test VP8/VP9 .webm videos.

Could you also please point to the specification of the supported rotations in these videos?

sdk/tests/conformance/textures/misc/video-rotation.html
nihohit: done

nihohit:
Hi Ken, thanks for the corrections! I just pushed a round of fixes. However, I still can't run the test, or other video tests (like `text-image-with-video-test.js`), due to CORS issues. Should the tests/resources be hosted in some specific way?

AFAIS, webm and VP9 don't support rotations or other arbitrary transformations by their metadata. Do you know of something that I missed?
https://www.webmproject.org/vp9/mp4/
http://wiki.webmproject.org/webm-metadata/global-metadata

nihohit:
> Could you also please point to the specification of the supported rotations in these videos?

I'm sorry, but I couldn't find the specs for mp4/h264 metadata online. The `rotate` tag is fairly common in video captured on mobile phones. Here's a non-official post about it - https://blog.addpipe.com/mp4-rotation-metadata-in-mobile-video-files/

kainino0x:
> Hi Ken, thanks for the corrections! I just pushed a round of fixes. However, I still can't run the test, or other video tests (like `text-image-with-video-test.js`), due to CORS issues. Should the tests/resources be hosted in some specific way?

IIRC, they should work without any special hosting, but you do have to use a server (rather than `file://`), such as `serve_localhost.py` in this repository.

nihohit:
Thanks @kainino0x!

The tests now pass on chrome, which means that they don't replicate the bug I found. I'll try to switch the files and see what I did wrong with them.

nihohit:
found the problem: the bug reproduces with videos encoded in yuv420, but not in yuv444.

kenrussell (changes requested):
Thanks @kainino0x for helping and @nihohit for revising the tests!

Could you please document how you created these test files in a comment in video-rotation.html?

video-rotation-0.mp4 doesn't play in either Firefox or Safari (also tested Firefox Nightly and Safari Technology Preview). That must be resolved before this can be merged. Commenting out the use of video-rotation-0.mp4 in video-rotation.html makes this test pass in Safari (nice work @grorg !) and run, though fail, in Firefox and Firefox Nightly (cc @jdashg ).

nihohit:
added comment. 
I'm not near a MacOS machine, so couldn't test Safari, but FF managed to run and fail.

sdk/tests/conformance/textures/misc/video-rotation.html
```diff
@@ -104,6 +104,9 @@
             gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
             gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
 
+            // These files were created by converting exif-orientation-test-1-normal.jpg to mp4
+            // files, rotating them using the transpose filter, and adding rotate metadata, all
+            // using the ffmpeg CLI.
```
kenrussell, line 109: Could you please include the complete command lines you used? If we have to recreate these resources in the future we don't want to have to start from scratch. Thanks.

kenrussell (changes requested):
Thanks for the fix to the 0-rotation file. I can confirm the test runs successfully in Safari Technology Preview now. May I please request one more update to the comments in the test?

nihohit:
done, sorry for not including this originally :)

jdashg (approved):

nihohit:
@kenrussell anything else I should fix?

kenrussell (approved):
Thanks, that's perfect! Merging now.

PR #3200, opened 2020-12-28
WEBGL_webcodecs_video_frame: Allow software video frame support

jchen10:
This lifts the restriction of importing harware video frames only.


kenrussell (approved):
Sounds OK to be able to use this extension more of the time. I do wonder whether this path will be any more efficient than the normal texImage2D path, and whether supporting software frames will add a surprising "performance cliff" to this extension that users may run into unexpectedly.

jchen10:
Thanks Ken.
There could be 2 paths for normal texImage2D. With this extension, 
1. VideoFrame->ImageBitmap->texImage2D:  for this path, VideoFrame(yuv)->ImageBItmap(rgb) copy can be saved. 
2. VideoFrame->Plane->ArrayBuffer->texImage2D: for this path, Plane->ArrayBuffer copy can be saved.

PR #3202, opened 2021-01-05
Change WEBGL_multi_draw examples for variety.

jdashg:
Try not to imply that they only work with UNSIGNED_SHORT indices.


shrekshao (approved):
LGTM. Thanks for the change

PR #3204, opened 2021-01-06
Update macOS and iOS support links

outofcontrol:
As per comments on https://github.com/KhronosGroup/WebGL/pull/3203
- Change webkit to Safari Technology Preview
- Add Chrome.com links for iOS URLs

PR #3205, opened 2021-01-12
Fix a typo in troubleshooting script

vednoc:
The author of #3204 forgot to add `,` and the script fails to execute. This patch fixes that problem.


kenrussell (approved):
Yikes. Thanks, good catch.

kenrussell:
@outofcontrol FYI.

outofcontrol:
Yes, good catch, I should know better. Completely missed the red flag in VS Code too.

PR #3206, opened 2021-01-14
Don't generate INVALID_VALUE in shaderSource.

jdashg:
Compilation requires proper charset, but only after preprocessing as per the GLSL specs.


sdk/tests/conformance/glsl/bugs/character-set.html
```diff
@@ -1,138 +1,178 @@
...
+// some comment: \n\\r\n
+void main() {
+  gl_FragColor = vec4(1, 0, 0, 1);
+}
+</script>
+
+<script id="fs-cleared-by-preprocessing-valid" type="x-shader/x-fragment">
+precision mediump float;
+#if defined(GL_GOOGLE_cpp_style_line_directive)
+#line 9 "foo.txt"
```
kenrussell, line 75: It looks like this particular case - the #line directive containing quotes - was lost in the test rewrite. Could this be added back in? (I realize it might be redundant with the `#if 0 .. "  #endif` test.)

kenrussell (commented):
Thanks for revising the test. Looks good overall; could you please address a couple of small comments? Before merging I'd like to try modifying Chromium's implementation to make sure it passes the test; will aim to do this tomorrow.

sdk/tests/conformance/glsl/bugs/character-set.html
kenrussell: Resolved in the latest update.

sdk/tests/conformance/glsl/bugs/character-set.html
```diff
@@ -1,138 +1,178 @@
...
+<meta charset="utf-8">
+<title>Character Set</title>
+<link rel="stylesheet" href="../../../resources/js-test-style.css"/>
+<script src="../../../js/js-test-pre.js"></script>
+<script src="../../../js/webgl-test-utils.js"></script>
+<script src="../../../js/glsl-conformance-test.js"></script>
+</head>
+<body>
+
+<script id="fs-ident" type="x-shader/x-fragment">
```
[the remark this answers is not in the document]
kenrussell: Resolved in the latest update.

kenrussell (approved):
Thanks @jdashg , this looks great. Let me take a couple of hours to try to confirm this is passable with revisions to Chromium's shaderSource validation before merging.

kenrussell:
Note that this is a follow-on to #3124 and #3128 .

kenrussell:
Confirmed with prototype changes to both Chromium and ANGLE under http://crbug.com/1171506 that this revised test passes. Thank you @jdashg for motivating these changes, which significantly simplify the processing of shaders' sources!

PR #3208, opened 2021-01-28
Revise invalid-passed-params test.

kenrussell:
After #3206 , the top-of-tree test must expect that shaderSource no
longer generates INVALID_VALUE. This portion of the test is removed in
the 1.0.3 and 2.0.0 snapshots to keep them passing on both old and new
browsers.


jdashg (approved):
Great, thanks!

Issue #3209, opened 2021-01-28
shader-with-define-line-continuation.frag test fails after #3206


kenrussell:
After the specification and implementation change from #3206, the test conformance/glsl/misc/shader-with-define-line-continuation.frag.html begins to fail.

ANGLE's shader preprocessor doesn't reject the use of the line continuation character in ESSL 1.00 shaders.

This test applies only to WebGL 1.0. Should we invalidate the test? Changing ANGLE's preprocessor only to pass this test sounds like a lot of unneeded effort.

PR #3212, opened 2021-01-29
Add normative text about supporting line continuation character.

kenrussell:
Follow-on to #3206 and #3209 .

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -610,6 +610,10 @@ image.src = "http://other-domain.com/image.jpg";
         a function, variable, structure name, or structure field starting with these prefixes must
         not be allowed to load.
     </p>
+    <p>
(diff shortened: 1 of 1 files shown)


jdashg (commented):
Should we just support it in general?
In Firefox's engine, we can't tell what is or isn't a preproc line easily, and I'm not sure we should bother trying. Universal `s/\\\n/ /`, with extra newlines injected after the line (for line numbers) is probably not so bad?

kenrussell:
> Should we just support it in general?
> In Firefox's engine, we can't tell what is or isn't a preproc line easily, and I'm not sure we should bother trying. Universal `s/\\\n/ /`, with extra newlines injected after the line (for line numbers) is probably not so bad?

Yes, good point - revised.

jdashg (approved):

PR #3213, opened 2021-01-30
Change attribute / uniform conflict rules.

### LINKED ISSUE #3201: In GLES, attribute names can be the same as fragment-only uniform names
See this ANGLE bug: http://anglebug.com/5497

In the GLSL ES Spec 3.00.3, section 4.3.5:

```
  If a uniform variable name is declared in one stage (e.g., a vertex shader)
  but not in another (e.g., a fragment shader), then that name is still
  available in the other stage for a different use
```

WebGL explicitly prohibits this behaviour:

https://github.com/KhronosGroup/WebGL/blob/master/sdk/tests/conformance/glsl/misc/shaders-with-name-conflicts.html#L60

```
Note: This was brought up on the OpenGL ES working group and confirmed
that the spec required these conflicts to fail to link.

Though most drivers allow this to work, some drivers to do no therefore
WebGL implementation must enforce this restriction to provide consistent
behavior.
```

Is there any link to this conversation? Seems like that is inconsistent with the language in the spec.

kenrussell:
The OpenGL ES working group has confirmed that even in the GLSL ES
1.00 specification, the intent is that the vertex and fragment shaders
have distinct namespaces. Attributes in the vertex shader do not
conflict with uniforms in the fragment shader.

Remove the now-invalid test from the 1.0.3 and 2.0.0 snapshots.

Fixes #3201.


sdk/tests/conformance/glsl/misc/shaders-with-name-conflicts.html
```diff
@@ -56,25 +60,27 @@
...
+GLSL ES 3.00 later confirmed that pipeline stages have distinct
+namespaces. The OpenGL ES working group confirmed in discussions
+https://github.com/KhronosGroup/WebGL/issues/3201 that this was the
+intended behavior for GLSL ES 1.00.
 
-Though most drivers allow this to work, some drivers to do no therefore
-WebGL implementation must enforce this restriction to provide consistent
-behavior.
+For this reason, this test asserts that such usage does not generate a
+conflict.
```
jdashg, line 74: This diff seems sorta garbled?

jdashg (changes requested):

sdk/tests/conformance/glsl/misc/shaders-with-name-conflicts.html
kenrussell: I left some history in the comment. The changes are intentional as is.

jdashg: I find this confusing to read, but I see it now.
The section is really just too big to be inlined like this. Consider

```
This test previously concluded that an attribute in a vertex shader
would conflict with a uniform of conflict across shaders the same name
in the fragment shader, based on the following sections of the GLSL
ES 1.0.17 specification:

---

Section 4.2.6
[...]
With the exception of uniform declarations, vertex and fragment shaders
have separate name spaces.  Functions and global variables declared in a
vertex shader cannot be referenced by a fragment shader and vice versa.
Uniforms have a single name space.  Uniforms declared with the same name
must have matching types and precisions.

Section 4.3.3
Attribute variables are required to have global scope

Section 4.3.4
The uniform qualifier is used to declare global variables

---

GLSL ES 3.00 later confirmed...
```

jdashg (approved):
I'm not going to block this on readability, but it would be nice if this were easier to read.

sdk/tests/conformance/glsl/misc/shaders-with-name-conflicts.html
kenrussell: Thanks, good suggestion. Have restructured it in this way.

kenrussell:
I'm positive this doesn't introduce any lint errors and we're likely blocked on some other Khronos repository's long-running CI job. Squashing and merging.

PR #3214, opened 2021-01-30
Expose bug in SH_SCALARIZE_VEC_AND_MAT_CONSTRUCTOR_ARGS workaround.

kenrussell:
Test case adapted from Shadertoy by Fabrice Neyret.

Also added a test of a shader whose compilation was previously broken
by this workaround.

Regression test for http://crbug.com/1165751 .

### DIFF
--- a/sdk/tests/conformance/glsl/bugs/00_test_list.txt
+++ b/sdk/tests/conformance/glsl/bugs/00_test_list.txt
@@ -47,5 +47,6 @@
 --min-version 1.0.4 undefined-index-should-not-crash.html
 --min-version 1.0.3 uniforms-should-not-lose-values.html
 --min-version 1.0.4 varying-arrays-should-not-be-reversed.html
+--min-version 1.0.4 vector-matrix-constructor-scalarization.html
 --min-version 1.0.4 vector-scalar-arithmetic-inside-loop.html
 --min-version 1.0.4 vector-scalar-arithmetic-inside-loop-complex.html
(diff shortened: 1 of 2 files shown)


kenrussell:
Note that on macOS this test currently fails in Chrome, Firefox and Safari. In Firefox the test can be made to pass on dual-GPU MacBook Pros by activating the discrete GPU as discussed in http://crbug.com/1165751 . An ANGLE change is needed in order to fix the bug in Safari, and more recent versions of Chrome.

kenrussell:
@jdashg  / @shrekshao : please review. Thanks.

sdk/tests/conformance/glsl/bugs/00_test_list.txt
```diff
@@ -47,5 +47,6 @@
 --min-version 1.0.4 undefined-index-should-not-crash.html
 --min-version 1.0.3 uniforms-should-not-lose-values.html
 --min-version 1.0.4 varying-arrays-should-not-be-reversed.html
+--min-version 1.0.4 vector-matrix-constructor-scalarization.html
 --min-version 1.0.4 vector-scalar-arithmetic-inside-loop.html
---min-version 1.0.4 vector-scalar-arithmetic-inside-loop-complex.html
\ No newline at end of file
+--min-version 1.0.4 vector-scalar-arithmetic-inside-loop-complex.html
```
kenrussell, line 52: Note: I think this was just adding a newline at the end of the file.

jdashg (approved):

kenrussell:
Thanks for your review @jdashg . Going ahead and merging this now; happy to make any follow-on changes later.

PR #3215, opened 2021-02-03
On shader compile error, add infolog div.

jdashg:


sdk/tests/js/glsl-conformance-test.js
```diff
@@ -132,24 +132,33 @@ function runOneTest(gl, info) {
...
   if (!vShader) {
-    vShader = wtu.loadShader(gl, vSource, gl.VERTEX_SHADER);
+    vShader = wtu.loadShader(gl, vSource, gl.VERTEX_SHADER, null, null, null, null, true);
+    let compiledVShader = vShader;
+    if (vShader && !gl.getShaderParameter(vShader, gl.COMPILE_STATUS)) {
+      compiledVShader = null;
+    }
     if (info.vShaderTest) {
-      if (!info.vShaderTest(vShader)) {
+      if (!info.vShaderTest(compiledVShader)) {
```
kenrussell, line 148: What happens if a null shader is passed to vShaderTest (and fShaderTest, below)?

sdk/tests/js/glsl-conformance-test.js
```diff
@@ -132,24 +132,33 @@ function runOneTest(gl, info) {
...
         testFailed("[vertex shader test] " + passMsg);
         return;
       }
     }
     // As per GLSL 1.0.17 10.27 we can only check for success on
     // compileShader, not failure.
-    if (!info.ignoreResults && info.vShaderSuccess && !vShader) {
+    if (!info.ignoreResults && info.vShaderSuccess && !compiledVShader) {
```
kenrussell, line 155: compiledVShader might be non-null when the shader's supposed to fail compilation because of the GLES spec's relaxed guarantees about when shader compilation succeeds - is that correct? It seems then that this might accidentally squelch test failures. Same comment about the fragment shader test below.

sdk/tests/js/glsl-conformance-test.js
```diff
@@ -132,24 +132,33 @@ function runOneTest(gl, info) {
...
-  if (!quietMode())
+  if (!quietMode()) {
     wtu.addShaderSource(consoleDiv, vLabel, vSource);
+  }
 
   // Reuse identical shaders so we test shared shader.
   var vShader = vShaderDB[vSource];
   if (!vShader) {
-    vShader = wtu.loadShader(gl, vSource, gl.VERTEX_SHADER);
+    vShader = wtu.loadShader(gl, vSource, gl.VERTEX_SHADER, null, null, null, null, true);
```
kenrussell, line 142: Ah, I understand the difference now. Could you add a comment here and for the fragment shaders below "explicitly skip the internal compile status check"?

kenrussell (approved):

sdk/tests/js/glsl-conformance-test.js
jdashg: See above! Basically this patch is first s/vShader/compiledVShader/, and then fixups

PR #3216, opened 2021-02-10
Increase tolerance to 8 from 3.

kenrussell:
Found necessary on multiple platforms in http://crbug.com/1165751 .

Follow-up to #3214 .


kenrussell:
@jdashg or @shrekshao : please review. Thanks.

kenrussell:
Or @kainino0x ? Thanks.

shrekshao (approved):

kenrussell:
Thanks @shrekshao for your review. I don't think this is controversial - merging now.

kenrussell:
BTW, slight differences in the rendered output between these two different mathematical expressions - which use constants / variables differently - have appeared on multiple platforms and GPUs.

jdashg (commented):
This is probably fine though I would have preferred a working hypothesis for why.

kenrussell:
I'm pretty sure it's due to differing optimizations being done by the driver's shader compiler in the two code paths.

PR #3217, opened 2021-02-11
Extra test for attribute divisor reset to 0 in angle-instanced-arrays.html

shrekshao:
Related to workaround for latest Mac Intel opengl driver bug http://crbug.com/1144207

Test description: we want to make sure webgl does instanced draw correctly when the non-zero divisor of an attribute is reset to zero. The attribute with the new divisor == 0 should be treated as a regular vertex attribute.

The test turns divisor of color attribute to zero. with `drawArraysInstancedANGLE(gl.TRIANGLES, 3, 3, instanceCount);`, it's the last three yellow vec4 that would be accessed by vertices. The test should draw 4 yellow triangles.

Tested locally on Mac AMD (passes chrome/firefox, patched chrome), Mac Intel (fails chrome/firefox, passes patched chrome)

### DIFF
--- a/sdk/tests/conformance/extensions/angle-instanced-arrays.html
+++ b/sdk/tests/conformance/extensions/angle-instanced-arrays.html
@@ -196,6 +196,9 @@ function runOutputTests() {
         0.0, 1.0, 0.0, 1.0, // Green
         0.0, 0.0, 1.0, 1.0, // Blue
         1.0, 1.0, 0.0, 1.0, // Yellow
+        // extra data when colorLoc divisor is set back to 0
+        1.0, 1.0, 0.0, 1.0, // Yellow
+        1.0, 1.0, 0.0, 1.0, // Yellow
     ]);
     var colorBuffer = gl.createBuffer();
     gl.bindBuffer(gl.ARRAY_BUFFER, colorBuffer);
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance/extensions/angle-instanced-arrays.html
```diff
@@ -277,6 +280,17 @@
     wtu.checkCanvasRect(gl, Math.ceil(0.25*canvas.width), 0, w, h, [0, 0, 255, 255]);
     wtu.checkCanvasRect(gl, Math.ceil(0.75*canvas.width), 0, w, h, [255, 255, 0, 255]);
 
+    debug("");
+    debug("Testing drawArraysInstancedANGLE with attributes 'divisor' reset to 0");
+    ext.vertexAttribDivisorANGLE(colorLoc, 0);
```
kenrussell, line 285: Could you please add another debug() call saying what the correct output is, and what the incorrect rendering looked like when this driver bug was occurring?

kenrussell (approved):
Great job Shrek working around this fundamental driver bug. New test section looks good; one small request. Please feel free to merge yourself afterward.

kenrussell (commented):
One more thought - could you add similar code to conformance2/rendering/instanced-arrays.html ?

shrekshao:
Updated similiar section for webgl2 instanced-arrays.html

sdk/tests/conformance2/rendering/instanced-arrays.html
```diff
@@ -174,6 +177,45 @@
     gl.drawArraysInstanced(desktopGL['POLYGON'], 0, 6, instanceCount);
     wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "drawArraysInstanced with POLYGON should return INVALID_ENUM");
 
+    debug("Testing drawArraysInstancedANGLE with param 'first' > 0");
```
kenrussell, line 180: drawArraysInstancedANGLE -> drawArraysInstanced (here and throughout)

kenrussell (approved):
Thanks for focusing so well on this problem and the test Shrek - great work. One tiny comment.

kenrussell:
By the way - if you can think of a refactoring into a separate .js file that would allow these two tests to share more code, that would be great.

shrekshao:
I think we can defer merge into one js to a later pull request

PR #3218, opened 2021-02-12
Fix negative compilation tests in glsl-conformance-test.js.

kenrussell:
Addresses regressions in the following tests:
  conformance/glsl/misc/boolean_precision.html
  conformance2/glsl3/switch-case.html
  conformance2/glsl3/texture-offset-non-constant-offset.html
  conformance2/glsl3/uniform-block-layouts.html

Follow-up to #3215 .

### DIFF
--- a/sdk/tests/js/glsl-conformance-test.js
+++ b/sdk/tests/js/glsl-conformance-test.js
@@ -162,8 +162,10 @@ function runOneTest(gl, info) {
       }
     }
     // Save the shaders so we test shared shader.
-    if (vShader) {
-      vShaderDB[vSource] = vShader;
+    if (compiledVShader) {
+      vShaderDB[vSource] = compiledVShader;
+    } else {
+      vShader = null;
     }
   }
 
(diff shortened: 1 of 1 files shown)


kenrussell:
Blocking roll of WebGL conformance suite into Chromium under http://crbug.com/1165751 / https://chromium-review.googlesource.com/2690219 . Taking the liberty of merging myself.

jdashg:
Hey so I understand that because of how it's currently implemented, you end up in situations like this, but this PR should really not block that roll. You need a way to choose non-tip for import.

sdk/tests/js/glsl-conformance-test.js
```diff
@@ -162,8 +162,10 @@ function runOneTest(gl, info) {
       }
     }
     // Save the shaders so we test shared shader.
-    if (vShader) {
-      vShaderDB[vSource] = vShader;
+    if (compiledVShader) {
+      vShaderDB[vSource] = compiledVShader;
+    } else {
+      vShader = null;
```
jdashg, line 168: Use compiledVShader wherever vShader is used below instead. The meaning of variables should avoid changing over the course of a function.

kenrussell: compiledVShader and compiledFShader are scoped within their blocks and are inaccessible to the code below. Feel free to refactor further as long as the tests continue to pass.

PR #3224, opened 2021-03-04
Replace non-ascii chars where unnecessary or accidental in extensions/*.

jdashg:
Some of these aren't even valid utf-8, and python's open(foo,'r').read()
raises UnicodeDecodeErrors.


kenrussell (approved):
Thanks for catching these!

jdashg:
Actually these might have been utf-8, and python's open() might have been trying to decode into ascii or latin1 instead.
Regardless, we didn't need these!

PR #3225, opened 2021-03-05
Fix html encodings

jdashg:
More non-utf8 files in conformance-suites snapshots.
These make it trickier to operate on these snapshots, even though we shouldn't really be touching them much.
This is minimally invasive, and should be the correct equivalent to what we had before, so I'd like to take it against the snapshots anyway.

@kenrussell thoughts?


conformance-suites/1.0.1/conformance/glsl/misc/non-ascii.vert.html
```diff
@@ -20,9 +20,9 @@
 <script id="vertexShader" type="text/something-not-javascript">
 // Non ascii data in source should fail
 // See GLSL ES Spec 1.0.17 section 3.1 and 3.2
-// ����͂`�r�b�h�h�ł͂Ȃ��ł��B
-// �s�������@�h���@�m�����@�`�r�b�h�h
-uniform mat4 �m�����`�r�b�h�h;
+// ‚±‚ê‚Í‚`‚r‚b‚h‚h‚Å‚Í‚È‚¢‚Å‚·B
+// ‚s‚ˆ‚‰‚“@‚h‚“@‚m‚‚”@‚`‚r‚b‚h‚h
+uniform mat4 ‚m‚‚”‚`‚r‚b‚h‚h;
```
kenrussell, line 25: What exactly was the conversion done to these files? (I'm not an expert on file encodings.) Were they UTF-16 before and have been converted to UTF-8 now? Are we sure that the characters are interpreted the same way, for example when this shader is loaded? I see the meta tag above indicates the character set is UTF-8 so was the browser just being permissive before in loading UTF-16 characters?

jdashg: Oh yeah maybe? I can see here that it says "NotASCII" with weird codepoints, so that comes from *somewhere*!

PR #3227, opened 2021-03-05
s/void/undefined/ in specs/latest/*.

jdashg:
This is the bulk of the changes we need to modernize our webidl, so this gets us most of the way there.


jdashg:
(I only did 1.0 and 2.0, not 2.0-compute)

toji (approved):
LGTM. This is a really straightforward search-and-replace style change to bring the spec up to date with current WebIDL syntax and introduces no behavioral changes. 👍

kenrussell (approved):
Looks perfect!

PR #3257, opened 2021-03-11
Replace VideoTrackReader

tguilbert-google:
VideoTrackReader is deprecated (see crbug.com/1157610), and it needs be replaced by the equivalent MediaStreamTrackProcessor.

### DIFF
--- a/extensions/proposals/WEBGL_webcodecs_video_frame/webgl_webcodecs_video_frame.js
+++ b/extensions/proposals/WEBGL_webcodecs_video_frame/webgl_webcodecs_video_frame.js
@@ -329,7 +329,7 @@ function requestWebGLVideoFrameHandler(canvas) {
         // Immediately schedule rendering of the next frame
         setTimeout(renderFrame, 0);
         ext.releaseVideoFrame(videoFrameHandle);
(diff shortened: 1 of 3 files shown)


kenrussell:
Yes, in Chrome you need either the `--enable-webgl-draft-extensions` command line flag or to set that in `about:flags`.

Could you please try that and verify these changes? Please comment here when you've verified the test. Thanks.

tguilbert-google:
I've confirmed that they work with the latest commit.

However, performance/index.html freezes after a while (~200-~400 frames). This might be a chromium issue. I will take a look tomorrow.

tguilbert-google:
FWIW, VideoTrackReader's removal has just landed [1], so it might be worth landing the test changes regardless of performance/index.html's stall?

[1] https://chromium.googlesource.com/chromium/src/+/8f0d7359a21af5216a8ace69a4a300be4c7de58c

kenrussell (approved):
Thanks for verifying these changes. Yes, let's land these and follow up on the hang of the performance test.

tguilbert-google:
Sounds good. Thanks!

PR #3258, opened 2021-03-12
Add simple NPOT tests to S3TC

### LINKED ISSUE #3117: Relax implicit power-of-two restrictions for mipmapped textures in WEBGL_compressed_texture_s3tc
Context:
https://github.com/BinomialLLC/basis_universal/issues/154
https://github.com/KhronosGroup/KTX-Software/issues/254
https://github.com/KhronosGroup/KTX-Specification/issues/123

https://bugs.chromium.org/p/angleproject/issues/detail?id=4841
https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_s3tc/
https://www.khronos.org/registry/OpenGL/extensions/ANGLE/ANGLE_texture_compression_dxt.txt

It appears that longstanding restrictions in the ANGLE_texture_compression_dxt extension, which were lifted into WebGL's WEBGL_compressed_texture_s3tc extension, have had the following consequences:

1) Because all texture levels beyond those of width/height 1 or 2 have to have width/height divisible by 4, this implies that all levels have to be power-of-two sized.

2) Because this restriction exists for S3TC, and because this is one possible transcoding target of Basis Universal textures, this implies that all textures processed with this pipeline have to obey the power-of-two restriction in case they are transcoded at runtime to S3TC.

The WebGL WG must figure out a solution to this problem. https://bugs.chromium.org/p/angleproject/issues/detail?id=4841
 tracks investigation at the ANGLE layer, which will inform a specification-level solution.

lexaknyazev:
Fixes #3117.

The updated tests pass on Chrome Stable .

Related to http://crbug.com/angleproject/4841 .

/cc @kenrussell @vonture

### DIFF
--- a/sdk/tests/conformance/extensions/ext-texture-compression-bptc.html
+++ b/sdk/tests/conformance/extensions/ext-texture-compression-bptc.html
@@ -60,7 +60,7 @@ function runTestExtension() {
   // * <height> is not a multiple of four, and <height> plus <yoffset> is
   //           not equal to TEXTURE_HEIGHT; or
   // * <xoffset> or <yoffset> is not a multiple of four.
-  ctu.testTexSubImageDimensions(gl, validFormats, expectedByteLength, getBlockDimensions,
+  ctu.testTexSubImageDimensions(gl, ext, validFormats, expectedByteLength, getBlockDimensions,
     16, 16, [
       { xoffset: 0, yoffset: 0, width: 4, height: 3,
         expectation: gl.INVALID_OPERATION, message: "height is not a multiple of 4" },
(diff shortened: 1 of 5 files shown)


kenrussell:
Thanks very much Alexey for these contributions. Would like to ask other working group members to help review and test on more browsers / platforms.

Please note I slightly edited the issue description above; when squashing & merging please make sure the latest version of the text is included in the commit message. Thanks.

kenrussell (approved):
Apologies again @lexaknyazev for the long delay reviewing.

Tested this locally on macOS and all tests pass both with and without ANGLE as the WebGL backend; however this platform doesn't support bptc so I can't verify the changes to that test.

@jdarpinian would you be able to test locally on Windows?

jdarpinian (approved):
The BPTC test passes for me on Windows.

kenrussell:
Super. Thanks James for testing. @lexaknyazev thanks again for this contribution. Merging now.

PR #3259, opened 2021-03-12
Add video element playback error detection, use it to fail test.

dalecurtis:
Some tests will silently hang when playback errors occur, so add
an error listener so the failure reason is more obvious.

### DIFF
--- a/conformance-suites/1.0.3/conformance/resources/webgl-test-utils.js
+++ b/conformance-suites/1.0.3/conformance/resources/webgl-test-utils.js
@@ -2594,6 +2594,15 @@ var runSteps = function(steps) {
  *        video is ready.
  */
 var startPlayingAndWaitForVideo = function(video, callback) {
+  if (video.error) {
+    testFailed("Video playback failed: " + e.message);
+    return;
+  }
+
+  video.addEventListener(
+      'error', e => { testFailed("Video playback failed: " + e.message); },
+      true);
+
   var rvfc = getRequestVidFrameCallback();
   if (rvfc === undefined) {
     var gotPlaying = false;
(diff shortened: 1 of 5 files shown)


jdarpinian (approved):
LGTM

jdashg (changes requested):
Please test if `try {await video.play() } catch { testFailed }` works instead.

dalecurtis:
@jdashg are you suggesting that in place of the "if (video.error) {}" block? That seems fine for demuxer and first frame errors, but it can't replace the addEventListener() for decode errors which occur later. 

One side note is that 'await' may change the timing flow here, but seems fine to try if nothing explodes.

jdashg:
Did you test this? I don't expect await to work in a non-`async function` context

conformance-suites/1.0.3/conformance/resources/webgl-test-utils.js
```diff
@@ -2641,7 +2636,11 @@ var startPlayingAndWaitForVideo = function(video, callback) {
 
   video.loop = true;
   video.muted = true;
-  video.play();
+  try {
+    await video.play();
+  } catch (error) {
+    testFailed('Video failed to play(): ' + error);
+  }
```
jdashg, line 2643: Let's not backport this, since it's a little risky, and we're just doing it to try to make things better for ourselves on tip

dalecurtis: Sorry I tested the first patch, but the not the second one. Yes it needs async function. I'll use async on tip and then restore to my previous if (video.error) check on earlier versions.

dalecurtis:
@jdashg any further comments?

kenrussell (approved):
For what it's worth this looks good to me too.

jdashg (approved):

jdashg:
Nope, thanks!

PR #3262, opened 2021-03-17
Increase video-rotation test tolerance, add vp9 version.

dalecurtis:
Speed focused software yuv conversion libraries may be off by more
than 5 points in a channel for values near the channel maximum.

There's also no reason we can't have a vp9 version of this test, so
add encodes for clients who want to test w/o proprietary codecs.


sdk/tests/conformance/textures/misc/video-rotation.html
```diff
@@ -89,11 +89,15 @@
                 return;
             }
 
-            const type = 'video/mp4; codecs="avc1.42E01E, mp4a.40.2"';
-            if (!video.canPlayType(type).replace(/no/, '')) {
-                debug(type + " unsupported");
-                return;
-            };
+            let prefer_vp9 = false;
```
kenrussell, line 92: Please have the test test each video format if it's supported. Thanks.

kenrussell (changes requested):
Thanks for this update. One comment then lgtm.

sdk/tests/conformance/textures/misc/video-rotation.html
dalecurtis: Done.

sdk/tests/conformance/textures/misc/video-rotation.html
```diff
@@ -141,9 +138,14 @@
...
-                fn += (prefer_vp9 ? ".vp9" : "" ) + ".mp4";
-                await testVideoElement(fn);
+            if (supports_h264) {
+                for (let fn of filenames)
+                    await testVideoElement(fn + ".mp4");
+            }
+
+            if (supports_vp9) {
+                for (let fn of filenames)
+                    await testVideoElement(fn + ".vp9.mp4");
```
kenrussell, line 148: .vp9?

kenrussell (changes requested):
Thanks very much for the update. One more typo?

sdk/tests/conformance/textures/misc/video-rotation.html
dalecurtis: No, this looks right? Sorry if I'm blind.

kenrussell: Ah ok, sorry, I didn't look carefully and didn't realize these were still in an .mp4 container.

kenrussell (approved):
Sorry for my confusion! Thanks, looks good!

PR #3263, opened 2021-03-19
Don't use await for video.play() promise.

dalecurtis:
Some tests are finishing as soon as the first frame has come in
and triggering a pause() or other condition which thorws a
'AssertionError: Video failed to play(): AbortError: The play()
request was interrupted by a call to pause().' error.

So just check for no sources and use the error handler like we
do on older WebGL test versions.

Addresses new failures seen while rolling forward WebGL
conformance in Chromium:
https://chromium-review.googlesource.com/2775887/
for http://crbug.com/1136205 .


kenrussell (approved):
Thanks for the fix Dale. LGTM.

Edited the description to reference the Chromium-side conformance roll-forward and bug.

PR #3264, opened 2021-03-24
Increase tolerance for video-rotation.html test.

dalecurtis:
Follow-on to #3199 and #3262 .

Associated with http://crbug.com/1191456 .


kenrussell (approved):
Thanks, looks good.

PR #3265, opened 2021-03-25
Fix whitespace by rerunning IDL extraction scripts.

kenrussell:
Somehow many changes were introduced recently that should be fixed
separately from other updates.


kenrussell:
Whitespace changes only that should be done before other pull requests are reviewed. Merging myself.

kenrussell:
cc @jdashg @lexaknyazev as FYI.

PR #3260, opened 2021-03-15
[webgl1] Align with Web IDL specification

autokagami:
This is an automated pull request to align the spec with the latest Web IDL specification.

The following is the Web IDL validation message, which may help understanding this PR:

```
Validation error at line 2 in webgl1,13, inside `interface WebGLContextEvent`:
 Constructor(DOMString type,
 ^ Constructors should now be represented as a `constructor()` operation on the interface instead of `[Constructor]` extended attribute. Refer to the [WebIDL spec section on constructor operations](https://heycam.github.io/webidl/#idl-constructors) for more information.
```

Currently this autofix might introduce awkward code formatting, and feel free to manually fix it whenever it happens.

Please file an issue at https://github.com/saschanaz/webidl-updater/issues/new if you think this PR is invalid or should be enhanced.

PR #3266, opened 2021-03-25
Align WebGLContextEvent with Web IDL spec.

kenrussell:
Proposed by @autokagami in #3260 .


jdashg (approved):

jdashg:
I think you need to --reset-authors the commit from the bot

kenrussell:
Thanks @jdashg for your review and suggestion.

PR #3267, opened 2021-03-26
Add getInternalformatParameter for EXT_texture_norm16

shrekshao:
Add a regression test as chrome is missing return for formats added in `EXT_texture_norm16` in `getInternalformatParameter`
`R16`, `RG16`, `RGBA16` should be all multisample-support

### DIFF
--- a/sdk/tests/conformance2/extensions/ext-texture-norm16.html
+++ b/sdk/tests/conformance2/extensions/ext-texture-norm16.html
@@ -153,6 +153,23 @@ function testSnorm16Unrenderable(internalFormat, format, type) {
                                 "framebuffer should not be complete with SNORM16 texture attached");
 }
 
+function runInternalFormatQueryTest()
+{
+    debug("");
+    debug("testing the internal format query");
+
+    var maxSamples = gl.getParameter(gl.MAX_SAMPLES);
+    var formats = new Array(ext.R16_EXT, ext.RG16_EXT, ext.RGBA16_EXT);
+    for (var fmt = 0; fmt < formats.length; ++fmt) {
+        var samples = gl.getInternalformatParameter(gl.RENDERBUFFER, formats[fmt], gl.SAMPLES);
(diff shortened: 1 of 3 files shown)


sdk/tests/conformance2/extensions/ext-texture-norm16.html
```diff
@@ -153,6 +153,23 @@
                                 "framebuffer should not be complete with SNORM16 texture attached");
 }
 
+function runInternalFormatQueryTest()
+{
+    debug("");
+    debug("testing the internal format query");
+
+    var maxSamples = gl.getParameter(gl.MAX_SAMPLES);
+    var formats = new Array(ext.R16_EXT, ext.RG16_EXT, ext.RGBA16_EXT);
```
jdashg, line 162: Was this copied from somewhere?
`const formats = [ext.R16_EXT, ext.RG16_EXT, ext.RGBA16_EXT];`

sdk/tests/conformance2/extensions/ext-texture-norm16.html
```diff
@@ -153,6 +153,23 @@
...
+    for (var fmt = 0; fmt < formats.length; ++fmt) {
```
jdashg, line 163: `for (const format of formats) {`

jdashg (approved):

sdk/tests/conformance2/extensions/ext-texture-norm16.html
shrekshao: Yes. Fixed the copy src with the same style

sdk/tests/conformance2/extensions/ext-texture-norm16.html
shrekshao: only fixed this for loop style. The copy src is using idx for some logic so left unchanged

jdashg:
Thanks!

kenrussell (commented):
LGTM after the fact. Nice work extending this test.

PR #3269, opened 2021-04-01
Allow either bad-size=>INVALID_OP or bad-stride=>INVALID_VAL for vert…

jdashg:
…exAttribPointer.

Bad size is specified as INVALID_OPERATION here:
https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.14.10

Too-large stride is specified as INVALID_VALUE here:
https://www.khronos.org/registry/webgl/specs/latest/1.0/#6.13

### DIFF
--- a/sdk/tests/conformance/attribs/gl-vertexattribpointer.html
+++ b/sdk/tests/conformance/attribs/gl-vertexattribpointer.html
@@ -152,7 +152,7 @@ if (!gl) {
               gl, size < info.minSize ? gl.INVALID_OPERATION : gl.NO_ERROR, "at stride limit",
               size, info.type, false, stride, offset);
           checkVertexAttribPointer(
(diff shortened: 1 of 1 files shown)


kenrussell (approved):
Sorry for the delay reviewing; had to revisit all the associated specs.

Just to clarify my understanding, is the problem with this particular test that both the size and stride are illegal, so it's ambiguous which one's validation causes the error (and the error differs depending on which one failed validation)?

jdashg:
That's right, though even this subtest's stated goal is to test too-large stride, which we spec'd in webgl to be INVALID_VALUE.

kenrussell:
I see - thanks for clarifying. Had a hard time finding again the too-large stride spec. Adjusted your message above to point to the two spec sections. Merging now.

PR #3270, opened 2021-04-01
Modify LICENSE.TXT

Some1NamedNate:
Add copyright year and holder.


kenrussell (approved):
Thanks. Looking at https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/licensing-a-repository it does look like this has been missing.

PR #3272, opened 2021-04-02
In EXT_disjoint_timer_query examples, delete query objects.

kenrussell:
Follows best practices for WebGL objects. Also modernize the
JavaScript in these examples (var -> let).

### DIFF
--- a/extensions/EXT_disjoint_timer_query/extension.xml
+++ b/extensions/EXT_disjoint_timer_query/extension.xml
@@ -224,6 +224,7 @@ interface EXT_disjoint_timer_query {
         ext.endQueryEXT(ext.TIME_ELAPSED_EXT);
 
         // ...at some point in the future, after returning control to the browser and being called again:
+        // (Note that this code might be called multiple times)
         if (query) {
           let available = ext.getQueryObjectEXT(query, ext.QUERY_RESULT_AVAILABLE_EXT);
           let disjoint = gl.getParameter(ext.GPU_DISJOINT_EXT);
@@ -260,6 +261,7 @@ interface EXT_disjoint_timer_query {
         ext.queryCounterEXT(endQuery, ext.TIMESTAMP_EXT);
 
         // ...at some point in the future, after returning control to the browser and being called again:
+        // (Note that this code might be called multiple times)
         if (startQuery) {
           let available = ext.getQueryObjectEXT(endQuery, ext.QUERY_RESULT_AVAILABLE_EXT);
           let disjoint = gl.getParameter(ext.GPU_DISJOINT_EXT);
@@ -324,6 +326,7 @@ interface EXT_disjoint_timer_query {
         }
 
         // ...at some point in the future, after returning control to the browser and being called again:
+        // (Note that this code might be called multiple times)
         if (startQuery || endQuery || timeElapsedQuery) {
           let disjoint = gl.getParameter(ext.GPU_DISJOINT_EXT);
           let available;
--- a/extensions/EXT_disjoint_timer_query/extension.xml
+++ b/extensions/EXT_disjoint_timer_query/extension.xml
@@ -224,21 +224,26 @@ interface EXT_disjoint_timer_query {
         ext.endQueryEXT(ext.TIME_ELAPSED_EXT);
 
         // ...at some point in the future, after returning control to the browser and being called again:
-        let available = ext.getQueryObjectEXT(query, ext.QUERY_RESULT_AVAILABLE_EXT);
-        let disjoint = gl.getParameter(ext.GPU_DISJOINT_EXT);
+        if (query) {
+          let available = ext.getQueryObjectEXT(query, ext.QUERY_RESULT_AVAILABLE_EXT);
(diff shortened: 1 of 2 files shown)


kenrussell:
@jdashg or @jdarpinian or @kainino0x please review - thanks.

extensions/EXT_disjoint_timer_query/extension.xml
```diff
@@ -224,25 +224,29 @@ interface EXT_disjoint_timer_query {
...
-          var timeElapsed = ext.getQueryObjectEXT(query, ext.QUERY_RESULT_EXT);
+          let timeElapsed = ext.getQueryObjectEXT(query, ext.QUERY_RESULT_EXT);
 
           // Do something useful with the time.  Note that care should be
           // taken to use all significant bits of the result, not just the
           // least significant 32 bits.
           adjustObjectLODBasedOnDrawTime(timeElapsed);
+
+          // Clean up the query object.
+          ext.deleteQueryEXT(query);
```
kainino0x, line 240: `createQueryEXT` was called regardless of this conditional, so shouldn't it be deleted unconditionally?

Same everywhere else.

extensions/EXT_disjoint_timer_query/extension.xml
```diff
@@ -214,8 +214,8 @@ interface EXT_disjoint_timer_query {
   <samplecode xml:space="preserve">
     <pre>
         // Example (1) -- uses beginQueryEXT/endQueryEXT.
-        var ext = gl.getExtension('EXT_disjoint_timer_query');
-        var query = ext.createQueryEXT();
+        let ext = gl.getExtension('EXT_disjoint_timer_query');
```
kainino0x, line 217: I think most of these `let`s could be `const`s instead (though this is a less important improvement than `var` &rarr; `let`).

extensions/EXT_disjoint_timer_query/extension.xml
```diff
@@ -303,23 +313,23 @@ interface EXT_disjoint_timer_query {
...
         } else {
-          var available;
+          let available;
           if (useTimestamps) {
             available = ext.getQueryObjectEXT(endQuery, ext.QUERY_RESULT_AVAILABLE_EXT);
           } else {
             available = ext.getQueryObjectEXT(timeElapsedQuery, ext.QUERY_RESULT_AVAILABLE_EXT);
           }
-
+1
```
kainino0x, line 326: accidental?

extensions/EXT_disjoint_timer_query/extension.xml
kenrussell: Looking at the example code (it's hard to do this without merging in these changes and building the extensions), it's at this point, inside this if-clause, that the query object's finally resolved, and the application's done with it. I think it would be confusing to put the deletion of the object anywhere else, since this code's basically run in a polling loop.

extensions/EXT_disjoint_timer_query/extension.xml
kenrussell: Yes, probably - I looked at doing that but didn't want to make mistakes and also didn't really have the time to copy out the example code and double-check my work.

extensions/EXT_disjoint_timer_query/extension.xml
kenrussell: Oops, yes - thanks for catching that. Fixed.

kenrussell:
@kainino0x thanks for your review. Could you please take another look and see whether you agree with the comments above or would like to see code changes in response? Thanks.

extensions/EXT_disjoint_timer_query/extension.xml
kenrussell: Ah, hang on - I see what you mean. Yes, let me revise this.

kenrussell: Code substantially reworked in the latest commit. Please re-review. Thanks.

extensions/EXT_disjoint_timer_query/extension.xml
kenrussell: At this point the query variables are all mutable, so can't be const. The extension objects could be. Let me know if you'd prefer I make that change throughout.

kainino0x: I don't think it's important.

extensions/EXT_disjoint_timer_query/extension.xml
```diff
@@ -224,25 +224,34 @@ interface EXT_disjoint_timer_query {
         ext.endQueryEXT(ext.TIME_ELAPSED_EXT);
 
         // ...at some point in the future, after returning control to the browser and being called again:
```
kainino0x, line 226: ah, I wasn't clear that this code would be called multiple times - perhaps just mention that in this comment

kainino0x (approved):

extensions/EXT_disjoint_timer_query/extension.xml
kenrussell: Good idea. Done throughout.

kenrussell:
Thanks for your reviews!

PR #3273, opened 2021-04-02
Split up deqp/functional/gles3/multisample.html .

### LINKED ISSUE #3271: Split up slow deqp/functional/gles3/multisample.html test
The WebGL 2.0 conformance test deqp/functional/gles3/multisample.html takes a very long time to run. In the past it's caused timeouts in the automated test infrastructure of some browsers, and is doing so again in Chromium; see http://crbug.com/1114780 .

The test should be split up into smaller pieces so each one can run in a more reasonable time period.

kenrussell:
This test is so large that it times out on some browsers' CI systems.
Split it up into its functional pieces.

Associated with Chromium bug http://crbug.com/1114780 .

Fixes #3271 .


kenrussell:
@jdarpinian or @jdashg or @kainino0x : please review.

(Please let me squash and merge since I got the commit message wrong and it'll need editing upon merging - thanks.)

Verified manually that the same tests are run. Locally, this decreases the total run time of all of these tests by 25%. Part of the problem looks like the O(n^2) addition of lots of DOM elements to the end of the page.

sdk/tests/deqp/functional/gles3/multisample/00_test_list.txt
```diff
@@ -0,0 +1,4 @@
+default_fbo.html
```
kainino0x, line 1: nit: Autogenerated file warning, if comments are allowed in 00_test_list.txt

sdk/tests/deqp/functional/gles3/multisample/multisample_test_generator.py
```diff
@@ -0,0 +1,92 @@
...
+var wtu = WebGLTestUtils;
+var gl = wtu.create3DContext('canvas', null, 2);
+
+functional.gles3.es3fMultisampleTests.run(gl, [%(start)s, %(end)s]);
+</script>
+</body>
+</html>
+"""
+
+_GROUPS = [
```
kainino0x, line 49: nit: Note that this must match CaseType is es3fMultisampleTests

kainino0x (approved):
Awesome this gives a net improvement too! Obvious in retrospect (given I've watched this test run and it slows down over time for that reason), but awesome.

jdarpinian (approved):
LGTM

sdk/tests/deqp/functional/gles3/multisample/00_test_list.txt
kenrussell: Thanks, added.

sdk/tests/deqp/functional/gles3/multisample/multisample_test_generator.py
kenrussell: Thanks, added.

kenrussell:
Thanks for your reviews!

PR #3274, opened 2021-04-06
Migrate Travis CI to Github Actions.

kenrussell:
Khronos' Travis CI instance often blocks on other repositories'
long-running jobs. WebGL's only takes a few seconds, so migrate it to
Github Actions.

This migration has not been tested yet. Some adjustment will likely be
necessary.


kenrussell:
@jdarpinian or @jdashg : please take a look.

If this seems OK I'll merge and debug it in the workspace.

@outofcontrol FYI.

.travis.yml
```diff
@@ -1,14 +0,0 @@
...
-sudo: required
-
-language: python
-
-python:
-  - "3.7"
-
-before_script:
-  - sudo apt-get install libxml2-utils
-  - pip install --upgrade setuptools
```
jdarpinian, line 11: I don't see this in the new script. Is it necessary?

jdarpinian:
Is it not possible to debug this in the pull request? I haven't used Github Actions before. Seems like the Checks tab should show the new check.

.travis.yml
kenrussell: Good question; I thought setuptools would be sufficiently up-to-date but perhaps not. Have added this to lint.yaml.

kenrussell:
> Is it not possible to debug this in the pull request? I haven't used Github Actions before. Seems like the Checks tab should show the new check.

I'm not sure - haven't used Github Actions before either. The Checks tab is empty for this pull request, though the Travis invocation is gone - which was part of the intent.

My assumption was that this would have to be merged and then temporary PRs uploaded in order to test and debug it.

jdarpinian (approved):
I'm surprised if it requires merging this to master just to test it. There may be something wrong with the yaml file. But we can try and see what happens.

jdashg (approved):

PR #3278, opened 2021-04-14
Remove gl_BaseVertex/gl_BaseInstance from WEBGL_draw_instanced_base_vertex_base_instance.

jdashg:
Update test.
Expand test with easy-to-debug 1x1 expected value tests.

I found these to be extra work in Gecko's implementation, not all
applications need them, and I needed further changes to ANGLE's shader compiler.
Additionally, they complicated the testing matrix.

Arguably, we should now pivot the extension from
ANGLE_draw_instanced_base_vertex_base_instance to EXT_base_instance.

In GL, these GLSL builtin vars are mediated by a different extension
(ARB_shader_draw_parameters) and we could go that direction for WebGL as
well.


jdashg:
If we decide not to go this way, I can still extract my new test additions, but I think it's simpler this way.

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
```diff
@@ -269,9 +255,248 @@
...
+    gl.useProgram(vertid_prog);
+    if (!desc.draw(desc)) {
+      //debug('Not supported for draw func, skipping...');
+      continue;
+    }
+    debug('\ndesc: ' + JSON.stringify(desc));
+
+    wtu.glErrorAssert(gl, 0);
+    if (!desc.vert_count || !desc.inst_count) {
+      expect_pixel(desc, 'vertid_prog', [255, 0, 0, 255]);
```
shrekshao, line 486: Q: when instanceCount is 0, is it a no-op? (testing on windows chrome this fails and I remember once read somewhere D3D treats instanceCount =0 and 1 the same? But likely I did it wrong)

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
```diff
@@ -269,9 +255,248 @@
...
+    })();
+
+  const instid_prog = (() => {
+      const vert_src = `\
+#version 300 es
+#line 425
+layout(location = 0) in int a_vertex_id; // Same as gl_VertexID
+layout(location = 1) in int a_instance_div1; // Subtly not gl_InstanceID
+layout(location = 2) in int a_instance_div2; // Subtly not gl_InstanceID/2
+layout(location = 3) in int a_instance_div3; // Subtly not gl_InstanceID/2
```
shrekshao, line 335: gl_InstanceID/3 based on later code

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
```diff
@@ -269,9 +255,248 @@
...
+  gl.disable(gl.STENCIL_TEST);
+  gl.disable(gl.BLEND);
+
+  for (const desc of TEST_DESCS) {
+    gl.disable(gl.SCISSOR_TEST);
+    gl.clearBufferfv(gl.COLOR, 0, [1,0,0,1]);
+
+    const last_vert_id = desc.base_vert + desc.vert_count - 1;
+    const last_inst_id = desc.inst_count - 1;
+    const last_inst = desc.inst_count - 1;
```
shrekshao, line 472: nit: `last_inst_id` and `last_inst` being the same but not changed?

shrekshao (commented):
Thanks for the update.

Will you draft a proposal for WebGL equivalent of `ARB_shader_draw_parameters` that brings in shader variable `gl_BaseVertex`, `gl_BaseInstance` (say `WEBGL_base_vertex_base_instance_parameters`)? Then if we go this path, it will be easier on chrome side to automatically enable this new shader param extension together with `WEBGL_[multi]_draw_instanced_base_vertex_base_instance` instead of some big implementation to unbind the shader params.

CC'ing @kenrussell for inputs on extension decision.

kenrussell (commented):
Thanks for your thought on these extensions and tests.

Could you please provide more details about:
- The extra work involved in supporting these shader variables
- What additional changes to ANGLE's shader compiler were needed
- Why the testing matrix was complicated by the addition of these shader variables

The ANGLE_base_vertex_base_instance extension was designed to include these, via emulation if required, on all platforms, as a developer convenience. Emulating them isn't (or shouldn't be, as I understand it) difficult. Removing part of that extension's functionality implies that the extension really needs to be split into multiple ones, because it's not feasible to disable just part of it in WebGL implementations based on ANGLE.

We deliberately subsetted the entry points in ANGLE_base_vertex_base_instance rather than doing a parallel of EXT_base_instance to reduce the combinatorial explosion of entry points. In particular, DrawElementsInstancedBaseInstanceEXT, and the multi-draw variant thereof, was removed.

cc @vonture who reviewed this extension for any thoughts.

jdashg:
Sure!
I started implementing this as specified, but I couldn't activate the extension without enabling the fake-gl_BaseVertex/Instance transform, which I didn't necessarily always want. (For proof-of-concept, I wanted to run on systems with native support for these draw funcs) I poked at trying to decouple these but didn't have immediate success.

Additionally, in trying to debug throughout this experimentation, it was not clear when things were rasterizing properly, but with the wrong glsl builtin var values. The testing is made more complicated by having to test not just these new draw commands, but also the multidraw ones, and also the glsl vars. Dropping the glsl vars makes testing simpler: The tests effectively synthesize reference versions of gl_BaseVertex and gl_BaseInstance, and compare against these references. That's great, but it also demonstrates how to synthesize the values as needed by authers, rather than doing so unconditionally. (Or handling elision of this overhead if the shader uses the extension, but doesn't use the vars)

Overall, I found this simpler to implement and test without the GLSL vars, and we know the GLSL vars are optional, and how to fairly simply implement them when desired in userland. (as we do for the test) 

> We deliberately subsetted the entry points in ANGLE_base_vertex_base_instance rather than doing a parallel of EXT_base_instance to reduce the combinatorial explosion of entry points. In particular, DrawElementsInstancedBaseInstanceEXT, and the multi-draw variant thereof, was removed.

That's fine, and we can still do that by subsetting EXT_base_instance instead of ANGLE_base_vertex_base_instance, since my proposal for this extension is closer to the former than the latter after removing the glsl aspects.

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
jdashg: instanceCount=0 draws nothing, but I think it's not something we can early-out from (it can still generate errors), so I hesitate to call it a no-op.

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
```diff
@@ -121,7 +102,12 @@
 
 const wtu = WebGLTestUtils;
 const canvas = document.getElementById("canvas");
-const gl = wtu.create3DContext(canvas, null, 2);
+canvas.style.backgroundColor = '#000';
+canvas.style.imageRendering = 'crisp-edges';
```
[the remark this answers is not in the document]
jdashg: Sure, I'm slightly surprised that works (discards invalid set calls) but relieved that it does.
MDN claims that everyone supports crisp-edges: https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering#browser_compatibility

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
```diff
@@ -269,9 +255,248 @@
...
+    },
+  }, {
+    name: 'drawElementsInstancedBaseVertexBaseInstanceWEBGL',
+    draw: desc => {
+      if (!EXT) return false;
+      EXT.drawElementsInstancedBaseVertexBaseInstanceWEBGL(
+          gl[desc.mode], desc.vert_count, gl.UNSIGNED_INT, 0,
+          desc.inst_count, desc.base_vert, desc.base_inst);
+      return true;
+    },
```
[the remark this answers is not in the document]
jdashg: This can be done in a later patch. I don't think I have interest writing these unless I want to add multidraw after all.

shrekshao:
> Additionally, in trying to debug throughout this experimentation, it was not clear when things were rasterizing properly, but with the wrong glsl builtin var values. The testing is made more complicated by having to test not just these new draw commands, but also the multidraw ones, and also the glsl vars. Dropping the glsl vars makes testing simpler: The tests effectively synthesize reference versions of gl_BaseVertex and gl_BaseInstance, and compare against these references. That's great, but it also demonstrates how to synthesize the values as needed by authers, rather than doing so unconditionally. (Or handling elision of this overhead if the shader uses the extension, but doesn't use the vars)

In terms of the newly added test code, I think this patch is certainly an improvement by adding the section of printing vertex_id, instance_count, base_vertex, base_instance info and testing them individually, regardless of whether we are splitting the extension into GLSL part and API part.

Having the added test section doesn't necessarily mean we have to drop the support of the GLSL params though. Also dropping the support wouldn't make the extension availability better as they can be emulated without `ARB_shader_draw_parameters` support. In terms of performance splitting the GLSL params would save webgl devs some uniforms if they don't need it. For webgl devs who want them, it means they would setup their own uniform or require one extra extension (if we decide to make a new extension for shader vars). Not taken into account of chrome implementation status, I think it's a valid move to split the GLSL params part for a new extension.

(though I would prefer to not have `expectTrue(!badProgram, "cannot compile program with #extension GL_ANGLE_base_vertex_base_instance");` at least for now which will make my life much easier...)

> subsetting `EXT_base_instance` instead of `ANGLE_base_vertex_base_instance`

This would be related to our early discussion on "emulation + fewer entrypoints" vs "native support + more entrypoints". We chose the former based on low availability of `EXT_base_instance` native extension on mobile devices and more new functions introduced ( basevertex[baseinstance] x [multidraw] ).

patch LGTM. I'd like we have a discussion at WG for the extension part before merge.

shrekshao (approved):

shrekshao:
a con point: `gl_DrawID` introduced in `WEBGL_multi_draw` is also from ES extension `ARB_shader_draw_parameters`. Missing BaseVertex/Instance in this extension could be a bit inconsistent

shrekshao:
Based on WG discussion on 2021/05/06 we agreed to remove gl_BaseVertex/gl_BaseInstance. The pull request looks good. @jdashg do you still want to add out-of-bounds tests before merge?

shrekshao:
conflicts is introduced by https://github.com/KhronosGroup/WebGL/pull/3286
`[AllowShared]`

kenrussell (approved):
Resolved in the WebGL working group to move forward with these changes.

PR #3279, opened 2021-04-15
s/NoInterfaceObject/Exposed=Window, LegacyNoInterfaceObject/ in exten…

jdashg:
…sions/.

### DIFF
--- a/extensions/EXT_texture_compression_bptc/extension.xml
+++ b/extensions/EXT_texture_compression_bptc/extension.xml
@@ -75,7 +75,7 @@
     </features>
   </overview>
   <idl xml:space="preserve">
-[NoInterfaceObject]
+[Exposed=(Window,Worker), LegacyNoInterfaceObject]
 interface EXT_texture_compression_bptc {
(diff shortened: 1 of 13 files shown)


foolip (approved):
Yay! (Don't have write access, just expressing support.)

extensions/EXT_texture_compression_bptc/extension.xml
```diff
@@ -75,7 +75,7 @@
     </features>
   </overview>
   <idl xml:space="preserve">
-[NoInterfaceObject]
+[Exposed=Window, LegacyNoInterfaceObject]
```
kenrussell, line 78: WebGL can be used on workers; throughout this needs to be:
  Exposed=(Window,Worker)
like in the WebGL 1.0 and 2.0 specs.

kenrussell (changes requested):
Thanks for taking care of this; one mechanical change needed throughout.

kenrussell (approved):
Thanks for taking care of this; looks good!

PR #3281, opened 2021-05-04
Allow makeXRCompatible on context.

jdarpinian:
This method comes from the WebXR spec.


kenrussell (changes requested):
Thanks - could you please put this in the 2.0.0 snapshot too?

This subsumes Dean's https://github.com/KhronosGroup/WebGL/pull/3198 ; I didn't get around to backporting the change to the snapshots.

jdashg (approved):
I swear I did this recently but maybe it was something else...

PR #3282, opened 2021-05-04
Verify that 'webgl' and 'webgl2' context types are distinct.

kenrussell:
Test case for https://bugs.webkit.org/show_bug.cgi?id=222758 .
Already passes in Chrome and Firefox.


jdarpinian (approved):
Would it be worth also checking '2d' and/or 'gpupresent'?

kenrussell:
> Would it be worth also checking '2d' and/or 'gpupresent'?

It should be pretty clear from the HTML spec that those can't be confused with any other ones. We can consider adding a more comprehensive test later.

Thanks for your review.

Issue #3283, opened 2021-05-12
Improve video-rotation test


kenrussell:
In http://crbug.com/1136205#c31 a scaling bug with rotated videos was found. Theoretically this could have been caught by the conformance/textures/misc/video-rotation.html test added in #3199 , but in order to do so the test needs to be improved by using rectangular rather than square videos.

CC'ing @dalecurtis in case it's possible for him or a team member to generate new video assets for this test, and adjust the test slightly to compensate.

PR #3284, opened 2021-05-13
Use rectangular test clips for the rotation test.

dalecurtis:
Square ones don't expose all the possible issues that can occur with rotation. 

Fixes: https://github.com/KhronosGroup/WebGL/issues/3283


kenrussell (approved):
Thanks! Verified that the test both catches the original bug and is fixed by the fix for http://crbug.com/1136205#c31 .

PR #3287, opened 2021-05-26
Verify that loops in ESSL 3.00 fragment shaders compile and link.

kenrussell:
Issues seen on some Adreno drivers from Qualcomm. Fixed in their
latest driver.

Regression test for http://crbug.com/1076188 .


kenrussell:
@kainino0x or @shrekshao : please review. Thanks.

kainino0x (approved):

kenrussell:
Going ahead with this - the Travis CI checks are unreasonably slow right now.

PR #3288, opened 2021-06-01
Add WebCodecs VideoFrame as a TexImageSource.

dalecurtis:
Adding VideoFrame as a TexImageSource allows texImage*() methods to
interoperate with VideoFrames. VideoFrames work with texImage*() the
same way the existing point-in-time capture for HTMLVideoElement does.

This adds a non-normative reference on the WebCodecs spec for the
VideoFrame interface and associated properties to accomplish this.

Note: VideoFrame objects are always same-origin, so I did not add
any text around origin restrictions for video frames.

https://www.w3.org/TR/webcodecs/#videoframe-interface

Bug: w3c/webcodecs#158

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -3924,7 +3924,7 @@ sub-rectangle updated by <code>texSubImage2D</code> are determined based on the 
             The width and height of the texture are set to the current values of the width and
             height properties of the <code>HTMLCanvasElement</code> or <code>OffscreenCanvas</code>
             object.
-        <dt><code>source</code> of type <code>HTMLVideoElement</code>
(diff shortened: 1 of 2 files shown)


kenrussell (changes requested):
Thanks for adding this!

Could you please add a reference at the bottom of the "references" section (maybe you intended to do this; the pull request description indicates so) and cite it at the first mention of VideoFrame in the spec?

dalecurtis:
Whoops, done!

kenrussell (approved):
Looks good to me! Let us discuss this at the WebGL working group meeting tomorrow before merging.

jdashg (approved):

PR #3286, opened 2021-05-18
Add [AllowShared] to WEBGL_multi_draw APIs' typed array arguments.

kenrussell:
Parallels changes made in #2387 to the core WebGL 1.0 and 2.0
specifications to support SharedArrayBuffer, and implicitly,
multithreaded WebAssembly.

Fixes #3219 .

PR #3289, opened 2021-06-09
Add WEBGL_multi_draw tests using SharedArrayBuffer.

kenrussell:
These are likely to require additional browser command-line arguments
since otherwise this functionality requires special configuration of
the web server hosting these tests.

Regression test for #3286.

Associated with http://crbug.com/1211185 .

### DIFF
--- a/sdk/tests/conformance/extensions/webgl-multi-draw.html
+++ b/sdk/tests/conformance/extensions/webgl-multi-draw.html
@@ -371,7 +371,7 @@ function doTest(ext, instanced) {
     wtu.checkCanvas(gl, [0, 255, 0, 255], "gl_DrawID is 0 for non-Multi* draw calls", 0);
   }
 
-  function runPixelTests(bufferUsage) {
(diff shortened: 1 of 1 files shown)


kenrussell:
@shrekshao : please review.

@jdashg , others: any comments welcome.

kenrussell:
Note that the duplication slightly increases the run time of this test, but even with this addition, it's still around 2.2 seconds on my work laptop.

shrekshao (approved):
LGTM.
(The reason for only testing STATIC_DRAW with SharedArrayBuffer is?)

kenrussell:
> LGTM.
> (The reason for only testing STATIC_DRAW with SharedArrayBuffer is?)

To reduce the overhead and duplication in the test.

kenrussell:
Thanks for your review @shrekshao . I'll wait until tomorrow to merge this in case Jeff has feedback.

kenrussell:
Merging now.

PR #3291, opened 2021-06-11
Increase video-rotation tolerance to 45 for VP9.

kenrussell:
Hardware accelerated video decoders on some GPUs on Windows require
this increased tolerance.

Associated with http://crbug.com/1219015 .


kenrussell:
@shrekshao please review. @dalecurtis FYI.

shrekshao (approved):
LGTM

dalecurtis:
45 is large enough that it sounds wrong. Really anything more than 10 per channel seems like an error.

kenrussell:
Fair point, but in the failing cases, the correct color channel (red, green or blue) is populated, just with values that are off.

The differences look to me perhaps like sRGB colorspace conversion happening, vs. not.

Can you think of reasons why this might have become flaky recently? Let's discuss on http://crbug.com/1219015 and http://crbug.com/1219625 .

PR #3292, opened 2021-06-15
Add drawingBufferColorSpace/unpackColorSpace attributes to WebGLRenderingContextBase.

kenrussell:
Define their behavior both for presenting the rendering context's
results, and uploading textures via texImage2D / texSubImage2D from
DOM inputs (the TexImageSource union type).

Follows https://github.com/whatwg/html/pull/6562 , in which the
PredefinedColorSpace enum was defined, and its behavior for 2D canvas
contexts specified.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -2696,6 +2696,7 @@ WebGLRenderingContext includes WebGLRenderingContextOverloads;
             Implementation experience revealed that it was beneficial to
             perform <code>ImageBitmap</code>s' color space conversion as late as possible when
             uploading to WebGL textures.
+
             </div>
 
             Next, the source image data is converted to the data type and format specified by
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -1721,7 +1721,8 @@ interface mixin <dfn id="WebGLRenderingContextBase">WebGLRenderingContextBase</d
     [Exposed=Worker] readonly attribute OffscreenCanvas canvas;
     readonly attribute GLsizei drawingBufferWidth;
     readonly attribute GLsizei drawingBufferHeight;
-    attribute PredefinedColorSpace colorSpace = "srgb";
+    attribute PredefinedColorSpace drawingBufferColorSpace = "srgb";
+    attribute PredefinedColorSpace unpackColorSpace = "srgb";
 
     [WebGLHandlesContextLoss] WebGLContextAttributes? getContextAttributes();
     [WebGLHandlesContextLoss] boolean isContextLost();
@@ -1982,21 +1983,31 @@ WebGLRenderingContext includes WebGLRenderingContextOverloads;
             the implementation is unable to satisfy the requested width or height.
         <dt>
             <code class=attribute-name>
-                <a id="DOM-WebGLRenderingContext-colorSpace">
-                    colorSpace
+                <a id="DOM-WebGLRenderingContext-drawingBufferColorSpace">
+                    drawingBufferColorSpace
                 </a>
             </code>
             of type <code><a href="https://html.spec.whatwg.org/multipage/canvas.html#predefinedcolorspace">PredefinedColorSpace</a></code>
             <a href="#refsPREDEFINEDCOLORSPACE">(specification)</a>
         <dd>
             The color space in which the drawing buffer's contents are assumed to
-            have been produced. Changing this attribute affects the on-screen display
-            of the rendering context, but does not change the drawing buffer's
-            contents. This attribute also affects how <code>TexImageSource</code>
-            sources are uploaded to textures in this context, and how the contents of
-            this context's drawing buffer are interpreted when they are drawn to a 2D
-            canvas context, or uploaded to textures in another WebGL
-            or <a href="#refsWEBGPU">WebGPU</a> rendering context.
+            have been produced. Changing this attribute causes the drawing buffer to
+            be reallocated; its current contents are lost. This attribute affects the
+            on-screen display of the rendering context, and the interpretation of
+            this context's drawing buffer when it is drawn to a 2D canvas context, or
+            uploaded to textures in another WebGL or <a href="#refsWEBGPU">WebGPU</a>
+            rendering context.
+        <dt>
+            <code class=attribute-name>
+                <a id="DOM-WebGLRenderingContext-unpackColorSpace">
+                    unpackColorSpace
+                </a>
+            </code>
+            of type <code><a href="https://html.spec.whatwg.org/multipage/canvas.html#predefinedcolorspace">PredefinedColorSpace</a></code>
+            <a href="#refsPREDEFINEDCOLORSPACE">(specification)</a>
+        <dd>
+            The color space into which <code>TexImageSource</code> sources are
+            converted when uploading them to textures in this context.
     </dl>
 
 <!-- ======================================================================================================= -->
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -1993,7 +1993,10 @@ WebGLRenderingContext includes WebGLRenderingContextOverloads;
             have been produced. Changing this attribute affects the on-screen display
             of the rendering context, but does not change the drawing buffer's
             contents. This attribute also affects how <code>TexImageSource</code>
-            sources are uploaded to textures.
+            sources are uploaded to textures in this context, and how the contents of
+            this context's drawing buffer are interpreted when they are drawn to a 2D
+            canvas context, or uploaded to textures in another WebGL
+            or <a href="#refsWEBGPU">WebGPU</a> rendering context.
     </dl>
 
 <!-- ======================================================================================================= -->
@@ -4583,6 +4586,11 @@ extensions.
         <dd><cite><a href="https://html.spec.whatwg.org/multipage/canvas.html#predefinedcolorspace">
             HTML Living Standard - The PredefinedColorSpace enum</a></cite>,
             WHATWG.
+        </dd>
+        <dt id="refsWEBGPU">[WEBGPU]</dt>
+        <dd><cite><a href="https://gpuweb.github.io/gpuweb/">
+            WebGPU Editor's Draft</a></cite>,
+            WebGPU Community Group.
         </dd>
     </dl>
 
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -2683,7 +2683,7 @@ WebGLRenderingContext includes WebGLRenderingContextOverloads;
             <a href="#TEXTURE_UPLOAD_SIZE">Texture Upload Width and Height</a>.<br><br>
 
             First, the source image data is conceptually converted to the color space
-            specified by the <a href="#DOM-WebGLRenderingContext-colorSpace">colorSpace</a>
+            specified by the <a href="#DOM-WebGLRenderingContext-unpackColorSpace">unpackColorSpace</a>
             attribute, except in the following situations:
 
             <ul>
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -1990,13 +1990,13 @@ WebGLRenderingContext includes WebGLRenderingContextOverloads;
             of type <code><a href="https://html.spec.whatwg.org/multipage/canvas.html#predefinedcolorspace">PredefinedColorSpace</a></code>
             <a href="#refsPREDEFINEDCOLORSPACE">(specification)</a>
         <dd>
-            The color space in which the drawing buffer's contents are assumed to
-            have been produced. Changing this attribute causes the drawing buffer to
-            be reallocated; its current contents are lost. This attribute affects the
-            on-screen display of the rendering context, and the interpretation of
-            this context's drawing buffer when it is drawn to a 2D canvas context, or
-            uploaded to textures in another WebGL or <a href="#refsWEBGPU">WebGPU</a>
-            rendering context.
+            The color space used to interpret the drawing buffer's content values.
+            Changing this attribute causes the drawing buffer to be reallocated; its
+            current contents are lost. This attribute affects the on-screen display
+            of the rendering context, and the interpretation of this context's
+            drawing buffer when it is drawn to a 2D canvas context, or uploaded to
+            textures in another WebGL or <a href="#refsWEBGPU">WebGPU</a> rendering
+            context.
         <dt>
             <code class=attribute-name>
                 <a id="DOM-WebGLRenderingContext-unpackColorSpace">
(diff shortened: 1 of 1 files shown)


kenrussell:
@jdashg @ccameron-chromium @RafaelCintron in particular (active participants in the ColorWeb CG), please take a look.

specs/latest/1.0/index.html
```diff
@@ -1979,6 +1980,20 @@ <h4>Attributes</h4>
...
+                <a id="DOM-WebGLRenderingContext-colorSpace">
+                    colorSpace
+                </a>
+            </code>
+            of type <code><a href="https://html.spec.whatwg.org/multipage/canvas.html#predefinedcolorspace">PredefinedColorSpace</a></code>
+            <a href="#refsPREDEFINEDCOLORSPACE">(specification)</a>
+        <dd>
+            The color space in which the drawing buffer's contents are assumed to
+            have been produced. Changing this attribute affects the on-screen display
+            of the rendering context, but does not change the drawing buffer's
```
kainino0x, line 1994: Does it also change the color space tag of this context's canvas when used as a source for other calls (like drawImage, texImage2D on another context, etc.)?

specs/latest/1.0/index.html
```diff
@@ -1979,6 +1980,20 @@ <h4>Attributes</h4>
...
+            contents. This attribute also affects how <code>TexImageSource</code>
+            sources are uploaded to textures.
```
kainino0x, line 1996: It's kind of surprising to me that this attribute changes upload behavior as well. In that way, it's one knob that's changing both the canvas behavior (for display) and the GL context behavior (for upload). In the latter case it's GL-style global state, where that state affects the behavior of particular GL calls. A usage pattern I would expect to emerge is that applications do things like:

```js
const oldColorSpace = gl.colorSpace;
gl.colorSpace = 'srgb';
// upload an sRGB JPG; keep it in sRGB (rather than converting it now), for internal processing by the app
gl.texImage2D(/* ... */);
gl.colorSpace = oldColorSpace; // set the color space back to the previous value for canvas output
```

In this particular example the app would probably be more likely to set `UNPACK_COLORSPACE_CONVERSION_WEBGL` to `NONE` instead, but I think it illustrates the point: can we have `gl.colorSpace` only affect the canvas output, and have a separate GL state for input? (`UNPACK_COLORSPACE_WEBGL`)

The shape of this API seems to imply that `colorSpace` sets the "working color space" for the entire WebGL context. But since it's mutable, that interpretation probably doesn't make sense. I think that interpretation makes sense in 2d canvas - if I'm reading the HTML spec correctly (please verify), the 2d canvas color space cannot change. But in WebGL, color spaces only matter at input and output points, so there should be no "working color space".

specs/latest/1.0/index.html
kenrussell: Good point - yes, it does change that behavior. Adding that in the next follow-up.

specs/latest/1.0/index.html
kenrussell: These are fair points. @jdashg, @ccameron-chromium and I discussed this, and our opinion was that it would be most useful to WebGL applications to be able to set the `colorSpace` on the fly - especially given #3222, which will allow reallocation of the drawing buffer with a specific format. The use case differs from the 2D rendering context, where the canvas is fully color managed - that's why the color space is fixed at context creation time in that API. In WebGL, color management is basically up to the application, and feature detection is needed to understand the supported drawing buffer formats, so `colorSpace` needs to be mutable.

Since the `colorSpace` is mutable in WebGL, and it will be inexpensive to change it, it seemed unnecessary to specify new pixel unpack state for the color space of incoming texture data. It would also be a bit confusing, since it's expected that nearly all applications that would change this setting would want their texture data converted to the canvas's working color space at the time of upload.

Given this background, what do you think? Note, I'm pushing a revision to this spec based on your feedback above and hope this conversation doesn't get lost.

specs/latest/1.0/index.html
kenrussell: Added now.

specs/latest/1.0/index.html
kainino0x: I do agree the colorSpace should be mutable. But to me, as written, it _looks_ like a "working" color space, when it is really just a knob that controls both input and output from WebGL. 

> It would also be a bit confusing, since it's expected that nearly all applications that would change this setting would want their texture data converted to the canvas's working color space at the time of upload.

I'm still not sure about this, although I don't have a specific, sufficiently-common counterexample in mind. To me it seems less confusing if the two things are controlled by two knobs instead of by one that you have to set and reset.

Most examples I can think of with more advanced usage would just use `UNPACK_COLORSPACE_CONVERSION_WEBGL` `NONE`. Here's the best one I have: a 3D application with a linear physical rendering pipeline wants to output with `colorSpace = 'linear-srgb'`, but upload assets, with conversion to `'srgb'`, into an SRGB-encoded texture format.

kenrussell: The new text's been written to try to make it clear that changing `colorSpace` has no effect on the contents of the drawing buffer. Would a rewording make it clearer?

Regarding the texture upload behavior: consider the case of a WebGL image viewer application using the `display-p3` color space. The application author will just want all incoming assets converted by the browser into that color space. That will be the behavior with this PR. @jdashg , @ccameron-chromium and I couldn't think of many (or any) situations where you'd want the browser to convert incoming assets to a different color space than the one being used for the canvas's display. For this reason I think saving and restoring the `colorSpace` property will be rare.

`UNPACK_COLORSPACE_CONVERSION_WEBGL` has limited applicability; it only works for HTMLImageElements, and ImageBitmap created from a Blob is a better alternative for that path nowadays (in Chrome, at least - Firefox engineers will disagree). It has no effect for video elements, for example.  This is another reason we didn't want to add a new pixel unpack state to specify a destination color space for texture uploads.

kainino0x: I think the wording is clear, it's more about the clarity of the API surface. People are going to need documentation and experimentation anyway though, so as long as it's consistent and well tested, it's probably not that important. I'm not that opposed to the API as written, so if others prefer it this way, it seems OK.

I forgot `UNPACK_COLORSPACE_CONVERSION_WEBGL` wouldn't consistently apply in the same paths that a `UNPACK_COLORSPACE_WEBGL` would. That indeed would be confusing.

specs/latest/1.0/index.html
```diff
@@ -2653,10 +2671,23 @@ <h4>Texture objects</h4>
...
-            If a packed pixel format is specified which would imply loss of bits of precision from the
+            First, the source image data is conceptually converted to the color space
+            specified by the <a href="#DOM-WebGLRenderingContext-colorSpace">colorSpace</a>
+            attribute, except in the following situations:
+
+            <ul>
+            <li> the source image data is an <code>HTMLImageElement</code>, and
+                 the <code>UNPACK_COLORSPACE_CONVERSION_WEBGL</code> pixel storage
+                 parameter is set to <code>NONE</code>. </li>
+            <li> the source image data is an <code>ImageBitmap</code>, in which case
```
ccameron-chromium, line 2682: I think the scheme will be to have ImageBitmapOptions have an optional colorSpace parameter. If the parameter was specified, and matches the rendering context's color space, then the conversion is a no-op by definition.

ccameron-chromium: (so I'd remove this second exception, or make it non-normative)

kainino0x: Just to check, there wouldn't be any special cases for ImageBitmaps created _without_ the optional colorSpace parameter set, right? I.e. they'll always be tagged with a particular colorspace (srgb by default I think)

> If the parameter was specified, and matches the rendering context's color space, then the conversion is a no-op by definition.

This is also what I've done in WebGPU, including the non-normative notes. https://gpuweb.github.io/gpuweb/#gpu-image-copy-texture-tagged

ccameron-chromium: ImageBitmaps created without a colorSpace won't be special (they'll be just like, say, an Image).

If an ImageBitmap is created without the colorSpace parameter specified, it will have its own internal color space (e.g, specified by an ICC profile, or video properties, etc). 
* There is no interface through which the actual pixel values (and their color space) can be queried.
* If that ImageBitmap is put into an ImageBitmapRenderingContext, it will appear in full fidelity exactly as it would if its source were put into an element (e.g, a rec2020, HLG-HDR video's frame will appear as rec2020 HLG-HDR, etc).
* If that ImageBitmap is imported into WebGL/GPU, it will be converted to the target color space at the moment of that import.

Let me know if this didn't hit the mark.

kainino0x: All sounds right to me, I just wasn't sure if it would be in the source space or in srgb if no conversion options were set. Source space does seem right, and I think there are still reasonable knobs apps can set to avoid unnecessary conversion steps.

specs/latest/1.0/index.html
```diff
@@ -1979,6 +1980,23 @@ <h4>Attributes</h4>
...
+                    colorSpace
+                </a>
+            </code>
+            of type <code><a href="https://html.spec.whatwg.org/multipage/canvas.html#predefinedcolorspace">PredefinedColorSpace</a></code>
+            <a href="#refsPREDEFINEDCOLORSPACE">(specification)</a>
+        <dd>
+            The color space in which the drawing buffer's contents are assumed to
+            have been produced. Changing this attribute affects the on-screen display
+            of the rendering context, but does not change the drawing buffer's
+            contents. This attribute also affects how <code>TexImageSource</code>
```
ccameron-chromium, line 1995: Several of the optimized backends require that the color space for the buffer be "latched" when the buffer is allocated (e.g, CAMetalLayer does this). Allowing the color space to be changed after the buffer was allocated ends up significantly complicating the implementation (we have to somehow override the color space, or do a blit to a buffer of the right color space at the end of the frame).

kenrussell: OK, let's change this to conform better to the per-platform implementations. How about:

`Changing this attribute causes the drawing buffer to be reallocated; its current contents are lost.`

?

ccameron-chromium: That sgtm.

Would we want to also want to separate this into drawingBufferColorSpace and unpackColorSpace, so that the developer can change the texture upload behavior without changing the drawing buffer behavior?

kenrussell: Sure, it makes sense to retain flexibility in this API. I'll push an update to this PR making those adjustments.

specs/latest/1.0/index.html
kenrussell: Per initial prototyping in Chromium in https://chromium-review.googlesource.com/c/chromium/src/+/3560989 - this section needs to be updated. It turns out it is beneficial to perform the colorspace conversion for ImageBitmaps as late as possible, during uploading to the WebGL context, and taking into account `unpackColorSpace`.

specs/latest/1.0/index.html
ccameron-chromium: Having had a bit more discussion about this, I think we don't want to expose a color space attribute in ImageBitmapOptions. A few reasons

1. Just adding a color space would still be under-specified, because we still want to specify a pixel format. E.g, an application can want to import an P3 image as srgb-linear float16. If we default to converting to 8-bit, this would clamp and quantize results inappropriately. If we were to add a pixel format, this would have to be an ImageBitmap-specific pixel format, with a compatibility matrix for WebGL and WebGPU.

2. Converting color space in ImageBitmap is probably not the most efficient path for importing to WebGL. The most efficient format, for JPEG among other formats, is to 
* decode to YUV (most CPU-efficient, b/c no need for YUV->RGB conversion)
* upload YUV (most efficient, because is smaller upload size than RGB)
* convert YUV->RGB and do color space conversion while importing to WebGL
The color conversion is almost-for-free compared to just doing the YUV->RGB conversion, and is done on the GPU where it's most efficient.

3. If we want move various steps (e.g, the CPU->GPU upload) to be done asynchronously, we would need to add to ImageBitmapOptions a parameter specifying the target device and the target API (e.g, send to WebGL on GPU 0).

Given all of that, I think it's best to keep the image in as close to its "native" format for as long as possible. It may be possible to do an eager upload to the GPU (when we know which GPU that will be).

specs/latest/1.0/index.html
```diff
@@ -1979,6 +1981,33 @@ <h4>Attributes</h4>
...
+            <code class=attribute-name>
+                <a id="DOM-WebGLRenderingContext-drawingBufferColorSpace">
+                    drawingBufferColorSpace
+                </a>
+            </code>
+            of type <code><a href="https://html.spec.whatwg.org/multipage/canvas.html#predefinedcolorspace">PredefinedColorSpace</a></code>
+            <a href="#refsPREDEFINEDCOLORSPACE">(specification)</a>
+        <dd>
+            The color space in which the drawing buffer's contents are assumed to
+            have been produced. Changing this attribute causes the drawing buffer to
```
kdashg, line 1994:
```suggestion
            The color space used to interpret the drawing buffer's content values. Changing this attribute causes the drawing buffer to
```

kdashg: We don't really care how it's "assumed to have been produced", just that this is how we should interpret the values.

kdashg (approved):

kdashg:
For ease of use, we should consider taking a context creation attrib for colorSpace, and filling these attribs in from that initial value. We should probably exclude it from getContextAttributes, though, since those are immutable results.

ccameron-chromium:
> For ease of use, we should consider taking a context creation attrib for colorSpace, and filling these attribs in from that initial value. We should probably exclude it from getContextAttributes, though, since those are immutable results.

That SGTM.

specs/latest/1.0/index.html
kenrussell: Good point and thanks for the suggestion. Have incorporated it.

kenrussell:
> > For ease of use, we should consider taking a context creation attrib for colorSpace, and filling these attribs in from that initial value. We should probably exclude it from getContextAttributes, though, since those are immutable results.
> 
> That SGTM.

@kdashg @ccameron-chromium sounds fine; let's do that in a follow-up since we have agreement on the rest of this PR at the moment.

specs/latest/1.0/index.html
kenrussell: Current version of this PR contains the drawingBufferColorSpace / unpackColorSpace split.

specs/latest/1.0/index.html
kenrussell: Have updated this PR to specify that ImageBitmaps are affected by unpackColorSpace, and provided rationale.

specs/latest/1.0/index.html
kenrussell: All good points. Have changed this PR to have unpackColorSpace apply to ImageBitmaps.

kenrussell:
All conversations resolved, and approved - squashing and merging.

PR #3294, opened 2021-06-22
Markup definitions with their type

dontcallmedom:
As defined in the [definition markup contract](https://tabatkins.github.io/bikeshed/#dfn-contract) that many Web specs are adhering to - these conventions are not specific to bikeshed (they are [being moved to a document on their own](https://github.com/tabatkins/bikeshed/issues/2088#issuecomment-864169459)).

Adhering to these conventions makes it easier for other specifications to reference items defined in WebGL, through projects such as [webref](github.com/w3c/webref/) and [shepherd](http://hg.csswg.org/dev/shepherd/) which themseves are integrated in bikeshed and respec, widely used by other specs.

If this type of changes are acceptable and welcome, I would be happy to provide additional patches for other IDL items, at least those that are known to be referenced from other specs - if so, I might need guidance on what convention to follow to assign `id`s to these definitions.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -743,7 +743,7 @@ enum WebGLPowerPreference { "default", "low-power", "high-performance" };
     <!-- The capitalization rules here match those of the crossOrigin
     attribute, which was added to HTMLMediaElement on behalf of
     WebGL. -->
-    <pre class="idl">dictionary <dfn id="WebGLContextAttributes">WebGLContextAttributes</dfn> {
+    <pre class="idl">dictionary <dfn data-dfn-type="dictionary" id="WebGLContextAttributes">WebGLContextAttributes</dfn> {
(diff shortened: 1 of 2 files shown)


kenrussell (changes requested):
Thanks for these improvements.

We'd be happy to integrate them but they should ideally be applied to the editors' drafts:

specs/latest/1.0/index.html
specs/latest/2.0/index.html

Would you make these changes there instead? Thanks in advance.

kenrussell (approved):
Thanks, looks good to me. Spec editor @jdashg should review this as well.

jdashg (approved):
Looks good to me!

jdashg (changes requested):
Could you fix your commit message? An \`id` got executed/expanded in bash...

jdashg (approved):

jdashg:
Thanks!

PR #3295, opened 2021-06-25
Add test case for reciprocal sqrt of sum of squares crash.

kenrussell:
This was occurring when the driver's GLSL compiler was identifying an
expression as containing the reciprocal square root of a sum of squares.

Regression test for https://crbug.com/1079309 and NVIDIA bug 2978301 .

### DIFF
--- a/sdk/tests/conformance2/glsl3/reciprocal-sqrt-of-sum-of-squares-crash.html
+++ b/sdk/tests/conformance2/glsl3/reciprocal-sqrt-of-sum-of-squares-crash.html
@@ -1,0 +1,64 @@
+<!--
+Copyright (c) 2021 The Khronos Group Inc.
+Use of this source code is governed by an MIT-style license that can be
+found in the LICENSE.txt file.
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>Shader identified as containing reciprocal square root of sum of squares should not crash</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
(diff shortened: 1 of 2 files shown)


jdashg (approved):
Wow, what the heck! Good find!

sdk/tests/conformance2/glsl3/reciprocal-sqrt-of-sum-of-squares-crash.html
```diff
@@ -0,0 +1,64 @@
...
+<script id="fshader" type="x-shader/x-fragment">#version 300 es
+precision highp float;
+
+#define CRASH 1
+
+out vec4 fragmentColor;
+void main()
+{
+    vec2 p = gl_FragCoord.xy;
+    float d = 1.0 / length(p);
```
shrekshao, line 35: nit: would be helpful with a comment stating how this expression meets reciprocal of the square root of a sum of squares

shrekshao (approved):

sdk/tests/conformance2/glsl3/reciprocal-sqrt-of-sum-of-squares-crash.html
kenrussell: Thanks for your feedback. have added a comment.

kenrussell:
Thanks @jdashg and @shrekshao for your reviews. Squashing / merging now.

aleino-nv:
> Wow, what the heck! Good find!

It's a compiler optimization in the driver that's crashing.
Longer list of example expressions that crash/don't crash: https://bugs.chromium.org/p/chromium/issues/detail?id=1079309#c21

PR #3296, opened 2021-07-03
Simplify WebGLRenderingContextBase#canvas attribute

saschanaz:
Web IDL should automatically drop the HTMLCanvasElement part from workers, [as in ImageBitmapRenderingContext](https://html.spec.whatwg.org/multipage/canvas.html#the-imagebitmaprenderingcontext-interface):

```webidl
[Exposed=(Window,Worker)]
interface ImageBitmapRenderingContext {
  readonly attribute (HTMLCanvasElement or OffscreenCanvas) canvas;
  undefined transferFromImageBitmap(ImageBitmap? bitmap);
};
```

Thus there is no need to process it manually. (Doing so makes things hard for IDL processing tools actually, e.g. https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1050.)


kenrussell (approved):
Thanks, looks good to me. Spec editor @jdashg should review too.

jdashg (approved):

PR #3304, opened 2021-08-03
Spec and test w,h constraints for BPTC, RGTC, and S3TC.

### LINKED ISSUE #3117: Relax implicit power-of-two restrictions for mipmapped textures in WEBGL_compressed_texture_s3tc

jdashg:
`w << level` and `h << level` must be multiples of 4.
This way, we forbid image allocations that are never valid, such as
1x1 for level 1. (would only be valid with a 2x2 base level, which is
forbidden)

Resolves #3117.


extensions/EXT_texture_compression_bptc/extension.xml
```diff
@@ -64,7 +64,7 @@
           <dt>COMPRESSED_RGB_BPTC_SIGNED_FLOAT_EXT</dt>
           <dt>COMPRESSED_RGB_BPTC_UNSIGNED_FLOAT_EXT</dt>
           <dd>
-          <p>When <code>level</code> equals zero <code>width</code> and <code>height</code>
+          <p><code>width &lt;&lt; level</code> and <code>height &lt;&lt; level</code>
           must be a multiple of 4.</p>
```
kenrussell, line 68: Here and throughout: I'm not sure this encodes the constraints discussed in #3117. Rereading the discussion there, it sounds like two branches are needed.
 - For both WebGL 1.0 and 2.0 contexts, the base level's width and height must each be a multiple of 4.
 - For WebGL 1.0 contexts, the base level's width and height must each be a power of two.

This should apply to all of these texture formats. DXT in particular needs to retain the power-of-two restriction for WebGL 1.0 contexts. WebGL's BPTC and RGTC extensions are written against the 1.0 spec even though WebGL 1.0 doesn't have RG textures, so theoretically they should retain the same restrictions.

sdk/tests/conformance/extensions/ext-texture-compression-bptc.html
```diff
@@ -87,10 +87,16 @@
...
+        expectation: gl.INVALID_OPERATION, message: "implied base mip 2x4 is invalid" },
+      { level: 1, width: 2, height: 1,
+        expectation: gl.INVALID_OPERATION, message: "implied base mip 4x2 is invalid" },
       { level: 1, width: 2, height: 2,
-        expectation: gl.NO_ERROR, message: "level > 0, is valid" },
-      { level: 2, width: 1, height: 1,
-        expectation: gl.NO_ERROR, message: "level > 0, is valid" },
+        expectation: gl.NO_ERROR, message: "implied base mip 4x4 is valid" },
+      { level: 2, width: 1, height: 3,
+        expectation: gl.NO_ERROR, message: "implied base mip 4x12 is valid" },
```
kenrussell, line 99: I think the base mip of 4x12 should only be valid in a WebGL 2.0 context.

sdk/tests/conformance/extensions/ext-texture-compression-rgtc.html
```diff
@@ -92,10 +92,16 @@
...
+        expectation: gl.INVALID_OPERATION, message: "implied base mip 2x4 is invalid" },
+      { level: 1, width: 2, height: 1,
+        expectation: gl.INVALID_OPERATION, message: "implied base mip 4x2 is invalid" },
       { level: 1, width: 2, height: 2,
-        expectation: gl.NO_ERROR, message: "level > 0, is valid" },
-      { level: 2, width: 1, height: 1,
-        expectation: gl.NO_ERROR, message: "level > 0, is valid" },
+        expectation: gl.NO_ERROR, message: "implied base mip 4x4 is valid" },
+      { level: 2, width: 1, height: 3,
+        expectation: gl.NO_ERROR, message: "implied base mip 4x12 is valid" },
```
kenrussell, line 104: Similarly here, I think this is only valid in a WebGL 2.0 context.

kenrussell (commented):
Thanks for revisiting and improving this. A few comments.

kenrussell:
@lexaknyazev could you also please take a look at these changes?

extensions/EXT_texture_compression_bptc/extension.xml
jdashg: I'm not really convinced by the webgl1 pot requirement, which I don't think in enforced or tested, and isn't spec'd. Patches welcome but I don't want to tackle that here.

kenrussell: OK, I see, even though we discussed this in the earlier issue it wasn't enforced. These changes are fine then.

sdk/tests/conformance/extensions/ext-texture-compression-rgtc.html
kenrussell: Never mind, per above.

sdk/tests/conformance/extensions/ext-texture-compression-bptc.html
kenrussell: Never mind, this wasn't enforced previously.

kenrussell:
Let's go ahead with these changes. @lexaknyazev if you have concerns we can address them in follow-on changes.

PR #3305, opened 2021-08-03
Require S3TC || ETC2 extensions for webgl2.

jdashg:


specs/latest/2.0/index.html
```diff
@@ -4196,6 +4196,21 @@ <h3><code>deleteQuery</code> implicitly calls <code>endQuery</code> if the query
...
+        Implementations must support at least one of the following:
+        <ul>
+            <li><a href="https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_etc/">WEBGL_compressed_texture_etc</a>
+            <li><a href="https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_s3tc/">WEBGL_compressed_texture_s3tc</a>
+        <ul>
+    </p>
+    <div class="note rationale">
+        <p>
+            To best support our ecosystem, we require implementations to support either S3TC formats (universal on desktop) or ETC2/EAC formats (universal on mobile).
+            This allows for authors to supply only two formats for resources and expect all implementations and devices to support running their content.
```
kenrussell, line 4210: Suggestion:

This allows authors to supply at most two formats...

kenrussell (approved):
This is a great step forward. CC @lexaknyazev in particular for any additional feedback.

lexaknyazev:
I'd raise the bar a little bit higher by setting the condition to `(S3TC && S3TC_sRGB) || ETC` because:
- S3TC sRGB is supported on all desktops that support WebGL 2;
- ETC extension includes sRGB formats.

jdashg:
@kenrussell PTAL w/s3tc_srgb formats

lexaknyazev:
Well, it looks like we can go even further, up to `(S3TC && S3TC_sRGB && RGTC) || ETC` because of the same reasons:
* RGTC (BC4, BC5) formats are supported on all desktops that support WebGL 2.0, namely:
  * They are always supported on ANGLE/D3D11.
  * They are required in the OpenGL 3.0 desktop spec.
  * They are in the same feature bucket as S3TC formats in Vulkan and Metal.
* ETC extension includes dedicated R/RG signed/unsigned formats.

jdashg:
@lexaknyazev That's a great analysis, thanks for including that. I think it would help for devs if we included that in the spec, so I'll look at adding a table.

lexaknyazev:
There's one caveat though. On D3D backends, DXT1 is always punch-through (i.e., with 1-bit alpha). It is mentioned in the corresponding WebGL extension specs. On all other backends, this table is accurate.

specs/latest/2.0/index.html
```diff
@@ -4196,6 +4196,81 @@ <h3><code>deleteQuery</code> implicitly calls <code>endQuery</code> if the query
...
+    <div class="note">
+        <p>
+          There are roughly equivalent formats in each suite for the following uses:
+          <table>
+            <tr>
+              <th>Usage</th>
+              <th>S3TC/RGTC option (desktop)</th>
+              <th>ETC2/EAC option (mobile)</th>
+            </tr><tr>
+              <td>R11 unsigned</td>
```
lexaknyazev, line 4230: It's a bit misleading to call the usage "R11" for both options here. In the EAC enum name, "11" refers to the format's implementation detail that it uses 11 bits of precision during interpolation. This is not applicable to RGTC.

Same comment for the next three entries.

jdashg: I think it's "usefully mostly true" (and it's indeed non-normative) in that it imparts the right instinct for how the format behaves.
The spec says:
> We will now go into detail how the decoding is done. The result will be an
11-bit fixed point number where 0 represents 0.0 and 2047 represents 1.0. This is
the exact representation for the decoded value. However, some implementations
may use, e.g., 16-bits of accuracy for filtering.

It's as true as the RGB8 compressed texture formats, which also aren't *really* 8-bit.
I think it's a useful approximation to make, even if it may not be fully true.

We can see RGTC has the same block-size requirements as its EAC counterparts, so I expect similar there. Reading through the decoding in the RGTC extension, we can see that max precision there is 8+3=11 bits as well.

lexaknyazev: RGTC (and S3TC, btw) precision is very loosely defined (and varies between GPU vendors), so any specific number is somewhat speculative. Assuming that readers will get only the "instinct" from this table, let's keep it this way.

kenrussell (approved):
Looks good to me!

lexaknyazev:
The same requirement doesn't look feasible for WebGL 1.0:
- **ETC2/EAC**: no guarantees as it's supported only on ES 3.0+ devices;
- **RGTC**: no support on D3D9;
- **S3TC**: should work on all desktops;
- **S3TC_sRGB**: should work on all desktops with top-of-tree ANGLE;
- **ETC1**: should work on most mobiles;
- **PVRTC1**: should work on some mobiles but will likely be deprecated soon.

Something like `(S3TC && S3TC_sRGB) || ETC1 || PVRTC1` would probably pass but this condition does not provide a nice feature parity like the table above.

PR #3306, opened 2021-08-03
Fix invalid texStorage allocations

lexaknyazev:


kenrussell (approved):
Looks good, thanks. Please always include context in the issue description. I'll add it here while squashing and merging.

PR #3307, opened 2021-08-05
Update mipmapped NPOT compressed texture expectations.

kenrussell:
WebGL 1.0's NPOT texture support is limited to textures without any
mip levels. Expect that the newly added RGTC and BPTC NPOT test from
PR #3304 generates an error on WebGL 1.0, and no error on WebGL 2.0.

Also change the expectation of an S3TC and S3TC-sRGB subtest which was
not following the new constraints that the size of the implied base
mip level must be a multiple of the block size.

Associated with ANGLE bug https://crbug.com/angleproject/6245 and
Issue #3117.


kenrussell:
Note: this ANGLE CL:
https://chromium-review.googlesource.com/3072568

makes the test changes from #3304 (largely, modulo these changes) pass in Chromium.

jdashg (approved):

jdashg:
Thanks!

PR #3308, opened 2021-08-06
Add test completion summary to top of test page results.

jdashg:
TEST COMPLETE: 42 PASS, 6 FAIL


kenrussell (approved):
Great idea! It's always painful to have to scroll through the tests when run outside the harness to see whether they succeeded. Looks good!

PR #3311, opened 2021-08-11
Add failure if "Duplicate notifyFinishedToHarness()".

jdashg:
This ends up being an error in Firefox's harness, so let's prevent it
upstream too.


kenrussell (approved):
Sounds good.

PR #3312, opened 2021-08-11
Add PNG Image types test

### LINKED ISSUE #2814: Add conformance tests to cover DOM uploads update (#2805)
As requested by @kenrussell in https://github.com/KhronosGroup/WebGL/pull/2805#issuecomment-468061152.

Current tests (`sdk/tests/js/tests/tex-image-and-sub-image-2d-with-image.js`) use the same `red-green.png` image (that contains RGB8 data) to test all WebGL internal formats. To cover updated behavior, we need to have two more source images and a new set of expected values.

Uploads from `HTMLCanvasElement`, `HTMLVideoElement`, `ImageBitmap`, and `ImageData`, are unaffected by the spec change, because these sources always have at least 3 channels, right? Is it OK to assume that SVG images are never internally rasterized to L or LA, so they are unaffected as well?

/cc @jdashg

lexaknyazev:
Partly reverts #2805 and fixes #2814.

### DIFF
--- a/sdk/tests/conformance/textures/misc/00_test_list.txt
+++ b/sdk/tests/conformance/textures/misc/00_test_list.txt
@@ -53,3 +53,4 @@ texture-transparent-pixels-initialized.html
 --min-version 1.0.4 texture-with-flip-y-and-premultiply-alpha.html
 --min-version 1.0.4 upload-from-srcset-with-empty-data.html
 --min-version 1.0.4 video-rotation.html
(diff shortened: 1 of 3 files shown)


jdashg (changes requested):
I would like to avoid .then, and instead use async/await for better clarity on control flow. PTAL: https://github.com/lexaknyazev/WebGL/pull/1

jdashg (approved):
Awesome, thanks!

PR #3314, opened 2021-08-12
Add test for e.g. getFragDataLocation('gl_FragColor').

jdashg:
This came up in a Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1657189
Apparent bug in Mesa: https://gitlab.freedesktop.org/mesa/mesa/-/issues/5221
So let's test it in WebGL!


kenrussell (approved):
Looks good. Tested minimally locally; test passes.

PR #3318, opened 2021-08-17
Fix and expand test for OES_draw_buffers_indexed.

jdashg:


kenrussell (approved):
I reviewed the changes manually by merging them locally and running `git diff -b main`. The new tests look well principled but I didn't review each individual assertion against the spec. @lexaknyazev could you please do a thorough review?

Would it be possible to commit a separate change doing the DOS->Unix line ending conversion, and then rebase this on top of it, so the pull request is cleaner?

lexaknyazev:
The new tests look good. Couple nits:
- Should the non-indexed part of `testColorMaskDrawNoOp` be also added to the core CTS?
- Could the tests be repeated for other attachment formats (float and integer, in particular)?

There's also a known Intel/Mac/GL issue with integer attachments and min/max blending equation. It is reproducible even without this extension, see `BlendIntegerTest.MRT*` end2end tests in ANGLE.

kenrussell:
> There's also a known Intel/Mac/GL issue with integer attachments and min/max blending equation. It is reproducible even without this extension, see `BlendIntegerTest.MRT*` end2end tests in ANGLE.

Small request: if there's a Mac OpenGL driver bug in this area then could we please leave that sub-test disabled for the moment until we can move to Metal? Hopefully this will resolve this issue. Otherwise in Chromium we'll have to disable the whole oes-draw-buffers-indexed.html test on Mac/Intel, allowing regressions to slip in.

lexaknyazev:
> Otherwise in Chromium we'll have to disable the whole oes-draw-buffers-indexed.html test on Mac/Intel, allowing regressions to slip in.

The Mac/Intel/GL issue is not limited to this extension, so it could live in a separate WebGL 2 test.

jdashg:
Yeah, it would be good to have the non-indexed subtests pulled out into conformance/rendering somewhere, eventually.
Tests for floatish/uint/sint would be good too, to expand on what's already here.

PR #3319, opened 2021-08-18
Simplify startPlayingAndWaitForVideo with modern async/await.

jdashg:


kenrussell:
@dalecurtis do you think this mechanism will work reliably in Chromium? We've iterated on this primitive over and over in order to eliminate flakiness.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -3137,36 +3137,20 @@ var runSteps = function(steps) {
...
-    rvfc.call(video, _ => { callback(video) });
-  }
-
-  video.loop = true;
-  video.muted = true;
-  // See whether setting the preload flag de-flakes video-related tests.
-  video.preload = 'auto';
-  video.play();
+  (async () => {
+    video.load(); // reset it
```
dalecurtis, line 3141: This will trigger a significant amount of unnecessary work since a load is already in flight.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -3137,36 +3137,20 @@ var runSteps = function(steps) {
...
-        callback(video);
-      } else {
-        requestAnimFrame.call(window, timeWatcher);
-      }
-    };
-
-    timeWatcher();
-  } else {
-    // Calls video.requestVideoFrameCallback(_ => { callback(video) })
-    rvfc.call(video, _ => { callback(video) });
```
dalecurtis, line 3162: This needs to be preserved since the play promise resolving doesn't mean that a frame can be drawn from it yet.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -3137,36 +3137,20 @@ var runSteps = function(steps) {
...
-  video.preload = 'auto';
-  video.play();
+  (async () => {
+    video.load(); // reset it
+    video.loop = true;
+    video.muted = true;
+    // See whether setting the preload flag de-flakes video-related tests.
+    video.preload = 'auto';
+    try {
+      await video.play();
```
dalecurtis, line 3147: I don't think this is safe, see https://github.com/KhronosGroup/WebGL/pull/3263 where I reverted this previously.

dalecurtis (commented):
I don't think this is safe.

kenrussell (commented):
Thanks @dalecurtis for your review.

At the specification level:
https://html.spec.whatwg.org/multipage/media.html#dom-media-play

is one issue that the `play()` Promise can be resolved when the network state is still `HAVE_FUTURE_DATA`? In other words, per spec, it's not guaranteed that data's available for the first frame when this Promise resolves?

jdashg:
`HAVE_FUTURE_DATA` seems to mean that there's data for the next frame:
> Data for the immediate current playback position is available, as well as enough data for the user agent to advance the current playback position in the direction of playback at least a little without immediately reverting to the HAVE_METADATA state, and the text tracks are ready. For example, in video this corresponds to the user agent having data for at least the current frame and the next frame when the current playback position is at the instant in time between the two frames, or to the user agent having the video data for the current frame and audio data to keep playing at least a little when the current playback position is in the middle of a frame. The user agent cannot be in this state if playback has ended, as the current playback position can never advance in this case.

Do we need to await on `timeupdate` then? `resize`? There needs to be a way to do this, otherwise we can't expect websites to get it right either. :)

jdashg:
I think HAVE_FUTURE_DATA ought to be enough to ensure we can texImage(video). At worst this might causes a wait-for-decode, but the data is ready at least.

sdk/tests/js/webgl-test-utils.js
jdashg: I think it probably does mean this? (see comments in main thread)

kenrussell:
Big thanks for finding #3322 . The failures that were seen previously were more widespread, and included widespread conformance/textures/video failures:

https://chromium-review.googlesource.com/c/chromium/src/+/2775887/5

e.g.

https://ci.chromium.org/ui/p/chromium/builders/try/linux-rel/644735/overview
https://ci.chromium.org/ui/p/chromium/builders/try/mac_optional_gpu_tests_rel/56041/overview
https://ci.chromium.org/ui/p/chromium/builders/try/win10_chromium_x64_rel_ng/816557/overview

so unfortunately this change is incompatible with Chromium's implementation of the HTMLVideoElement.play() spec, at least.

jdashg:
I do think we sort of fall short here of defining what should happen and-then-therefore testing that. We need to define at what point calling texImage(video) should succeed, and what should happen otherwise.

My intuition is that we should try to define it such that `await video.play(); texImage(video)` should reliably succeed. How implementable does that sound?

dalecurtis:
We've generally interpreted `readyState` to indicate that the encoded data for a given playback position is available, not that we _have_ rendered the frame yet. While it _generally_ does mean that we have rendered a frame, it's not 100% accurate -- especially in rapid load/test cycles like seen in web platform and conformance tests.

> Media elements have a ready state, which describes to what degree they are _ready to be_ rendered at the current playback position.

Emphasis on "ready to be" mine. If we interpret that to mean the decoded frame must have been rendered, then significant load deferral optimizations for offscreen and/or autoplay-rejected videos are not possible. Which turns into a chicken or the egg problem since some sites won't try to play an element until HAVE_FUTURE_DATA.

`requestVideoFrameCallback()` is one way to ensure the frame has actually been presented. I think we could defer play promise resolution until the first frame is rendered, but it would need a deeper analysis and real world testing.

jdashg:
I'm not sure we actually need the first frame to have been rendered in the general sense, though we should test that for sure. (gotta make sure the fastpaths work)

For instance, we need to support videos that are *never* rendered to the screen, and only exist floating loose from the DOM, yet still play and can be uploaded into textures.

dalecurtis:
By rendered I just meant available to script. rVFC() works for offscreen elements. In Chromium's implementation we let `texImage(VideoElement)` drive the clock for offscreen elements, but its done asynchronously since we'd otherwise need to block on a background thread from the main thread.

jdashg:
I think that `await video.play(); texImage(video)` should succeed. Is this implementable?

dalecurtis:
Probably, I don't see any immediate perils, but as noted in https://github.com/KhronosGroup/WebGL/pull/3319#issuecomment-901513639 it'll need some real world testing. Certainly simpler sounding things have ended up as very hairy yaks in the past :|

kenrussell:
@grorg @kimmok what do you think about the feasibility of making the Promise returned from `HTMLVideoElement.play` obey these guarantees in Safari?

kainino0x:
FYI I recently ported startPlayingAndWaitForVideo into the WebGPU CTS. If any changes end up landing here I'd like to port them over too.
https://github.com/gpuweb/cts/blob/main/src/webgpu/web_platform/util.ts

jdashg:
@dalecurtis PTAL
I sort of combined the two, and use rvfc where supported.
`await video.play()` seems to fix some flakiness in Firefox's CI.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -3136,38 +3128,30 @@ var runSteps = function(steps) {
...
+
+  try {
+    await video.play();
+  } catch (e) {
+    testFailed('video.play failed: ' + e);
+    return;
+  }
+
+  if (video.requestVideoFrameCallback) {
+    await new Promise(go => video.requestVideoFrameCallback(go));
```
kenrussell, line 3150: @dalecurtis do you think this will have the same basic semantics as the earlier code in Chromium?

sdk/tests/js/webgl-test-utils.js
```diff
@@ -3136,38 +3128,30 @@ var runSteps = function(steps) {
...
  *        video is ready.
  */
-function startPlayingAndWaitForVideo(video, callback) {
+async function startPlayingAndWaitForVideo(video, callback) {
   if (video.error) {
     testFailed('Video failed to load: ' + video.error);
     return;
   }
 
-  video.addEventListener(
```
dalecurtis, line 3145: I think you still want this, otherwise an error after playback startup won't be caught.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -3136,38 +3128,30 @@ var runSteps = function(steps) {
...
-};
+
+  try {
+    await video.play();
+  } catch (e) {
+    testFailed('video.play failed: ' + e);
+    return;
+  }
+
+  if (video.requestVideoFrameCallback) {
```
dalecurtis, line 3149: This is probably fine, but may slightly slow down test execution in Chromium. The old code is better for rVFC, since it may fire before play() is called. In an ideal world this function would set the rVFC callback then the `video.src` and never even bother calling `play()` since that's just a source of flakiness as far as the conformance pixel tests are concerned.

Without keeping the rest of the error checking though, it's hard to keep the rVFC path since the promise rejection is pulling double duty for error detection and play denial.

I think it's fine to land this and revert if we see issues (or improvements!), but defer to @kenrussell. If we want to avoid even minor test slow down risk, the rVFC path should set rVFC before `play()` and only `try { video.play(); } catch (e) {...}` without the `await`.

You might at least remove the `await new Promise(...)` in favor of: 

```
if (video.requestVideoFrameCallback) {
  video.requestVideoFrameCallback(_ => { callback(video); });
  return;
}
```

dalecurtis (approved):

sdk/tests/js/webgl-test-utils.js
jdashg: Having a unified codepath that optionally awaits for the rvfc is nicer than divergent behavior. Also using rvfc to carry the actual callback means the function as a whole isn't awaitable anymore.

I can't imagine any serious perf differences between rvfc before play or not. (Really, I don't think rvfc should be spec-allowed to callback before await video.play does)

Thanks for the feedback!

sdk/tests/js/webgl-test-utils.js
jdashg: Honestly I think it's more likely to cause problems to have errors that might happen later. We just added assertions that errors did not happen after finishTest() is called, and this seems like it risks that. `await video.play()` should throw in any case that we care about here.

jdashg:
Thanks all!

sdk/tests/js/webgl-test-utils.js
dalecurtis: Possibly, I'd probably at least keep a console log or something then. The risk is only to debuggability in case the playback failure leads to conformance test failures.

sdk/tests/js/webgl-test-utils.js
dalecurtis: It will have different semantics as I note in the comment thread below, but it should be okay % potential for minor slowness.

sdk/tests/js/webgl-test-utils.js
jdashg: I feel like this ought to be handled by warnings that the browser generates, or otherwise should instead happen following an unexpected failure in some later texImage(video).

PR #3322, opened 2021-08-18
Await runImageBitmapTest to fix racing image-bitmap-from-video tests.

jdashg:


kenrussell (approved):
Yikes. Thanks for catching this.

Should we mark runImageBitmapTest in sdk/tests/js/tests/tex-image-and-sub-image-with-image-bitmap-utils.js async at the same time? It's using old-school Promise APIs rather than async/await.

jdashg:
I have a follow-up that refactors it, but I want to get this minimal fix in first!

PR #3323, opened 2021-08-19
[video-rotation] Remember to finishTest() after e.g. !supports_vp9.

jdashg:


kenrussell (approved):
Great catch and nice refactoring.

PR #3324, opened 2021-08-20
Expose the clearing canvas with scissor issue on Intel D3D drivers

Jiawei-Shao:
This patch adds a conformance test to expose an Intel D3D driver
bug about clearing canvas with scissor test in WebGL.

See https://crbug.com/1206763 for more details.

### DIFF
--- a/sdk/tests/conformance/rendering/clear-default-framebuffer-with-scissor-test.html
+++ b/sdk/tests/conformance/rendering/clear-default-framebuffer-with-scissor-test.html
@@ -21,7 +21,7 @@ found in the LICENSE.txt file.
 
 <script>
 "use strict";
-// This test exposes an Intel D3D driver issue.
+// This test exposes an issue in older Intel D3D drivers.
 var wtu = WebGLTestUtils;
 
 function test_clear_with_scissor_on_canvas()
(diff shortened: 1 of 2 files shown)


Jiawei-Shao:
@kenrussell @kainino0x  PTAL, thanks!

sdk/tests/conformance/rendering/clear-default-framebuffer-with-scissor-test.html
```diff
@@ -0,0 +1,68 @@
...
+      if (data[pixelIndex * 4 + colorIndex] !== expectedColor[colorIndex]) {
+        var y = Math.floor(pixelIndex / scissorRectSize);
+        var x = pixelIndex % scissorRectSize;
+        testFailed("The canvas color at (" + x + ", " + y + ") is not expected");
+        return;
+      }
+    }
+  }
+}
+
```
jdashg, line 58: I think this can be `wtu.checkCanvasRect(0, 0, scissorRectSize, scissorRectSize, [0, 255, 0, 255]);`

jdashg (approved):

jdashg:
@Jiawei-Shao What do you think? Happy to merge as-is and I can simplify it later.

sdk/tests/conformance/rendering/clear-default-framebuffer-with-scissor-test.html
Jiawei-Shao: Oh I found we cannot use `wtu.checkCanvasRect(0, 0, scissorRectSize, scissorRectSize, [0, 255, 0, 255]);` here because we have already created a `2d` context from `canvas2D`, while `wtu.checkCanvasRect()` requires a WebGL context as its first parameter.

The issue only occurs when the Chromium compositor composes the WebGL default framebuffer (a texture) into the whole webpage - actually the content in the original WebGL default framebuffer is correct, so I have to use a `2D` canvas to trigger this issue.

kenrussell:
Very sorry @Jiawei-Shao for not reviewing this promptly! Thanks for the new test; it looks good!

Issue #3275, opened 2021-04-07
Use inclusive terminology in WebGL test suite


kimmok:
Use inclusive terminology in WebGL test suite

- blacklist -> blocklist
- whitelist -> allowlist

PR #3325, opened 2021-08-23
Use inclusive terminology in WebGL test suite #3275

kkinnunen-apple:
Use:
 blacklist -> blocklist
 whitelist -> allowlist


toji:
Thank you!

Issue #3277, opened 2021-04-07
Test suite should test that depth test does not affect default framebuffer when context is created with stencil=true, depth=false


kimmok:
Test suite should test that depth test does not affect default framebuffer when context is created with stencil=true, depth=false

https://bugs.webkit.org/show_bug.cgi?id=223734

kenrussell:
Fixed by #3326 .

PR #3326, opened 2021-08-23
Test suite should test that depth test does not affect default frameb…

kkinnunen-apple:
…uffer when context is created with stencil=true, depth=false #3277

Change conformance/context/context-attributes-alpha-depth-stencil-antialias.html
to test all combinations of stencil, depth, antialias, alpha with the same test.

Make the test test all the properties:
 - drawing with stencil test affects result: yes/no
 - drawing with depth test affects result: yes/no
 - drawing with has aliasing: yes/no
 - drawing with alpha affects result: yes/no

Match the results of the draw function with the context creation configuration.

### DIFF
--- a/sdk/tests/conformance/context/context-attributes-alpha-depth-stencil-antialias.html
+++ b/sdk/tests/conformance/context/context-attributes-alpha-depth-stencil-antialias.html
@@ -47,6 +47,7 @@ var framebuffer;
 var fbHasColor;
 var fbHasDepth;
 var fbHasStencil;
+var contextVersion = wtu.getDefault3DContextVersion();
 
 function init()
 {
@@ -54,6 +55,23 @@ function init()
 
     runTest();
 }
+
+var vertices = new Float32Array([
+    1.0, 1.0, 0.0,
+    -1.0, 1.0, 0.0,
+    -1.0, -1.0, 0.0,
+    1.0, 1.0, 0.0,
+    -1.0, -1.0, 0.0,
+    1.0, -1.0, 0.0]);
+
+var colors = new Uint8Array([
+    255, 0, 0, 255,
+    255, 0, 0, 255,
+    255, 0, 0, 255,
+    255, 0, 0, 255,
+    255, 0, 0, 255,
+    255, 0, 0, 255]);
+
 
 function getWebGL(canvasWidth, canvasHeight, contextAttribs, clearColor, clearDepth, clearStencil)
 {
@@ -63,7 +81,7 @@ function getWebGL(canvasWidth, canvasHeight, contextAttribs, clearColor, clearDe
     canvas.width = canvasWidth;
     canvas.height = canvasHeight;
 
-    gl = wtu.create3DContext(canvas, contextAttribs);
+    gl = wtu.create3DContext(canvas, contextAttribs, contextVersion);
     if (!gl)
         return null;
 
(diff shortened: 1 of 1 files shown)


kenrussell (approved):
Thanks for this contribution and sorry for taking so long to review it. Couple of questions:

1) Could you improve the error messages a little bit? Chromium for example is failing 4 of the new test combinations and it would be great to have them more clearly explain what the browser's doing wrong.

2) Could you please click through the CLA? Apple's already a Khronos member; this is effectively a no-op.

Thanks.

kenrussell:
Looking at the testing results, I think this revised test may be too strict for WebGL 1.0 implementations, where `depth`, `stencil` and `antialias` are requests, not requirements: https://www.khronos.org/registry/webgl/specs/latest/1.0/#2.2 .

However, Chrome fails these new tests even in WebGL 2.0 mode, so we could potentially agree to adopt this new, stricter test given that Safari and Firefox are passing it.

Thoughts?

kkinnunen-apple:
> 1. Could you improve the error messages a little bit? Chromium for example is failing 4 of the new test combinations and it would be great to have them more clearly explain what the browser's doing wrong.

I tried to formulate the test so that in fact all the info is in the asserts that are printed.

```
PASS gl = getWebGL(2, 2, { depth: false, stencil: true, antialias: false, alpha: false }, [ 0, 0, 0, 0 ], 1, 0) is non-null.
PASS contextAttribs = gl.getContextAttributes() is non-null.
PASS gl.getParameter(gl.RED_BITS) is >= 8
PASS gl.getParameter(gl.GREEN_BITS) is >= 8
PASS gl.getParameter(gl.BLUE_BITS) is >= 8
PASS contextAttribs.alpha is false
PASS contextAttribs.depth is false
PASS contextAttribs.antialias is false
PASS gl.getParameter(gl.ALPHA_BITS) is 0
PASS gl.getParameter(gl.STENCIL_BITS) is >= 8
PASS gl.getParameter(gl.DEPTH_BITS) is 0
PASS should be 0,0,0,255
Testing default framebuffer.
Testing that stencil affects the rendering.
PASS should be 0,0,0,255
PASS should be 0,0,0,255
Testing that depth does not affect the rendering.
FAIL should be 255,0,0,255
```

E.g. the only obscure piece is that it does not say "Now starts case with depth:false, stencil:true"., it says "gl = getWebGL(2, 
2, { depth: false, stencil: true, ...". I thought this was what the existing test also intended to give out as context. If this is not in line with the existing standards, I can of course remove that and add a some sort of header printout. However, if you mean that the info should be present at the exact location chrome fails today, I  don't think that makes sense. It has to be in some sort of header and not part of any specific nor all subheaders of the test.

If I see correctly it's in line with existing tests, and the messages give the exact case. (FWIW, in very many existing tests I've used to fix our bugs, the result itself really doesn't give the full context in the assertions)

> 2. Could you please click through the CLA? Apple's already a Khronos member; this is effectively a no-op.

I think it says I've done that. ("All committers have signed the CLA.")

> Looking at the testing results, I think this revised test may be too strict for WebGL 1.0 implementations, where `depth`, `stencil` and `antialias` are requests, not requirements: https://www.khronos.org/registry/webgl/specs/latest/1.0/#2.2 .
> 
> However, Chrome fails these new tests even in WebGL 2.0 mode, so we could potentially agree to adopt this new, stricter test given that Safari and Firefox are passing it.

Yeah, I think the test should be revised to be consistent with WebGL 1.0 spec. I try to do this.

kenrussell:
> E.g. the only obscure piece is that it does not say "Now starts case with depth:false, stencil:true"., it says "gl = getWebGL(2,
> 2, { depth: false, stencil: true, ...". I thought this was what the existing test also intended to give out as context. If this is not in line with the existing standards, I can of course remove that and add a some sort of header printout. However, if you mean that the info should be present at the exact location chrome fails today, I don't think that makes sense. It has to be in some sort of header and not part of any specific nor all subheaders of the test.

Fair enough, I agree with you. Could you still see whether it's feasible to print a newline via debug(...) just before the new configuration's tests start?

> Yeah, I think the test should be revised to be consistent with WebGL 1.0 spec. I try to do this.

Thanks, appreciate it.

kenrussell (approved):
Re-ran the revised test. Firefox nightly passes in both WebGL 1.0 and 2.0 modes. Chrome and Safari fail the `depth:false, stencil:true` cases in both WebGL 1.0 and 2.0 modes. Since it's clear the test is passable, merging now and we'll fix the failing browsers' behaviors.

PR #3328, opened 2021-08-27
[oes-draw-buffers-indexed.html] Reset colorMask and blend disable.

jdashg:
I sure forgot to set this.


kenrussell (approved):
Looks good to me.

kenrussell:
Note: follow-on to #3318 .

PR #3329, opened 2021-08-27
Disallow emulated compressed formats

lexaknyazev:
This statement is added to prevent WebGL implementations from exposing ETC* formats on platforms that advertise them regardless of hardware support, most notably desktop OpenGL 4.3+ or any older version combined with `GL_ARB_ES3_compatibility`.

Among "desktop" GPUs, only Intel newer than Haswell support ETC* in hardware when using OpenGL.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -4073,7 +4073,7 @@ extensions.
 </p>
 
 <p>
-    Unless a WebGL implementation uses a software rasterizer, only hardware-supported compressed texture extensions may be enabled.
+    WebGL implementations should only expose compressed texture formats that are more efficient than the uncompressed form.
 </p>
 
     <h3><a name="MAX_GLSL_TOKEN_SIZE">Maximum GLSL Token Size</a></h3>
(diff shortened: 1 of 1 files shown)


jdashg:
We'll need to work on the language here (probably not 'hardware' and 'software' unfortunately). I have some feelings that this should be "SHOULD NOT" not "MUST NOT", but we'll talk it over in meeting.

Thanks for proposing this!

MarkCallow:
> We'll need to work on the language here (probably not 'hardware' and 'software' unfortunately).

Probably something like "do not expose if inflated prior to texture sampling". Inflation during sampling is what is wanted.

lexaknyazev:
When a WebGL implementation uses a software rasterizer such as SwiftShader, inflation inevitably happens before sampling.

MarkCallow:
> When a WebGL implementation uses a software rasterizer such as SwiftShader, inflation inevitably happens before sampling.

2 things:
1. The proposed language already has an exception for software rasterizers which I expected to remain.
2. Why is it inevitable? There are software block unpackers for all block-compressed formats. Maybe there shouldn't be an exception.

lexaknyazev:
@jdashg PTAL

jdashg (approved):
Thanks!

PR #3333, opened 2021-09-13
Fix several compatibility issues in tool

gyagp:
platform.dist() is deprecated for python 3.9, so lsb_release is used
instead to get Linux version.
Browsers on Windows should be 64-bit, so PROGRAMFILES, instead of
PROGRAMFILES(X86), is used for their paths.
options, instead of chrome_options, is used as parameter of
selenium.webdriver.Chrome() to suppress the deprecation warning.

### DIFF
--- a/tools/conformance.py
+++ b/tools/conformance.py
@@ -15,6 +15,11 @@ import sys
 import time
 
 try:
+    import lsb_release
+except ImportError:
+    pass
+
+try:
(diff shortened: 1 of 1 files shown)


gyagp:
@kenrussell, PTAL!

kenrussell (approved):
Thanks Yang for the fixes! They look good!

PR #3335, opened 2021-09-17
Test double indirect row_matrix access.

greggman:
Looking at the code in one of the backends I see this is problematic
so it seemed there should be a test.


kenrussell (approved):
Sorry for the long delay reviewing.

Looks good! I see that failures in this test are already suppressed on Mac/OpenGL, and Linux/Intel/OpenGL in Chrome...and that your recent fixes to ANGLE's Metal backend make both portions of this test pass on Mac/Metal! Great work!

kenrussell:
Small comment: if these tests are associated with Chromium or ANGLE bugs please add a URL to the bug in the extended commit message since this makes it easier to track things when rolling forward WebGL conformance into Chromium; thanks.

PR #3336, opened 2021-09-17
[gl-object-get-calls] Spin event loop occasionally to avoid freezing …

jdashg:
…page.

This test is *really* slow in Firefox for architectural reasons, and for robustness reasons, I don't want to special-case error generation here just to make this synthetic use-case faster.


greggman:
Would it be too slow to do this?

```
const pause = _ => new Promise(resolve => setTimeout(resolve, _));

// then in any loop in an async function

for (...) {
  await pause();
  ... stuff...
}
```

Less changes.

jdashg:
I think it's better with slightly more changes that include printing how far along it is.
Also, using setTimeout like that counts as "nested setTimeout" and so *does* incurs throttling: https://jsfiddle.net/d2b3ok59/
That's just too slow.

PR #3342, opened 2021-10-10
Initial submit of WEBGL_multisampled_render_to_texture proposal

cabanier:


extensions/proposals/WEBGL_multisampled_render_to_texture/extension.xml
```diff
@@ -0,0 +1,160 @@
+<?xml version="1.0"?>
+
+<extension href="WEBGL_multisampled_render_to_texture/">
```
kenrussell, line 3: Please run "make" under the `extensions/` directory and fix the errors generated. At least this href is incorrect; there may be others.

sdk/tests/conformance2/extensions/webgl_multisampled_render_to_texture.html
```diff
@@ -0,0 +1,181 @@
...
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<script>
+"use strict";
+
+let wtu = WebGLTestUtils;
+let gl = wtu.create3DContext(null, null, 2);
```
kenrussell, line 23: Since the extension is written against the WebGL 1.0 specification, this test should be placed in `conformance/extensions/` rather than `conformance2/extensions/`. In that case the `create3DContext` call should not specify a context version as its third argument.

sdk/tests/conformance2/extensions/webgl_multisampled_render_to_texture.html
```diff
@@ -0,0 +1,181 @@
+<!--
+Copyright (c) 2019 The Khronos Group Inc.
```
kenrussell, line 2: 2021

sdk/tests/conformance2/extensions/webgl_multisampled_render_to_texture.html
```diff
@@ -0,0 +1,181 @@
+<!--
```
kenrussell, line 1: Please add any new tests to `00_test_list.txt` in the same directory.

For WebGL 1.0 tests, use `--min-version 1.0.4`.

For WebGL 2.0 tests, use `--min-version 2.0.1`.

kenrussell (changes requested):
Thanks for this contribution. A few issues to be addressed.

jdashg:
I would prefer not to add more WebGL 1 extensions at this time, now that all UAs have WebGL2. (And any machine that we support this ext on would support webgl2)

PR #3356, opened 2021-11-04
Disallow point rendering with undefined point size

### LINKED ISSUE #2818: Require gl_PointSize for points rendering or define a default value

lexaknyazev:
Fixes #2818.

Should the CTS update be included in this PR before merging?

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -3286,6 +3286,8 @@ WebGLRenderingContext includes WebGLRenderingContextOverloads;
             <dd>
             If <em>first</em> is negative, an <code>INVALID_VALUE</code> error will be generated.
             If the CURRENT_PROGRAM is null, an <code>INVALID_OPERATION</code> error will be generated.
+            If the <em>mode</em> is <code>POINTS</code> and the current program does not set
+            <code>gl_PointSize</code> value, an <code>INVALID_OPERATION</code> error will be generated.
         <dt class="idl-code">void drawElements(GLenum mode, GLsizei count, GLenum type, GLintptr offset)
             <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/2.0/es_full_spec_2.0.25.pdf#nameddest=section-2.8">OpenGL ES 2.0 &sect;2.8</a>, <a class="nonnormative" href="http://www.khronos.org/opengles/sdk/2.0/docs/man/xhtml/glDrawElements.xml">man page</a>)</span>
         <dd>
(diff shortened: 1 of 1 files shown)


specs/latest/1.0/index.html
```diff
@@ -4455,6 +4459,17 @@ <h3>Wide point primitive clipping</h3>
...
 </div>
 
+<h3>Undefined Point Size</h3>
+<p>
+    Draw calls with <code>POINTS</code> primitives are rejected when the bound program does not set
+    <code>gl_PointSize</code>.
+</p>
+<div class="note rationale">
+    <p>
+        Point rendering with undefined point size is undefined in GLES and may be unsupported in other APIs.
```
kenrussell, line 4469: From concall: please mention something more precise about why this doesn't work.

specs/latest/1.0/index.html
```diff
@@ -3286,6 +3286,8 @@ <h4><a name="WRITING_TO_THE_DRAWING_BUFFER">Writing to the drawing buffer</a></h
             <dd>
             If <em>first</em> is negative, an <code>INVALID_VALUE</code> error will be generated.
             If the CURRENT_PROGRAM is null, an <code>INVALID_OPERATION</code> error will be generated.
+            If the <em>mode</em> is <code>POINTS</code> and the current program does not set
+            <code>gl_PointSize</code> value, an <code>INVALID_OPERATION</code> error will be generated.
```
kenrussell, line 3290: ...set the gl_PointSize variable? Here and below.

kenrussell (commented):
Yes, it would be ideal to update the conformance tests at the same time.

kenrussell (changes requested):
@lexaknyazev is it feasible for you to make the requested changes?

specs/latest/1.0/index.html
lexaknyazev: After looking into this again, it seems there are two underlying issues there:
- D3D11 does not support point rendering with point size greater than `1.0`, so ANGLE generates a geometry shader to handle that.
- The presence of `gl_PointSize` in the original ESSL source non-trivially affects ANGLE's shader translator (e.g., how varyings are packed), so simply prepending all shaders with `gl_PointSize = 1.0;` may cause unexpected issues. Those may be caused by unknown bugs in the translator.

So this issue looks like an ANGLE limitation, which may remain indefinitely.

What would be the appropriate language for the spec here?

greggman:
Just so this doesn't get lost, I'm not sure this should happen. See https://github.com/KhronosGroup/WebGL/pull/2822#issuecomment-984949138

lexaknyazev:
Closed in favor of #3370.

PR #3357, opened 2021-11-05
Add format-filterable-renderable test.

jdashg:
Checks all texture formats for filterable and color-renderable.
WebGL 1 and 2.
WebGL 1's RGB8 doesn't look like it's spec-required. Should it be?
Every device and browser I tested supports it, and it's color-renderable
in ES3/webgl2.

### DIFF
--- a/sdk/tests/conformance/textures/misc/format-filterable-renderable.html
+++ b/sdk/tests/conformance/textures/misc/format-filterable-renderable.html
@@ -1,0 +1,374 @@
+<!--
+Copyright (c) 2021 The Khronos Group Inc.
+Use of this source code is governed by an MIT-style license that can be
+found in the LICENSE.txt file.
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<link rel="stylesheet" href="../../../resources/js-test-style.css"/>
+<script src="../../../js/js-test-pre.js"></script>
+<script src="../../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<script>
+"use strict";
+
+const wtu = WebGLTestUtils;
+description();
+
+const gl = wtu.create3DContext();
+gl.canvas.width = gl.canvas.height = 1;
+
+function makeTexImage(format, unpackFormat, unpackType, data) {
+    data = data || null;
+
+    const tex = gl.createTexture();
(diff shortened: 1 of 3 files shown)


sdk/tests/conformance/textures/misc/format-filterable-renderable.html
```diff
@@ -0,0 +1,374 @@
...
+    const bytes = new Uint8Array(4);
+    gl.readPixels(0,0,1,1,gl.RGBA, gl.UNSIGNED_BYTE, bytes);
+    return [].map.call(bytes, x => x/255.0);
+}
+
+// GLES 2.0.25 p63
+const FORMAT_INFO_WEBGL1 = {
+    RGBA8  : { filter: true, render: true , unpack: ['RGBA', 'UNSIGNED_BYTE'] },
+    RGB8   : { filter: true, render: true , unpack: ['RGB', 'UNSIGNED_BYTE'] },
+    RGBA4  : { filter: true, render: true , unpack: ['RGBA', 'UNSIGNED_SHORT_4_4_4_4'] },
```
greggman, line 102: IIRC the only color format required to be able to be rendered to is RGBA8 and that requirement is from the WebGL spec, the OpenGL ES 2.0 spec requires no formats to be able to be rendered to

jdashg: This is true, so this might be stretching a bit.
However, RGBA4/RGB5_A1/RGB656 are all required *renderbuffer* formats in webgl1, even if the corresponding textures aren't required.
Having them maybe-renderable is a portability hazard it'd be cool to chip away at, and one way for us to start is to try to over-test, and see if we find any failing systems.

greggman: I'm not against adding them but you should probably update the spec first before adding a test that requires them.

jdashg: If we didn't have such a long history of kinda-sorta-front-running via the CTS, I would be more aggressive about speculative specification :)

jdashg: If you insist, I will strike those subtests, but I'm not going to push for speccing this myself.

kenrussell (approved):
The revised test looks good to me. Thanks for putting together this comprehensive test.

greggman (approved):
lgtm

PR #3358, opened 2021-11-09
Not for commit: test glGetUniformBlockIndex + KHR_parallel_shader_com…

kenrussell:
…pile.

Attempted reduced test case for https://crbug.com/1267322 . Can't
reproduce the problem in isolation yet.


kenrussell:
@magcius FYI.

PR #3361, opened 2021-11-12
Track main branch of WebGLDeveloperTools submodule.

kenrussell:
This will make submodule initialization behave more intuitively.


kenrussell:
Merging myself.

PR #3362, opened 2021-11-12
Update sdk/devtools submodule to current top-of-tree.

kenrussell:


kenrussell:
This should actually be the correct update. Merging myself.

PR #3363, opened 2021-11-16
Fix gl_VertexID value in base-vertex-base-instance extension test

shrekshao:
> The index of any element transferred to the GL by DrawArraysOneInstance
is referred to as its vertex ID, and may be read by a vertex shader as gl_VertexID.
The vertex ID of the ith element transferred is first + i.

From the [OpenGL ES spec section 10.5](https://www.khronos.org/registry/OpenGL/specs/es/3.2/es_spec_3.2.pdf)

Related to #3278

### DIFF
--- a/sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
+++ b/sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
@@ -310,7 +310,7 @@ function testGlslBuiltins() {
       const vert_src = `\
 #version 300 es
 #line 405
-layout(location = 0) in int a_vertex_id; // COUNT_UP_DATA
+layout(location = 0) in int a_vertex_id; // Same as gl_VertexID
 out vec4 v_color;
 
 void main() {
@@ -330,7 +330,7 @@ void main() {
       const vert_src = `\
 #version 300 es
 #line 425
-layout(location = 0) in int a_vertex_id; // COUNT_UP_DATA
+layout(location = 0) in int a_vertex_id; // Same as gl_VertexID
 layout(location = 1) in int a_instance_div1; // Same as base_instance+gl_InstanceID
 layout(location = 2) in int a_instance_div2; // Same as base_instance+floor(gl_InstanceID/2)
 layout(location = 3) in int a_instance_div3; // Same as base_instance+floor(gl_InstanceID/3)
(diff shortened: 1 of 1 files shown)


shrekshao:
@jdashg does this change look good to you?

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
```diff
@@ -310,7 +310,7 @@
       const vert_src = `\
 #version 300 es
 #line 405
-layout(location = 0) in int a_vertex_id; // Same as first_vert+gl_VertexID
+layout(location = 0) in int a_vertex_id; // COUNT_UP_DATA
```
jdashg, line 313: "Same as gl_VertexID". I'm not sure what COUNT_UP_DATA is

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
```diff
@@ -469,8 +469,8 @@
     gl.disable(gl.SCISSOR_TEST);
     gl.clearBufferfv(gl.COLOR, 0, [1,0,0,1]);
 
-    const last_gl_vert_id = desc.base_vert + desc.vert_count - 1;
-    const last_vert_id = desc.first_vert + last_gl_vert_id;
+    const last_gl_vert_id = desc.first_vert + desc.base_vert + desc.vert_count - 1;
```
jdashg, line 472: `desc.base_vert + desc.first_vert`

jdashg (changes requested):

jdashg:
Oh, and *add that spec reference to the test*!!

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
shrekshao: vert_count can be 0, 1, 2 so they are not equal?

shrekshao: Oh I think you mean flip the order...
```diff
- const last_gl_vert_id = desc.first_vert + desc.base_vert + desc.vert_count - 1;
+ const last_gl_vert_id = desc.base_vert + desc.first_vert + desc.vert_count - 1;
```

jdashg (approved):

jdashg:
Thanks!

PR #3364, opened 2021-11-18
Require rgb renderable webgl1

jdashg:
Depends on #3357

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -3892,9 +3892,19 @@ framebuffer attachment complete, non-zero, and have the same width and height, m
 framebuffer being framebuffer complete:
 
 <ul>
-<li> <code>COLOR_ATTACHMENT0</code> = <code>RGBA/UNSIGNED_BYTE</code> texture
-<li> <code>COLOR_ATTACHMENT0</code> = <code>RGBA/UNSIGNED_BYTE</code> texture + <code>DEPTH_ATTACHMENT</code> = <code>DEPTH_COMPONENT16</code> renderbuffer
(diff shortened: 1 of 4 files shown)


specs/latest/1.0/index.html
```diff
@@ -3892,9 +3892,19 @@ <h3><a name="FBO_ATTACHMENTS">Framebuffer Object Attachments</a></h3>
...
+</ul>
+
+A texture is color-renderable if it has any of the following format-type pairs:
+
+<ul>
+<li> <code>RGBA</code>, <code>UNSIGNED_BYTE</code>
+<li> <code>RGB</code>, <code>UNSIGNED_BYTE</code>
+<li> <code>RGBA</code>, <code>UNSIGNED_SHORT_4_4_4_4</code>
+<li> <code>RGBA</code>, <code>UNSIGNED_SHORT_4_4_4_4</code>
+<li> <code>RGB</code>, <code>UNSIGNED_SHORT_5_6_5</code>
```
kenrussell, line 3907: A couple of recollections:
* Early ES 2.0 implementations didn't guarantee renderability of these formats.
* I recall that ES 3.0 revoked renderability of some RGB formats, but looking again at the spec, RGB8 is in fact guaranteed to be color-renderable (ES 3.0.6 p.130).

What are the concerns we should be thinking of when asserting these formats are renderable in both WebGL 1.0 and 2.0?

greggman:
FYI: I wrote a C++ test to test these formats and 5551 fails on NVidia Mac

greggman:
So I wrote a C++ test and 5_5_5_1 fails on NVidia MacOS

kenrussell:
Not surprising. Note though that ANGLE would likely emulate that format on top of RGBA8 as a driver bug workaround, since it's legal for the implementation to provide more bits of precision.

PR #3368, opened 2021-12-04
Set gl_PointSize in primitive restart tests for GL_POINTS.

kenrussell:
Failing to set this was relying on currently-undefined behavior and
causing the tests to fail on Apple's M1.

Related to #2818, #2822 and #3356.


kenrussell:
@greggman or @jdashg please review - thanks.

This has been confirmed to fix these tests on Apple's M1.

greggman (approved):
lgtm

jdashg (approved):

PR #3369, opened 2021-12-15
Require ETC1 support to match ETC2 on WebGL 2.0

### LINKED ISSUE #3014: ETC1/ETC2 detection
ETC2/EAC formats are a part of OpenGL ES 3.0 core spec, but they were deliberately omitted from WebGL 2.0 as there's no support for them on many target platforms (e.g. D3D11).

The situation with OpenGL drivers is quite interesting. According to the spec, desktop OpenGL drivers can advertise support for these formats through `GL_ARB_ES3_compatibility` extension or simply by supporting OpenGL 4.3+ core. __This does not mean that these formats are actually supported in hardware.__

For this reason, Chromium/ANGLE has a quite sophisticated feature detection.

- **ANGLE**
  - When running on OpenGL 4.3 (or higher) or `GL_ARB_ES3_compatibility` extension is present, ANGLE enables these formats and exposes them using dedicated extension strings like `OES_compressed_ETC2_*_texture` and `OES_compressed_EAC_*_texture`.
  - When running on native OpenGL ES 3.0+, ANGLE additionally exposes `GL_ANGLE_compressed_texture_etc` extension string.
- **Chromium**
  - Completely ignores context version or dedicated format strings and checks only for `GL_ANGLE_compressed_texture_etc` so that the `WEBGL_compressed_texture_etc` extension is enabled only on OpenGL ES contexts.
  - By the way, this does not apply to `WEBGL_compressed_texture_etc1` extension. It is exposed in Chromium on all OpenGL contexts (bug?).

---

On the other hand, Firefox simply checks for OpenGL 4.3 or `GL_ARB_ES3_compatibility` and enables `WEBGL_compressed_texture_etc` if either is available.

This leads to at least two sub-optimal cases:
- When an application tries to guess what compressed format to use at runtime (either out of pre-encoded variants or by relying on Basis transcoding), it may see `_etc*` extension and use it despite the lack of hardware support.
- Some hardware (e.g. recent Intel GPUs) actually has proper ETC2 support in their OpenGL (and even Vulkan) drivers. There seems to be no way to reliably query that.

lexaknyazev:
This is as much as we could do on the WebGL CTS side to fix #3014. Another part of the fix was added in #3329 as a guidance for the implementors.

We could also add an extra test for WebGL 1.0 to catch a similar issue there:
```js
shouldBeTrue("!has_etc || has_etc1");
```

### DIFF
--- a/sdk/tests/conformance2/extensions/required-extensions.html
+++ b/sdk/tests/conformance2/extensions/required-extensions.html
@@ -43,7 +43,13 @@ debug(`has_etc: ${has_etc}`);
 
 shouldBeTrue("((has_s3tc && has_s3tc_srgb && has_rgtc) || has_etc)");
 
-debug("")
+// ETC1 extension must not be exposed on WebGL 2.0 contexts without ETC2.
+debug("");
+const has_etc1 = hasExt('WEBGL_compressed_texture_etc1');
+debug(`has_etc1: ${has_etc1}`);
+shouldBeTrue("has_etc1 == has_etc");
+
+debug("");
 var successfullyParsed = true;
 </script>
 


kenrussell (approved):
Strengthening these checks sounds good to me!

jdashg (approved):
I guess this is fine?
If there were hardware that didn't match this, I wouldn't want this change, but if we can have this, I guess I won't block it.

PR #3371, opened 2021-12-21
Immediately invalidate current program if its relink fails.

kenrussell:
Rather than continuing to preserve the current executable as specified
in the OpenGL ES API, generate INVALID_OPERATION for subsequent draw
calls if the currently bound program is relinked, and that relink fails.

Update program-test.html to follow the new behavior.

Agreed upon in the WebGL working group meeting of 2021-10-21.

### DIFF
--- a/sdk/tests/conformance/programs/program-test.html
+++ b/sdk/tests/conformance/programs/program-test.html
@@ -312,9 +312,9 @@ function go() {
     assertMsg(gl.getProgramParameter(progGood2, gl.LINK_STATUS) == false,
               "linking should fail with in-use formerly good program, with new bad shader attached");
 
-    // Invalid link leaves previous valid program intact.
+    // In WebGL, an invalid link immediately invalidates the previous valid program.
     gl.drawArrays(gl.TRIANGLES, 0, 3);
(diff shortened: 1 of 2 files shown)


kenrussell:
@jdashg or @greggman in particular please review - thanks.

Any other suggestions welcome.

greggman (approved):
lgtm

sdk/tests/conformance/programs/program-test.html
```diff
@@ -312,9 +312,9 @@
     assertMsg(gl.getProgramParameter(progGood2, gl.LINK_STATUS) == false,
               "linking should fail with in-use formerly good program, with new bad shader attached");
 
-    // Invalid link leaves previous valid program intact.
+    // In WebGL, an invalid link immediately invalidates the previous valid program.
```
kdashg, line 315:
```suggestion
    // In WebGL, an unsuccessful link immediately invalidates the previous valid program.
```

kdashg (approved):

sdk/tests/conformance/programs/program-test.html
kenrussell: Thanks, applied this change.

PR #3374, opened 2022-01-03
Fix broken / redirected links in References

SuperSodaSea:
Fixing [GLES30], [GLES30GLSL] and [WEBIDL]. Should fix #3373.

~~**HELP WANTED**: The link for **[GLES30]**:~~

> ~~http://www.khronos.org/registry/gles/specs/3.0/es_spec_3.0.4.pdf~~

~~refers to GLES **3.0.4**, but I can only find the latest pdf for GLES **3.0.6** on Khronos' pages~~

> ~~https://www.khronos.org/registry/OpenGL/specs/es/3.0/es_spec_3.0.pdf~~

~~which I think cannot be replaced directly.~~
~~Do we need to find an old **3.0.4** version of the pdf, or change the spec to fit the latest **3.0.6** version?~~

All references to GLES 3.0.4 are upgraded to 3.0.6.

PR #3375, opened 2022-01-08
Revise [WEBIDL] specification reference.

kenrussell:
Mention that it is the living standard, and update the author list.


kenrussell:
Follow-on to #3374.

kdashg (approved):

PR #3377, opened 2022-01-31
Delete unintended copy of San Angeles sources.

kenrussell:
This has been present for a number of years and contained Windows
binaries we didn't mean to carry forward.

Associated with Chromium bug 1290867 (restricted view, sorry).


kenrussell:
This just deletes a subdirectory which wasn't intended to be there in the first place. Merging myself.

PR #3378, opened 2022-02-01
Reject proposed extension EXT_shader_framebuffer_fetch.

kdashg:
At this point, this functionality should be pursued in WebGPU instead.


kenrussell (approved):
This change was discussed and agreed upon in the WebGL working group. It was recommended to pursue EXT_shader_pixel_local_storage instead, implementing it on top of this extension, in #3385 .

PR #3382, opened 2022-02-12
Reject WEBGL_subarray_uploads proposal.

kdashg:
Per 2022-01-27 meeting.


kenrussell (approved):
Really sorry for taking so long - approved.

PR #3383, opened 2022-02-12
Reject WEBGL_texture_multisample.

kdashg:
As per 2022-01-27 meeting.


kenrussell (approved):
Really sorry for not looking at this until now, but thank you, approved.

PR #3390, opened 2022-03-08
Test the canvas' drawingbuffer keeps its contents when antialiased

greggman:
The Apple Metal backend had an issue that blitting a multisample
renderbuffer could lose its content. Whether or not the contents
is lost is up to the driver. This repoed on M1 Macs.

https://bugs.webkit.org/show_bug.cgi?id=237113


kenrussell:
Very sorry for the long delay reviewing.

Confirmed by building ToT Chromium with and without the fix that this test catches the bug on M1 hardware.

PR #3393, opened 2022-03-21
Expand gl-scissor-fbo-test to cover antialias:true.

kenrussell:
Regression test for https://bugs.webkit.org/show_bug.cgi?id=237906 .

Confirmed with Safari 15.4 and Safari Technology Preview Release 141
(Safari 15.4, WebKit 17614.1.3.8) with "WebGL via Metal" that this
test catches the underlying bug. The bug has been fixed in top-of-tree
ANGLE and passes on Chrome Canary 102.0.4955.0 with --use-angle=metal.


kenrussell:
@greggman or @kdashg : please review. Thanks.

Confirmed locally that this catches the described bug on M1 hardware.

kdashg (approved):

PR #3395, opened 2022-03-24
Test depth buffer commands in a multisampling framebuffer

brendan-duncan:
Recent updates to WebKit caused a series of WebGL2 regressions. This test reproduces behavior found in several applications, where rendering to a multisampling framebuffer can break if depth_test is modified during the render.
https://brendan-duncan.github.io/webgl_bug_tests/msaa_depth_buffer.html

The test creates a multisampling framebuffer. It clears the framebuffer with a red color, disables depth_test, draws a large yellow quad, enables depth_test, and draws a smaller blue quad.

A series of conditions causes the behavior:
1) Render from requestAnimationFrame. The first render is correct. Subsequent frames fail.
2) Render into a mutlisampling framebuffer with a color and depth buffer.
3) Disable DEPTH_TEST and depthMask.
4) Render a quad.
5) Enable DEPTH_TEST and depthMask.
6) Render a quad.
7) Resolve the multisampling framebuffer and display the resolved texture.

On WebKit with recent ANGLE updates, the blue quad is not rendered.

### DIFF
--- a/sdk/tests/conformance2/rendering/00_test_list.txt
+++ b/sdk/tests/conformance2/rendering/00_test_list.txt
@@ -38,6 +38,7 @@ instanced-arrays.html
 --min-version 2.0.1 instanced-rendering-bug.html
 --min-version 2.0.1 instanced-rendering-large-divisor.html
 --min-version 2.0.1 line-rendering-quality.html
+--min-version 2.0.1 multisampling-depth-resolve.html
 --min-version 2.0.1 multisampling-fragment-evaluation.html
 out-of-bounds-index-buffers-after-copying.html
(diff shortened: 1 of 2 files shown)


kenrussell (approved):
Thank you Brendan; this is excellent.

I confirm that this fails in current Safari Technology Preview Release 142 (Safari 15.4, WebKit 17614.1.5.16). It passes with top-of-tree WebKit; note on the associated bug https://bugs.webkit.org/show_bug.cgi?id=238118 that there is a bug in one of the scripts for launching the browser against built WebKit. It also passes in Chrome Canary with ANGLE's Metal renderer enabled.

PR #3396, opened 2022-03-25
Revise program-handling test to follow new spec behavior.

kenrussell:
In PR #2879, referencing Issue #2842, this test was added in order to
verify the behavior of failed relinks.

Since then, the WebGL specification was changed in PR #3371 to
immediately invalidate the current program upon failed relink. This
was motivated by the discovery of multiple bugs with inconsistent
program states during relinking, including http://crbug.com/1241123
and http://crbug.com/angleproject/6358 .

Revise program-handling.html to expect the current program to be
immediately invalidated upon failed relink, with associated behavioral
changes. Split the test into two cases to retain existing coverage:
failed relinks, and successful drawing after removal of the program's
shaders.

Associated with Chromium bug http://crbug.com/953120 .

### DIFF
--- a/sdk/tests/conformance/programs/program-handling.html
+++ b/sdk/tests/conformance/programs/program-handling.html
@@ -63,56 +63,76 @@ if (!gl) {
     testFailed("context does not exist");
 }
 
-var vs = wtu.loadShaderFromScript(gl, "vshader");
-var fs = wtu.loadShaderFromScript(gl, "fshader");
-var prg = wtu.createProgram(gl, vs, fs);
-gl.useProgram(prg);
-const positionLoc = gl.getAttribLocation(prg, 'a_position');
-const colorLoc = gl.getUniformLocation(prg, 'u_color');
+function testProgramInvalidation() {
+    debug('');
+    debug('== Testing invalidation of the current program ==');
+    var vs = wtu.loadShaderFromScript(gl, "vshader");
+    var fs = wtu.loadShaderFromScript(gl, "fshader");
+    var prg = wtu.createProgram(gl, vs, fs);
+    gl.useProgram(prg);
+    const positionLoc = gl.getAttribLocation(prg, 'a_position');
+    const colorLoc = gl.getUniformLocation(prg, 'u_color');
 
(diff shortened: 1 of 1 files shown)


kenrussell:
@kdashg : please review. Thanks.

@greggman : please feel free to review too.

Hopefully this won't be controversial; it's being updated to cover already-agreed-upon specification changes.

Tested locally on Chrome Canary on macOS; all cases pass.

greggman (approved):
lgtm

sdk/tests/conformance/programs/program-handling.html
```diff
@@ -63,56 +63,76 @@
...
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be no errors before relink");
+
+    debug("Try linking");
+    gl.linkProgram(prg);
+    assertMsg(gl.getProgramParameter(prg, gl.LINK_STATUS) == false, "link should fail");
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be no errors after linkProgram");
+    debug("Attempt to draw green; because link failed, in WebGL, the draw should generate INVALID_OPERATION");
+    gl.uniform4fv(colorLoc, [0, 1, 0, 1]);
+    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "because the current program has been invalidated, uniform* calls generate INVALID_OPERATION");
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be no errors before draw");
```
kdashg, line 98: You probably don't need this second no-error check after just checking the error.
I think that if we suspect late or multiple stacked errors, we should modify glErrorShouldBe to abstract them away.

kdashg (approved):

sdk/tests/conformance/programs/program-handling.html
kenrussell: Thanks for your feedback; I see your point. For the time being I'm going to leave in this second error check out of an abundance of caution, but will think about a follow-up PR to `WebGLTestUtils.glErrorShouldBe` that will capture all errors and ensure that there are none outside the passed set. Also please feel free to file an issue about that.

PR #3397, opened 2022-03-29
Fix recently-rejected extensions' URLs.

kenrussell:
These are breaking the build of the extensions/ directory.


kenrussell:
@kdashg FYI. Merging myself to fix the build of extensions/ - broken on Khronos' server as well.

kdashg:
Thanks!

PR #3398, opened 2022-03-29
Deprecate the WEBGL_compressed_texture_pvrtc extension.

kenrussell:
Guide developers to prefer the ETC or ASTC compressed texture formats
instead, which are more broadly supported and offer better control
over compression fidelity.

Supersedes #2816 - on request from Apple.

### DIFF
--- a/extensions/WEBGL_compressed_texture_pvrtc/extension.xml
+++ b/extensions/WEBGL_compressed_texture_pvrtc/extension.xml
@@ -14,6 +14,13 @@
   </depends>
   <overview>
     <p>
+      <b>Note: this extension is deprecated.</b> On systems that support this compressed texture
+      format, please consider using the <a
+      href="../WEBGL_compressed_texture_etc">WEBGL_compressed_texture_etc</a> or <a
+      href="WEBGL_compressed_texture_astc">WEBGL_compressed_texture_astc</a> formats instead. They
+      are more widely supported and offer a larger range of quality controls.
+    </p>
+    <p>
       This extension exposes the compressed texture formats defined in the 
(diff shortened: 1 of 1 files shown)


kenrussell:
@kdashg after discussion with @lexaknyazev and @kkinnunen-apple on Slack, they've asked for this resolution to the PVRTC extension rather than #2816 . Would you please review and offer your thoughts?

kenrussell:
Now asking @kkinnunen-apple officially for review. Thanks.

kkinnunen-apple (approved):

extensions/WEBGL_compressed_texture_pvrtc/extension.xml
```diff
@@ -13,6 +13,13 @@
     <api version="1.0"/>
   </depends>
   <overview>
+    <p>
+      <b>Note: this extension is deprecated.</b> On systems that support this compressed texture
+      format, please consider using the <a
+      href="../WEBGL_compressed_texture_etc">WEBGL_compressed_texture_etc</a> or <a
+      href="WEBGL_compressed_texture_astc">WEBGL_compressed_texture_astc</a> formats instead. They
```
lexaknyazev, line 20: Should the ASTC link start with `../`?

kenrussell: Yes - thanks for catching that. Fixed, and tested the built HTML locally.

lexaknyazev (approved):

kdashg (approved):
Sure, we can do it this way.

kenrussell:
Great. Thanks everyone for your reviews.

PR #3402, opened 2022-04-12
gl_VertexID test with a large count

greggman:
This test calls gl.drawArrays with a count of 4194304 which happens to reveal bugs in certain WebGL implementations.

### DIFF
--- a/sdk/tests/conformance2/rendering/vertex-id-large-count.html
+++ b/sdk/tests/conformance2/rendering/vertex-id-large-count.html
@@ -1,0 +1,127 @@
+﻿<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL 2 gl_VertexID Large Count Tests</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/desktop-gl-constants.js"></script>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<!-- Shaders for testing instanced draws -->
+<script id="vs" type="text/plain">
+#version 300 es
+
+uniform ivec2 resolution;
+void main() {
+    ivec2 pixel = ivec2(
+        gl_VertexID % resolution.x,
+        gl_VertexID / resolution.x);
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/rendering/vertex-id-large-count.html
```diff
@@ -0,0 +1,121 @@
...
+
+<script>
+"use strict";
+description("Test gl_VertexID");
+
+debug("");
+
+const wtu = WebGLTestUtils;
+const canvas = document.createElement("canvas");
+const size = 1024;
```
kdashg, line 47: It would be cool to test even bigger as well, but this is a great addition!

sdk/tests/conformance2/rendering/vertex-id-large-count.html
```diff
@@ -0,0 +1,121 @@
...
+canvas.width = size;
+canvas.height = size;
+const gl = wtu.create3DContext(canvas, null, 2);
+gl.viewport(0, 0, size, size);
```
kdashg, line 51: This *shouldn't* be necessary

sdk/tests/conformance2/rendering/vertex-id-large-count.html
```diff
@@ -0,0 +1,121 @@
...
+        gl.clear(gl.COLOR_BUFFER_BIT);
+        gl.uniform4fv(colorLoc, color);
+        gl.uniform2i(resolutionLoc, size, size);
+        gl.drawArrays(gl.POINTS, first, count);
+        wtu.glErrorShouldBe(gl, gl.NO_ERROR);
+        const y = first / size;
+        const height = count / size;
+        wtu.checkCanvasRect(gl, 0, y, size, height, u8Color(color));
+        if (y > 0) {
+          wtu.checkCanvasRect(gl, 0, 0, size, size - height, transparentBlack);
```
kdashg, line 99: Can you describe in a comment what the general behavior is?

kdashg (approved):

sdk/tests/conformance2/rendering/vertex-id-large-count.html
greggman: WebGL2 requires 2048 and 2048x2048 as probably okay on all WebGL2 devices so bumped to 2048

sdk/tests/conformance2/rendering/vertex-id-large-count.html
greggman: added

PR #3404, opened 2022-04-27
Add unpackColorSpace tests to tex-image-and-sub-image-2d-with-video

ccameron-chromium:
Change the test to use named colors, such as 'Red' and 'Green', and add the utility function namedColorInColorSpace to provide color space conversion.

Several tests replicate the behavior of converting a color from an 8-bit RGBA value to how that color will appear when sampled with a given internal format. Move this functionality to a colorAsSampledWithInternalFormat function.

Some of the tests started failing when using RGBA4 as an internal format, because the values being tested were no longer 0 or 255. Add a tolerance utility function to compute the appropriate for a given internalformat, format, and type.

To all test cases in tex-image-and-sub-image-2d-with-video.js, add an unpackColorSpace parameter. Cross-product all tests with the set of all color spaces. Add a utility function to perform this cross-product.

### DIFF
--- a/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-video.js
+++ b/sdk/tests/js/tests/tex-image-and-sub-image-2d-with-video.js
@@ -20,8 +20,6 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
     var tiu = TexImageUtils;
     var gl = null;
     var successfullyParsed = false;
-    var redColor = [255, 0, 0];
-    var greenColor = [0, 255, 0];
 
     // Test each format separately because many browsers implement each
     // differently. Some might be GPU accelerated, some might not. Etc...
@@ -45,31 +43,13 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
             return;
         }
 
-        switch (gl[pixelFormat]) {
-          case gl.RED:
-          case gl.RED_INTEGER:
-            greenColor = [0, 0, 0];
-            break;
-          case gl.LUMINANCE:
-          case gl.LUMINANCE_ALPHA:
-            redColor = [255, 255, 255];
-            greenColor = [0, 0, 0];
-            break;
-          case gl.ALPHA:
-            redColor = [0, 0, 0];
-            greenColor = [0, 0, 0];
-            break;
-          default:
-            break;
-        }
-
         gl.clearColor(0,0,0,1);
         gl.clearDepth(1);
 
         runTest();
     }
 
-    function runOneIteration(videoElement, useTexSubImage2D, flipY, topColor, bottomColor, sourceSubRectangle, program, bindingTarget)
+    function runOneIteration(videoElement, unpackColorSpace, useTexSubImage2D, flipY, topColorName, bottomColorName, sourceSubRectangle, program, bindingTarget)
     {
         sourceSubRectangleString = '';
         if (sourceSubRectangle) {
@@ -101,6 +81,10 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
                        gl.TEXTURE_CUBE_MAP_NEGATIVE_Y,
                        gl.TEXTURE_CUBE_MAP_POSITIVE_Z,
                        gl.TEXTURE_CUBE_MAP_NEGATIVE_Z];
+        }
+        // Handle target color space.
+        if (unpackColorSpace) { 
+          gl.unpackColorSpace = unpackColorSpace;
         }
         // Handle the source sub-rectangle if specified (WebGL 2.0 only)
         if (sourceSubRectangle) {
@@ -165,6 +149,12 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
             loc = gl.getUniformLocation(program, "face");
         }
 
+        const topColor = wtu.colorAsSampledWithInternalFormat(
+            wtu.namedColorInColorSpace(topColorName, unpackColorSpace),
+            internalFormat);
+        const bottomColor = wtu.colorAsSampledWithInternalFormat(
+            wtu.namedColorInColorSpace(bottomColorName, unpackColorSpace),
+            internalFormat);
         for (var tt = 0; tt < targets.length; ++tt) {
             if (bindingTarget == gl.TEXTURE_CUBE_MAP) {
                 gl.uniform1i(loc, targets[tt]);
@@ -173,7 +163,7 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
             wtu.clearAndDrawUnitQuad(gl, [0, 0, 0, 255]);
             // Check a few pixels near the top and bottom and make sure they have
             // the right color.
-            const tolerance = 6;
+            const tolerance = Math.max(6, tiu.tolerance(internalFormat, pixelFormat, pixelType));
             debug("Checking lower left corner");
             wtu.checkCanvasRect(gl, 4, 4, 2, 2, bottomColor,
                                 "shouldBe " + bottomColor, tolerance);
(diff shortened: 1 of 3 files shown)


sdk/tests/js/tests/tex-image-and-sub-image-utils.js
```diff
@@ -791,11 +791,71 @@ var TexImageUtils = (function() {
...
+      return [null];
+  }
+
+  /**
+   * For each entry in unpackColorSpaces, duplicate all of cases, adding an
+   * unpackColorSpace key with its value set to that entry to each case.
+   */
+  var crossProductTestCasesWithUnpackColorSpaces = function(cases, unpackColorSpaces)
+  {
+    var caseWithColorSpace = function(c, cs) { return Object.assign({}, c, {unpackColorSpace:cs}); }
```
greggman, line 812:
```
return Object.assign({}, c, {unpackColorSpace:cs})
```

can be shortened to

```
return {...c, ...{unpackColorSpace:cs}};
```

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
```diff
@@ -791,11 +791,71 @@ var TexImageUtils = (function() {
...
+    var casesList = unpackColorSpaces.map(cs => cases.map(c => caseWithColorSpace(c, cs)));
+    return [].concat.apply([], casesList);
```
greggman, line 814: Could this just be

```
return caseList;
```

?

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-video.js
```diff
@@ -102,6 +82,10 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
                        gl.TEXTURE_CUBE_MAP_POSITIVE_Z,
                        gl.TEXTURE_CUBE_MAP_NEGATIVE_Z];
         }
+        // Handle target color space.
+        if (unpackColorSpace) {
+          gl.unpackColorSpace = unpackColorSpace;
```
greggman, line 87: where can I find out more about this? Not much comes up when I search

https://www.google.com/search?q=%22unpackColorSpace%22

In any case, is the path of not specifying a color space still tested if color spaces are supported?

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
ccameron-chromium: This didn't work when I tested against FF, so I left it unchanged.

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
ccameron-chromium: casesList ends up being
  [[list of cases with srgb], [list of cases with display-p3]]
and so the concat moves them all into a single list.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-video.js
ccameron-chromium: This is to test the PR in https://github.com/KhronosGroup/WebGL/pull/3292

The unpackColorSpace is an attribute that should have a default value of 'srgb' (another test should verify that indeed this is the default value). If unpackColorSpace is not present, then the feature isn't implemented (and everything should be converted to sRGB, which is the same as setting unpackColorSpace to 'srgb').

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
greggman: How did you test? Works in Firefox just fine AFAIK

https://jsgist.org/?src=c212a3fa63ff82bc55c88b73fd996b1b

I've been using it for years

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
greggman: In that case you can use `flat`

```
return caseList.flat()
```

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
ccameron-chromium: Hmm, no idea what I did wrong. Now it works fine. Updated.

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
ccameron-chromium: Updated.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-video.js
greggman: Maybe I missing it. To me there are 3 cases

1. set unpackColorSpace to 'srgb'
2. set unpackColorSpace to 'display-p3'
3. don't set unpackColorSpace (supposed to be the same as 1)

If 3 is not tested, how do we actually know it's the same as 1?

greggman: re-reading your comment, I agree with you. If unpackColorSpace exists then checking it defaults to srgb should be enought

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
```diff
@@ -791,11 +791,71 @@ var TexImageUtils = (function() {
...
+      return [null];
+  }
+
+  /**
+   * For each entry in unpackColorSpaces, duplicate all of cases, adding an
+   * unpackColorSpace key with its value set to that entry to each case.
+   */
+  var crossProductTestCasesWithUnpackColorSpaces = function(cases, unpackColorSpaces)
+  {
+    var caseWithColorSpace = function(c, cs) { return {...c, ...{unpackColorSpace:cs}}; }
```
kenrussell, line 812: Seeing the earlier discussion with @greggman , having done some reading, and not having seen this trick before - could you please add a comment like:

```
// Merge test case and unpackColorSpace using spread operator
```
? I spent some time wondering if two iterables with multiple items each were being folded into a dictionary.

sdk/tests/js/webgl-test-utils.js
```diff
@@ -3296,6 +3296,95 @@ function linearChannelToSRGB(value) {
...
+          return [0, 255, 0, 255];
+          break;
+        default:
+          throw 'unexpected color name: ' + colorName;
+      };
+      break;
+    case 'display-p3':
+      switch(colorName) {
+        case 'Red':
+          return [234, 51, 35, 255];
```
kenrussell, line 3331: In a comment could you point to documentation of how this is derived? Or if the math already exists elsewhere in this repository, could a helper function be called to construct this value?

sdk/tests/js/webgl-test-utils.js
```diff
@@ -3296,6 +3296,95 @@ function linearChannelToSRGB(value) {
...
+          break;
+        case 'Green':
+          return [117, 251, 76, 255];
```
kenrussell, line 3334: Similarly here? I notice this isn't just a swizzle of the display-p3 channel values for sRGB red.

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
```diff
@@ -791,11 +791,71 @@ var TexImageUtils = (function() {
...
+  /**
+   * Return a list of unpack color spaces to test, supported by the specified
+   * WebGLRenderingContext.
+   */
+  var unpackColorSpacesToTest = function(gl)
+  {
+    if ('unpackColorSpace' in gl)
+      return ['srgb', 'display-p3'];
+    else
+      return [null];
```
kenrussell, line 803: Would it be better to use `undefined` rather than `null` here? (Maybe not.)

sdk/tests/js/webgl-test-utils.js
```diff
@@ -3296,6 +3296,95 @@ function linearChannelToSRGB(value) {
...
+ *            null, which is treated as sRGB
+ *            'srgb'
+ *            'display-p3'.
+ * @return {!Array.<number>} color The color in the specified color
+ *        space as an 8-bit RGBA array with unpremultiplied alpha.
+ */
+var namedColorInColorSpace = function(colorName, colorSpace) {
+  var result;
+  switch (colorSpace) {
+    case null:
```
kenrussell, line 3316: If `null` is changed to `undefined` in `unpackColorSpacesToTest`, above, then this would have to be changed too.

kenrussell (approved):
Thanks @ccameron-chromium for writing this - looks like a solid improvement to start handling color spaces. I verified that this is a no-op in current Chrome Canary on macOS, but didn't test it with experimental web features turned on.

Approving now; a few small comments and questions throughout.

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
greggman: I want to respond constructively. Using the spread operator in 2022 is normal idiomatic JavaScript. Does it really need it described? I get that you didn't know it but now that you do if you browse most modern JavaScript you'll see it everywhere. For example [here's the react docs using it](https://reactjs.org/docs/higher-order-components.html)

If you really feel it's important then sure, keep the comment but then it seems like every use of the spread operator will need to be explained by a comment since you can't assume devs will find this one explanation. 

Is it just this particular one off usage?

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
ccameron-chromium: changed to undefined.

sdk/tests/js/webgl-test-utils.js
ccameron-chromium: done

sdk/tests/js/webgl-test-utils.js
ccameron-chromium: I've added a link (at the top of the function) to the section of CSS color level 4 that goes over the conversion formulas. It may be that we will end up wanting to add functions to do the conversion at some point (it depends on what we end up encountering as we add more tests that cover this functionality).

I also specified the colors involved as CSS colors (in the new syntax that only WebKit supports, but everyone will get there).

sdk/tests/js/tests/tex-image-and-sub-image-utils.js
ccameron-chromium: I'm certainly not helping the situation by using 1 and 2 character variable names almost exclusively :)

I've updated the variable names to be more self-documenting.

sdk/tests/js/tests/tex-image-and-sub-image-2d-with-video.js
kenrussell: Let's call this resolved then.

sdk/tests/js/webgl-test-utils.js
kenrussell: Thanks.

sdk/tests/js/webgl-test-utils.js
kenrussell: Resolved with comment above.

kenrussell (approved):
Looks good!

If run time of these tests appears to be a problem then we can revisit these changes.

PR #3405, opened 2022-05-03
Test case for as-yet-unreported Metal shader bug from Skia team.

kenrussell:
operator/= between a matrix and scalar (float) argument on macOS 10.15
seems to be polyfilled incorrectly by the Metal standard library.


brianosman:
This bug includes a description of the machine and OS version where it happened, as well as a dump from Skia & ANGLE showing the exact GLSL shader that triggered the failure: skbug.com/13290

PR #3406, opened 2022-05-06
Add Occlusion Query Scissor Test

greggman:
The dEQP occlusionquery_strict test failed on Chrome but
not on Safari. This test provokes the same bug in both
Chrome and Safari.

See https://anglebug.com/7157

### DIFF
--- a/sdk/tests/conformance2/query/occlusion-query-scissor.html
+++ b/sdk/tests/conformance2/query/occlusion-query-scissor.html
@@ -1,0 +1,117 @@
+<!--
+Copyright (c) 2022 The Khronos Group Inc.
+Use of this source code is governed by an MIT-style license that can be
+found in the LICENSE.txt file.
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL Occlusion Query w/Scissor Conformance Tests</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/query/occlusion-query-scissor.html
```diff
@@ -0,0 +1,116 @@
+<!--
+Copyright (c) 2019 The Khronos Group Inc.
```
kenrussell, line 2: Please update to 2022.

sdk/tests/conformance2/query/occlusion-query-scissor.html
```diff
@@ -0,0 +1,116 @@
...
+<script src="../../js/webgl-test-utils.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<canvas id="canvas" style="width: 50px; height: 50px;"> </canvas>
+<div id="console"></div>
+<script>
+"use strict";
+description("This test verifies the functionality of occlusion query objects with the scissor test.");
+
```
kenrussell, line 23: Could you add a reference to a bug filed somewhere? For example:

`debug('Regression test for <a href="http://anglebug.com/xyzw">http://anglebug.com/xyzw</a>');`

kenrussell (commented):
Super work, thanks for distilling this into a reproducible test. Before we merge I'd like to ensure the commit message refers to a bug ID, so can I request a small change in that vein?

greggman:
added bug references

kenrussell (approved):
Sorry I dropped this on the floor! The bug reference looks good.

Every 16th iteration is failing on M1, on Chrome Canary with ANGLE's Metal backend. Let's investigate why offline.

PR #3408, opened 2022-05-12
Add drawingBufferColorSpace and unpackColorSpace to constants-and-properties

ccameron-chromium:


kenrussell (approved):
Sorry for taking so long to review this; looks good. While the comments indicate the "ignoredProperties" list was intended for properties added in later versions of the spec, "otherProperties" seems to do the same thing and be more precise.

PR #3409, opened 2022-05-16
Replace RGBA with LUMINANCE in invalid texture format test

sunnyps:
ANGLE is going to support GL_EXT_texture_norm16 on ES2/D3D11 since Chromium's Skia context uses ES2 and R16/RG16 support is needed for texturing from P010 HDR videos. Unfortunately, texture-with-flip-y-and-premultiply-alpha.html expects RGBA/UNSIGNED_SHORT to be an invalid format/type combination which is no longer the case after texture_norm16 is enabled on ES2. Use LUMINANCE/UNSIGNED_SHORT as the expected invalid format/type since that's always invalid irrespective of extensions.

See anglebug.com/7322 for context.

### DIFF
--- a/sdk/tests/conformance/textures/misc/texture-with-flip-y-and-premultiply-alpha.html
+++ b/sdk/tests/conformance/textures/misc/texture-with-flip-y-and-premultiply-alpha.html
@@ -30,8 +30,8 @@ var tex = gl.createTexture();
 gl.bindTexture(gl.TEXTURE_2D, tex);
 
 function testUpload() {
-  // Because WEBGL_depth_texture is enabled, UNSIGNED_SHORT becomes a valid type, but RGBA/UNSIGNED_SHORT is invalid.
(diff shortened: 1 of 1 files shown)


sunnyps:
Let me know if this looks ok - I can probably make ANGLE only enable texture_norm16 on non-WebGL ES2 contexts only if it's a bad idea to enable GL_EXT_texture_norm16 on WebGL1.

kenrussell:
In ANGLE in general, all extensions that can affect WebGL's behavior need to be "enable-able" extensions - not enabled by default. EXT_texture_norm16 in particular is one which is actually exposed at the WebGL level, and since Chrome's using ANGLE on nearly all platforms now, that extension needs to be in that group.

Will it be possible to expose (and enable) EXT_texture_norm16 on non-WebGL contexts, while leaving it as an enable-able extension for WebGL contexts?

sunnyps:
I kinda reached the same conclusion after I made this PR. I've changed the workaround in ANGLE so that it only applies to non-WebGL ES2 contexts. Also, note that EXT_texture_norm16 is only enableable on WebGL2/ES3 not WebGL1, but yes it would be weird if EXT_texture_norm16 was enabled by default on WebGL1 but only requestable on WebGL2. Going to close this.

PR #3410, opened 2022-05-17
Update constants-and-properties-2

ccameron-chromium:
Update constants-and-properties-2 to include drawingBufferColorSpace and unpackColorSpace.


kenrussell (approved):
Looks good.

PR #3411, opened 2022-05-17
Add test of comma operator in constant-expression context.

kenrussell:
In GLSL 3.00 section 12.43 "Sequence operator and constant
expressions" this is disallowed.

Associated with http://crbug.com/1324747 .


cndv3996:
`The rest of the shader code after that line is just boilerplate from my shader-building tool and doesn't add value to the test.`

So, maybe this just good?
`void main() {
    int array[(2, 3)];  // <-- no error
    out_color = vec4(1.0);
}`

johnstiles-google:
Yes, that looks better.

PR #3414, opened 2022-05-20
Add tolerance to multisample-draws-between-blits.html.

kenrussell:
conformance2/renderbuffers/multisample-draws-between-blits.html needs
a tolerance of 1 in its image comparisons on certain GPUs (NVIDIA
specifically); add a tolerance of 2 to accommodate other GPUs.

Follow-on to #3390 and #3412 .

Associated with https://crbug.com/1326332 .


greggman (approved):
lgtm

kenrussell:
Thanks. Merging now so we can remove these test suppressions.

PR #3415, opened 2022-05-21
Fix a variable reference in red-format video-to-texture tests.

kenrussell:
Follow-on to #3404 .

Associated with http://crbug.com/1208480 .


kenrussell:
@ccameron-chromium fyi.

kenrussell:
Found while rolling these tests into Chromium in https://chromium-review.googlesource.com/c/chromium/src/+/3656580 .

greggman (approved):
lgtm

PR #3419, opened 2022-06-01
Test the bias parameter to the texture GLSL function

greggman:
In GLSL there's a function `texture(sampler, texcoords, bias)`.
This test tests that `bias` works and that it is limited to
its advertised limits from `MAX_LOD_BIAS`

I believe ANGLE fails this test on several platforms as it returns a `MAX_LOD_BIAS` of `2.0` but actually allows the bias to go past the limit

### DIFF
--- a/sdk/tests/conformance2/glsl3/00_test_list.txt
+++ b/sdk/tests/conformance2/glsl3/00_test_list.txt
@@ -41,6 +41,7 @@ shader-with-invalid-characters.html
 shader-with-mis-matching-uniform-block.html
 short-circuiting-in-loop-condition.html
 --min-version 2.0.1 switch-case.html
+--min-version 2.0.1 texture-bias.html
 --min-version 2.0.1 texture-offset-non-constant-offset.html
 texture-offset-out-of-range.html
 --min-version 2.0.1 texture-offset-uniform-texture-coordinate.html
--- a/sdk/tests/conformance2/glsl3/texture-bias.html
+++ b/sdk/tests/conformance2/glsl3/texture-bias.html
@@ -1,0 +1,146 @@
+<!--
+Copyright (c) 2022 The Khronos Group Inc.
+Use of this source code is governed by an MIT-style license that can be
+found in the LICENSE.txt file.
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>GLSL texture bias test</title>
+<link rel="stylesheet" href="../../resources/js-test-style.css"/>
+<script src="../../js/js-test-pre.js"></script>
+<script src="../../js/webgl-test-utils.js"></script>
+<script src="../../js/glsl-conformance-test.js"></script>
+</head>
+<body>
+<div id="description"></div>
+<div id="console"></div>
+<script>
+"use strict";
+description("Texture bias should both function and respect limits.");
+
+function runTest(gl) {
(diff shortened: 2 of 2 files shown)


greggman:
I thought I'd mention this test seems to fail on Firefox for a different reason. See

https://bugzilla.mozilla.org/show_bug.cgi?id=1772197

cndv3996:
For a long time havn't worked on conformance tests. Brought back some miserable memories!

I'm trying to understand your algorithms.
Say you draw the LOD level (say 15) to respective mipmaps.

As we go through pixels. This line:
`vec4 texColor = texture(tex, v_uv, (gl_FragCoord.x - 0.5) * biasMult);  // the color we care about`

Will use LOD bias from say 0 to 14?
And 2.0 should be the function really takes in.

And we'll never expect the pixel we read out exceed 3?
Which means we only should read from mipmap #0 to #2? Never lower resolution?

greggman:
> For a long time havn't worked on conformance tests. Brought back some miserable memories!
> 
> I'm trying to understand your algorithms. Say you draw the LOD level (say 15) to respective mipmaps.
> 
> As we go through pixels. This line: `vec4 texColor = texture(tex, v_uv, (gl_FragCoord.x - 0.5) * biasMult); // the color we care about`
> 

As it says in the comments

```
// Draw each mip. Result should be [mip0, mip1, mip2, ...]
```

And

```
// Draw each mip. Result should be [mipMax, mipMax - 1, mipMax - 2, ...]
```

> Will use LOD bias from say 0 to 14? 

yes

> And 2.0 should be the function really takes in.

I don't understand this comment. 

> And we'll never expect the pixel we read out exceed 3? Which means we only should read frommipmap #0 to #2? Never lower resolution?

We should not be able to read past +/- whatever `gl.getParameter(gl.MAX_TEXTURE_LOD_BIAS)` returns for a given computed mip following the lod formula from the spec. In other words, based on the texture coordinates mip level will be chosen. The bias is added after that calculation. The code setups the UV coords first so that calculation will choose the largest mip level

```
// set the UVs so we'd get mip level 0 for every pixel
```

Then it adds biases 0 to maxLod

Then

```
// set the UVs so we'd get highest mip level (the 1x1 level mip) for every pixel
```

then it subtracts biases 0 to maxLod

cndv3996:
Thanks, I think I get it.

Draw 2 times:
1. At mip level 0,
As we draw a pixel to color buffer,  we read first from mip levels (starting from 0) of texture ascendingly.
We'll give biases from 0 to maxLod and even larger as you mentioned.

But texture() should acctually read from 0 to maxLod and not larger levels.

2. Vice Versa.

Great to learn it, thanks!

kenrussell (commented):
Sorry for the long delay reviewing. This test passes in both Chromium and Safari; they're using different backends (OpenGL vs. Metal), so it looks like the failures seen in Firefox Nightly are in the browser and not the test.

kenrussell (approved):

kenrussell:
Note: tested on M1 MacBook Pro.

kenrussell:
Correction: per Kelsey's feedback on https://bugzilla.mozilla.org/show_bug.cgi?id=1772197 , there was a bug in the computation of maxLODs. Fixing that makes the test pass on all browsers.

kenrussell (approved):
The test's now passing on all browsers on the M1. Merging.

PR #3420, opened 2022-06-02
Spec shadow sampler filterability as implementation-defined.

### LINKED ISSUE #3359: Disallowing filtering when using sampler2DShadow?
> The texture is not multisample; either the magnification filter is not NEAREST, or the minification filter is neither NEAREST nor NEAREST_MIPMAP_NEAREST; and...the effective internal format specified for the texture arrays is a sized internal depth or depth and stencil format (see table 8.11), and the value of TEXTURE_COMPARE_MODE is NONE.

Apparently filtering of depth textures is not allowed but most (all?) implementations are allowing it if the sampler is `sampler2DShadow`

https://jsgist.org/?src=1d175927a891af5fdc203a3cccd3964e

Stable Windows/Macos/Firefox/Safari/Chrome all look like this
chrome-macos-sampler2DShadow

iOS Safari they are not filtered when set to LINEAR but they do render

kdashg:
Closes #3359.

### DIFF
--- a/specs/2.0.0/index.html
+++ b/specs/2.0.0/index.html
@@ -3858,16 +3858,6 @@ WebGL2RenderingContext includes WebGL2RenderingContextBase;
            <code>WEBGL_compressed_texture_etc</code>.
     </div>
 
-    <h3><a name="MAYBE_DEPTH_COMPARE_FILTERED">E.g. <code>sampler2DShadow<code> does not guarantee e.g. <code>LINEAR<code> filtering</a></h3>
-    <p>When TEXTURE_COMPARE_MODE is set to COMPARE_REF_TO_TEXTURE, a texture is always complete
-       even if it uses a LINEAR filtering without the required extensions. However, it is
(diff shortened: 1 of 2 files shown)


specs/2.0.0/index.html
```diff
@@ -3858,6 +3858,16 @@ <h3><a name="NO_ETC2_EAC">No ETC2 and EAC compressed texture formats</a></h3>
            <code>WEBGL_compressed_texture_etc</code>.
     </div>
 
+    <h3><a name="MAYBE_DEPTH_COMPARE_FILTERED">E.g. <code>sampler2DShadow<code> does not guarantee e.g. <code>LINEAR<code> filtering</a></h3>
```
kenrussell, line 3861: This new text should be added to specs/latest/2.0/index.html rather than this file - possible to make that change?

kenrussell (approved):
Thanks for adding this - given that the behavior isn't easily queryable in either OpenGL ES or Vulkan, I'd like to make this spec change. @greggman, what do you think?

kenrussell:
Moved the spec update to the top-of-tree 2.0/index.html; merging.

PR #3421, opened 2022-06-03
Warn in two video tests if the tab was backgrounded.

kenrussell:
There is a hypothesis that this is occurring on some platforms when
running the conformance tests in an automated test harness. Attempt to
confirm this.

Associated with http://crbug.com/1293967 .


kenrussell:
@greggman or @kdashg : please review. Thanks.

greggman (approved):
lgtm

does the testing infra record console.log or does it record DOM output (or both)

kdashg (approved):

PR #3422, opened 2022-06-03
Max uniform blocks

greggman:


kenrussell:
> This tests actually using the maximum number of uniform blocks. I didn't see any tests that do this. I tested on MacOS M1, MacOS Intel, iOS, Windows NVidia in both Firefox and Chrome and didn't see any issues.

Is there an ANGLE or Chromium bug that the commit message can reference?

greggman:
no there's no bug. it just wasn't tested AFAICT.

kenrussell:
Sounds fine. Still referenced https://anglebug.com/7280 in the commit message.

kenrussell:
Attempting to roll forward WebGL conformance in Chromium to incorporate this test, it's failing on Android, macOS and Windows:

https://ci.chromium.org/ui/p/chromium/builders/try/android_optional_gpu_tests_rel/b8809128773665191121/overview
https://ci.chromium.org/ui/p/chromium/builders/try/mac_optional_gpu_tests_rel/84568/overview
https://ci.chromium.org/ui/p/chromium/builders/try/win_optional_gpu_tests_rel/86989/overview

The test seems incorrect in that it assumes the shader compilation and program linking succeed. Per specification, these are allowed to fail for essentially any reason. It does seem that there are bugs in implementations - an assert is hit in ANGLE's D3D backend which is the cause of the Windows failure - but given that the test has this issue, I'm going to revert the test while this is investigated asynchronously.

PR #3423, opened 2022-06-04
Use SVG logos

lexaknyazev:


kdashg (approved):

kenrussell:
Thanks, looks good in Firefox, Safari and Chrome.

Wondering whether we should update image references in the most recent spec snapshots too, since the CSS picks up the higher-resolution Khronos Specification sidebar but not the logos.

kenrussell (approved):

PR #3424, opened 2022-06-06
Add more cases to invalidate-framebuffer test

greggman:
These new cases fail on the Metal backend

See Issue: https://anglebug.com/7397


kenrussell (approved):
Nice work on these.

Could you please edit the commit message and point to a bug that this PR is associated with? For example,

"Associated with http://crbug.com/angleproject/xyzw ."

Thanks.

PR #3183, opened 2020-11-17
Verify that WebGL's implicit clears obey the rasterizer discard state.

kenrussell:
Regression test for http://crbug.com/1149176 .

PR #3430, opened 2022-06-08
Test compositing with webgl-multi-draw-instanced-base-vertex-base-instance

greggman:
See:
http://crbug.com/1333620
https://bugs.webkit.org/show_bug.cgi?id=241347

I think this tests this feature. At least locally the default draw functions pass and the extension draw functions do not.

### DIFF
--- a/sdk/tests/conformance/extensions/webgl-multi-draw.html
+++ b/sdk/tests/conformance/extensions/webgl-multi-draw.html
@@ -7,6 +7,7 @@
 <script src="../../js/desktop-gl-constants.js"></script>
 <script src="../../js/js-test-pre.js"></script>
 <script src="../../js/webgl-test-utils.js"></script>
+<script src="../../js/tests/compositing-test.js"></script>
 </head>
 <body>
 <script id="vshaderIllegalDrawID" type="x-shader/x-vertex">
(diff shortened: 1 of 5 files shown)


lexaknyazev:
Confirmed that new tests in `WEBGL_multi_draw` fail on top-of-tree WebKit as it has the same implementation issue as Chromium.

greggman:
I will update with more tests for ANGLE_instance_arrays

lexaknyazev:
Seems like another related test was added in #3183. Should its assertions be ported here as well? Also, we should ensure that `drawRangeElements` is covered.

lexaknyazev:
Seems like another related test was added in #3183. Should its assertions be ported here as well? Also, we should ensure that `drawRangeElements` is covered.

greggman:
Ok, so I think the test is correct now. I saw some strange behavior in that it looks like sometimes no draw is happneing. One of the tests does

1. draw entire canvas red
2. read-pixels verify red
3. wait for composite
4. check the canvas is cleared (preserve is false)

The test fails on most webgl-multi-draw and newer extensions. I thought the issue was that it was not drawing but, debugging the code, the issue is whatever those functions are doing their bookeeping is wrong and between steps 1 and 2 they are clearing the canvas 😱

This happens on the 2nd test in those contexts. Meaning if there are 5 tests, the first will pass, the rest will fail because of this erroneous clear. If remove the first test the same pattern (first test passes, rest fail)

I'm guessing making the extension draw functions match the non-extension draw functions will fix this issue.

I moved the compositing test into a separate file (compositing-test.js) and then made all the compositing tests and extension that need to test compositing use it.

Added tests for 

* ANGLE_instanced_arrays
* WebGL2 (drawArraysInstanced, drawElementInstanced, drawRangeElements) 
* WEBGL_multi_draw
* WEBGL_draw_instanced_base_vertex_base_instance
* WEBGL_multi_draw_instanced_base_vertex_base_instance

sdk/tests/js/tests/compositing-test.js
```diff
@@ -0,0 +1,136 @@
...
+    await testPreserveDrawingBufferFalse(glPreserveDrawingBufferFalse, drawFn, false);
+    await testPreserveDrawingBufferFalse(glPreserveDrawingBufferFalse, drawFn, true);
+
+    await testPreserveDrawingBufferTrue(glPreserveDrawingBufferTrue, drawFn, false);
+    await testPreserveDrawingBufferTrue(glPreserveDrawingBufferTrue, drawFn, true);
+  };
+}
+
+return createCompositingTestFn;
+}());
```
kenrussell, line 136: Add newline at end?

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
```diff
@@ -896,10 +897,107 @@
...
+    function drawArraysInstancedBaseInstanceWEBGL(gl) {
+      const ext = gl.getExtension('WEBGL_draw_instanced_base_vertex_base_instance');
+      if (!ext) {
+        return true;
+      }
+
+      ext.drawArraysInstancedBaseInstanceWEBGL(gl.TRIANGLES, 0, 6, 1, 0);
+    }
+
+    function drawElementsInstancedBaseInstanceWEBGL(gl) {
```
kenrussell, line 948: Is it worth making this match the name of the draw call being invoked (drawElementsInstancedBaseVertexBaseInstanceWEBGL)?

kenrussell (approved):
Looks great @greggman . Couple of small questions - please feel free to merge yourself. (I prefer the "squash and merge" workflow because it preserves the pull request numbers in the commit.)

sdk/tests/conformance/extensions/webgl-multi-draw.html
```diff
@@ -846,10 +847,181 @@
   }
 }
 
-runTest();
+function waitForComposite() {
+  debug('wait for composite');
+  return new Promise(resolve => wtu.waitForComposite(resolve));
```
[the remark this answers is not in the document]
greggman: I have not seen it be intermittent but ... the canvas has to be visible. That's why I set it to `position: fixed; top: 0`, so that regardless of where the page is scrolled the canvas is visible.  Before I added that it would not repo if offscreen.

PR #3431, opened 2022-06-08
Clarify working of EXT_texture_norm16, and expand tests.

### LINKED ISSUE #3417: Clarify EXT_texture_norm16
- The WebGL extension spec lists support for `UNSIGNED_SHORT` textures as FBO attachments as a WebGL-specific feature although this seems to be already included in the GLES extension.
- The WebGL extension spec lists only `texImage2D` and `texSubImage2D` as the extended entrypoints while the GLES extension explicitly includes also 3D, immutable, and renderbuffer functions. To avoid any ambiguity, the WebGL extension should either only link to the GLES extension or have the same explicit list. On top of that, the missing functions are not covered by the CTS.

kdashg:
Closes #3417.

In test:
* Forbid RGBA/RGBA/UNSIGNED_SHORT (and SHORT)
* Forbid renderability of RGB16.
* Add more/better debug info for what each subtest is testing.

### DIFF
--- a/sdk/tests/conformance2/extensions/ext-texture-norm16.html
+++ b/sdk/tests/conformance2/extensions/ext-texture-norm16.html
@@ -49,12 +49,13 @@ function generateFormatColor(format, value, alpha) {
   }
 }
 
-function testNorm16Texture(internalFormat, format, type) {
+function testNorm16Texture(internalFormat, format, type, error="NO_ERROR") {
+  debug(`\ntestNorm16Texture(${[].slice.call(arguments).join(", ")})`);
   let pixelValue;
   let expectedValue;
   let imageData;
 
-  switch(type) {
+  switch(gl[type]) {
     case gl.SHORT:
       pixelValue = 0x7fff;
       expectedValue = 0xff;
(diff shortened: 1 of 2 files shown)


sdk/tests/conformance2/extensions/ext-texture-norm16.html
```diff
@@ -73,14 +74,16 @@
   // Texture sampled from
   gl.activeTexture(gl.TEXTURE0);
   gl.bindTexture(gl.TEXTURE_2D, textures[0]);
-  gl.texImage2D(gl.TEXTURE_2D, 0, internalFormat, 1, 1, 0, format, type, imageData);
+  gl.texImage2D(gl.TEXTURE_2D, 0, ext[internalFormat] || gl[internalFormat],
+        1, 1, 0, gl[format], gl[type], imageData);
 
-  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "texture bindings succeed");
+  wtu.glErrorShouldBe(gl, gl[error], `texImage should generate error:${error}`);
```
lexaknyazev, line 80: To ensure that implementations enable the extension for all affected entry points, could other `texImage` functions be added here and throughout? Could we also test `renderbufferStorageMultisample` in `testNorm16Render`?

kdashg: I spent a few minutes on it but it was taking more time than I have, so follow-ups welcome.

kenrussell (approved):
The new restrictions seem reasonable. The revised tests pass in both WebKit and Chromium; Firefox Nightly doesn't currently expose EXT_texture_norm16. Follow-ups welcome.

PR #3432, opened 2022-06-08
Clarify blendColor clamps iff unextended webgl1.

### LINKED ISSUE #3413: Revisit constant blend color clamping test
The `conformance/rendering/blending.html` test and ANGLE do not agree whether the constant blend color should be clamped when floating-point color buffers are not supported.

Specifically, ANGLE clamps the color value on store on non-extended ES 2.0 contexts (likely following the example [here](https://chromium-review.googlesource.com/c/angle/angle/+/1541723/comments/36d38fdf_067eb122)). While the WebGL test has different expectations.

/cc @kdashg @kenrussell

kdashg:
Update test to test this.
s/GLclampf/GLfloat/ for blendColor args.
Resolves #3413.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -2034,8 +2034,10 @@ WebGLRenderingContext includes WebGLRenderingContextOverloads;
     <dl class="methods">
         <dt class="idl-code">void activeTexture(GLenum texture)
             <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/2.0/es_full_spec_2.0.25.pdf#nameddest=section-3.7">OpenGL ES 2.0 &sect;3.7</a>, <a class="nonnormative" href="http://www.khronos.org/opengles/sdk/2.0/docs/man/xhtml/glActiveTexture.xml">man page</a>)</span>
-        <dt class="idl-code">void blendColor(GLclampf red, GLclampf green, GLclampf blue, GLclampf alpha)
+        <dt class="idl-code">void blendColor(GLfloat red, GLfloat green, GLfloat blue, GLfloat alpha)
             <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/2.0/es_full_spec_2.0.25.pdf#nameddest=section-4.1.6">OpenGL ES 2.0 &sect;4.1.6</a>, <a class="nonnormative" href="http://www.khronos.org/opengles/sdk/2.0/docs/man/xhtml/glBlendColor.xml">man page</a>)</span>
+            <dd>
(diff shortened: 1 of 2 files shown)


specs/latest/1.0/index.html
```diff
@@ -2034,8 +2034,10 @@ <h4>Setting and getting state</h4>
     <dl class="methods">
         <dt class="idl-code">void activeTexture(GLenum texture)
             <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/2.0/es_full_spec_2.0.25.pdf#nameddest=section-3.7">OpenGL ES 2.0 &sect;3.7</a>, <a class="nonnormative" href="http://www.khronos.org/opengles/sdk/2.0/docs/man/xhtml/glActiveTexture.xml">man page</a>)</span>
-        <dt class="idl-code">void blendColor(GLclampf red, GLclampf green, GLclampf blue, GLclampf alpha)
+        <dt class="idl-code">void blendColor(GLfloat red, GLfloat green, GLfloat blue, GLfloat alpha)
             <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/2.0/es_full_spec_2.0.25.pdf#nameddest=section-4.1.6">OpenGL ES 2.0 &sect;4.1.6</a>, <a class="nonnormative" href="http://www.khronos.org/opengles/sdk/2.0/docs/man/xhtml/glBlendColor.xml">man page</a>)</span>
+            <dd>
+                <code>blendColor</code> clamps its arguments to the range 0 to 1 in WebGL 1, unless EXT_color_buffer_half_float or WEBGL_color_buffer_float are enabled.
```
lexaknyazev, line 2040: For consistency, please consider using "WebGL 1.0" instead of "WebGL 1".

lexaknyazev (commented):
Since this test is exclusively focused on constant blend color clamping, could we rename it to something more specific than `blending.html` to make test expectations and results in implementations' CI more descriptive?

kdashg:
> Since this test is exclusively focused on constant blend color clamping, could we rename it to something more specific than `blending.html` to make test expectations and results in implementations' CI more descriptive?

Maybe, but getting this perfect is low priority for me. PRs welcome!

lexaknyazev:
https://github.com/KhronosGroup/WebGL/blob/2f1b7f49963614451b205a93e10ba643cce60e50/sdk/tests/conformance/rendering/blending.html#L163-L171

Should this part use `EXT_color_buffer_half_float` instead?

lexaknyazev:
Both RGBA16F and RGBA32F cases fail on WebGL 1.0 because the blend color is not re-set after the extensions are enabled.

lexaknyazev (approved):
LGTM. Will follow-up with the remaining fixes.

kenrussell:
Merging per agreement in yesterday's WebGL concall. @lexaknyazev will follow up with a revision for the WebGL 1.0 + extension cases.

Issue #3434, opened 2022-06-10
Enhance CTS for missing VAO bindings


lexaknyazev:
From WebGL 1.0, Section 6.2:
> If a vertex attribute is enabled as an array via `enableVertexAttribArray` but no buffer is bound to that attribute (generally via `bindBuffer` and `vertexAttribPointer`), then draw commands (`drawArrays` or `drawElements`) will generate an `INVALID_OPERATION` error.

The CTS should check that all draw commands enforce this condition: `drawArrays`, `drawElements`, `drawArraysInstancedANGLE`, `drawElementsInstancedANGLE`, `drawArraysInstanced`, `drawElementsInstanced`, `drawRangeElements`, `multiDraw*`.

PR #3437, opened 2022-06-13
Test all draw functions fail if attributes are invalid

greggman:
#3434 

ATM this test only tests if an attribute is enabled but
no buffer is bound that each draw function generates
INVALID_OPERATION

Not sure what else needs to be test. Maybe mis-matching attributes (an ivec4 attribute in shader but float attribute in vao)

### DIFF
--- a/sdk/tests/conformance/extensions/webgl-multi-draw.html
+++ b/sdk/tests/conformance/extensions/webgl-multi-draw.html
@@ -8,6 +8,7 @@
 <script src="../../js/js-test-pre.js"></script>
 <script src="../../js/webgl-test-utils.js"></script>
 <script src="../../js/tests/compositing-test.js"></script>
+<script src="../../js/tests/invalid-vertex-attrib-test.js"></script>
 </head>
 <body>
 <script id="vshaderIllegalDrawID" type="x-shader/x-vertex">
@@ -910,27 +911,7 @@ async function doCompositingTests([glPreserveDrawingBufferFalse, glPreserveDrawi
   await testPreserveDrawingBufferTrue(glPreserveDrawingBufferTrue, drawFn);
 }
 
-async function runCompositingTests() {
-  const compositingTestFn = createCompositingTestFn({
-    webglVersion: 1,
-    shadersFn(gl) {
-      const vs = `\
-      //#extension GL_ANGLE_multi_draw : enable
-      attribute vec4 position;
-      void main() {
-        gl_Position = position;
-      }
-      `;
-      const fs = `\
-      precision mediump float;
-      void main() {
-        gl_FragColor = vec4(1, 0, 0, 1);
-      }
-      `;
-      return [vs, fs];
-    },
-  });
-
+async function runDrawTests(testFn) {
   function drawArrays(gl) {
     gl.drawArrays(gl.TRIANGLES, 0, 6);
   }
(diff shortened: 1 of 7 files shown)


sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
```diff
@@ -979,19 +959,50 @@
...
-    await compositingTestFn(drawArrays);             // sanity check
-    await compositingTestFn(drawElements);           // sanity check
-    await compositingTestFn(drawArraysInstanced);    // sanity check
-    await compositingTestFn(drawElementsInstanced);  // sanity check
+    await testFn(drawArrays);             // sanity check
+    await testFn(drawElements);           // sanity check
+    await testFn(drawArraysInstanced);    // sanity check
+    await testFn(drawElementsInstanced);  // sanity check
+
+    await testFn(drawArraysInstancedBaseInstanceWEBGL);
```
kenrussell, line 967: These have to be short-circuited at the top level in order for the test to pass when the extension isn't available, because the invalid vertex attrib tests expect the drawFn to have an effect.

sdk/tests/conformance/extensions/webgl-multi-draw.html
```diff
@@ -1005,17 +986,46 @@
     );
   }
 
-  await compositingTestFn(drawArrays);             // sanity check
-  await compositingTestFn(drawElements);           // sanity check
+  await testFn(drawArrays);             // sanity check
+  await testFn(drawElements);           // sanity check
+
+  await testFn(multiDrawArraysWEBGL);
```
kenrussell, line 992: These have to be short-circuited at the top level in order for the test to pass when the extension isn't available, because the invalid vertex attrib tests expect the drawFn to have an effect.

kenrussell (commented):
Nice work @greggman figuring out how to refactor these tests to apply the invalid vertex attribute tests to all draw functions.

In testing this locally I found a couple of bugs that happen when extensions weren't supported. I'll push a commit to this PR shortly which fixes these, the merge conflicts, and one other bug, for you to review.

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
```diff
@@ -948,24 +928,24 @@
     function drawElementsInstancedBaseVertexBaseInstanceWEBGL(gl) {
       const ext = gl.getExtension('WEBGL_draw_instanced_base_vertex_base_instance');
       if (!ext) {
-        return true;
+        throw 'Should not have run this test without WEBGL_draw_instanced_base_vertex_base_instance';
       }
 
-      ext.drawElementsInstancedBaseVertexBaseInstanceWEBGL(gl.TRIANGLES, 6, gl.UNSIGNED_BYTE, 0, 1, 0, 1);
+      ext.drawElementsInstancedBaseVertexBaseInstanceWEBGL(gl.TRIANGLES, 6, gl.UNSIGNED_BYTE, 0, 1, 0, 0);
```
kenrussell, line 934: Note: it looked to me like the original code here was wrong - specifying 1 rather than 0 for the base instance. However, the tests pass with either version of this code.

kenrussell (approved):
Thanks again @greggman for this improvement and apologies for taking so long to review it. These revised tests pass in Chrome, Firefox and Safari, modulo a couple of WEBGL_multi_draw failures in Safari Technology Preview which I believe @lexaknyazev already fixed in https://github.com/WebKit/WebKit/pull/1452 .

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
```diff
@@ -979,19 +959,50 @@
...
-    await compositingTestFn(multiDrawArraysInstancedBaseInstanceWEBGL);
-    await compositingTestFn(multiDrawElementsInstancedBaseVertexBaseInstanceWEBGL);
+async function runInvalidAttribTests(gl) {
+  const invalidAttribTestFn = createInvalidAttribTestFn(gl);
+  await runDrawTests(invalidAttribTestFn);
 }
 
 async function main() {
-  runTest();
+  //runTest();
```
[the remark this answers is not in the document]
kenrussell: Fixed in my revision to the PR.

sdk/tests/conformance/extensions/webgl-multi-draw.html
kenrussell: Fixed in my revision to the PR.

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
kenrussell: Fixed in my revision of the PR.

sdk/tests/conformance2/extensions/webgl-multi-draw-instanced-base-vertex-base-instance.html
greggman: lgtm your way

greggman (commented):
lgtm

PR #3438, opened 2022-06-19
Expand clearBuffer tests

lexaknyazev:
- Added a few more cases to this test, covering depth and stencil clears.
- Added additional test runs with scissor enabled.

Passes in Firefox, various failures in Chromium and WebKit.

WebKit fix: https://github.com/WebKit/WebKit/pull/1625


kenrussell (approved):
Looks good to me.

Verified that the needed code changes, when ported from WebKit to Chromium, fix the new test cases.

PR #3442, opened 2022-06-23
Propose extension EXT_provoking_vertex.

kdashg:


extensions/proposals/EXT_provoking_vertex/extension.xml
```diff
@@ -0,0 +1,57 @@
...
+  <contact>
+    <a href="https://www.khronos.org/webgl/public-mailing-list/">WebGL working group</a> (public_webgl 'at' khronos.org)
+  </contact>
+  <contributors>
+    <contributor>Kelsey Gilbert (Mozilla)</contributor>
+    <contributor>Members of the WebGL working group</contributor>
+  </contributors>
+  <number>48</number>
+  <depends>
+    <api version="1.0"/>
```
kenrussell, line 13: Should this depend on 2.0 rather than 1.0, since flat shading was only introduced in ESSL 300?

kenrussell (approved):
Looks fine to me.

I wonder whether we should call this WEBGL_provoking_vertex and diverge a bit - provide an API giving a hint about which is more efficient on the current backend?

extensions/proposals/EXT_provoking_vertex/extension.xml
kenrussell: Discussed on today's call - I'll try to edit in place.

kdashg, line 13:
```suggestion
    <api version="2.0"/>
```

kdashg: done!

kenrussell:
Thanks! Merging now.

PR #3443, opened 2022-06-27
Test side-effects of WEBGL_multi_draw

lexaknyazev:
The spec requires that `WEBGL_multi_draw` implicitly enables `ANGLE_instanced_arrays`. Test the only JS-visible behavior change.

### DIFF
--- a/sdk/tests/conformance/extensions/webgl-multi-draw.html
+++ b/sdk/tests/conformance/extensions/webgl-multi-draw.html
@@ -1014,8 +1014,30 @@ async function runCompositingTests() {
   await compositingTestFn(multiDrawElementsInstancedWEBGL);
 }
 
+function testSideEffects() {
+  debug("")
+  debug("Testing that ANGLE_instanced_arrays is implicitly enabled")
+  const canvas = document.createElement("canvas");
+  const gl = wtu.create3DContext(canvas, undefined, 1);
+  if (!gl) {
+    testFailed('WebGL context creation failed');
+    return;
+  }
+  gl.enableVertexAttribArray(0);
+  const VERTEX_ATTRIB_ARRAY_DIVISOR_ANGLE = 0x88FE;
+  gl.getVertexAttrib(0, VERTEX_ATTRIB_ARRAY_DIVISOR_ANGLE);
+  wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "divisor enum unknown");
+  const ext = gl.getExtension("WEBGL_multi_draw");
+  if (ext) {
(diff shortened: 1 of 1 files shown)


sdk/tests/conformance/extensions/webgl-multi-draw.html
```diff
@@ -1014,8 +1014,29 @@
...
+  debug("")
+  debug("Testing that ANGLE_instanced_arrays is implicitly enabled")
+  const canvas = document.createElement("canvas");
+  const gl = wtu.create3DContext(canvas, undefined, 1);
+  if (!gl) {
+    testFailed('WebGL context creation failed');
+    return;
+  }
+  gl.enableVertexAttribArray(0);
+  gl.getVertexAttrib(0, 0x88FE /* VERTEX_ATTRIB_ARRAY_DIVISOR_ANGLE */);
```
kkinnunen-apple, line 1027: suggestion: if you want to avoid the argument comment 
```
var local_VERTEX_ATTRIB_ARRAY_DIVISOR_ANGLE = 0x88FE;
```

sdk/tests/conformance/extensions/webgl-multi-draw.html
```diff
@@ -1014,8 +1014,29 @@
...
+  wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "divisor enum unknown");
+  const ext = gl.getExtension("WEBGL_multi_draw");
+  if (ext) {
+    debug("WEBGL_multi_draw enabled");
+    gl.getVertexAttrib(0, 0x88FE /* VERTEX_ATTRIB_ARRAY_DIVISOR_ANGLE */);
```
kkinnunen-apple, line 1032: Suggestion if you want to avoid the argument comment:
```
var ext = ...
shouldBe("local_VERTEX_ATTRIB_ARRAY_DIVISOR_ANGLE", "ext.VERTEX_ATTRIB_ARRAY_DIVISOR_ANGLE");
gl.getVertexAttrib(0, ext.VERTEX_ATTRIB_ARRAY_DIVISOR_ANGLE);
```

kkinnunen-apple (approved):

kenrussell (commented):
In the future could you please link to bugs filed against browser implementations in the pull request description, to make it easier to keep track of what may need to be updated in order to pass more rigorous tests?

In this case, the WebKit bug filed was:
Implicit extension enabling should use WebGL frontend
https://bugs.webkit.org/show_bug.cgi?id=241918

and the Chromium-side bug just filed was:
WebGL extensions must enable dependent ones
https://crbug.com/1343108

Thanks.

PR #3444, opened 2022-06-29
Expand & fix exif-orientation test

lexaknyazev:
- The `imageOrientation` option is a string-based enum, not a boolean; it's now supported everywhere.
- Added tests for uploading `ImageBitmap` objects created from `Image` elements.

Passes in Firefox, various failures in WebKit and Chromium.


sdk/tests/conformance/textures/misc/exif-orientation.html
```diff
@@ -56,24 +56,9 @@
...
-        try {
-            bitmap = await createImageBitmap(blob, {imageOrientation: flipY});
-        } catch (e) {
-            output("  (createImageBitmap options not supported - skipping flipY=true case)");
-            return;
-        }
-    } else {
-        bitmap = await createImageBitmap(blob);
-    }
+    const bitmap = await createImageBitmap(source, {imageOrientation: flipY ? 'flipY' : 'none'});
```
kkinnunen-apple, line 61: you could also do
`flipY ? {imageOrientation: flipY } : undefined`

sdk/tests/conformance/textures/misc/exif-orientation.html
```diff
@@ -93,8 +78,14 @@
     let blob = await response.blob();
     output("----------------------------------------------------------------");
     output("Testing " + filename + " via ImageBitmap from Blob");
-    await testImageBitmapFromBlobWithFlipY(blob, true);
-    await testImageBitmapFromBlobWithFlipY(blob, false);
+    await testImageBitmapWithFlipY(blob, true);
```
kkinnunen-apple, line 81: in these you would do
await testImageBitmapWithFlipY(blob, undefined);
await testImageBitmapWithFlipY(blob, "flipY");
await testImageBitmapWithFlipY(blob, "none");

lexaknyazev: Done.

kenrussell (approved):
Apologies for the long delay approving this. I'd hoped to fix the underlying bug in Chromium before strengthening this test but that hasn't been possible.

PR #3446, opened 2022-06-30
Split shader-with-non-reserved-words test

greggman:
Apparently this test took > 4 minutes on Windows debug builds
as it created > 1600 shaders.

I was originally going to split it into HLSL, GLSL 4.20,
GLSL 4.20 future but the majorty of words are HLSL so that
didn't really help.

I opted to just keep the test mostly as is and just split it
by N parts.

See: https://crbug.com/1337430


kenrussell (approved):
Thanks @greggman for doing this. Looks good - I checked the slicing math in shader-with-non-reserved-words.js.

PR #3451, opened 2022-07-02
Move script tag from head to body.

kenrussell:
In the tests:
  conformance/rendering/color-mask-preserved-during-implicit-clears.html
  conformance/rendering/draw-webgl-to-canvas-2d-repeatedly.html

on very slow devices, it seems that the script tag can run before the
document is fully loaded, causing the description() call to fail to
look up the named div.

Associated with https://crbug.com/1341423 .


kenrussell:
@kdashg and/or @greggman : please review. Thanks.

Tested these changes locally; the tests still run as expected.

kdashg (approved):

PR #3454, opened 2022-07-05
Relax texture error codes

### LINKED ISSUE #3453: Specify copyTexImage errors
The OpenGL ES 3.0 spec does not precisely define error codes for unsupported source/target combinations. The updated ES 3.2 suggests generating `INVALID_OPERATION` for not guaranteed conversions and `INVALID_ENUM` if "an invalid value is specified for _internalformat_".

On top of that, [ES 3.0 ref page](https://www.khronos.org/registry/OpenGL-Refpages/es3.0/html/glCopyTexImage2D.xhtml) provides a list of allowed internal formats with `INVALID_ENUM` error for others, while dEQP seems to expect `INVALID_OPERATION`.

The immediate issue is ANGLE returning `INVALID_OPERATION` for SNORM target formats while `conformance2/textures/misc/copy-texture-image.html` expects `INVALID_ENUM`.

lexaknyazev:
ANGLE returns `INVALID_VALUE` only for never-valid `internalformat` values regardless of enabled extensions.

Fixes #3453.

### DIFF
--- a/sdk/tests/conformance/extensions/ext-texture-compression-bptc.html
+++ b/sdk/tests/conformance/extensions/ext-texture-compression-bptc.html
@@ -115,17 +115,17 @@ function runTestExtension() {
     gl.bindTexture(gl.TEXTURE_2D, tex);
 
     gl.texImage2D(gl.TEXTURE_2D, 0, ext.COMPRESSED_RGBA_BPTC_UNORM_EXT, 4, 4, 0, gl.RGBA, gl.UNSIGNED_BYTE, null);
-    wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "COMPRESSED_RGBA_BPTC_UNORM_EXT fails with texImage2D");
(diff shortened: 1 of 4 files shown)


lexaknyazev:
`conformance2/textures/misc/tex-input-validation.html` has two similar assertions for `texImage`:
```
  { target: gl.TEXTURE_2D,
    internalFormat: 0x1903, // GL_RED
    border: 0,
    format: 0x1903, // GL_RED
    type: gl.UNSIGNED_BYTE,
    expectedError: [gl.INVALID_ENUM, gl.INVALID_VALUE] },
```
and
```
  { target: gl.TEXTURE_3D,
    internalFormat: gl.RG,
    border: 0,
    format: gl.RGBA,
    type: gl.UNSIGNED_BYTE,
    expectedError: gl.INVALID_VALUE},
```

ANGLE may support `EXT_texture_rg` so it does not generate `INVALID_VALUE` for `GL_RED` and `GL_RG` used as internal format. Both cases above produce only `INVALID_OPERATION` there.

kenrussell (approved):
Sounds fine. Might want to consider backporting these to the most recent snapshots in `conformance-suites/`.

PR #3455, opened 2022-07-06
Expand FBO DS attachment test

lexaknyazev:
Assert all three attachment points after each state change.


kenrussell (approved):
Looks good. Passes in Chrome and Firefox; presumably also in WebKit with the fix for https://bugs.webkit.org/show_bug.cgi?id=223337 .

PR #3458, opened 2022-07-10
Tighten OES_vertex_array_object spec and tests

lexaknyazev:
Specify and test the same object lifecycle semantics as used everywhere else. The new test passes in Firefox, fails in WebKit and Chromium.

https://bugs.webkit.org/show_bug.cgi?id=242573


kenrussell:
I'm reviewing this now - please hold off merging. Thanks.

kenrussell (approved):
LGTM. Verified that the appropriate implementation fixes make the new tests pass.

PR #3460, opened 2022-07-11
Revert "Test using the maximum number of uniform blocks. (#3422)"

kenrussell:
This reverts commit ce7c74d0a6cc00934705602a43cef1d5898a6c52.

Per comment on #3422, this test is assuming that the shader
compilation and program linking succeed. This is not guaranteed per
specification. The test has been observed to fail on Android, macOS
and Windows - on some platforms possibly due to implementation bugs,
but this should be investigated before re-integration.


kenrussell:
Merging myself to unblock browsers' successful integrations of the top-of-tree WebGL conformance suite.

PR #3462, opened 2022-07-12
Expand transform feedback tests

lexaknyazev:
Fails in WebKit & Chromium. Passes in Firefox.

WebKit bug: https://bugs.webkit.org/show_bug.cgi?id=242632
WebKit PR: https://github.com/WebKit/WebKit/pull/2335

Chromium bug: https://crbug.com/1343722


kenrussell (approved):
LGTM. Confirmed locally that a small change to Chromium's validation - which we agree with - makes all of the new tests pass.

PR #3463, opened 2022-07-12
Log visibility upon load in two video tests.

kenrussell:
The visibility of these tests isn't changing at runtime; log the
visibility state upon load as well.

Follow-on to #3421.

Associated with http://crbug.com/1293967 .


greggman (commented):
lgtm

fyi: in the future you can also use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) which is more idomatic for JS in 2022

PR #3464, opened 2022-07-16
Fix & expand blend color clamping test

lexaknyazev:
This PR updates the test to match the spec. Notable failures:
1. Firefox does not clamp constant blend color on store on unextended WebGL 1.0.
2. Adreno drivers clamp constant blend color internally even on ES 3.2 contexts and correctly reflect that via `getFloatv(GL_BLEND_COLOR, ...)` query.

Instead of permanently suppressing the test on [2], we may consider making this behavior implementation-dependent.

### DIFF
--- a/extensions/EXT_color_buffer_float/extension.xml
+++ b/extensions/EXT_color_buffer_float/extension.xml
@@ -53,10 +53,6 @@
         <li>Fragment shader outputs to buffers with these internal formats are
         not clamped.</li>
 
-        <li>Colors specified with <code>clearColor</code> and
-        <code>blendColor</code> are not clamped when applied to buffers with
-        these internal formats.</li>
-
         <li>The format and type combination <code>RGBA</code> and
         <code>UNSIGNED_BYTE</code> cannot be used for reading from a
         floating-point color buffer.</li>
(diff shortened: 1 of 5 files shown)


kdashg (approved):

lexaknyazev:
Given the Adreno driver bug (reproducible on Android 12), we have three options:
1. Suppress the test permanently there (the worst option).
2. Restrict the WebGL specs so that constant blend color is always clamped (functional regression compared to the current state).
3. Make it implementation-defined and only require `gl.getParameter(gl.BLEND_COLOR)` result to be consistent with rendering, i.e., clamp on store if the driver does the same.

kenrussell:
I'd prefer (3) - given where we are in WebGL's lifecycle, let's avoid breaking existing working web sites. Is it feasible to test this behavior? I suspect that Chromium caches the user's specified blend color rather than re-querying it from the driver (though we can fix that if so).

lexaknyazev:
> Is it feasible to test this behavior?

Seems simple: set a large value, query it back, and draw something. If the state query returns unclamped value, expect that it was also unclamped during rendering. I can update the specs and this test to assert that. ANGLE update would be trivial.

kenrussell (approved):
Looks good. This new behavior is easily explainable to end users and seems readily implementable.

kenrussell:
BTW, I verified that the WebGL 1.0 and 2.0 versions of conformance/rendering/blending.html pass in Chromium and WebKit on macOS on a M1 GPU.

kdashg:
@lexaknyazev This is just blocked by the merge conflict

kenrussell:
Kelsey seems busy, and this change has been verified to fix the issue on affected GPUs, so merging now.

PR #3467, opened 2022-08-02
Add HEVC version of transparent video test

lexaknyazev:
WebKit does not support alpha channel in WebM videos.


kenrussell (approved):
Looks good to me.

Does WebKit not support WebM at all, or does it not support transparent WebM? The latter would sound more difficult to feature-detect.

lexaknyazev:
- WebKit does support non-transparent WebM in official builds, alpha channel is ignored.
- When both WebM and HEVC video sources are available, it chooses the latter.
- The goal of this test should be to assert alpha channel in video texture uploads in general, so having extra flexibility around supported codecs seems fine.

kenrussell:
More flexibility sounds fine. I'd like to ensure the commit message includes your first two bullet points, so if I could squash and merge (and adjust the commit message) after @kkinnunen-apple approves I'd appreciate it.

kkinnunen-apple (approved):
Thanks!

PR #3469, opened 2022-08-04
Specify which operations are "Draw Operations" and "Read Operations".

### LINKED ISSUE #3436: Clarify implicit clears with blitFramebuffer and clearBuffer*
From WebGL 1.0, Section 2.2:
> WebGL presents its drawing buffer to the HTML page compositor immediately before a compositing operation, but only if at least one of the following has occurred since the previous compositing operation:
>
> - Context creation
> - Canvas resize
> - `clear`, `drawArrays`, or `drawElements` has been called while the drawing buffer is the currently bound framebuffer

With WebGL 2.0, one could use `blitFramebuffer` and/or `clearBuffer*` with the default back buffer as a destination. Does this operation trigger an implicit clear? Whatever the answer is, the CTS should cover it.

kdashg:
Specify it categorically based on "are allowed to read/write" behavior,
and provide lists and annotations less normatively.

Fixes #3436.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -348,14 +348,15 @@
     </p>
     <p>
         WebGL presents its drawing buffer to the HTML page compositor immediately before a
-        compositing operation, but only if at least one of the following has occurred since the
+        compositing operation, but only if at least one of the following have been called since the
         previous compositing operation:
         <ul>
             <li>Context creation</li>
             <li>Canvas resize</li>
             <li>
-                <code>clear</code>, <code>drawArrays</code>, or <code>drawElements</code> has been
-                called while the drawing buffer is the currently bound framebuffer
+                Any of the
+                <a href=#DRAW_OPERATIONS>[Draw Operations]</a>,
+                called while the drawing buffer is the currently bound (draw) framebuffer.
             </li>
         </ul>
     </p>
@@ -2575,6 +2576,9 @@ WebGLRenderingContext includes WebGLRenderingContextOverloads;
                         GLint border)
             <span class="gl-spec">(<a href="http://www.khronos.org/registry/gles/specs/2.0/es_full_spec_2.0.25.pdf#nameddest=section-3.7.2">OpenGL ES 2.0 &sect;3.7.2</a>, <a class="nonnormative" href="http://www.khronos.org/opengles/sdk/2.0/docs/man/xhtml/glCopyTexImage2D.xml">man page</a>)</span>
         <dd>
+            A <a href=#READ_OPERATIONS>[Read Operation]</a>.
+            <br>
+
             If an attempt is made to call this function with no WebGLTexture bound (see above), an
             <code>INVALID_OPERATION</code> error is generated. <br><br>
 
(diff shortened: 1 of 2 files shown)


specs/latest/1.0/index.html
```diff
@@ -3316,43 +3323,66 @@ <h4>Uniforms and attributes</h4>
...
     <p>
-        If any one of these 3 calls attempts to draw to a missing attachment of a complete framebuffer,
+        These include, but (due to additions in extensions or in WebGL 2) are not limited to:
+        <ul>
+            <li><code>clear</code></li>
+            <li><code>drawArrays</code></li>
+            <li><code>drawElements</code></li>
+        </ul>
+        For instance,
+        <a href="https://registry.khronos.org/webgl/extensions/ANGLE_instanced_arrays/">ANGLE_instanced_arrays</a>
```
lexaknyazev, line 3348: Consider using a relative path since it's a part of the same repo.

specs/latest/1.0/index.html
```diff
@@ -3316,43 +3323,66 @@ <h4>Uniforms and attributes</h4>
...
+        adds <code>drawArraysInstancedANGLE</code>, and
+        <a href="https://registry.khronos.org/webgl/specs/latest/2.0/">WebGL 2.0</a>
```
lexaknyazev, line 3350: Same.

specs/latest/1.0/index.html
```diff
@@ -3316,43 +3323,66 @@ <h4>Uniforms and attributes</h4>
...
         be directed to the drawing buffer or to a Framebuffer object. When rendering is
-        directed to the drawing buffer, making any of the 3 rendering calls shall
+        directed to the drawing buffer, making any of the rendering calls shall
         cause the drawing buffer to be presented to the HTML page compositor at the start
         of the next compositing operation.
     </p>
-
     <p>
-        If any one of these 3 calls attempts to draw to a missing attachment of a complete framebuffer,
+        These include, but (due to additions in extensions or in WebGL 2) are not limited to:
```
lexaknyazev, line 3341: s/2/2.0/

specs/latest/1.0/index.html
```diff
@@ -3365,17 +3395,33 @@ <h4><a name="WRITING_TO_THE_DRAWING_BUFFER">Writing to the drawing buffer</a></h
...
+    <h4><a id=READ_OPERATIONS>Read Operations</a></h4>
 
     <p>
-        Pixels in the current framebuffer can be read back into an ArrayBufferView object.
+        OpenGL ES 2.0 has several calls which which are allowed to
+        read from the currently bound (read) framebuffer.
+        WebGL categorizes all such calls as <a href=#READ_OPERATIONS>[Read Operations]</a>.
+    </p>
+    <p>
+        These include, but (due to additions in extensions or in WebGL 2) are not limited to:
```
lexaknyazev, line 3406: s/2/2.0/

lexaknyazev (approved):
Looks good with a few editorial suggestions.

kenrussell (approved):
Nice refactoring and clarification. Looks good to me modulo @lexaknyazev 's suggested edits.

specs/latest/1.0/index.html
kdashg: Most urls in that document are absolute also. With relative paths, it would become tricky when we snapshot the spec, because it will move in the directory heirarchy:
```
$ find specs -name index.html
specs/1.0.0/index.html
specs/1.0.1/index.html
specs/1.0.2/index.html
specs/1.0.3/index.html
specs/2.0.0/index.html
specs/latest/1.0/index.html
specs/latest/2.0/index.html
specs/latest/index.html
```
We could of course make changes such that it could maintain the same depth when snapshotted, but that's part of the context.

PR #3470, opened 2022-08-05
Clarify TRANSFORM_FEEDBACK_BUFFER bindings

### LINKED ISSUE #3468: Clarify TRANSFORM_FEEDBACK_BUFFER_BINDING state
OpenGL ES 3.2, `BindTransformFeedback`:
> The `TRANSFORM_FEEDBACK_BUFFER_BINDING` generic buffer binding point has now been moved to context state, rather than per-object state, and should be unaffected by this command. It is possible that some drivers do not yet implement this change.

Based on the implementations, it looks like the [Differences Between WebGL and OpenGL ES 3.0](https://registry.khronos.org/webgl/specs/latest/2.0/#5) section must contain a similar statement since OpenGL ES 3.0 defines the generic binding point per-object.

Moreover, the existing [Preventing undefined behavior with Transform Feedback](https://registry.khronos.org/webgl/specs/latest/2.0/#5.3) section should clearly state which bindings (generic, indexed, or both) it refers to.

lexaknyazev:
Fixes #3468.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -3637,8 +3637,9 @@ WebGL2RenderingContext includes WebGL2RenderingContextOverloads;
     <h3>Preventing undefined behavior with Transform Feedback</h3>
 
     <p>
-        A buffer which is simultaneously bound to a TRANSFORM_FEEDBACK_BUFFER binding point in the currently
-        bound transform feedback object and any other binding point in the WebGL API cannot be used. Any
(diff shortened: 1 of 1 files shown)


kenrussell:
Do the conformance tests already cover this specification change? Do browsers already implement it?

lexaknyazev:
The updated description should match ANGLE and the existing `transform_feedback/simultaneous_binding.html` test.

kenrussell (approved):
Sounds fine. I'll clarify the description while squashing and merging.

PR #3473, opened 2022-08-08
getBufferSubData with offset ArrayBufferView

djg:
Add tests where the destination for the data returned by `getBufferSubData` does
not coincide with the beginning of array storage.


kenrussell (changes requested):
Thanks for these nice additions. Could you please also add a couple of tests of `getBufferSubData(gl.ARRAY_BUFFER, n, ...)` where `n > 0` as were added in https://github.com/WebKit/WebKit/pull/3036 ? That seems to be untested in this file. Thanks.

kenrussell:
BTW - have uploaded #3474 to try to fix the broken lint check.

djg:
> Thanks for these nice additions. Could you please also add a couple of tests of `getBufferSubData(gl.ARRAY_BUFFER, n, ...)` where `n > 0` as were added in [WebKit/WebKit#3036](https://github.com/WebKit/WebKit/pull/3036) ? That seems to be untested in this file. Thanks.

Updated, now with extra spicy `srcByteOffset` tests.

PR #3474, opened 2022-08-08
Run "sudo apt update" before installing dependencies.

kenrussell:
Suggested on:
https://github.com/orgs/community/discussions/26649

as solution for failing Github action in:
https://github.com/KhronosGroup/WebGL/runs/7718272204?check_suite_focus=true


kenrussell:
This seems to be working - the lint check against this PR ran the update step successfully - so merging myself.

PR #3476, opened 2022-08-08
Expand simultaneous XFB binding test

lexaknyazev:
When a `bufferData` call is rejected due to simultaneous bindings, the bound buffer size must not reset. The new test fails in Chromium, passes in WebKit and Firefox.

### DIFF
--- a/sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
+++ b/sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
@@ -304,6 +304,17 @@ for (let genericBindPointValue of genericBindPointValues) {
   gl.copyBufferSubData(gl.COPY_READ_BUFFER, gl.COPY_WRITE_BUFFER, 0, 0, 1);
   wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "copyBufferSubData with double bound buffer");
 
+  debug("<hr/>Test that rejected operations do not change the bound buffer size");
+
+  gl.bindBuffer(gl.ARRAY_BUFFER, tfBuffer);
+  gl.bufferData(gl.ARRAY_BUFFER, 8, gl.STATIC_DRAW);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "bufferData with double bound buffer");
+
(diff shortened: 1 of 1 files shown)


kenrussell (approved):
Was there an associated WebKit bug fix for this new test? I haven't run the new test yet but it looks like on all platforms which use ANGLE this error should be caught:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/angle/src/libANGLE/validationES2.cpp;l=3570?q=validationES2.cpp

lexaknyazev:
WebKit was failing the test even before this PR; fixed here: https://commits.webkit.org/253210@main.

Chromium resets the buffer size before calling into ANGLE:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc;l=2171

kenrussell:
Upon further review, I see the Chromium-side cache that's being updated incorrectly during this test and am fixing it; would still appreciate pointers to any WebKit bug fixes. Thanks.

sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
```diff
@@ -304,6 +304,17 @@
...
 
+  debug("<hr/>Test that rejected operations do not change the bound buffer size");
+
+  gl.bindBuffer(gl.ARRAY_BUFFER, tfBuffer);
+  gl.bufferData(gl.ARRAY_BUFFER, 8, gl.STATIC_DRAW);
+  wtu.glErrorShouldBe(gl, gl.INVALID_OPERATION, "bufferData with double bound buffer");
+
+  gl.bindTransformFeedback(gl.TRANSFORM_FEEDBACK, null);
+  gl.bufferSubData(gl.ARRAY_BUFFER, 0, new Uint8Array(16));
+  wtu.glErrorShouldBe(gl, gl.NO_ERROR, "bufferSubData should succeed");
```
kdashg, line 315: "should succeed now that not double-bound"

kdashg (approved):

sdk/tests/conformance2/transform_feedback/simultaneous_binding.html
lexaknyazev: Done in #3675.

PR #3480, opened 2022-08-11
Allow makeXRCompatible in OffscreenCanvas contexts

### LINKED ISSUE #3479: Allow extra IDL members when appropriate
Some tests assert that WebGL context objects do not have any unknown members. This approach fails when a browser implements another specification that extends `WebGLRenderingContextBase`.

For example, `conformance/offscreencanvas/methods.html` fails in Chromium because of [`makeXRCompatible`](https://immersive-web.github.io/webxr/#dom-webglrenderingcontextbase-makexrcompatible).

lexaknyazev:
Fixes #3479.

### DIFF
--- a/sdk/tests/js/tests/canvas-tests-utils.js
+++ b/sdk/tests/js/tests/canvas-tests-utils.js
@@ -340,6 +340,7 @@ function testAPIs(contextType) {
       passed = passed && r;
     }
 
+    methods.push(...["makeXRCompatible"]);
     var extended = false;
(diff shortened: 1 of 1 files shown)


kenrussell (approved):
Looks good to me.

If you wouldn't mind, improving the error logging in these tests (`conformance/offscreencanvas/methods.html` and `conformance/offscreencanvas/methods-worker.html`) in a follow-up would be greatly appreciated.

PR #3482, opened 2022-08-12
Split gles2 conversions and gles2/3 swizzle tests

greggman:
I have no objection to splitting these tests but... the tests
don't actually appear to be slow under normal conditions. They take
7 to 15 seconds without splitting.

Instead, if I understand correctly, the issue is some debug builds
of chrome run these too slowly. That would seem to be an issue
with the debug builds doing something slow too low-level.

At the same time I don't see any real harm in splitting these tests.
The splitting is automated and also unlikely to change.


kenrussell (approved):
Thanks for doing this split. Hopefully this will address the issues seen on Chromium's debug bots without having to investigate deeply why debug builds of the browser run so much more slowly.

Tested locally and verified that the WebGL 1.0 and 2.0 versions of the tests run correctly.

Merging now and referring to http://crbug.com/1143424 .

Issue #3481, opened 2022-08-12
Problems when running conformance.py


didiHe:

didiHe:
I download the code from website,and install the packages needed.
Then I try to run the test:
python3 conformance.py --browser-name firefox --version 2.0.0
but error happened:
![image](https://user-images.githubusercontent.com/98087129/184311540-708c1b0f-fe08-457f-8148-9a6e45a18732.png)
the OS is ubuntu18.04
the python version is Python 3.8.10

kenrussell:
@gyagp any chance you can investigate this? Unfortunately our team's not using this harness.

gyagp:
Firefox was not supported on Linux. I just uploaded a PR to add the support, and also fixed some compatibility issue with new Selenium as it deprecated some APIs. Sorry that I don't have enough time for all the browsers, so only Firefox and Chrome are taken care of. 
BTW, I encountered a problem with Firefox from snap, as it reported profile couldn't be loaded. I followed the steps (https://www.omgubuntu.co.uk/2022/04/how-to-install-firefox-deb-apt-ubuntu-22-04) to install a .deb file instead.

PR #3483, opened 2022-08-15
tools: Add Firefox support on Linux and adapt Selenium changes

gyagp:
Add Firefox support on Linux
Selenium has deprecated some old APIs, so this PR also adapts the
changes for Firefox and Chrome.

Bug: #3481


gyagp:
@kenrussell PTAL

kenrussell (approved):
Very nice - thanks Yang for these updates.

PR #3484, opened 2022-08-17
Add test of gl_VertexID and gl_InstanceID.

kdashg:
Findings:
* MacOS does not handle gl_InstanceID reliably.
* No pure-GL drivers appear to have implemented ES3.1 gl_VertexID behavior for DrawElements*.
  * Instead, gl_VertexID is always 0 for DrawElements*.

Passes in Chrome and Safari, various failures in Firefox.


kdashg:
This is https://bugzilla.mozilla.org/show_bug.cgi?id=1779800

kdashg:
@kenrussell You can see the line that we might need to uncomment for now, depending on the state of the ANGLE-on-Metal rollout:
```
//ERRATA.IGNORE_GL_INSTANCE_ID = true; // Chrome on ANGLE-on-Mac-GL needs this.
```

Safari seems to pass even with that errata commented out, as does patched Firefox.

kenrussell (changes requested):
This doesn't run in Chrome Canary or Firefox Nightly as is:

```
Uncaught (in promise) TypeError: Failed to execute 'getAttachedShaders' on 'WebGLRenderingContext': parameter 1 is not of type 'WebGLProgram'.
    at fn (builtin-vert-attribs.html:234:36)
    at builtin-vert-attribs.html:247:28
```

Could you please investigate and revise?

sdk/tests/conformance2/rendering/builtin-vert-attribs.html
```diff
@@ -0,0 +1,408 @@
...
+(async () => {
+    for (const desc of TESTS) {
+        await wtu.dispatchPromise(); // Yield, for responsiveness.
+        debug('');
+        debug('---------------------');
+        debug(`desc: ${JSON.stringify(desc)}`);
+
+        let fn = (vs) => {
+            //console.log({vs});
+            const prog = wtu.setupProgram(gl, [vs, FS]);
```
cndv3996, line 229: Here, looks vs, fs should be ids of DOM elements, instead of source shaders, I think.

cndv3996: Oh, sorry, how careless, missed other branches of if

sdk/tests/conformance2/rendering/builtin-vert-attribs.html
```diff
@@ -0,0 +1,408 @@
...
+
+## gl_InstanceID random for DrawArrays calls
+Use ERRATA.IGNORE_GL_INSTANCE_ID to cause these tests to pass.
+
+## Adds `first` to user-attrib instanced fetch ids in DrawArrays calls.
+Use ERRATA.FIRST_ADDS_TO_INSTANCE to cause these tests to pass.
+*/
+
+const wtu = WebGLTestUtils;
+const gl = wtu.create3DContext('e_canvas');
```
cndv3996, line 84: I think, we should create a webgl2 context here, maybe

PR #3488, opened 2022-08-23
Test compressed formats rejection with texStorage

lexaknyazev:
In `GLES2DecoderPassthroughImpl::Initialize`, Chromium implicitly enables many requestable extensions, including those that provide compressed texture formats; the `texStorage` calls defer all validation to ANGLE, and that's why the tests fail there for ASTC, ETC1, S3TC, and S3TC sRGB extensions.

~~On top of that, ANGLE currently behaves as if ETC2/EAC formats are always supported for ES 3.x clients, even in WebGL compatibility mode. WebKit also defers all `texStorage` validation to ANGLE, so ETC2/EAC tests fail both in Chromium and WebKit with this PR.~~


kdashg (approved):

kenrussell:
Have filed http://crbug.com/1405252 about this. Will try that change with this PR and approve/merge if it's passing.

kenrussell (approved):
Thanks, approved. Fixing the Chromium issue under http://crbug.com/1405252 .

I'd like to try editing the new failure message a little bit during the merge process.

PR #3489, opened 2022-08-24
Fix quiet mode test reporting

bsheedy-work:
Fixes testPassedOptions and testSkippedOptions only reporting
results when actually adding the <span> to the page. The test
harness should be notified when a test is done regardless of
whether the page contents are going to be modified.

### DIFF
--- a/sdk/tests/js/js-test-pre.js
+++ b/sdk/tests/js/js-test-pre.js
@@ -268,9 +268,9 @@ function getCurrentTestName()
  */
 function testPassedOptions(msg, addSpan)
 {
+    reportTestResultsToHarness(true, _currentTestName + ": " + msg);
     if (addSpan && !quietMode())
     {
-        reportTestResultsToHarness(true, _currentTestName + ": " + msg);
         _addSpan('<span><span class="pass">PASS</span> ' + escapeHTML(_currentTestName) + ": " + escapeHTML(msg) + '</span>');
     }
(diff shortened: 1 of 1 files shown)


kenrussell (approved):
Could you do a quick local run of a couple of the larger dEQP tests with and without these changes and see if there's any run time impact? If there is, perhaps we can test `quietMode()` and pass "" as the string instead of doing the string concatenation.

Thanks.

bsheedy-work:
Sample times on my local Linux machine with Chromium:

deqp/data/gles2/shaders/swizzles.html
  With change
    78s
    75s
    74s
  Without change
    74s
    74s
    75s

I've managed to get my machine into a bad state where Chromium crashes on launch now, so I'll get more data after I reboot. Doesn't look like there's much of a change, though.

bsheedy-work:
deqp/data/gles3/shaders/conversions_vector_combine.html
With change
  54s
  56s
  52s
Without change
  53s
  51s
  48s

So it looks like there isn't a significant difference. For completeness, though, let's try with using an empty string if we're in quiet mode:

deqp/data/gles3/shaders/conversions_vector_combine.html
  53s
  56s
  52s

deqp/data/gles2/shaders/swizzles.html
  77s
  74s
  77s

kenrussell (approved):
Thanks for doing the runs - the performance impact looks minimal. Merging.

FYI @greggman @toji .

PR #3490, opened 2022-08-26
Add EXT_provoking_vertex test

lexaknyazev:


kkinnunen-apple (approved):

kenrussell (approved):
Very nice test Alexey.

You mentioned on Slack that the test partly fails because of an ANGLE bug. Could you please file that bug so that it can be referenced in the commit description when this is merged? Thanks.

PR #3491, opened 2022-08-29
Cache window values

bsheedy-work:
Adds caching for window.location.pathname and
window.parent.webglTestHarness. These values are accessed every
time reportTestResultsToHarness is called and accessing a variable
via window can be significantly slower than a local copy.

As an extreme example,
deqp/functional/gles3/vertexarrays/multiple_attributes.output.html
reports 33,013,505 results. On my local machine with Chromium, the
test takes ~20 seconds with this fix and ~215 seconds without it.

Related to crbug.com/1357073.

### DIFF
--- a/sdk/tests/js/js-test-pre.js
+++ b/sdk/tests/js/js-test-pre.js
@@ -93,20 +93,25 @@ const RESULTS = {
   fail: 0,
 };
 
+// We cache these values since they will potentially be accessed many (100k+)
+// times and accessing window can be significantly slower than a local variable.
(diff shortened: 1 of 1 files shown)


bsheedy-work:
For completeness, I've confirmed that the other tests that were having issues on Mac in the linked Chromium bug also run in under ~20 seconds now:

deqp/functional/gles3/vertexarrays/multiple_attributes.stride.html
deqp/functional/gles3/vertexarrays/single_attribute.first.html

kenrussell (approved):
Looks fine to me.

PR #3495, opened 2022-10-18
Add conformance/textures/misc/texture-srgb-upload.html.

kdashg:
Test for both webgl1+ext and webgl2.
Test uploads from ArrayBuffer and (inlined) video. Ensure that srgb textures fetch/decode 0x7f as 0.21 not 0.5.

Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1696693


kenrussell (approved):
Looks good to me. Added a couple of early-outs which make the test run to completion on Safari Technology Preview, and removed some commented-out code.

PR #3499, opened 2022-11-01
Fix broken extension links

lexaknyazev:


kenrussell (approved):
Thanks for fixing these!

PR #3500, opened 2022-11-01
Use WEBGL prefix for the provoking vertex extension

lexaknyazev:
WebGL extensions with non-`WEBGL` prefixes mirror OpenGL ES extensions with similar names.

Since `GL_EXT_provoking_vertex` is defined only for desktop OpenGL (unlike other `EXT` extensions referenced by WebGL extension specs), the WebGL provoking vertex extension should have a `WEBGL` prefix.


kenrussell (approved):
This was discussed at the last WebGL working group meeting on 2022-10-20 and it was agreed upon to rename this extension from EXT_ to WEBGL_.

PR #3501, opened 2022-11-23
Cleanup WEBGL_provoking_vertex

lexaknyazev:
- Reference an ANGLE extension that does not include quads and other desktop GL features.
- Use the same parameter name as all upstream extensions and specs for consistency.


kdashg:
Thanks! I think I can just merge this.

PR #3502, opened 2022-11-24
Test calling getUniform from non-current program

greggman:
It's possible this is already tested but I didn't see anything obvious and Firefox fails this test

There's 3yr old bug to fix it here: https://bugzilla.mozilla.org/show_bug.cgi?id=1645092

Turns out the bug made [this webgl-state-diagram](https://webglfundamentals.org/webgl/lessons/resources/webgl-state-diagram.html?exampleId=use-2-programs#no-help) fail in Firefox. I've since [hacked in a workaround](https://github.com/gfxfundamentals/webgl-fundamentals/commit/e6d75517941ff1a9895a1a4d051ac32b08b97933) but it would be nice if that hadn't been needed.

### DIFF
--- a/sdk/tests/conformance/uniforms/00_test_list.txt
+++ b/sdk/tests/conformance/uniforms/00_test_list.txt
@@ -2,6 +2,7 @@ gl-uniform-arrays.html
 # This test is no longer valid with the new packing restrictions
 #--min-version 1.0.02 gl-uniform-unused-array-elements-get-truncated.html
 gl-uniform-bool.html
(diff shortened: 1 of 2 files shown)


kdashg (approved):

kdashg:
@kenrussell Do you still want to block sign-off/merge on checking status in Chrome?

kenrussell (approved):
Thanks for checking. LGTM.

kdashg:
This is fixed in Nightly too now.

PR #3505, opened 2022-12-01
Adjust EXT_sRGB spec and extend its test

### LINKED ISSUE #3475: RGB(A)8 renderbuffers on WebGL 1.0
This looks like a quirk existing due to historical reasons but may still be worth clarifying.
1. WebGL 1.0, Section 2.2 guarantees that the default renderbuffer has at least 8 bits per component.
2. OpenGL ES 2.0, Section 4.4.3 allows only `RGBA4`/`RGB5_A1`/`RGB565` internal formats for renderbuffer color attachments. 8-bit renderbuffers are provided only by [`OES_rgb8_rgba8`](https://registry.khronos.org/OpenGL/extensions/OES/OES_rgb8_rgba8.txt).
3. WebGL [`EXT_sRGB`](https://registry.khronos.org/webgl/extensions/EXT_sRGB/) extension is based on [`GL_EXT_sRGB`](https://registry.khronos.org/OpenGL/extensions/EXT/EXT_sRGB.txt) that requires `OES_rgb8_rgba8`.

Given 1-3, it seems quite counter-intuitive that WebGL 1.0 renderbuffers can have `SRGB8_ALPHA8_EXT` or floating-point formats but not `RGBA8`.

At the very least, we should add a note to `EXT_sRGB` clarifying the ignored dependency. Optionally, WebGL 1.0 renderbuffers may be allowed to have `RGB8`/ `RGBA8` internal formats (aliased as their unsized counterparts, similar to `DEPTH_STENCIL`) as if `OES_rgb8_rgba8` is always enabled.

Maybe related: #3364.

/cc @kenrussell @kdashg

lexaknyazev:
- Fixes #3475.
- Exposes a WebKit bug with `getFramebufferAttachmentParameter`.

### DIFF
--- a/extensions/EXT_sRGB/extension.xml
+++ b/extensions/EXT_sRGB/extension.xml
@@ -22,6 +22,9 @@
   <overview>
     <mirrors href="http://www.khronos.org/registry/gles/extensions/EXT/EXT_sRGB.txt"
              name="EXT_sRGB">
+      <addendum>
+        Support for <code>OES_rgb8_rgba8</code> or equivalent functionality is not required.
+      </addendum>
     </mirrors>
 
     <features>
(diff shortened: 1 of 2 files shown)


kenrussell (approved):
Looks good.

PR #3507, opened 2022-12-07
Add an extension proposal for ANGLE_shader_pixel_local_storage

csmartdalton:
This brings the ANGLE_shader_pixel_local_storage extension to WebGL.

### DIFF
--- a/extensions/proposals/ANGLE_shader_pixel_local_storage/extension.xml
+++ b/extensions/proposals/ANGLE_shader_pixel_local_storage/extension.xml
@@ -1,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<proposal href="proposals/ANGLE_shader_pixel_local_storage/">
+  <name>ANGLE_shader_pixel_local_storage</name>
+
+  <contact> <a href="https://www.khronos.org/webgl/public-mailing-list/">WebGL working group</a> (public_webgl 'at' khronos.org) </contact>
+
+  <contributors>
+    <contributor>Chris Dalton, Rive</contributor>
+    <contributor>Kenneth Russell, Google Inc.</contributor>
+    <contributor>Shahbaz Youssefi, Google Inc.</contributor>
+    <contributor>Kelsey Gilbert, Mozilla Corp.</contributor>
+    <contributor>Geoff Lang, Google Inc.</contributor>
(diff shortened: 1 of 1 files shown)


kenrussell (approved):
Thanks, looks great on first glance.

Have you run `make` in your local `extensions` directory and verified this produces valid HTML?

csmartdalton:
> Have you run `make` in your local `extensions` directory and verified this produces valid HTML?

I did! and the `index.html` looks good.

Can I get some help with running the workflow? I can't kick it off since I'm a first-time contributor.

kenrussell:
Thanks for confirming. The only workflow we have right now is a lint check and for some reason it looks like it's not running - nothing in the UI I have available offers a way to start it. I'll squash and merge now and we can iterate on this.

kenrussell:
FYI, Khronos has a Github hook that runs after each commit, which regenerates the extension registry on their website.

csmartdalton:
Cool! I see here under the list of proposed extensions:
https://registry.khronos.org/webgl/extensions/

And here's the IDL:
https://registry.khronos.org/webgl/extensions/proposals/ANGLE_shader_pixel_local_storage/

csmartdalton:
Hi all,

I’m adding pixel local storage bindings to emscripten and would appreciate
some feedback on the best way to advertise coherent pixel local storage
support.

We currently advertise coherency in two different ways:

   1. Adding a new extension string:
   "ANGLE_shader_pixel_local_storage_coherent"
   2. A new call in the WebGL extension interface: "isCoherent()"

Both of these are challenging for emscripten:

   1. Emscripten could append "ANGLE_shader_pixel_local_storage_coherent"
   to the extensions string, when applicable
      - BUT this requires enabling the WebGL extension in glGetString* in
      order to call isCoherent() internally, even if the app doesn't
want to use
      it
   2. Emscripten could create a new function that maps directly to the
   interface's "isCoherent()" method:  e.g.,
   "emscripten_webgl_pixel_local_storage_is_coherent()"

Alternatively, we could just update everything to do it the same way:

3. "glGetBooleanv(GL_PIXEL_LOCAL_STORAGE_COHERENT_ANGLE, &isCoherent)"

Does anyone have any opinions on which route to take? If not I'd lean
toward #3, to make code written for ANGLE and WEBGL more compatible, but
it's not a strong preference.

Thanks for all your help!
-Chris

PR #3509, opened 2022-12-24
Add a 'webglcontextrecreationready' event

davychxn:
This event is designed under the background of surging online visits to more and more extremely high traffic WebGL only websites. Nowadays daily visits at the level of 10 million to a single WebGL dependent webpage are not rare. (Theorically at least a visit per 8.6 milliseconds) User client GPUs are often shared between applications. If another GPU dependent application causes temporary hardware issues, like running long shaders. A certain percentage of website users can run into such situations, hence fail to create WebGL context. We lack such an context independent event to inform users when is the most applicable timepoint to recreate context.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -3773,6 +3773,81 @@ if(!gl) {
         errorInfo);
 }</pre>
     </div>
+    
+    <h4><a name="CONTEXT_RECREATION_READY">The Context Recreation Ready Event</a></h4>
+
+    <p>
+        When the user agent is to <b><a name="fire-a-webgl-context-recreation-ready">fire a WebGL
+        context recreation ready</a></b> at a <em>canvas</em>, it must perform the following steps:
+
+        <ol class="nestedlist">
+
+        <li> <a href="#fire-a-webgl-context-event">Fire a WebGL context event</a> named
+        "webglcontextrecreationready" at <em>canvas</em>, if the initial context creation failed caused by temporary hardware issues. 
+        You might see such issues at a certain rate by extremely high trafficing WebGL only websites. </li>
+
+        </ol>
+
+    </p>
+    
+    <div class="note">
+        This event is designed under the background of surging online visits to more 
+        and more extremely high traffic WebGL only websites. Nowadays daily visits at 
+        the level of 10 million to a single WebGL dependent webpage are not rare. 
+        (Theorically at least a visit per 8.6 milliseconds)
+        User client GPUs are often shared between applications. If another GPU dependent 
+        application causes temporary hardware issues, like running long shaders. 
+        A certain percentage of website users can run into such situations, hence fail to 
+        create WebGL context. We lack such an context independent event to inform users 
+        when is the most applicable timepoint to recreate context.
+    </div>
+
+    <div class="example">
+
+        The following code illustrates how an application can recreate context if the initial creation failed:
+
+<pre>var gl = null;
+var errorInfo = "";
+function createContext() {
+
+    var gl = canvas.getContext("webgl");
+    if(!gl) {
+      alert("A WebGL context could not be created at the moment.\nReason: " +
+            errorInfo + 
+            "A event will be fired if hardware is ready again. Possibly in less than 30 seconds depending on specific hardwares.");
+    }
+    
+    return gl;
+}
+
+function onContextCreationError(event) {
+
+  canvas.removeEventListener(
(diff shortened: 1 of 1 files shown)


HarelM:
+1
From a MapLibre maintainer (a library that heavily uses webgl in the browser) point of view there's a need to reduce the webgl creation failure rate. I believe this will help.

greggman:
I don't think I understand what you expect to happen. The browser as no info about when is a good time to create/re-create a context.

If you get a contextlost event, my suggestion would be to tell the user either directly or indirectly and let them try again. If you can't create a context at all then you can either ask the user "try again?" or you can fall back to something else. (Google maps for example falls back to 2D canvas)

To be slight more critical of the proposal

> This event is designed under the background of surging online visits to more 
> and more extremely high traffic WebGL only websites. Nowadays daily visits at 
> the level of 10 million to a single WebGL dependent webpage are not rare. 
> (Theorically at least a visit per 8.6 milliseconds)

What does any of this have to do with browser or WebGL. A browser has no idea how much traffic any individual site gets. The fact that some sites get 10 million hits has nothing to do with WebGL.

> "webglcontextrecreationready" at <em>canvas</em>, if the initial context creation failed caused by temporary hardware issues.

The browser doesn't have any knowledge of "temporary hardware issues"

> You might see such issues at a certain rate by extremely high trafficing WebGL only websites.

That a site has high traffic is unrelated to WebGL.

cndv3996:
> I don't think I understand what you expect to happen. The browser as no info about when is a good time to create/re-create a context.
> 
> If you get a contextlost event, my suggestion would be to tell the user either directly or indirectly and let them try again. If you can't create a context at all then you can either ask the user "try again?" or you can fall back to something else. (Google maps for example falls back to 2D canvas)
> 
> To be slight more critical of the proposal
> 
> > This event is designed under the background of surging online visits to more
> > and more extremely high traffic WebGL only websites. Nowadays daily visits at
> > the level of 10 million to a single WebGL dependent webpage are not rare.
> > (Theorically at least a visit per 8.6 milliseconds)
> 
> What does any of this have to do with browser or WebGL. A browser has no idea how much traffic any individual site gets. The fact that some sites get 10 million hits has nothing to do with WebGL.
> 
> > "webglcontextrecreationready" at _canvas_, if the initial context creation failed caused by temporary hardware issues.
> 
> The browser doesn't have any knowledge of "temporary hardware issues"
> 
> > You might see such issues at a certain rate by extremely high trafficing WebGL only websites.
> 
> That a site has high traffic is unrelated to WebGL.

Hi Gregg, merry Christmas! How is the weather in Tokyo?

Your background really awesome. You started development job since 1991? In that year I got my first Family Computer of Nintendo. My dad found that for me from one of his close friends.

I think you are right, but I still want to explain my understanding a bit more.

> If you get a contextlost event, my suggestion would be to tell the user either directly or indirectly and let them try again. If you can't create a context at all then you can either ask the user "try again?" or you can fall back to something else. (Google maps for example falls back to 2D canvas)

If your products (Like Bing Maps or Google Maps) have users coming in almost every milliseconds, I think the service providers don't want them to see a popup telling them try again themselves from time to time? The products revenue heavily depend on user experiences.

Yeah, falls back to 2D canvas should be a good idea for a long time. But something like online Maps with worldwide service, most of them doing vectorgraph rendering. And Maps will have more and more details and information integrated into data sources. WebGL only rendering should be more beneficial for the users and industry? (Like the product I saw @HarelM was mentioning?)

> What does any of this have to do with browser or WebGL. A browser has no idea how much traffic any individual site gets. The fact that some sites get 10 million hits has nothing to do with WebGL.

Yeah,  a browser doesn't need to know that or be able to know that, I agree. But if we introduce the event and have GPU drivers implement it. A browser can help more users prevent that? With less efforts? And more secure or reliable? Otherwise users will need to implement their own retry mechanisms. A event can make things easier.

> The browser doesn't have any knowledge of "temporary hardware issues"

I'm not sure about this part. I didn't contribute any code to Chromium, but I would like to if I find the chance.
If we already have a 'webglcontextrestored' event, maybe we can some way? But this 'webglcontextrestored' depends on a existing context. If we never have one, maybe we still want a similar event (Independent on a context but dependent on a canvas)?

Do I have a chance to become a member? But I don't want to pay a member fee higher than 199 USD a year!

cndv3996:
Need more plans on the issue. Close it for now.

davychxn:
Reopen it for future reference. I still believe the proposal could be hepful.

PR #3511, opened 2023-01-12
Update clip / cull distance specs and tests

### LINKED ISSUE #3109: Consider moving EXT_clip_cull_distance forward
As pointed out on the public_webgl mailing list here:
https://www.khronos.org/webgl/public-mailing-list/public_webgl/2006/msg00007.php

a use case for the EXT_clip_cull_distance extension is 2D rendering. From Kevin Rogovin:

"""
The main place is for 2D (UI rendering). One example is in a public
repo, https://github.com/intel/fastuidraw. There, I use HW-clip planes
together with the depth buffer to make clipping changes (severe
changes, like clip against a filled path, rotated rects and so on)
really cheap and almost entirely handled by the GPU. I also find it
really useful for an offscreen rendering technique I use in a current
(but closed source) project. I can emulate with discard, but tests
show that this has a large negative performance impact.
"""

The working group should consider moving this extension forward - first by providing one implementation.

lexaknyazev:
Fixes #3109.


extensions/WEBGL_clip_cull_distance/extension.xml
```diff
@@ -0,0 +1,118 @@
...
+  </contributors>
+
+  <number>43</number>
+
+  <depends>
+    <api version="2.0"/>
+  </depends>
+
+  <overview>
+    <mirrors href="https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_clip_cull_distance.txt"
```
kenrussell, line 20: Should this mirror https://chromium.googlesource.com/angle/angle/+/HEAD/extensions/ANGLE_clip_cull_distance.txt now? Since the extension text below mentions `ANGLE_clip_cull_distance`, that seems more appropriate.

kenrussell (commented):
Looks good overall - especially nice work on the conformance test. One question before approving.

extensions/WEBGL_clip_cull_distance/extension.xml
lexaknyazev: The ANGLE extension is not self-contained and plain text files do not have hyperlinks, so it seems to me that linking directly to the original EXT extension is more useful to readers. All WebGL-specific differences are mentioned right here anyway.

kenrussell: Would you link to https://chromium.googlesource.com/angle/angle/+/HEAD/extensions/ANGLE_clip_cull_distance.txt in a separate hyperlink in this extension spec instead?

lexaknyazev: Done.

kenrussell: Thanks.

kenrussell (approved):

PR #3512, opened 2023-01-12
Clarify FBO queries for floating-point extensions

lexaknyazev:
Both `EXT_color_buffer_half_float` and `WEBGL_color_buffer_float` say:
> The component types of framebuffer object attachments can be queried.

This PR clarifies and tests such queries. Seems like all browsers do not implement them.


kdashg (approved):

kenrussell (approved):
Thank you for these careful spec and test updates, and also for the patch addressing them in Chromium! Tested everything locally, looks good.

PR #3517, opened 2023-02-09
Move EXT_polygon_offset_clamp to Draft, add tests

lexaknyazev:


sdk/tests/conformance/extensions/ext-polygon-offset-clamp.html
```diff
@@ -0,0 +1,172 @@
...
+    gl.drawArrays(gl.TRIANGLE_STRIP, 0, 4);
+
+    wtu.checkCanvasRect(
+        gl,
+        0, 0, 3 * w / 4 - 1, h,
+        [0, 255, 0, 255], "should be green");
+
+    wtu.checkCanvasRect(
+        gl,
+        3 * w / 4 + 1, 0, w / 4 - 1, h,
```
kenrussell, line 134: Do you think all implementations are likely accurate enough to have this cutover region within a +/-1 pixel region in the rendered canvas?

kenrussell (approved):
Nice work on the test in particular. One question; please feel free to merge yourself.

sdk/tests/conformance/extensions/ext-polygon-offset-clamp.html
lexaknyazev: Yes. A similar test passes on ANGLE CI.

PR #3521, opened 2023-02-17
Promote ANGLE_shader_pixel_local_storage to draft status

csmartdalton:


lexaknyazev:
- If the PLS functionality is expected to be implementable without ANGLE and/or if WebGL-specific differences would cause observable compatibility issues with native apps ported to WebGL, the extension prefix should probably be `WEBGL_`.
- There should be a "New Functions" section with the new functions and a "New Tokens" section with the new tokens accepted for `getParameter` and their return types (similar to `EXT_disjoint_timer_query`).
- The `backingtexture` parameter of `framebufferTexturePixelLocalStorageANGLE` should have a type of `WebGLTexture`. Might as well rename it to `texture` for consistency with the core `framebufferTexture*` functions. The description should have the following extra error defined:
  > If `texture` was generated by a different `WebGL2RenderingContext` than this one, generates an `INVALID_OPERATION` error.
- If both coherent and non-coherent versions could be exposed, add a new WebGL-specific API call to query that, e.g. `GLboolean isCoherent()`. Exposing two distinct yet almost identical interfaces would be much less convenient. 
- Please provide rationale why memoryless planes are not supported.

csmartdalton:
Thanks for the review

> * If the PLS functionality is expected to be implementable without ANGLE and/or if WebGL-specific differences would cause observable compatibility issues with native apps ported to WebGL, the extension prefix should probably be `WEBGL_`.
> * There should be a "New Functions" section with the new functions and a "New Tokens" section with the new tokens accepted for `getParameter` and their return types (similar to `EXT_disjoint_timer_query`).
> * The `backingtexture` parameter of `framebufferTexturePixelLocalStorageANGLE` should have a type of `WebGLTexture`. Might as well rename it to `texture` for consistency with the core `framebufferTexture*` functions. The description should have the following extra error defined:
>   > If `texture` was generated by a different `WebGL2RenderingContext` than this one, generates an `INVALID_OPERATION` error.
> * If both coherent and non-coherent versions could be exposed, add a new WebGL-specific API call to query that, e.g. `GLboolean isCoherent()`. Exposing two distinct yet almost identical interfaces would be much less convenient.
> * Please provide rationale why memoryless planes are not supported.

Thanks for the suggestions Alexey!

I did end up renaming to `WEBGL_` (for now) because of the changes to the interface. This is no longer compatible with the ANGLE extension, and the the same C code would behave very differently for the same extension, depending on whether it was compiled with Emscripten or not.

This brings up a tough question though. The shading language features all have an ANGLE suffix:

`#extension ANGLE_shader_pixel_local_storage : require`
`pixelLocalLoadANGLE()`
etc.

Would it be weird to expose a `WEBGL` extension with `ANGLE` shading language suffixes?

lexaknyazev:
> Would it be weird to expose a `WEBGL` extension with `ANGLE` shading language suffixes?

This happened before with
- https://registry.khronos.org/webgl/extensions/WEBGL_clip_cull_distance/ (draft)
- https://registry.khronos.org/webgl/extensions/WEBGL_draw_buffers/
- https://registry.khronos.org/webgl/extensions/WEBGL_multi_draw/
- https://registry.khronos.org/webgl/extensions/WEBGL_multi_draw_instanced_base_vertex_base_instance/ (draft)

kenrussell (approved):
Looks good to me. It looks like this has been revised taking into consideration the feedback discussed in this week's WebGL conference call, so let's merge it. Verified locally that the XML validates and builds.

PR #3528, opened 2023-03-15
Update WEBGL_shader_pixel_local_storage

csmartdalton:
* Delete LOAD_OP_DISABLE_WEBGL. This has been removed from the ANGLE extension.

* Add sequence<> and typed array overloads for setting clear values.

* Add a 'src_offset' parameter to the typed array overloads for setting clear values. This is convenient for WASM, since it enables passing HEAPF32 et. al.

* Add a note about implicitly enabled extensions.

* Add a demo, sdk/demos/rive/bubbles.html, that implements the "advanced" blend equations using pixel local storage.

* Minor tweaks.


extensions/WEBGL_shader_pixel_local_storage/extension.xml
```diff
@@ -37,9 +37,20 @@
...
+        </div>
+      </addendum>
+      <addendum>
+        <div class="note">
+          When <code>WEBGL_shader_pixel_local_storage</code> is enabled, the following WebGL
+          extensions are enabled implicitly:
+          <p>
+            <code>OES_draw_buffers_indexed</code><br/>
+            <code>EXT_color_buffer_float</code><br/>
+            <code>EXT_color_buffer_half_float</code>
```
lexaknyazev, line 52: Note for the future: WebGL PLS CTS should explicitly check for this (e.g., by trying to create a renderbuffer).

csmartdalton: Sounds good. Where are the CTS tests? Can I start porting the ANGLE tests to CTS?

lexaknyazev: PLS extension should be in `sdk/tests/conformance2/extensions`, added as a new `.html` file and a new entry in `00_test_list.txt` there.

csmartdalton: Awesome, thanks!

@kenrussell , if I add tests to `sdk/tests/conformance2/extensions` will they automatically run on Chrome when I add the JS bindings?

kenrussell: @csmartdalton yes, if it's structured similarly to the other extension tests, then as soon as a browser implements the extension, the test will start to run. Note the importance of adding it to `00_test_list.txt`.

In Chrome, we'll need to roll forward the WebGL conformance suite in order to pick up the new test on the automated bots.

extensions/WEBGL_shader_pixel_local_storage/extension.xml
```diff
@@ -109,24 +135,75 @@ interface WEBGL_shader_pixel_local_storage {
...
+      Sets the unsigned integer clear value for the given plane.
+      <div class="note">
+        If <code>value</code> has less than <code>4</code> elements, generates an
+        <code>INVALID_OPERATION</code> error.
+      </div>
+    </function>
+  </newfun>
+
+  <newfun>
+    <function name="framebufferPixelLocalClearValuefvWEBGL" type="undefined">
```
kenrussell, line 172: Since these entry points take in only a small amount of data and aren't reading back values to JavaScript, the pattern in the WebGL 2.0 spec for the `uniform*` entry points seems more applicable - take in e.g. Float32List (e.g. either a Typed Array or a sequence of floats) and have the src_offset be an optional argument defaulting to 0. To save time, I'll merge this and propose a follow-on PR.

kenrussell (approved):
Great work Chris, looks good overall. Merging now and will propose a follow-on PR to address my one concern about the framebufferPixelLocalClearValue entry points.

extensions/WEBGL_shader_pixel_local_storage/extension.xml
csmartdalton: @lexaknyazev is there an easy way to check if an extension is enabled? I can't call `getExtension` since that will automatically enable it. Or do I have to try and use the dependency extensions and make sure they are on?

lexaknyazev: 1. Unextended WebGL 2.0 does not support float and half-float renderbuffers. This could be observed by trying to create one and checking for a generated error.
2. The PLS extension implicitly enables float and/or half-float color buffer extensions, therefore float and/or half-float renderbuffer creation may succeed after enabling the PLS extension.

However this seems inconsistent with the ANGLE PLS extension spec:
- Half-float formats are not mentioned at all
- Float format (R32F) is not optional; should the relevant extension be required?

csmartdalton: You are correct. This is not in the spec yet, but it's one of the issues in the current working doc:

https://docs.google.com/document/d/1BsXxar1gbJ_w-Xvy76VbgmLOgB3qON6BY6YB5LVOqwo/edit

The spec needs to be updated to mark float formats as dependent on EXT_color_buffer_float

PR #3529, opened 2023-03-16
Simplify framebufferPixelLocalClearValue* overloads.

kenrussell:
Use Float32List, Int32List and Uint32List (defined in the WebGL 1.0 and 2.0 IDL) to allow typed array and ECMAScript arrays to be passed in identically. This matches the style of WebGL 2.0's uniform* entry points.

Update extensions' XSL transform to support optional arguments in function parameter lists.

### DIFF
--- a/extensions/WEBGL_shader_pixel_local_storage/extension.xml
+++ b/extensions/WEBGL_shader_pixel_local_storage/extension.xml
@@ -80,19 +80,13 @@ interface WEBGL_shader_pixel_local_storage {
                                                      GLint level,
                                                      GLint layer);
   undefined framebufferPixelLocalClearValuefvWEBGL(GLint plane,
-                                                   sequence&lt;GLfloat&gt; value);
-  undefined framebufferPixelLocalClearValueivWEBGL(GLint plane,
-                                                   sequence&lt;GLint&gt; value);
-  undefined framebufferPixelLocalClearValueuivWEBGL(GLint plane,
-                                                    sequence&lt;GLuint&gt; value);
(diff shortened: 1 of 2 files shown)


kenrussell:
@csmartdalton could you take a look? I tested these changes by building the extension registry locally and they look good to me. Also @lexaknyazev . Thanks.

extensions/WEBGL_shader_pixel_local_storage/extension.xml
```diff
@@ -135,44 +129,8 @@ interface WEBGL_shader_pixel_local_storage {
...
-    </function>
-  </newfun>
-
-  <newfun>
-    <function name="framebufferPixelLocalClearValuefvWEBGL" type="undefined">
-      <param name="plane" type="GLint"/>
-      <param name="value" type="Float32Array"/>
-      <param name="src_offset" type="GLuint"/>
+      <param name="value" type="Float32List"/>
+      <param name="src_offset" type="GLuint" default="0"/>
```
csmartdalton, line 133: Looks like I mixed up styles for `src_offset` and `srcOffset`. Is there a preferred convention?

Also, love the `default` property!

csmartdalton (approved):
Thanks for doing this Ken! LGTM, with one nit on a style issue I introduced previously

extensions/WEBGL_shader_pixel_local_storage/extension.xml
kenrussell: Could you change this to `srcOffset` in your follow-up to match the WebGL 2.0 spec? Thanks.

PR #3530, opened 2023-03-17
Initial conformance tests for WEBGL_shader_pixel_local_storage

csmartdalton:
Creates a set of tests that verify the non-normative WebGL behavior, as well as providing some scaffolding to verify the ANGLE extension is properly hooked up.

If these all look good, we can start porting tests from the ANGLE extension next.


csmartdalton:
@kenrussell @lexaknyazev please take a look.

Ken, this collides with your previous PR. I can rebase after you land.

I think this, combined with the ANGLE tests, ought to be enough coverage to land a draft implementation in Chrome!

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
```diff
@@ -0,0 +1,334 @@
...
+}
+
+// Check that the dependency extensions of WEBGL_shader_pixel_local_storage are properly enabled or
+// disabled.
+function checkDependencyExtensionsEnabled(enabled) {
+  if (wtu.getSupportedExtensionWithKnownPrefixes(gl, "OES_draw_buffers_indexed") !== undefined) {
+    gl.getIndexedParameter(gl.BLEND_EQUATION_RGB, 1);
+    wtu.glErrorShouldBe(gl, enabled ? gl.NONE : gl.INVALID_ENUM,
+                        "OES_draw_buffers_indexed not enabled or disabled as expected");
+    gl.bindRenderbuffer(gl.RENDERBUFFER, null);
```
lexaknyazev, line 102: Seems unnecessary in this block.

csmartdalton:
Closed by accident.

This is rebased and ready for review.

extensions/WEBGL_shader_pixel_local_storage/extension.xml
lexaknyazev, line 1: Please remove `class="note"` from error definitions as they are normative in WebGL.

lexaknyazev (commented):
Please add a test that asserts correct PLS behavior on page composition (`wtu.waitForComposite`) events without an app explicitly ending PLS. It should include app-observable state of indexed color masks (if they could be implicitly changed) and implicit clears of the default FBO.

extensions/WEBGL_shader_pixel_local_storage/extension.xml
kenrussell: @csmartdalton could you take care of this?

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
```diff
@@ -0,0 +1,334 @@
...
+  wtu.glErrorShouldBe(gl, gl.INVALID_VALUE);
+
+  // Check that srcOffset works properly.
+  const arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
+  pls.framebufferPixelLocalClearValuefvWEBGL(0, new Float32Array(arr), 1);
+  wtu.glErrorShouldBe(gl, gl.NONE);
+  shouldBeTrue(`arraysEqual(pls.getFramebufferPixelLocalStorageParameterWEBGL(
+                                0, pls.PIXEL_LOCAL_CLEAR_VALUE_FLOAT_WEBGL),
+                            new Float32Array([1, 2, 3, 4]))`);
+  pls.framebufferPixelLocalClearValuefvWEBGL(1, new Float32Array(arr), 2);
```
kenrussell, line 233: For each of these entry points, could you also test the overload wehre you pass in "arr" rather than a Typed Array created from it?

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
```diff
@@ -0,0 +1,334 @@
...
+  const fbo = gl.createFramebuffer();
+  gl.bindFramebuffer(gl.FRAMEBUFFER, fbo);
+  pls.framebufferTexturePixelLocalStorageWEBGL(0, tex, 0, 0);
+  wtu.glErrorShouldBe(gl, gl.NONE);
+
+  gl.viewport(0, 0, 10, 10);
+
+  const vs = `#version 300 es
+  void main() {
+      gl_Position.x = (gl_VertexID & 1) == 0 ? -1. : 1.;
```
kenrussell, line 279: Could you add a comment describing what geometry this renders?

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
```diff
@@ -0,0 +1,334 @@
...
+      gl_Position.y = (gl_VertexID & 2) == 0 ? -1. : 1.;
+      gl_Position.zw = vec2(0, 1);
+  }`;
+
+  const fs = `#version 300 es
+  #extension GL_ANGLE_shader_pixel_local_storage : require
+  precision lowp float;
```
kenrussell, line 286: Is lowp guaranteed to have sufficient precision for this test to pass on all GPUs?

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
```diff
@@ -0,0 +1,334 @@
...
+  wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "parameter unknown without enabling the extension");
+  shouldBeNull(
+    "gl.getParameter(0x96E3 /*PIXEL_LOCAL_STORAGE_ACTIVE_PLANES_WEBGL*/)");
+  wtu.glErrorShouldBe(gl, gl.INVALID_ENUM, "parameter unknown without enabling the extension");
+  wtu.glErrorShouldBe(gl, gl.NONE);
+}
+
+// Check that the dependency extensions of WEBGL_shader_pixel_local_storage are properly enabled or
+// disabled.
+function checkDependencyExtensionsEnabled(enabled) {
```
kenrussell, line 97: As the first thing in this function could you log something like:

`debug("Checking that dependent extensions are " + (enabled ? "enabled" : "disabled"));`

Otherwise if errors crop up below, it'll be difficult to tell from the logs which case failed.

kenrussell (changes requested):
Great work Chris on these tests! A few small revisions requested.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
csmartdalton: Yes, `lowp` is [specified](https://www.khronos.org/files/opengles_shading_language.pdf) to have a floating point precision of `2^-8`, which is enough to handle 8-bit colors precisely.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
csmartdalton: Done. Moved it to the top with a better name too.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
csmartdalton: Done. That was actually my intention, but I must have written this code too late at night 😂

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
csmartdalton: Done

extensions/WEBGL_shader_pixel_local_storage/extension.xml
csmartdalton: Done

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
csmartdalton: Done

csmartdalton (commented):
> Please add a test that asserts correct PLS behavior on page composition (wtu.waitForComposite) events without an app explicitly ending PLS.

This is done, but commented for now because it breaks Chrome. Good test idea!

> It should include app-observable state of indexed color masks (if they could be implicitly changed) and implicit clears of the default FBO.

ANGLE has a comprehensive test for indexed color mask. I think I can bring in all of the ANGLE validation tests with this PR. Did you have anything more specific in mind?

lexaknyazev:
> ANGLE has a comprehensive test for indexed color mask. I think I can bring in all of the ANGLE validation tests with this PR. Did you have anything more specific in mind?

Browsers may change color masks to enable all channels while doing an implicit clear; alpha mask may be forced to be false for contexts created with opaque default framebuffer (via `WebGLContextAttributes`).

The test should ensure that browser's and PLS' implicit color mask updates work correctly together and the last user-set state is restored.

csmartdalton:
> The test should ensure that browser's and PLS' implicit color mask updates work correctly together and the last user-set state is restored.

Got it, thanks!

https://chromium-review.googlesource.com/c/angle/angle/+/4355032 would enable us to easily handle the clear after compositing when thte app leaves PLS enabled. Do you think these spec changes look OK?

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
kenrussell: Did you force-push the updated version here? I'm not seeing these changes.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
kenrussell: Did you force-push the new revision here? I'm not seeing these changes.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
kenrussell: Thanks.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
kenrussell: Not seeing this update either - possible to re-check whether your latest version's been pushed here?

kenrussell (changes requested):

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
csmartdalton: Derp. Forgot to commit my local changes.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
csmartdalton: Done

csmartdalton (commented):
I had some local changes I had forgotten to commit before I did my last update 🤦.

Let's hold off until I finish building Chrome and can verify these tests are all correct and passing, but I think we should be just about ready.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
```diff
@@ -0,0 +1,408 @@
...
+<canvas id="canvas" width="128" height="128"> </canvas>
+<div id="console"></div>
+<script>
+"use strict";
+description("This test verifies the functionality of the WEBGL_shader_pixel_local_storage " +
+            "extension, if it is available.");
+
+const wtu = WebGLTestUtils;
+const canvas = document.getElementById("canvas");
+const gl = wtu.create3DContext(canvas, {alpha: false}, 2);
```
lexaknyazev, line 24: Could the test be repeated for the default (`alpha: true`) case? No-alpha composition may use different code paths.

lexaknyazev: Create a regular canvas from JS, get an alpha-enabled context, run tests, delete that canvas, repeat with alpha disabled.

It's a good idea to check PLS operations with `OffscreenCanvas` but it would better be a separate test.

csmartdalton: Done! Can't quite delete it yet though because it needs to go to sleep for compositing.

This test also captures PLS being active on two contexts at once.

OffscreenCanvas might be a good candidate for some of the future tests I can import from ANGLE.

Does this look good to land now?

extensions/WEBGL_shader_pixel_local_storage/extension.xml
```diff
@@ -115,53 +124,44 @@ interface WEBGL_shader_pixel_local_storage {
...
       <param name="value" type="Int32List"/>
-      <param name="src_offset" type="GLuint" default="0"/>
+      <param name="srcOffset" type="GLuint" default="0"/>
       Sets the integer clear value for the given plane.
-      <div class="note">
-        If <code>value</code> has less than <code>src_offset + 4</code> elements, generates an
-        <code>INVALID_OPERATION</code> error.
-      </div>
+      <p/>
+      If <code>value</code> has less than <code>srcOffset + 4</code> elements, generates an
```
kenrussell, line 154: I think a line was accidentally deleted here.

extensions/WEBGL_shader_pixel_local_storage/extension.xml
```diff
@@ -115,53 +124,44 @@ interface WEBGL_shader_pixel_local_storage {
...
       <param name="value" type="Uint32List"/>
-      <param name="src_offset" type="GLuint" default="0"/>
-      Sets the unsigned integer clear value for the given plane.
-      <div class="note">
-        If <code>value</code> has less than <code>src_offset + 4</code> elements, generates an
-        <code>INVALID_OPERATION</code> error.
-      </div>
+      <param name="srcOffset" type="GLuint" default="0"/>
+      <p/>
+      If <code>value</code> has less than <code>srcOffset + 4</code> elements, generates an
```
kenrussell, line 164: It looks like the documentation and INVALID_OPERATION error were lost from the docs here.

kenrussell (changes requested):
Thanks for revising this Chris. It looks like there were a couple of small mistakes in the extension.xml. Did you try running "make" in your extensions/ directory locally?

extensions/WEBGL_shader_pixel_local_storage/extension.xml
csmartdalton: Good catch!

extensions/WEBGL_shader_pixel_local_storage/extension.xml
csmartdalton: Done

extensions/WEBGL_shader_pixel_local_storage/extension.xml
csmartdalton: > Thanks for revising this Chris. It looks like there were a couple of small mistakes in the extension.xml. Did you try running "make" in your extensions/ directory locally?

I did run `make`, and just now ran it again. All looks right now!

csmartdalton:
> Thanks for revising this Chris. It looks like there were a couple of small mistakes in the extension.xml. Did you try running "make" in your extensions/ directory locally?

I think this should all be good now. I ran `make` locally and checked that the output reflected the most recent edits.

kenrussell (approved):
Thanks for the updates - looks good!

PR #3536, opened 2023-04-20
Fix slow tests runs for e.g. multiple_attributes.output.html when run via harness.

kdashg:
On my machine, the subtests in multiple_attributes.output.html still run over 100% slower when run in harness (at least in Firefox), but they do run in tolerable time now.

### DIFF
--- a/sdk/tests/deqp/framework/common/tcuTestCase.js
+++ b/sdk/tests/deqp/framework/common/tcuTestCase.js
@@ -453,7 +453,32 @@ goog.scope(function() {
 
                 if (inited) {
                     // Run the test, save the result.
+
+                    const debug = tcuTestCase._debug = tcuTestCase._debug || (() => {
+                        function LapStopwatch() {
+                            this.lap = function() {
+                                const now = performance.now();
+                                const ret = now - this.last;
+                                this.last = now;
+                                return ret;
+                            };
+                            this.lap();
+                        }
+                        return {
+                            stopwatch: new LapStopwatch(),
+                            testDoneCount: 0,
+                        };
+                    })();
+                    const overheadDur = debug.stopwatch.lap();
+
                     tcuTestCase.lastResult = state.currentTest.iterate();
+
+                    const testDur = debug.stopwatch.lap();
+                    debug.testDoneCount += 1;
+                    console.log(
+                        `[test ${debug.testDoneCount}] Ran in ${testDur}ms`,
+                        `(+ ${overheadDur}ms overhead)`,
+                    );
                 } else {
                     // Skip uninitialized test.
                     tcuTestCase.lastResult = tcuTestCase.IterateResult.STOP;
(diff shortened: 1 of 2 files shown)


kdashg:
As noted in a comment, that test in particular causes over 33M calls to reportResults. Firefox spends over 50% of the time in memset as it makes profiling markers for each setTimeout call, and dev tools web console gets pretty bogged down dealing with so many console.log prints.

sdk/tests/deqp/framework/common/tcuTestCase.js
```diff
@@ -453,7 +453,32 @@ goog.scope(function() {
 
                 if (inited) {
                     // Run the test, save the result.
+
+                    const debug = tcuTestCase._debug = tcuTestCase._debug || (() => {
```
kenrussell, line 457: Is the double assignment to `tcuTestCase._debug` intentional?

sdk/tests/js/webgl-test-harness.js
```diff
@@ -595,11 +616,15 @@ TestHarness.prototype.getTest = function(url) {
 TestHarness.prototype.reportResults = function(url, success, msg, skipped) {
   url = FilterURL(url);
   var test = this.getTest(url);
-  this.clearTimeout(test);
-  log((success ? "PASS" : "FAIL") + ": " + msg);
+  if (0) {
```
kenrussell, line 619: Does removing this significantly reduce the ability to debug test failures from the logs?

Would adjusting the specific test `deqp/functional/gles3/vertexarrays/multiple_attributes.output.html` be warranted instead, if so?

kenrussell (approved):
LGTM with a couple of questions.

sdk/tests/deqp/framework/common/tcuTestCase.js
kdashg: It's not double-assignment. The less terse way to write it is:
```
tcuTestCase._debug = tcuTestCase._debug || (() => { // executed if !tcuTestCase._debug };
const debug = tcuTestCase._debug;
```

sdk/tests/js/webgl-test-harness.js
kdashg: Once the harness is attached, the harness should choose if it wants to console.log or not. (and the main test runner UI should probably not spew to console.log)
Maybe the test should be adjusted, but that can be handled separately.

sdk/tests/deqp/framework/common/tcuTestCase.js
kenrussell: Thanks for explaining.

sdk/tests/js/webgl-test-harness.js
kenrussell: OK.

kenrussell (approved):

PR #3538, opened 2023-06-07
Fix XML syntax, typos, and formatting; expand XSL rules

lexaknyazev:


kenrussell (approved):
Very nice improvements!

PR #3541, opened 2023-06-19
Enhance timer query tests; update specs

lexaknyazev:


kenrussell (approved):
Looks good. Thanks for making these tests more rigorous.

PR #3546, opened 2023-06-29
Proposal: EXT_clip_control

lexaknyazev:
Due to historical reasons, WebGL uses [-1, +1] range for Z values. This leads to suboptimal depth buffer precision utilization because the best precision is allocated to mid-range values. When WebGL is implemented on top of Direct3D, Metal, or Vulkan, that range is rescaled to [0, +1] further increasing the loss and using extra per-vertex shader instructions.

The proposed extension would allow applications to directly use [0, +1] range.


kenrussell (approved):

PR #3547, opened 2023-06-29
Proposal: EXT_depth_clamp

lexaknyazev:
Vertices located closer than the near or farther than the far planes are always clipped. Various shadow rendering techniques greatly benefit from being able to squash (aka “pancake”) all geometry outside of the defined range instead of clipping it.

The proposed extension would allow applications to enable depth clamping while implicitly disabling near and far clipping planes.


kenrussell (approved):

PR #3548, opened 2023-06-29
Proposal: WEBGL_polygon_mode

lexaknyazev:
Rendering only triangle edges, i.e., wireframe, is incredibly useful when debugging GPU applications. For example:
  - Preventing an empty screen when the view is accidentally blocked
  - Confirming correct assembly of dynamically created geometry

Overlaying wireframe over filled triangles is useful for debugging fragment-based micro-geometry effects such as normal maps, bump maps, parallax maps, etc.

Simply switching primitive mode to LINES is not enough:
  - Incompatible vertex arrangement / index buffers
  - Depth bias is ignored for LINES, making correct overlay very difficult to achieve

The proposed extension would allow applications to toggle between solid and wireframe polygon rendering as well as enabling wireframe depth bias.


kenrussell (approved):

PR #3549, opened 2023-06-29
Proposal: EXT_conservative_depth

lexaknyazev:
Writing to `gl_FragDepth`, e.g., to render soft particles, disables implicit hardware optimizations and therefore increases the number of wasted fragment shader invocations.

The proposed extension would allow applications to hint hardware in what direction the written fragment depth is different from the rasterizer’s output. This re-enables implicit hardware optimizations and therefore prevents fragment shader invocations that are supposed to fail the depth test anyway.


kenrussell (approved):

PR #3550, opened 2023-06-29
Proposal: NV_shader_noperspective_interpolation

lexaknyazev:
- Perspective correction is always applied to non-flat varyings
- Extra computations are performed when not needed, e.g., for 2D apps
- Applications that need screen-space interpolation of 3D vectors have to implement workarounds to undo perspective correction

The proposed extension would allow applications to disable perspective correction of non-flat varyings.


kenrussell (approved):

PR #3551, opened 2023-06-29
Proposal: OES_sample_variables

lexaknyazev:
Although WebGL 2.0 supports multisampled rendering, it does not provide relevant fragment built-ins.

The proposed extension would allow applications to control per-fragment sample mask as well as specialize fragment outputs depending on the sample being processed.


kenrussell (approved):

PR #3552, opened 2023-06-29
Proposal: OES_shader_multisample_interpolation

lexaknyazev:
Even with multisampled rendering, fragment varyings are interpolated at a pixel center or centroid at best.

The proposed extension would allow applications to enable per-sample varying interpolation and to explicitly query interpolated values at centroid, at a specific sample, or at an arbitrary in-pixel location.


kenrussell (approved):

PR #3553, opened 2023-06-29
Proposal: EXT_blend_func_extended

lexaknyazev:
Regular blending operations do not allow outputting an extra RGBA factor that could be used for modulating existing color buffer content independently of the primary fragment output. Example use cases include dissolving objects, colored transparent surfaces, glyph rendering, decals, etc.

Although these rendering effects could be achieved via `WEBGL_shader_pixel_local_storage` functionality, dual-source blending typically has wider reach and equivalent or better performance.

The proposed restricted extension would allow applications to use dual-source blending in both WebGL 1.0 and 2.0 contexts.


kenrussell (approved):

PR #3554, opened 2023-06-29
Proposal: EXT_render_snorm

lexaknyazev:
Signed normalized formats are texturable but not renderable in WebGL 2.0. For values in [-1, +1] range, `snorm16` has much better precision and likely higher throughput than `float16`.

The proposed extension would allow applications to render into signed normalized pixel formats, including 16-bit formats when `EXT_texture_norm16` is enabled.


kenrussell (approved):

PR #3556, opened 2023-06-29
Proposal: WEBGL_render_shared_exponent

lexaknyazev:
Shared exponent floating-point pixel format, i.e., RGB9_E5 is texturable but not renderable in WebGL 2.0. The only other small-float alternative is R11F_G11F_B10F, which may exhibit hue shifts due to asymmetry in channel encodings.

The proposed extension would allow applications in memory-constrained environments to choose between two small-float formats depending on the acceptable tradeoffs.


kenrussell (approved):

PR #3557, opened 2023-06-29
Proposal: WEBGL_stencil_texturing

lexaknyazev:
It is not possible to directly fetch stencil values from a combined depth-stencil texture in WebGL 2.0.

The proposed extension would allow applications to toggle between depth and stencil values when accessing combined depth-stencil textures in shaders.


kenrussell (approved):

PR #3558, opened 2023-06-29
Proposal: EXT_texture_mirror_clamp_to_edge

lexaknyazev:
When a texture is symmetrical, only a part of it (as little as 1/8 in 3D case) could be stored in GPU memory. The existing MIRRORED_REPEAT mode is incompatible with certain effects that rely on the edge color being sampled for out-of-range texture coordinates.

The proposed extension would allow applications to use mirrored clamp-to-edge texture wrapping mode.


kenrussell (approved):

PR #3539, opened 2023-06-15
Editorial: account for SharedArrayBuffer change in Web IDL

annevk:
See https://github.com/whatwg/webidl/pull/1311 for context.

At the same time adjust extract-idl.py to hold Python correctly ("!=" instead of "is not").

PR #3560, opened 2023-06-29
Incorporate @annevk 's SharedArrayBuffer and other Web IDL changes.

kenrussell:
See whatwg/webidl#1311 for context.

At the same time adjust extract-idl.py to hold Python correctly ("!=" instead of "is not").

Thanks to @annevk for these changes in #3539 . They were contributed by @annevk and @grorg under Khronos' CLA; merging them separately to bypass the automated checks.


kenrussell:
@kdashg please review.

@annevk @grorg thanks for these edits and for explicitly granting permission to use them under Khronos' CLA in #3539 .

kdashg (approved):

PR #3562, opened 2023-07-06
Fix async portions of PLS test.

kenrussell:
Await the results of asynchronous functions, and explicitly call finishTest() at the end of runTest().

Follow-on to #3530 .

### DIFF
--- a/sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
+++ b/sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
@@ -46,7 +46,7 @@ function arraysEqual(a, b) {
   return true;
 }
 
-function runTest() {
+async function runTest() {
   if (!gl) {
     return function() {
       testFailed("WebGL2 context does not exist");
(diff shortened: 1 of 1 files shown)


kenrussell:
@csmartdalton please feel free to review too.

lexaknyazev (approved):

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
```diff
@@ -67,8 +67,10 @@
   checkInitialValues();
   checkWebGLNonNormativeBehavior();
 
-  checkRendering(gl);
-  checkRendering(gl_no_alpha);
+  await checkRendering(gl);
+  await checkRendering(gl_no_alpha);
+
+  finishTest();
```
csmartdalton, line 73: Just for my own understanding, how does the test harness know that the test is async now and that it should wait for `finishTest`? Before was it assuming the test was over as soon as it finished parsing?

csmartdalton (approved):
Thank you for putting this together Ken!

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
lexaknyazev: `js-test-post.js` notifies the test harness explicitly. When included statically, it's called as soon as the parsing is finished. With `finishTest`, the same script is added on the fly.

lexaknyazev:
@kenrussell Please add another `finishTest()` call when the extension is not supported to avoid test timeout.

kenrussell:
> @kenrussell Please add another `finishTest()` call when the extension is not supported to avoid test timeout.

Thanks for catching that. Fixing in https://github.com/KhronosGroup/WebGL/pull/3563 .

PR #3563, opened 2023-07-06
Call finishTest() when PLS isn't supported.

kenrussell:
Timeouts were accidentally introduced in the last commit by failing to do this. Follow-on to #3562 and #3530.

### DIFF
--- a/sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
+++ b/sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
@@ -50,6 +50,7 @@ async function runTest() {
   if (!gl) {
     testFailed("WebGL2 context does not exist");
     finishTest();
+    return;
   }
 
   debug("\nCheck the behavior surrounding WEBGL_shader_pixel_local_storage being enabled.");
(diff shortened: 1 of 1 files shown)


kenrussell:
@csmartdalton FYI.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
```diff
@@ -50,6 +50,7 @@
   if (!gl) {
     return function() {
       testFailed("WebGL2 context does not exist");
+      finishTest();
     }
   }
```
lexaknyazev, line 55: This part seems incorrect. Should probably be:
```js
if (!gl) {
  testFailed("WebGL context does not exist");
  finishTest();
  return;
}
```

kenrussell: Agreed, thanks, fixed.

lexaknyazev: An extra return from the block is needed to exit the function early.

kenrussell: Thanks again, fixed.

lexaknyazev (approved):

PR #3566, opened 2023-07-10
Reject WEBGL_webcodecs_video_frame extension.

kenrussell:
This extension informed WebGPU's zero-copy video frame importing functionality, but was difficult to use in WebGL. The prototype implementation is causing technical debt in Chromium per crbug.com/1462924, so moving this extension to rejected status.


kenrussell:
Note, I confirmed with @jchen10 offline that it's OK to reject this extension.

PR #3567, opened 2023-07-10
Reject WEBGL_texture_source_iframe extension.

kenrussell:
The security constraints needed for this extension were too difficult to define, despite a couple of attempts over a couple of years' time.


kenrussell:
Note, I confirmed with @zhenyao offline that it's OK to reject this extension.

kdashg (approved):

PR #3568, opened 2023-07-21
Move EXT_clip_control to draft; add tests

lexaknyazev:


kenrussell (approved):
Thanks, new test looks good and so does promoting the extension to draft.

PR #3569, opened 2023-07-21
Move EXT_depth_clamp to draft; add tests

lexaknyazev:


kenrussell (approved):
New test looks good, as does moving this to draft status.

PR #3570, opened 2023-07-21
Move WEBGL_polygon_mode to draft; add tests

lexaknyazev:


sdk/tests/conformance/extensions/webgl-polygon-mode.html
```diff
@@ -0,0 +1,184 @@
...
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be no errors");
+}
+
+function checkDiagonal(r, g, b) {
+    const pixels = new Uint8Array(w * h * 4);
+    gl.readPixels(0, 0, w, h, gl.RGBA, gl.UNSIGNED_BYTE, pixels);
+    for (let i = 0; i < w; i++)
+    {
+        const p = (i * w + i) * 4;
+        const c = pixels.slice(p, p + 4);
```
kenrussell, line 101: As a follow-up, this will allocate a lot of temporary Uint8Array objects. To optimize this would you consider calling `p` the `baseIndex`, or similar, and checking `pixels[baseIndex + 0]`, `+ 1`, and `+ 2`?

kenrussell (approved):
New test looks good, as does promoting this to draft. One comment to consider as a follow-up.

sdk/tests/conformance/extensions/webgl-polygon-mode.html
lexaknyazev: Fixed in #3574.

PR #3571, opened 2023-07-24
Move EXT_conservative_depth to draft; add tests

lexaknyazev:


kenrussell (approved):
Test looks good as does moving the extension to draft.

PR #3572, opened 2023-07-24
Move NV_shader_noperspective_interpolation to draft; add tests

lexaknyazev:


kenrussell (approved):
New test looks good - I'm a little surprised that the underlying spec will guarantee the two geometries raster exactly the same - and promoting the extension to draft looks good as well.

PR #3573, opened 2023-07-26
Move multisampling shader extensions to draft; add tests

lexaknyazev:


sdk/tests/conformance2/extensions/oes-sample-variables.html
```diff
@@ -0,0 +1,474 @@
...
+<div id="description"></div>
+<div id="console"></div>
+<script>
+"use strict";
+description("This test verifies the functionality of the OES_sample_variables extension, if it is available.");
+
+debug("");
+
+var wtu = WebGLTestUtils;
+var gl = wtu.create3DContext("c", null, 2);
```
kenrussell, line 27: The tests below strictly assume `{ antialias:false }`; even though it's the default in `wtu.create3DContext` I wonder whether it's worth explicitly specifying it here.

sdk/tests/conformance2/extensions/oes-sample-variables.html
```diff
@@ -0,0 +1,474 @@
...
+            }
+        }`;
+    gl.useProgram(wtu.setupProgram(gl, [wtu.simpleVertexShaderESSL300, frag]));
+
+    wtu.setupUnitQuad(gl);
+    wtu.drawUnitQuad(gl);
+
+    gl.bindFramebuffer(gl.READ_FRAMEBUFFER, fbo);
+    gl.bindFramebuffer(gl.DRAW_FRAMEBUFFER, null);
+    gl.blitFramebuffer(0, 0, 32, 32, 0, 0, 32, 32, gl.COLOR_BUFFER_BIT, gl.NEAREST);
```
kenrussell, line 153: This is what assumes the back buffer isn't antialiased.

kenrussell (approved):
New tests look good. I didn't understand all of the OES_shader_multisample_interpolation test from reading it, but let's integrate the tests and run them against a couple of implementations, as well as promote the extensions to draft.

sdk/tests/conformance2/extensions/oes-sample-variables.html
lexaknyazev: Fixed in #3574.

PR #3574, opened 2023-07-27
Address new tests feedback

lexaknyazev:
Follow-up to #3570 and #3573.


kenrussell (approved):
Thanks for these revisions; especially appreciate the details in the OES_shader_multisample_interpolation test.

PR #3576, opened 2023-07-31
Move texture extensions to draft; add tests

lexaknyazev:


sdk/tests/conformance/extensions/ext-texture-mirror-clamp-to-edge.html
```diff
@@ -0,0 +1,213 @@
...
+
+    function checkPixels() {
+        function checkPixel(x, y, color) {
+            const screen = (s, t) => s * (t * 0.5 + 0.5);
+            wtu.checkCanvasRect(gl, screen(w, x), screen(h, y), 1, 1, color,
+                                `(${x.toFixed(3)}, ${y.toFixed(3)}): ${color} `);
+        }
+        for (const signX of [+1, -1]) {
+            for (const signY of [+1, -1]) {
+                checkPixel(signX * 0.125, signY * 0.125, black);
```
kenrussell, line 139: The meaning of these coefficients is probably obvious when running the tests, but they're confusing when reviewing the code. Comments would help.

sdk/tests/conformance2/extensions/webgl-stencil-texturing.html
```diff
@@ -0,0 +1,270 @@
...
+        ];
+
+        const programs = [
+            [readDepthProgram, 'depth'],
+            [readStencilProgram, 'stencil']
+        ];
+
+        function validFilters(magFilter, minFilter) {
+            return magFilter == gl.NEAREST &&
+                   (minFilter == gl.NEAREST || minFilter == gl.NEAREST_MIPMAP_NEAREST);
```
kenrussell, line 197: Glad that you're testing that e.g. linear filtering isn't supported, and that the texture is incomplete if using other filters than these. Linear filtering was supported for depth textures as undefined behavior in core OpenGL ES, and some developers started relying on it and filed bugs about its absence on certain GPUs.

kenrussell (approved):
Looks good overall. One request for expanded comments in the texture mirroring test.

sdk/tests/conformance/extensions/ext-texture-mirror-clamp-to-edge.html
lexaknyazev: Done.

kenrussell: Thanks.

kenrussell (approved):
Thanks for the revised comment. Looks good.

PR #3578, opened 2023-08-14
Add smoke test that texImage2D doesn't lose bit depth

kainino0x:
I *believe* this is expected behavior for >8bpc image sources, although I do recall there was the EXT_texture_norm16_dom_source proposal and I'm not sure what that means for this behavior.

This is a port of https://chromium-review.googlesource.com/c/chromium/src/+/4772575 which added (a slightly weaker version of) this smoke test to Chrome for @ccameron-chromium's fix to preserve bit depth in cases like this.
Passes in Chrome Canary 118.0.5948.0.


lexaknyazev:
FWIW, WebKit truncates 16-bit PNGs to 8 bits:
https://github.com/WebKit/WebKit/blob/58a7098b60208c1ed69e6949a1d5984e829ad127/Source/WebCore/platform/image-decoders/png/PNGImageDecoder.cpp#L377-L378

/cc @kkinnunen-apple

kenrussell (approved):
In the WebGL WG meeting on 2023-09-07, Firefox, WebKit and Chromium representatives agreed this test should be landed, even if it's currently failing in some browsers. This can be a first step toward guaranteeing that some kinds of extended range / precision images are uploaded losslessly to WebGL.

Issue #3582, opened 2023-08-31
DEPTH_STENCIL_ATTACHMENT should be tested to behave as expected when the attachment object has been deleted


kkinnunen-apple:
DEPTH_STENCIL_ATTACHMENT should be tested to behave as expected when the attachment object has been deleted

PR #3583, opened 2023-08-31
Test that deleted fbo attachments behave same as undeleted wrt DEPTH_…

kkinnunen-apple:
…STENCIL_ATTACHMENT (#3582)

Test that deleted fbo attachments behave same as undeleted wrt DEPTH_STENCIL_ATTACHMENT.


kenrussell (approved):
Sorry for dropping this review on the floor; the changes here look good.

PR #3585, opened 2023-09-04
Move EXT_blend_func_extended to draft; add tests

lexaknyazev:


sdk/tests/conformance/extensions/ext-blend-func-extended.html
```diff
@@ -0,0 +1,573 @@
...
+    shouldBe("ext.MAX_DUAL_SOURCE_DRAW_BUFFERS_EXT", "0x88FC");
+}
+
+function runQueryTests() {
+    debug("");
+    debug("Testing getParameter");
+    shouldBeGreaterThanOrEqual("gl.getParameter(ext.MAX_DUAL_SOURCE_DRAW_BUFFERS_EXT)", "1");
+    wtu.glErrorShouldBe(gl, gl.NO_ERROR, "should be no errors");
+
+    if (contextVersion == 1) {
```
kenrussell, line 115: I find it confusing when a test under `conformance/` changes behavior when run as a WebGL 2.0 test. Would you consider refactoring this into a .js file which is included both by `conformance/extensions/ext-blend-func-extended.html` and `conformance2/extensions/ext-blend-func-extended.html`, so each one runs with the particular context version? `conformance/extensions/00_test_list.txt` would mark that version of the test expiring in version `1.9.9` so it doesn't run in the 2.0 conformance suite. There are several examples of similarly factored tests in `sdk/tests/js/tests/`.

sdk/tests/conformance/extensions/ext-blend-func-extended.html
```diff
@@ -0,0 +1,573 @@
...
+        out mediump vec4 my_FragColor;
+        void main() {
+        #ifdef GL_EXT_blend_func_extended
+            my_FragColor = vec4(1.0, 0.0, 0.0, 1.0);
+        #else
+            #error no GL_EXT_blend_func_extended;
+        #endif
+        }`;
+    shaderSets.push([wtu.simpleVertexShader, macro100]);
+    if (contextVersion == 2) {
```
kenrussell, line 219: It'd be fine to run the version 100 shader in the WebGL 2.0 test to make sure there are no regressions parsing the ESSL 100 form of the shaders.

kenrussell (approved):
Nice work on the tests; one comment about how their code is factored.

sdk/tests/conformance/extensions/ext-blend-func-extended.html
lexaknyazev: Done.

kenrussell (approved):
Thanks for refactoring - looks good.

PR #3591, opened 2023-09-14
Avoid very long expressions in tests

fujii:
Windows WebKit is observing stack overflow in ANGLE for the following
tests due to very long expressions containing additions of 16384
uniform variables. <https://webkit.org/b/261297>

- glsl/bugs/complex-glsl-does-not-crash.html
- glsl/misc/shader-uniform-packing-restrictions.html

Split the long expressions by using addition assignment operators.

### DIFF
--- a/sdk/tests/conformance/glsl/bugs/complex-glsl-does-not-crash.html
+++ b/sdk/tests/conformance/glsl/bugs/complex-glsl-does-not-crash.html
@@ -58,7 +58,9 @@ varying vec4 v_varying;
 $(uniforms)
 void main()
 {
-    v_varying = $(code);
+    vec4 t = [0, 0, 0, 0];
+    t += $(code);
+    v_varying = t;
     gl_Position = a_position;
 }
 </script>
@@ -67,7 +69,9 @@ precision mediump float;
 $(uniforms)
 void main()
 {
-    gl_FragColor = $(code);
+    vec4 t = [0, 0, 0, 0];
+    t += $(code);
+    gl_FragColor = t;
 }
 </script>
 <script>
@@ -168,7 +172,7 @@ for (var ss = 0; ss < shaderTypes.length; ++ss) {
       }
       return {
         uniforms: uniforms.join("\n"),
-        code: codes.join(" + \n            "),
+        code: codes.join(";\n    t += "),
       };
     };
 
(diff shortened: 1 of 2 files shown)


kdashg:
The point of these tests is kinda to ensure that this sort of case is handled gracefully!
Firefox did IIRC increase our stack size for threads where we call the shader compiler, but there should be protections built into ANGLE for this too I thought?

kenrussell (changes requested):
Yes, sorry, the WebGL working group doesn't make test changes to work around browser issues. These tests work in both Firefox and Chromium on Windows, which both use ANGLE's shader translator. Now, these browsers do have a separate GPU process, which perhaps doesn't exist in the Windows WebKit port. Can you change WebKit/Win's build process and just increase the per-thread stack size to e.g. 2 MB? https://learn.microsoft.com/en-us/cpp/build/reference/f-set-stack-size?view=msvc-170

We won't merge this pull request, but for the moment will leave it open in case you want to discuss this further.

fujii:
There is a test case long-expressions-should-not-crash.html of testing long expressions.
Do other tests also have to test long expressions?

And, the maximum expression of the test is 10000, smaller than 16384.
https://github.com/KhronosGroup/WebGL/blob/0ca2eac35bab941ec27b50e51d31036df03acf84/sdk/tests/conformance/glsl/bugs/long-expressions-should-not-crash.html#L71
Why not 100000? Because it crashes ANGLE.
This isn't  aWindows WebKit spefic problem.

kenrussell:
Similar situations to this one have arisen in the past, and there was strong pushback from other WebGL working group members that the conformance suite should not be modified in order to work around browser implementation issues. The conformance rules don't cover this specific situation, but please see these paragraphs:

https://github.com/KhronosGroup/WebGL/blob/main/sdk/tests/CONFORMANCE_RULES.txt#L118

https://github.com/KhronosGroup/WebGL/blob/main/sdk/tests/CONFORMANCE_RULES.txt#L123

I understand that the tests may have been crafted to be "reasonable", that they may be reaching boundary conditions of the Windows WebKit port, and that all browsers are generally affected. At the same time, these tests are passing on every other WebGL implementation, and it should be straightforward to modify the thread stack size in Windows WebKit's build process so that these tests can pass as written. Would you please investigate doing so?

fujii:
Got it.

PR #3592, opened 2023-09-20
Add tests for gl.readPixels(), gl.texImage2D() and gl.texSubImage2D() from Wasm Memories of 4GB and 16GB in size.

juj:
Add tests for gl.readPixels(), gl.texImage2D() and gl.texSubImage2D() from Wasm Memories of 4GB and 16GB in size.

CC @kenrussell

### DIFF
--- a/sdk/tests/conformance2/00_test_list.txt
+++ b/sdk/tests/conformance2/00_test_list.txt
@@ -18,3 +18,4 @@ textures/00_test_list.txt
 transform_feedback/00_test_list.txt
 uniforms/00_test_list.txt
 vertex_arrays/00_test_list.txt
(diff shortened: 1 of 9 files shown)


kenrussell (approved):
Thank you Jukka for these tests!

With a small modification to make them runnable directly without URL query parameters, the 4 GB tests pass in Chromium ToT and Safari Tech Preview. They happen to fail in Firefox for reasons discussed in this week's WebGL concall; this will be worked on by the Mozilla team.

The 16 GB tests fail on all browsers right now. This seems fine; better to have the tests in place and decide whether to run them as part of the next conformance test snapshot later.

PR #3594, opened 2023-09-22
In new WebGL 2-only Wasm tests, fetch WebGL 2.0 context directly.

kenrussell:
This makes them runnable by navigating to them outside of the test harness and without URL query parameters.

Follow-on to #3592.


kenrussell:
FYI @juj . Merging myself; have tested these changes locally.

PR #3593, opened 2023-09-20
Add drawingBufferStorage test

ccameron-chromium:

PR #3596, opened 2023-09-23
Add drawingBufferStorage test from @ccameron-chromium .

kenrussell:
Integrate changes from #3593 . Refactor into separate WebGL 1 and 2 tests, which
share code, and move to the top-of-tree conformance suite.

WebGL 1 test is disabled for now, as it isn't fully passing and it's unclear whether
this is a bug in the test.


kenrussell:
@ccameron-chromium FYI, as mentioned in #3593.

sdk/tests/js/tests/drawingbuffer-storage-test.js
```diff
@@ -0,0 +1,275 @@
...
+
+let gl;
+let oldViewport;
+let width;
+let height;
+let format;
+let hasDrawingBufferStorage;
+let maxRenderbufferSize;
+
+function runTest(contextVersion) {
```
dmitshur, line 17: This `contextVersion` parameter is unused, which prevents runTest(1) and runTest(2) from testing both WebGL 1 and 2 contexts. I filed #3719 for this and another issue related to RGBA16F that I spotted.

PR #3597, opened 2023-10-10
Test parallel compilation of many small programs.

kenrussell:


kenrussell:
Stresses a code path in Chromium that wasn't easily exercised.

lexaknyazev (approved):

PR #3606, opened 2023-12-20
Upgrade extract-idl.py for Python3; regenerate IDLs.

kenrussell:
Update to html5lib-1.1, and incorporate webencodings-0.5.1 dependency, both of which which are required in order to run the extract-idl.py script on python3. html5lib is MIT licensed and webencodings is BSD licensed; both are compatible with this repository.

Regenerate WebGL 1.0 and 2.0 IDL from current specs.

Follow-on to #3222 .


specs/latest/1.0/Makefile
```diff
@@ -7,4 +7,4 @@ webgl.idl: index.html extract-idl.py Makefile
 	echo "//" >> webgl.idl
 	echo "// WebGL IDL definitions scraped from the Khronos specification:" >> webgl.idl
 	echo "// https://www.khronos.org/registry/webgl/specs/latest/" >> webgl.idl
-	PYTHONPATH=../../../resources/html5lib/src python extract-idl.py index.html >> webgl.idl
```
kainino0x, line 10: I'm guessing that when this script was updated to python3, the author ran it directly (with system-installed dependencies). It's not really obvious that it's supposed to be run through the Makefile instead, especially since it's executable and has a shebang. What do you think about either `chmod -x extract-idl.py` so that it's not executable directly on \*nix (and/or changing the shebang to `#!/usr/bin/false # Run this via Makefile instead`) or printing a warning from extract-idl.py if PYTHONPATH is not set?

kainino0x (approved):
LGTM. Would be nice if something checked that the IDL files were up to date on CQ :)

Also a comment about how we could keep this working in the future.

specs/latest/1.0/Makefile
kenrussell: It's possible that someone rebuilt the IDL having had those dependencies installed locally. From talking with you and @greggman just now, it sounds like guaranteeing a hermetic environment is a little tricky. For now I'd like to merge this so that the IDL is updated, and will look into adding a check to the Github actions to ensure the IDL is up to date with every pull request. That should help.

PR #3610, opened 2024-01-12
Promote WEBGL_blend_func_extended to community approved

lexaknyazev:


kenrussell (approved):
LGTM - thank you for taking care of all of the renamings needed because of the differences from the native extension (specifically, removing the associated functions).

PR #3613, opened 2024-01-26
Canvases should reset their contents when width or height is set

greggman:
even if the sizes are the same.

Note: Canvas 2D does this. There are various issues related to requestAnimationFrame and ResizeObserver that this affects. It would probably be best of all the APIs did the same thing so that the workarounds for resizing the canvas are the same across APIs

Safari passes this test, Firefox and Chrome fail

### DIFF
--- a/sdk/tests/conformance/canvas/canvas-test.html
+++ b/sdk/tests/conformance/canvas/canvas-test.html
@@ -182,7 +182,33 @@ if (!gl) {
                   "gl.colorMask should not change after canvas resize");
         shouldBe('getViewport()', '"0,0,300,150"');
         checkCanvasContentIs(0, 0, 0, 0);
-      }
+
+        debug("");
+        debug("change the size of the canvas to the same size it already is and see that it gets cleared");
+        gl.colorMask(true, true, true, true);
+        gl.clearColor(0.25, 0.5, 0.75, 1);
+        gl.clear(gl.COLOR_BUFFER_BIT | gl.DEPTH_BUFFER_BIT);
+        // Check it's not cleared
+        checkCanvasContentIs(64, 128, 192, 255);
+
+        gl.canvas.width = gl.canvas.width;
+        err = gl.getError();
+        // Some implementations might lost the context when resizing
+        if (err != gl.CONTEXT_LOST_WEBGL) {
+          checkCanvasContentIs(0, 0, 0, 0);
+
+          gl.clear(gl.COLOR_BUFFER_BIT | gl.DEPTH_BUFFER_BIT);
+          // Check it's not cleared
+          checkCanvasContentIs(64, 128, 192, 255);
+
+          gl.canvas.height = gl.canvas.height;
+          err = gl.getError();
+          // Some implementations might lost the context when resizing
+          if (err != gl.CONTEXT_LOST_WEBGL) {
+            checkCanvasContentIs(0, 0, 0, 0);
+          }
+        }
+     }
 
       debug("");
       finishTest();


greggman:
Probably need a similar test for OffscreenCanvas?

sdk/tests/conformance/canvas/canvas-test.html
```diff
@@ -182,7 +182,33 @@
...
         checkCanvasContentIs(0, 0, 0, 0);
-      }
+
+        debug("");
+        debug("change the size of the canvas to the same size it already is and see that it gets cleared");
+        gl.colorMask(true, true, true, true);
+        gl.clearColor(0.25, 0.5, 0.75, 1);
+        gl.clear(gl.COLOR_BUFFER_BIT | gl.DEPTH_BUFFER_BIT);
+        // Check it's not cleared
+        checkCanvasContentIs(64, 128, 192, 255);
```
kkinnunen-apple, line 192: Would be nice to test that stencil and depth are cleared, too. That's of course cumbersome..

kkinnunen-apple (approved):

greggman:
> Probably need a similar test for OffscreenCanvas?

FYI, Offscreen canvas is apparently different. Safari, Firefox, and Chrome don't reset the WebGL canvas when the size is set to the same size.

kenrussell (approved):
Hopefully it'll still be possible for browsers to optimize the reallocation of the back buffer by batching up changes to width and height.

kdashg:
I believe we have a different test that requires the opposite, or otherwise that there is content that assumes the opposite, because I see this line in Firefox:
https://searchfox.org/mozilla-central/rev/2a867dd1ab015c3ef24b774a57709fb3b3dc4961/dom/canvas/ClientWebGLContext.cpp#797
```c=
if (size == curSize) return NS_OK;  // MUST skip no-op resize
```

greggman:
Safari doesn't do this so any site that is expecting a no-op is not getting a no-op

MarkCallow:
What is the purpose of resetting the contents when width or height is set to the same size? I'm pretty sure this kind of choice is behind most of the cases where I see the content I want to read then the screen/window/canvas goes black and  the same content is redrawn usually after an irritatingly long time. This is ugly and annoying to the user.

greggman:
It is already true and expected in Canvas 2D. It can't be fixed there because people actually use it as a shortcut to clear the canvas. 

[The interop between compositing the page and canvas updates has a bunch of issues](https://stackoverflow.com/questions/77842752/how-do-i-use-resizeobserver-with-requestanimationframe-correctly). 

IMO, it would best of all the canvas APIs did the same thing here so that the correct usage for one type of context ("2d", "webgl", "webgl2", "webgpu", "bitmaprender") is the same for all types of contexts.  Otherwise you get into this issue where you do things one way, they seem to work, you pass on that "this is how you use canvas" message on the internet. You post examples. People copy the examples. It spreads far and wide. But, it turns out it's wrong because different canvas APIs treat this differently.  If they all treat it the same then the "this is how you use canvas" issue has one correct answer.

kdashg:
I do not think there is enough benefit here to outweigh the backwards-compat hit.

greggman:
What backward compat hit? 

It already clears on Safari so you wanted to be compatible for the last 11 years you needed to deal with the fact that it clears.

kdashg:
When I accidentally made this change in Firefox ~5years ago, I reverted it for a reason. I regret that I did not write down what it was.
It is not clear to me that this has always been the case in Safari either. It could be a regression, or maybe people have started working around it such that I would not have needed to revert this change when I made it in Firefox years ago.
This is enough uncertainty that I would rather not risk it, because I don't think that aligning behavior that has been out of alignment for a decade is that important.

kenrussell:
Per WebGL working group meeting of 2024-02-08: to make progress on this, it's necessary to create a patch (perhaps against Chromium) which removes the early-out for resizing canvases to the same size, and send that patch through the trybots, which will run the entire WebGL CTS. This will help find whether there is a test which contradicts the one being added here.

PR #3617, opened 2024-02-02
Remove usage of the Theora codec

### LINKED ISSUE #3616: Remove usage of the Theora codec
The WebGL conformance test suite reference a few Theora file, and after years of service, this codec is being retired from the Web.

Some context, for the two implementations that have historically supported Theora, Chromium (except on Android) and Gecko (all platforms):

    https://chromestatus.com/feature/5158654475239424 (cc @dalecurtis)
    https://bugzilla.mozilla.org/show_bug.cgi?id=1860492 (I speak for Mozilla here, I'm handling this)

If need be, I can provide a media file that has exactly the same characteristics and switch it here with a different codec. However, my understanding is that what we care about here are primaries / transfer function / color space / color range and then pixel format / plane arrangement, and this seems to be covered by other codecs, so we should not lose any coverage here, as far as WebGPU is concerned.

See https://github.com/gpuweb/cts/pull/3358 for the WebGPU version of this.

padenot:
This fixes #3616.

I haven't touched anything in `conformance-suites/`, this looked like versioned and frozen test suite, but let me know if I should remove Theora resources there as well.


kenrussell (approved):
This looks perfect, thank you! Yes, we should avoid touching the `conformance-suites/` snapshots since they're supposed to represent points in time.

kenrussell:
Also @padenot could you click through the CLA to satisfy that check? (Mozilla's already a Khronos member, so this doesn't change anything except clearing up that automated check)

kenrussell:
Thank you!

PR #3619, opened 2024-02-06
Promote WEBGL_stencil_texturing to community approved

lexaknyazev:


kenrussell (approved):
As agreed upon in the WebGL working group.

PR #3625, opened 2024-02-22
Remove WebGLObject superinterface.

### LINKED ISSUE #3542: WebGLObject is specified in IDL but none of the implementations implement it
WebGLObject is specified in IDL but none of the implementations implement it

Maybe remove from IDL? It appears to be a specification vehicle, similar to *invalidated* flag.

Comes up in https://github.com/web-platform-tests/interop/issues/363

kenrussell:
Currently no browser exposes this type, causing failures of the automatically-generated Web IDL tests in the Interop suite.

This PR removes WebGLObject from the core WebGL spec and all extension specs.

Fixes #3542.


kenrussell:
After doing these edits I would rather not land them. While WebGLObject may be a specification vehicle as described in #3542 , it is also a useful mechanism to identify which Web IDL interfaces represent graphics resources.

Let's discuss in an upcoming working group conference call.

PR #3626, opened 2024-02-29
[webgl2 spec] "shader-output-incompatible" draw buffer validation.

### LINKED ISSUE #3623: Clarify draw calls with incompatible render targets
The WebGL 2.0 spec says the following:
> If the values written by the fragment shader do not match the format(s) of the corresponding color buffer(s) (for example, the output variable is an integer, but the corresponding color buffer has a floating-point format, or vice versa), and the color buffer states are not set to `NONE` by `DrawBuffers`, draws generate an `INVALID_OPERATION` error; if the color buffer states are set to `NONE` by `DrawBuffers`, they remain untouched.

> If any draw buffer with an attachment does not have a defined fragment shader output, draws generate `INVALID_OPERATION`, unless all 4 channels of `colorMask` are set to false.

1. The language above does not explicitly address the following case: a draw buffer is enabled, has an attachment of an incompatible format (e.g., an integer attachment for a floating-point fragment shader output), and all 4 channels of `colorMask` are set to false.
2. Are there any interactions with `OES_draw_buffers_indexed`? I.e., should indexed color masks be considered for these rules?

kdashg:
Fixes #3623.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -3720,16 +3720,26 @@ WebGL2RenderingContext includes WebGL2RenderingContextOverloads;
     </div>
 
     <p>
-        If the values written by the fragment shader do not match the format(s) of the corresponding
-        color buffer(s) (for example, the output variable is an integer, but the corresponding color
-        buffer has a floating-point format, or vice versa), and the color buffer states are not set to
-        <code>NONE</code> by <code>DrawBuffers</code>, draws generate an <code>INVALID_OPERATION</code> error;
-        if the color buffer states are set to <code>NONE</code> by <code>DrawBuffers</code>, they remain untouched.
+        A draw buffer is "shader-output-incompatible" with a fragment shader if:
+        <ul>
+          <li>the values written to it by the shader do not match the format of the corresponding color buffer
+            (For example, the output variable is an integer, but the corresponding color buffer has a floating-point format, or vice versa)
+          </li>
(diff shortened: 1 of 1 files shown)


kenrussell (approved):
Thank you for putting this together; looks good to me.

@lexaknyazev do you have any feedback on this spec update?

specs/latest/2.0/index.html
```diff
@@ -3720,16 +3720,26 @@ <h3>Draw Buffers</h3>
...
+            (For example, the output variable is an integer, but the corresponding color buffer has a floating-point format, or vice versa)
+          </li>
+          <li>OR there is no defined fragment shader output for its location.</li>
+        </ul>
     </p>
 
     <p>
-        If any draw buffer with an attachment does not have a defined fragment shader output, draws
-        generate <code>INVALID_OPERATION</code>, unless all 4 channels of <code>colorMask</code> are set to false.
+        If any draw buffers are "shader-output-incompatible", generate <code>INVALID_OPERATION</code> if:
```
lexaknyazev, line 3733: s/generate/draws generate

lexaknyazev (approved):
LGTM with a trivial nit.

specs/latest/2.0/index.html
kdashg, line 3733:
```suggestion
        If any draw buffers are "shader-output-incompatible", draws generate <code>INVALID_OPERATION</code> if:
```

PR #3627, opened 2024-02-29
build.py in specs/latest/2.0 as an alternative to Makefile.

kdashg:
It continues in 2024 to be kinda onerous to use make on Windows.
Simple Makefiles can become python scripts without much trouble, so let's consider trying that!


kenrussell (approved):
Nice!

Merging now - let's consider either copying this into the 1.0 directory or moving it up a level and having it work on both the 1.0 and 2.0 specs.

kenrussell:
P.S. Happy to delete the Makefiles, too.

PR #3630, opened 2024-03-06
Add tests for bufferData, bufferSubData and getBufferSubData from WASM Memories of 4GB and 16GB in size.

perryuwang:
Add tests for bufferData, bufferSubData and getBufferSubData from WASM Memories of 4GB and 16GB in size.


kenrussell (approved):
Thank you for these comprehensive tests! They look great!

PR #3632, opened 2024-03-08
Fix broken Chrome troubleshooting URL

pabloasenjo3:
I choose a very similar forum thread, pointing to an identical answer.


kenrussell:
Thanks, I can confirm the original support thread has been deleted. The new one seems reasonable. Thanks for this contribution.

PR #3635, opened 2024-04-03
Add epsilon to float comparison

bsheedy-work:
Adds an epsilon when comparing float values in
conformance/extensions/webgl-depth-texture.html. As-is, it is possible for tests to fail with output such as:

At 2,0, expected within [0.2,0.6], was 0.600
FAIL At 2,0, expected within [0.2,0.6], was 0.600

### DIFF
--- a/sdk/tests/conformance/extensions/webgl-depth-texture.html
+++ b/sdk/tests/conformance/extensions/webgl-depth-texture.html
@@ -323,21 +323,18 @@ function runTestExtension(unpackFlipY) {
                     d01, d01, d11, d11
                 ];
                 expectedMax = expectedMin.slice();
-
-                expectedMin = expectedMin.map(x => x - eps);
-                expectedMax = expectedMax.map(x => x + eps);
             } else {
                 expectedMin = [
-                    d00-eps, d00, d00, d10-eps,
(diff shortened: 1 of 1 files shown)


kdashg (approved):

kainino0x (commented):
Some of the expectations already have epsilons in them (where the expected min and max are equal). This will double epsilon for those.

It seems the original test was trying to avoid overapplying epsilons but it doesn't seem very important to me. So the epsilons could be removed above (lines 327-328, 330-341) and applied here only.

kdashg (approved):

bsheedy-work:
> Some of the expectations already have epsilons in them (where the expected min and max are equal). This will double epsilon for those.
> 
> It seems the original test was trying to avoid overapplying epsilons but it doesn't seem very important to me. So the epsilons could be removed above (lines 327-328, 330-341) and applied here only.

Removed.

kenrussell (commented):
Looks good to me too, but let's give @kainino0x a chance to review.

kainino0x (approved):
This is what I suggested, so LGTM, but I suggested it, so approving based on Ken's secondary review.

PR #3636, opened 2024-04-03
ext-color-buffer-half-float.html now webgl2 when requested.

kdashg:
Previously it was hardcoded to test webgl1.


kenrussell (approved):
Thanks for upgrading this test. LGTM

PR #3637, opened 2024-04-08
attribute USVString WebGLObject.label, application-provided.

kdashg:
This is similar to WebGPU's `GPUObjectBase.label`.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -992,10 +992,18 @@ var context = canvas.getContext('webgl',
     </p>
     <p>
         Each <code>WebGLObject</code> has
-        an <b><a name="webgl-object-invalidated-flag">invalidated</a></b> flag, which is initially unset.
+        an internal <b><a name="webgl-object-invalidated-flag">invalidated</a></b> flag, which is initially unset.
+    </p>
+    <p>
+        Each <code>WebGLObject</code> has a <code>.label</code> attribute, which defaults to <code>""</code>.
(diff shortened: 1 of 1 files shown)


kdashg:
https://www.w3.org/TR/webgpu/#gpuobjectbase
```js
interface mixin GPUObjectBase {
    attribute USVString label;
};
```

greggman:
The difference between WebGPU's label and WebGL is that, AFAIK, WebGPU always returns an object. WebGL does not

```
// on lost context
const b = gl.createBuffer();
b.label = "verts";              // crash!, can't set property label of `null`
```

WebGPU doesn't have this issue

kdashg:
I strongly prefer the ergonomics of .label, and would rather champion infallible allocations instead of reduced ergonomics.

greggman:
In the spirit of the fact that WebGL is supposed to make porting OpenGLES code easier, isn't  [`WEBGL_debug`](https://registry.khronos.org/webgl/extensions/proposals/WEBGL_debug/) the more WebGL-ish thing to support?

Of course it's doesn't have to be either/or but still.

kdashg:
I believe my proposal is the best way forward for the most users.

kkinnunen-apple (approved):
Looks good, thanks.

kenrussell (approved):
This direction sounds good. While it's a different API style than other WebGL extensions, it's expected that this would be implementable/supportable at the native level pretty much ubiquitously.

Issue #3639, opened 2024-04-10
offscreencanvas-sync.html hangs with buffering implementations


kkinnunen-apple:
offscreencanvas-sync.html hangs with buffering implementations

Does not flush the fence sync

kenrussell:
If I was wrong and this wasn't fixed by #3640 , please reopen - thanks.

PR #3640, opened 2024-04-10
offscreencanvas-sync.html Flush to insert fence sync to command strea…

kkinnunen-apple:
…m (#3639)

Flush the commands after inserting fence sync. Otherwise it is not guaranteed that the implementation processes the commands.

Remove the clear commands, their results are not used.


kenrussell (approved):
Looks good to me.

kdashg (approved):

PR #3641, opened 2024-04-11
Test I420 JPEG -> ImageDecoder -> VideoFrame -> WebGLTexture.

kenrussell:
This configuration (specifically, an I420 encoded JPEG) was broken in Chromium. Other red/green tests with different file formats can be added to the same test case later if desired.

Associated with Chromium bug crbug.com/328284177 .

PR #3652, opened 2024-05-04
Fix image-decoder-to-texture in WebGL 2 mode.

kenrussell:
The test was fetching nonexistent "width" and "height" properties on the VideoFrame, leading to a (0, 0)-sized texture upload. codedWidth and codedHeight are the best approximation without handling the full visibleRect in the test.

Follow-on to #3641 .

Associated with Chromium bug crbug.com/337904214 .


kenrussell:
@kdashg @kkinnunen-apple going to merge this myself because it was an obvious bug in the test (once I debugged into the browser :) ) and prevents it from running in WebGL 2.0 mode.

PR #3664, opened 2024-07-17
Clarify clear value for depth buffers is 1.0 for invalidateFramebuffe…

### LINKED ISSUE #3661: Clarify WebGL2 invalidateFramebuffer() depth clear value
> In OpenGL ES 3.0, after calling invalidateFramebuffer or invalidateSubFramebuffer, the affected region's contents become effectively undefined. In WebGL 2.0, it is required for the contents to either stay unchanged or become all 0.

If caller specifies GL_DEPTH_ATTACHMENT or GL_DEPTH_STENCIL_ATTACHMENT, the depth will be invalidated.

The spec now says that invalidated depth buffer is initialized to 0.0.

Contrast this to normal depth buffer initialization:

> If initial data is not provided to these calls, the WebGL implementation must initialize their contents to 0; depth resources must be cleared to the default 1.0 clear depth. 

Uninitialized depth buffer is specified to be initialized to 1.0.

Is this discrepancy intended?

kdashg:
…r and invalidateSubFramebuffer.

Fixes #3661.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -4076,7 +4076,8 @@ WebGL2RenderingContext includes WebGL2RenderingContextOverloads;
     <p>
         In OpenGL ES 3.0, after calling <code>invalidateFramebuffer</code> or <code>invalidateSubFramebuffer</code>,
         the affected region's contents become effectively undefined. In WebGL 2.0, it is required for the contents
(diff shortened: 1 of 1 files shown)


kenrussell (approved):
Thanks, looks good to me.

PR #3668, opened 2024-07-24
Add drawingBufferToneMapping

### LINKED ISSUE #3666: Extended brightness range rendering
The current spec does not clearly indicate what should be done with color values that land outside of the standard dynamic range gamut of the screen (colors that, when converted to the color space of the screen, land outside of the `[0, 1]` interval).

The following proposal indicates a default behavior (that matches existing implementations and WebGPU's default behavior), and provides a mechanism for opting in to extended-range brightness.
https://github.com/ccameron-chromium/webgl-hdr/blob/master/EXPLAINER.md

ccameron-chromium:
Fixes https://github.com/KhronosGroup/WebGL/issues/3666

See: https://github.com/ccameron-chromium/webgl-hdr/blob/master/EXPLAINER.md

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -378,6 +378,11 @@
         an <a href="https://html.spec.whatwg.org/multipage/webappapis.html#imagebitmap">ImageBitmap</a> <a href="#refsHTML">[HTML]</a>
         from this context's canvas.
     </p>
+    <p>
+        When presenting the drawing buffer, the pixel values in the drawing buffer shall be
+        converted from the color space of the drawing buffer to the color space of the screen, and
+        then shall be tone mapped as specified by <code>drawingBufferToneMapping</code>.
+    </p>
     <div class="note">
         <p>
             While it is sometimes desirable to preserve the drawing buffer, it can cause significant
@@ -443,6 +448,46 @@
             generate <code>INVALID_VALUE</code>.
 
             If allocation fails, generate <code>OUT_OF_MEMORY</code>.
+        </dd>
+
+        <dt class="idl-code">undefined drawingBufferToneMapping(WebGLToneMapping toneMapping);
+        <dd>
+            <p>
+              The value of <code>toneMapping.mode</code> indicates how the drawing buffer shall be
+              tone mapped by the HTML page compositor when presented.
+            </p>
+
+            <p>
+              The allowed values are:
+            </p>
+            <dl><dd>
+                <p>
+                  <em>standard</em>
+                </p>
+                <p>
+                  All colors that are inside of the standard dynamic range of the screen shall not
+                  be changed, and all colors that outside of the standard dynamic range of the
+                  screen shall be projected to the standard dynamic range of the screen.
+                </p>
+                <div class="note">
+                  This projection is often accomplished by clamping color values in the color
+                  space of the screen to the [0, 1] interval.
+                </div>
+                <p>
+                  <em>extended</em>
+                </p>
+                <p>
+                  All colors that are inside of the extended dynamic range of the screen shall not
+                  be changed, and all colors that outside of the extended dynamic range of the
+                  screen shall be projected to the extended dynamic range of the screen.
+                </p>
+                <div class="note">
+                  This projection is often accomplished by clamping color values in the color
+                  space of the screen to the interval of values that the screen is capable of
+                  displaying, which may include values greater than 1.
+                </div>
+            </dd></dl>
+        </dd>
     </dl>
 
 <!-- ======================================================================================================= -->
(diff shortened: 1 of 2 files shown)


specs/latest/1.0/index.html
```diff
@@ -443,6 +448,46 @@ <h3><a name="THE_DRAWING_BUFFER">The Drawing Buffer</a></h3>
             generate <code>INVALID_VALUE</code>.
 
             If allocation fails, generate <code>OUT_OF_MEMORY</code>.
+        </dd>
+
+        <dt class="idl-code">undefined drawingBufferToneMapping(WebGLToneMapping toneMapping);
+        <dd>
+            <p>
+              The value of <code>toneMapping.mode</code> indicates how the drawing buffer shall be
+              tone mapped by the HTML page compositor when presented.
```
kkinnunen-apple, line 457: For MSAA contexts, does the projection apply before or after the multisample resolve?

If I recall correctly, current implementations have the problem that already the resolve is done in incorrect colorspace (was it done in linear, even though the end result is used as srgb?). Not sure if that factors into the exact pixel equation tone mapping is about to modify...

specs/latest/1.0/index.html
```diff
@@ -378,6 +378,11 @@ <h3><a name="THE_DRAWING_BUFFER">The Drawing Buffer</a></h3>
         an <a href="https://html.spec.whatwg.org/multipage/webappapis.html#imagebitmap">ImageBitmap</a> <a href="#refsHTML">[HTML]</a>
         from this context's canvas.
     </p>
+    <p>
+        When presenting the drawing buffer, the pixel values in the drawing buffer shall be
+        converted from the color space of the drawing buffer to the color space of the screen, and
+        then shall be tone mapped as specified by <code>drawingBufferToneMapping</code>.
```
kkinnunen-apple, line 384: > Question 1: Why not use an attribute, like drawingBufferColorSpace?
> The IDL definition does not allow attributes to be dictionaries, so that is not an option.

Question 1b: Why WebGLToneMapping has to be a dictionary instead of the single mode entry it contains? Are there common tone mapping parameters that could be passed here in the future that could not be passed as a separate function?

specs/latest/1.0/index.html
ccameron-chromium: Short answer: Tone mapping is done after multisample resolve.

Long answer: The "bad behavior" sometimes seen in this issue is related to color space conversion (namely, color conversion happening after the resolve), not tone mapping (which is after-resolve). In the spec we indicate that tone mapping is done after conversion to the color space of the screen, which is always post-resolve.

Really long answer (about color space conversion, not tone mapping): The resolve is most correctly done in a linear space (a space that has a linear relationship to physical light quantities).

The resolve is done on raw pixel values (with one exception). This means:
* By default, if you're using `RGBA8` as the pixel format, and `"srgb"` or `"display-p3"` as your color space, then you will get "the wrong values", because the pixel values are not linearly related to physical light.
* If the framebuffer is `SRGB8_ALPHA8`, then (IIUC), the resolve will happen after a conversion to linear space, so, you get "the right values".
* If the framebuffer is `RGBA16F` and your color space is `"srgb"` or `"display-p3"`, then you will get "the wrong values".
* If you specify a linear color space (like the below-mentioned `"rec2100-linear"`), then you will get "the right values", because the pixels themselves correspond to linear light values. This will probably be the most common case for WebGL and WebGPU HDR applications.

I'd like for us to add `"rec2100-linear"` as a `PredefinedColorSpace` very soon, but when we do it, I'll want to make sure we get some extra details right (namely, when converted to an image, which will have to be fixed-point, it should be PQ encoded). I think there is now strong consensus on what's "right" here, but I want those issues to be very-well-ventilated in their own feature proposals. See some more comments below.

specs/latest/1.0/index.html
ccameron-chromium: There are tone mapping modes that we plan to add which would use more parameters.

Example: HDR10. We would like to add a mode that matches HDR10 rendering (The reason it's not in this specification is because it's usually framed in relation to what would be the `rec2100-pq` color space, and we want to make sure all of the details of that are [sorted out](https://github.com/whatwg/html/issues/9112) before introducing it. We also want to do it separately to make sure that all of those details are clearly exposed for scrutiny, rather than being put in a giant bundle including other features.)

In that situation, we would want to include content light level info (CLLI) and mastering display color volume (MDCV) metadata, which are standard for use with HDR10. Over in the [explainer](https://github.com/ccameron-chromium/webgpu-hdr/blob/main/EXPLAINER.md) for the WebGPU version (which has now [merged and shipped](https://github.com/gpuweb/gpuweb/issues/4239)), we have a worked example as follows (translated to this API):

```
  gl.drawingBufferStorage(gl.RGBA16F,
                          gl.drawingBufferWidth,
                          gl.drawingBufferHeight);
  gl.drawingBufferColorSpace = "rec2100-linear";
  gl.drawingBufferToneMapping({mode:"hdr10",
                               clli:{maxFALL:200, maxCLL:1200}})
```
On macOS, this would correspond to using [HDR10 CAEDRMetadata](https://developer.apple.com/documentation/quartzcore/caedrmetadata/3194382-hdr10metadatawithdisplayinfo?language=objc), and on Windows it would correspond to using [DXGI_HDR_METADATA_HDR10](https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_5/ns-dxgi1_5-dxgi_hdr_metadata_hdr10).

If such a canvas were to be then encoded to an HDR image or video stream, then the resulting image or video would be required to be include that metadata.

Vanilagy:
What's the blocker on this?

PR #3670, opened 2024-07-24
Add drawingBufferStorage to known context methods

lexaknyazev:
Follow-up to #3651.


kenrussell (approved):
Thanks, looks good. This is non-controversial so I'll merge myself.

PR #3671, opened 2024-07-24
Fix conformance/rendering/blending for WebGL 1.0

lexaknyazev:


sdk/tests/conformance/rendering/blending.html
```diff
@@ -146,7 +146,8 @@
         debug('Blending for RGBA8:');
 
         const gl = CreateContext();
-        fb = CreateValidFb(gl, [[gl.RGBA8, gl.RGBA, gl.UNSIGNED_BYTE]]);
+        // Do not use RGBA8 enum on WebGL 1.0
+        fb = CreateValidFb(gl, [[gl.texImage3D ? gl.RGBA8 : undefined, gl.RGBA, gl.UNSIGNED_BYTE]]);
```
kdashg, line 150:
```suggestion
        fb = CreateValidFb(gl, [[gl.texImage3D ? gl.RGBA8 : gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE]]);
```

kdashg: `undefined` works but I think it's nice to just ask for the right thing up-front.

kdashg (approved):

sdk/tests/conformance/rendering/blending.html
lexaknyazev: Should similar updates be done to `RGBA16F` and `RGBA32F` below? Those rely on `undefined` being a special case in `CreateValidFb`.

kenrussell (approved):
Looks good to me, either way the internalformat is passed to CreateValidFb.

PR #3675, opened 2024-07-25
Do not leak state between adjacent XFB tests

lexaknyazev:
Unbind the buffer from the COPY_WRITE_BUFFER target so that it is bound only as XFB for the next test.


lexaknyazev:
@kenrussell This is just a side-effect cleanup so that the actual issue (tested by the code after the new line) is better represented. Passes in Firefox and WebKit, fails in Chromium just like #3476.

kenrussell (approved):
Thanks for catching this test bug. Looks good.

PR #3679, opened 2024-07-29
Replace Makefiles with build.py in specs/*.

kdashg:


kenrussell (approved):
Thanks Kelsey for this rewrite - tested these locally, looks good.

PR #3681, opened 2024-08-02
Use ANGLE in Chrome test runner

daijh:
Chrome now relies solely on ANGLE, and removes the DesktopGL [1]. This commit update the Chrome flag to "--use-gl=angle" accordingly.

[1] https://issues.chromium.org/issues/40848940


daijh:
This PR is ready for review.

kenrussell (approved):
Thanks for this contribution. The `--use-gl` flag shouldn't be needed any more, but it seems fine to make this change.

PR #3682, opened 2024-08-06
Resize functionality for test view

### LINKED ISSUE #3680: Usability Issue With Testing UI
**Description:**

I would like to report a usability issue I encounter while using the testing framework.

**Problem:**
When utilizing the test user interface, I find it challenging to view the test information effectively. The layout places the test details on the side of the page, with no ability to customize the scale (other than small or large) which makes it difficult to see all the relevant information at once.

**Issues with Current UI:**
**Fixed Expanded View:** The expanded view cannot be adjusted to fit the developers needs, which is less ideal. Users cannot simultaneously view the list of tests and their results comfortably.

**Suggestions for Improvement:**
1. Consider implementing a more flexible layout where users can easily view test information alongside results.
2. Introduce adjustable panels that allows users to customize the visibility of test information according to their preferences.

Improving the usability of the testing framework would significantly enhance the developer experience and efficiency when running and reviewing tests.

registerspoint:
Closes #3680
This addresses a usability issue within the testing framework's user interface. Previously, users encountered difficulties when viewing test information due to a layout that did not support customization. The test details were fixed on the side of the page with limited resizing options.

### DIFF
--- a/sdk/tests/webgl-conformance-tests.html
+++ b/sdk/tests/webgl-conformance-tests.html
@@ -5,1301 +5,1347 @@ found in the LICENSE.txt file.
 -->
 <!DOCTYPE html>
 <html>
+
 <head>
-<meta charset="utf-8">
-<!-- Prevents Chrome from offering to translate tests which generate
+  <meta charset="utf-8">
+  <!-- Prevents Chrome from offering to translate tests which generate
      random characters for things like attribute names -->
-<meta name="google" value="notranslate">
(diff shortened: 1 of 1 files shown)


kdashg (changes requested):
Thanks for the submission!
Unfortunately it's hard to review the important changes when the diff is dominated by formatting changes.
So that we can review it better, let's do formatting changes in a separate PR from the functional changes you're proposing here!

PR #3688, opened 2024-09-05
Clarify GLES/WebGL difference with uninitialized compressed textures.

### LINKED ISSUE #3686: There is no way to initialize a compressed 2D or 3D texture with uninitialized contents
According to GLES 3.0 spec, when calling glCompressedTexImage2D or glCompressedTexImage3D, if one passes a null data pointer, then a texture with uninitialized contents should be allocated in GPU memory:

image

This behavior is the same as with glTexImage2D and glTexImage3D. E.g. to create a 2D texture in WebGL with uninitialized memory contents, one can call

`GL.texImage2D(target, level, internalFormat, width, height, border, format, type, null);`

However, it looks like creating an uninitialized compressed 2D texture is not possible in WebGL. Attempting to run

```html
<!DOCTYPE html><html><body><script>
var gl = document.createElement("canvas").getContext('webgl');
gl.bindTexture(gl.TEXTURE_2D, gl.createTexture());
var dxt1 = gl.getExtension('WEBGL_compressed_texture_s3tc').COMPRESSED_RGB_S3TC_DXT1_EXT;
gl.compressedTexImage2D(gl.TEXTURE_2D, 0, dxt1, 4, 4, 0, null);
</script></body></html>
```

generates errors in both Chrome and Firefox:

image

image

Reading the spec, in case of `texImage2D`, the parameter `pixels` is marked with a `?` to be optional:

image

However for `compressedTexImage2D`, this optionality is not present:

image

Reading section "Differences Between WebGL and OpenGL ES 2.0" on compressed textures, there is no mention of an intentional discrepancy between passing null to texImage2D() vs compressedTextImage2D():

image

suggesting that this may be an unintentional oversight? Or if there is a rationale, it might be good to note in the above section?

This was originally found in https://github.com/emscripten-core/emscripten/issues/19300.

kenrussell:
Fixes #3686.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -4476,8 +4476,10 @@ WebGL2RenderingContext includes WebGL2RenderingContextOverloads;
     <p>
       In OpenGL ES 3.0 it is explicitly allowed to create a compressed 2D texture with uninitialized
       data by passing <code>null</code> to <code>glCompressedTexImage2D</code>. In WebGL 2.0,
-      however, this is still explicitly disallowed. Users can call <code>texStorage2D</code> to work
-      around this functional difference.
+      however, this is still explicitly disallowed. Users can use <code>texStorage2D</code> with a
+      compressed internal format to allocate zero-initialized compressed textures; these textures
+      are however immutable, and not completely compatible with textures allocated
+      via <code>compressedTexImage2D</code>.
     </p>
 
 <!-- ======================================================================================================= -->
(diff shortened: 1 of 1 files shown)


kenrussell:
@juj any feedback welcome. This seemed most appropriate to add to the WebGL 2.0 spec because this has always been disallowed in WebGL 1.0.

lexaknyazev:
Using `texStorage` is not a functional replacement because it would prevent potential texture redefinitions. Assuming that we cannot retroactively allow `null`, the only option is to use `new Uint8Array(n)` with an appropriate size.

kenrussell:
Fair point; I've rewritten this paragraph. Would you please take another look?

juj (approved):
LGTM, thanks for the update!

kdashg (approved):

PR #3689, opened 2024-09-05
UNPACK_COLORSPACE_CONVERSION_WEBGL now applies to all inputs, not just images.

### LINKED ISSUE #3683: Make UNPACK_COLORSPACE_CONVERSION_WEBGL=FALSE apply to all content
For texImage2D, the spec says

> First, the source image data is conceptually converted to the color space specified by the [unpackColorSpace](https://registry.khronos.org/webgl/specs/latest/1.0/#DOM-WebGLRenderingContext-unpackColorSpace) attribute, except if the source image data is an HTMLImageElement, and the UNPACK_COLORSPACE_CONVERSION_WEBGL pixel storage parameter is set to NONE.

This proposal is to remove the "except if the source image data is an HTMLImageElement" part, and instead have this read:

> First, the source image data is conceptually converted to the color space specified by the [unpackColorSpace](https://registry.khronos.org/webgl/specs/latest/1.0/#DOM-WebGLRenderingContext-unpackColorSpace) attribute. If any of the following conditions apply then this color space conversion is skipped, and only YUV to RGB conversion is performed (if applicable):
> * The UNPACK_COLORSPACE_CONVERSION_WEBGL pixel storage parameter is set to NONE
> * The source image data is an ImageBitmap that was created with colorSpaceConversion set to "none".

Prior to this change, `ImageData`, `HTMLCanvasElement`, `HTMLVideoElement`, and `ImageBitmap` have no way to "opt out" of color conversion. With this change, they do.

The motivation for only having `HTMLImageElement` be affected by `UNPACK_COLORSPACE_CONVERSION_WEBGL=NONE` is discussed in [this comment](https://github.com/KhronosGroup/WebGL/issues/3662#issuecomment-2276375174). Those motivations are now obsolete.

We do want to be clear that any data that is a YUV format is *always* converted from YUV to RGB. This applies to images as well as videos (we often think of videos as being stored in YUV, but most images are, too).

kdashg:
Fixes #3683.

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -2769,7 +2769,7 @@ WebGLRenderingContext includes WebGLRenderingContextOverloads;
 
             First, the source image data is conceptually converted to the color space specified by
             the <a href="#DOM-WebGLRenderingContext-unpackColorSpace">unpackColorSpace</a>
-            attribute, except if the source image data is an <code>HTMLImageElement</code>, and
+            attribute, except if the
             the <code>UNPACK_COLORSPACE_CONVERSION_WEBGL</code> pixel storage parameter is set
             to <code>NONE</code>. <br><br>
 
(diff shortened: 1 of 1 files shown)


kenrussell (approved):
Thanks for putting this together! LGTM

PR #3690, opened 2024-09-18
Fix extra/multisample-corruption-stress.html test

kenrussell:


kenrussell:
Tested locally; merging myself.

PR #3691, opened 2024-09-18
Change canvas height each iteration in multisample-corruption test.

kenrussell:
Attempting to make this test a bit more stressful.


toji (approved):
Ah, yes. Because what we all need in life is for things to be more stressful. :wink: 

LGTM

PR #3692, opened 2024-09-19
Add missing constants to the enum tests

lexaknyazev:
Some of the used constants were undefined thus they were not actually tested. The fixed test fails on Chromium and WebKit.


lexaknyazev:
- For Chromium, remove these: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgl/webgl_rendering_context_base.cc;l=3710-3712;drc=47a6f1b0235a3d224745a644daff142614adf7b6
- For WebKit, remove these: https://github.com/WebKit/WebKit/blob/23c822d05dbde32880363d168f935b874af51af5/Source/WebCore/html/canvas/WebGLRenderingContextBase.cpp#L1969-L1970
 
/cc @kkinnunen-apple

kenrussell (approved):
Thanks for improving the test. Looks good to me.

I filed crbug.com/369766154 about this on the Chromium side.

PR #3693, opened 2024-09-20
Add conformance2/uniforms/uniform-block-idents-as-expr.

### LINKED ISSUE #3644: Clarify discarded expressions of  uniform block reference through instance name only
Consider uniform block declaration with a _instance-name_:

`uniform MyBlock { int x; int y; } b; `

The spec spends effort to disallow using `b`  without any field access. You aren't allowed to use `MyBlock`, so it is not easy to refer to plain `b` in a valid program.

Except, you can use it in a discarded statement expression, like `b;`  or in a comma expression like `somethingsomething, b;`

This is a problem because some rewrite steps don't anticipate uniform blocks, and ANGLE assert with these.

Would it be in the spirit of the spec to fail these programs, or should they pass?

As per my reading, that could affect the statically used uniform counts? However, I can't parse the spec to understand if individual members of an uniform block are affected with different static use counts. E.g. if I refer to b.x; does it mean that b.y can be optimized away if not used? Would b; in this case mean that all uniform variables of that block are statically used.

https://registry.khronos.org/OpenGL/specs/es/3.0/GLSL_ES_Specification_3.00.pdf
>For uniform blocks, the application uses the block name to identify the block. Block names have no other use within a shader beyond interface matching; it is an error to use a block name at global scope for anything other than as a block name (e.g., use of a block name for a global variable name or function name is currently reserved). It is a compile-time error to use the same block name for more than one block declaration in the same shader interface (as defined above) within one shader, even if the block contents are identical.

kdashg:
Fixes #3644.

Fails 4-6 subtests on top-of-tree Chrome and Firefox.

### DIFF
--- a/sdk/tests/conformance2/uniforms/00_test_list.txt
+++ b/sdk/tests/conformance2/uniforms/00_test_list.txt
@@ -1,9 +1,9 @@
 --min-version 2.0.1 draw-with-uniform-blocks.html
 --min-version 2.0.1 gl-uniform-arrays-sub-source.html
 --min-version 2.0.1 query-uniform-blocks-after-shader-detach.html
+--min-version 2.0.1 uniform-block-idents-as-expr.html
 --min-version 2.0.1 uniform-blocks-with-arrays.html
 --min-version 2.0.1 simple-buffer-change.html
 --min-version 2.0.1 dependent-buffer-change.html
 --min-version 2.0.1 incompatible-texture-type-for-sampler.html
 --min-version 2.0.1 large-uniform-buffers.html
-
--- a/sdk/tests/conformance2/uniforms/uniform-block-idents-as-expr.html
+++ b/sdk/tests/conformance2/uniforms/uniform-block-idents-as-expr.html
@@ -1,0 +1,107 @@
+<!--
+Copyright (c) 2024 The Khronos Group Inc.
+Use of this source code is governed by an MIT-style license that can be
+found in the LICENSE.txt file.
+-->
+
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>WebGL uniform block identifiers used as expressions</title>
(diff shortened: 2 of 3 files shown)


kdashg:
(I only tested on Windows)

kenrussell:
Thanks very much Kelsey for putting this together. On current Chromium, the failing cases are:

```
test[13]: {"programOk":false,"vsrc":"uniform BlockName { vec4 member; } InstanceName;\nvoid f() { InstanceName; }","fsrc":""}
FAIL Expected linked: false, was true

test[14]: {"programOk":false,"vsrc":"","fsrc":"uniform BlockName { vec4 member; } InstanceName;\nvoid f() { InstanceName; }"}
FAIL Expected linked: false, was true

test[23]: {"programOk":false,"vsrc":"uniform BlockName { vec4 member; } InstanceName[1];\nvoid f() { InstanceName; }","fsrc":""}
FAIL Expected linked: false, was true

test[24]: {"programOk":false,"vsrc":"","fsrc":"uniform BlockName { vec4 member; } InstanceName[1];\nvoid f() { InstanceName; }"}
FAIL Expected linked: false, was true

test[27]: {"programOk":false,"vsrc":"uniform BlockName { vec4 member; } InstanceName[1];\nvoid f() { InstanceName[0]; }","fsrc":""}
FAIL Expected linked: false, was true

test[28]: {"programOk":false,"vsrc":"","fsrc":"uniform BlockName { vec4 member; } InstanceName[1];\nvoid f() { InstanceName[0]; }"}
FAIL Expected linked: false, was true
```

What would be needed to pass these? Does ANGLE's shader translator need to fail compilation of these vertex / fragment shaders? Is more complex validation needed at link time?

kenrussell (approved):

kkinnunen-apple (approved):
Thank you.

lexaknyazev (approved):

kenrussell:
Thanks Kelsey for the test.

PR #3694, opened 2024-09-20
Annotate "[buffered] " logs, and flush on finishTest, not just testFailed.

kdashg:


kenrussell (approved):
Thanks for improving this and sorry for taking so long to review it!

PR #3696, opened 2024-10-02
Use BigInt parameter in 16gb-wasm-memory tests

backes:
The JS API for Memory was updated in https://github.com/WebAssembly/memory64/issues/68 to require BigInt parameters.

### DIFF
--- a/sdk/tests/conformance2/wasm/bufferdata-16gb-wasm-memory.html
+++ b/sdk/tests/conformance2/wasm/bufferdata-16gb-wasm-memory.html
@@ -29,7 +29,7 @@ const SIZE = 16 * 1024 * 1024 * 1024;
 (() => {
   let view;
   try {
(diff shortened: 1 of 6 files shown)


kenrussell (approved):
Thanks for this contribution. In the description, is it more accurate to say "...to require BigInt parameters when the Memory type is i64"?

If so - looks good to me as long as we update the message when this is merged.

Could we also say "Associated with Chromium bug crbug.com/370792790" in the description when merging? Welcome to mention any other browser bugs too.

Thanks in advance.

backes:
Ack, I updated the commit message. Feel free to merge this any time.

kenrussell:
Discussed this in the WebGL WG meeting of 2024-10-03; as long as this is following the WebAssembly spec and has been tested in one implementation it seems fine to merge.

PR #3699, opened 2024-10-18
Test uploading individual LOD levels to 3D textures from images

Jimbly:
See:
https://issues.chromium.org/issues/40910322


Jimbly:
Notes:
* (expected) Currently fails on Chrome on all platforms when uploading mipmaps of larger (128x128) 3D textures (see linked bug)
* (unsure) The `R[GBA]_INTEGER/UNSIGNED_BYTE` for `TEXTURE_3D` (not `TEXTURE_2D_ARRAY`) tests fail for me (on any browser), where it seems the sampling is wrapping around, or something similar - not sure if this is indicating I have something wrong with my tests, or if this is a GPU or implementation bug

sdk/tests/js/tests/tex-image-and-sub-image-3d-with-image.js
```diff
@@ -248,10 +270,111 @@ function generateTest(internalFormat, pixelFormat, pixelType, prologue, resource
...
+            if (!rCoordLocation) {
+                testFailed('Shader incorrectly set up; couldn\'t find uRCoord uniform');
+                return;
+            }
+            gl.uniform1f(rCoordLocation, rTextureCoord);
+
+            if (useMinMax) {
+                // Note: this seems like it should work, but fails on uint textures,
+                //   bug in test / test author, or bug in implementations?
+                gl.texParameterf(bindingTarget, gl.TEXTURE_MIN_LOD, lod);
```
Jimbly, line 301: Using `TEXTURE_MIN/MAX_LOD` here works in all tests except for the unsigned integer formats (where it appears to sample from the wrong level).  I'm not sure if that means there's something I don't understand, or if I've stumbled upon a bug.  I'd be happy to change this code to also test the alternative way to choose which lod levels are sampled from, or to remove this (currently unreachable) branch if desired.

PR #3665, opened 2024-07-18
Tolerate throws in wasm/* tests, add 2gb-in-4gb variants.

kdashg:
For browsers that don't support this large of wasm memory sizes, consider as PASS, so that browser features don't block webgl conformance.
For UAs that throw on use of large TypedArrays, FAIL but safely.

Also, Firefox *does* support ~4GB wasm memory sizes, but currently throws when >2GB TypedArray views are passed to entrypoints. However, Firefox does support <2GB-sized views of >2GB ArrayBuffers, so add tests for that.
This reflects the ability for apps to use large WASM heaps, so long as only smaller views are passed to webgl.

PR #3702, opened 2024-11-29
Fix memory64 tests

backes:
1) Use 'address' instead of 'index'; this was a last-minute change, see
   https://github.com/WebAssembly/memory64/pull/90 and
   https://github.com/WebAssembly/memory64/pull/92.
2) Only treat RangeError as passing test.

Note that the tests were passing before because engines threw a TypeError if the memory descriptor used bigints without specifying {address: 'i64'}.


backes:
@kenrussell PTAL
Note the previous #3696 which changed to using Bigints, and #3665 which allowed memory allocation to throw.

kdashg (approved):

PR #3711, opened 2025-02-06
Test invalid parameter indices for OES_draw_buffers_indexed

lexaknyazev:
The new test cases pass in Firefox but fail in Chromium and WebKit: they return zeros or `false`s instead of `null`s.


kkinnunen-apple (approved):

kenrussell (approved):
Thanks for expanding the tests; looks good to me.

PR #3713, opened 2025-02-07
Fix description of drawRangeElements out-of-range behavior.

### LINKED ISSUE #3710: WebGL 2.0 working draft, drawRangeElements section links to incorrect "out-of-range" description
- The link goes to invalid place in WebGL 1 spec.
- The WebGL 1 spec does not define how the out-of-range indices behave with respect to start, end. E.g. if something else than "discard start, end" is ok behavior, then it's not clear what it is. If "discard start,end" is the only valid behavior, then it's already described in sentence below, and the link to out-of-range description is not needed.
- Typo: "imeplementation"

> void drawRangeElements(GLenum mode, GLuint start, GLuint end, GLsizei count, GLenum type, GLintptr offset) ([OpenGL ES 3.0.6 §2.9.3](https://www.khronos.org/registry/OpenGL/specs/es/3.0/es_spec_3.0.pdf#nameddest=subsection.2.9.3), [man page](http://www.khronos.org/opengles/sdk/docs/man3/html/glDrawRangeElements.xhtml))
> A [[Draw Operation]](https://registry.khronos.org/webgl/specs/latest/1.0/#DRAW_OPERATIONS).
> Draw using the currently bound element array buffer. All error conditions specified for drawElements in the section [Writing to the drawing buffer](https://registry.khronos.org/webgl/specs/latest/1.0/index.html#WRITING_TO_THE_DRAWING_BUFFER) of the WebGL 1.0 specification apply.
> 
> In addition, if indices used to draw are outside the range of [start, end], an imeplementation can either guarantee the behaviors defined in [Out-of-Range Array Accesses](https://registry.khronos.org/webgl/specs/latest/1.0/index.html#OUT_OF_RANGE_ARRAY_ACCESSES), or simply discard the arguments start and end and call drawElements instead. In either situation, no GL errors should be generated for this cause.

kenrussell:
It should have linked to the WebGL 1.0 spec's description of enabled vertex array attributes and range checking.

Fixes #3710.

### DIFF
--- a/specs/latest/2.0/index.html
+++ b/specs/latest/2.0/index.html
@@ -2486,10 +2486,10 @@ WebGL2RenderingContext includes WebGL2RenderingContextOverloads;
         <code>drawElements</code> in the section <a href="../1.0/index.html#WRITING_TO_THE_DRAWING_BUFFER">Writing
         to the drawing buffer</a> of the WebGL 1.0 specification apply.<br><br>
 
-        In addition, if indices used to draw are outside the range of [<em>start</em>, <em>end</em>], an imeplementation
-        can either guarantee the behaviors defined in <a href="../1.0/index.html#OUT_OF_RANGE_ARRAY_ACCESSES">Out-of-Range
-        Array Accesses</a>, or simply discard the arguments <em>start</em> and <em>end</em> and call <code>drawElements</code>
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM

PR #3716, opened 2025-02-14
Test WEBGL_multi_draw with offsets 2 GB into a 4 GB typed array.

kenrussell:
Test case for Chromium bug https://crbug.com/395670641 and Emscripten bug https://github.com/emscripten-core/emscripten/issues/17539 .


kenrussell:
All 4 draw calls fail in current Chromium and pass with the fix for https://crbug.com/395670641 .

shrekshao (approved):

PR #3717, opened 2025-02-20
Add webgl2 tests to conformance/misc/shader-precision-format.html.

kdashg:
Passes in Mac Safari, fails in Mac Firefox.


kenrussell (approved):
I'd prefer to put the WebGL 2 specific parts in a new conformance2/misc/shader-precision-format.html to make it clearer that the test changes behavior under WebGL 2, but as you see fit.

kdashg:
In meeting I believe we agreed that this could land as-is, and PR fixups can be put in afterwards as our time allows.

PR #3721, opened 2025-03-15
Firefox should not call requestVideoFrameCallback to ensure videos are uploadable.

kdashg:
It seems to hang sometimes, maybe due to visibility testing. Fortunately, we believe `await video.play()` is enough in Firefox.

In diagnosing and fixing this, did clean-up refactor of tex-image-and-sub-image-*d-with-video.js to use async+await.


kenrussell (approved):
Looks good - will let you know if this change seems to have any issues when rolled into Chromium.

PR #3722, opened 2025-03-15
Add DRNG (Deterministic RNG) and use it to do fuzzing in attribs/gl-vertex-attrib.html.

kdashg:
Passes on at least macOS Firefox/Safari/Chrome.


kenrussell (approved):
Nice strengthening of the test. Will let you know if this raises any problems when this is rolled into Chromium; hopefully Safari/Edge folks will do the same.

PR #3725, opened 2025-03-21
Imply webgl2 within conformance2/ for getDefault3DContextVersion().

### LINKED ISSUE #3719: drawingBufferStorage Conformance Tests test webgl1 instead of webgl2, skip RGBA16F in webgl1 without EXT_color_buffer_float extension
Consider the following two pages:

- WebGL drawingBufferStorage Conformance Tests - https://registry.khronos.org/webgl/sdk/tests/conformance/canvas/drawingbuffer-storage-test.html
- WebGL2 drawingBufferStorage Conformance Tests - https://registry.khronos.org/webgl/sdk/tests/conformance2/canvas/drawingbuffer-storage-test.html

I'm spotting two distinct (but related) problems.

1.	The first page calls `runTest(1)` and the second page calls `runTest(2)`, intending to test WebGL 1 and WebGL 2 contexts. However, they both end up running the same tests with WebGL 1 because the `contextVersion` parameter in [drawingbuffer-storage-test.js](https://github.com/KhronosGroup/WebGL/blob/64d5c9e77b062c30996068410dd916a2337f7dc1/sdk/tests/js/tests/drawingbuffer-storage-test.js) is unused. Perhaps the intention was to set the default, something like:

	```diff
	 function runTest(contextVersion) {
	   […]
	   let wtu = WebGLTestUtils;
	+  wtu.setDefault3DContextVersion(contextVersion);
	   initialize();
	   if (hasDrawingBufferStorage) {
	     testClearColor();
	     testNoAlpha();
	     testMissingExtension();
	     testMaxSize();
	     testDrawToCanvas();
	   }
	 }
	```

	Or to pass it to the multiple individual `wtu.create3DContext` calls explicitly.

2.	In case of the WebGL 1, there's a comment noting that:

	```
	// WebGL 1 must use EXT_color_buffer_half_float for RGBA16F.
	```

	But it's located inside an `if (gl.getExtension('EXT_color_buffer_float')) {` block, so if the EXT_color_buffer_float extension isn't available, then the test ends up skipping RGBA16F instead of trying the EXT_color_buffer_half_float extension. Maybe something like the following, or a more explicit and/or strict version thereof was intended?

	```diff
	-if (gl.getExtension('EXT_color_buffer_float')) {
	+if (contextVersion == 1 || gl.getExtension('EXT_color_buffer_float')) {
	   // WebGL 1 must use EXT_color_buffer_half_float for RGBA16F.
	   let rgba16f = gl.RGBA16F;
	   if (!rgba16f) {
	     let ext = gl.getExtension('EXT_color_buffer_half_float');
	     if (ext) {
	       rgba16f = ext.RGBA16F_EXT;
	     }
	   }
	   […]
	```

	At least in Chrome 134.0.6998.45 and Chrome Canary 136.0.7059.0 on macOS 15.3.1, the EXT_color_buffer_half_float extension is available under both WebGL 1 and 2 contexts, but EXT_color_buffer_float is only available in a WebGL 2 context, so this causes RGBA16F to be skipped even though it could be tested. If I apply the change above, the test runs, though it doesn't pass for me.

	<details><summary>Results of Testing RGBA16F in a WebGL 1 context</summary>

	```
	Testing RGBA16F
	PASS gl.getError() is gl.NO_ERROR
	PASS gl.drawingBufferFormat is format
	PASS gl.drawingBufferWidth is width
	PASS gl.drawingBufferHeight is height
	FAIL approx equal: expected: 0.25,0.5,0.75,0.125, actual: 0,0,0,0, tolerance: 0.00001
	FAIL gl.getError() should be 0. Was 1280.
	```

	</details>

I hope this report is helpful. Thanks.

kdashg:
Fixes #3719.

### DIFF
--- a/sdk/tests/js/webgl-test-utils.js
+++ b/sdk/tests/js/webgl-test-utils.js
@@ -1581,7 +1581,7 @@ var getUrlOptions = (function() {
   }
 })();
 
-var default3DContextVersion = 1;
+var default3DContextVersion = null;
 
 /**
  * Set the default context version for create3DContext.
@@ -1598,7 +1598,26 @@ var setDefault3DContextVersion = function(version) {
  * then look at the global default3DContextVersion variable.
  */
 var getDefault3DContextVersion = function() {
-    return parseInt(getUrlOptions().webglVersion, 10) || default3DContextVersion;
+    const urlVersion = parseInt(getUrlOptions().webglVersion, 10);
+    if (urlVersion) return urlVersion;
+
+    if (!default3DContextVersion) {
+        const pathname = location.pathname;
+        const path_parts = pathname.split('/');
+        let in_conformance2_path = false;
+        while (path_parts.length) {
+            const part = path_parts.pop();
+            if (part.startsWith('conformance')) {
+                in_conformance2_path = (part == 'conformance2');
+                break;
+            }
+        }
+        // Found this first, so this must be a webgl2 test!
+
+        default3DContextVersion = in_conformance2_path ? 2 : 1;
+        console.warn(`getDefault3DContextVersion(): Implying webgl${default3DContextVersion} based on path.`);
+    }
+    return default3DContextVersion;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


kenrussell (approved):
Thanks for fixing this!

PR #3728, opened 2025-04-09
Add test for tex{Sub}Image2D from HTMLImageElement w/ SVG image w/o natural sizes

fsoder:
The case where an `HTMLImageElement` was referencing an SVG image with no natural width and height wasn't tested.

### DIFF
--- a/sdk/tests/conformance/textures/misc/00_test_list.txt
+++ b/sdk/tests/conformance/textures/misc/00_test_list.txt
@@ -20,7 +20,7 @@ origin-clean-conformance.html
 tex-image-and-sub-image-2d-with-array-buffer-view.html
 tex-image-and-uniform-binding-bugs.html
 --min-version 1.0.3 tex-image-canvas-corruption.html
-tex-image-svg-image-no-natural-width-and-height.html
+--min-version 1.0.4 tex-image-svg-image-no-natural-width-and-height.html
 --min-version 1.0.2 tex-image-webgl.html
 tex-image-with-format-and-type.html
 tex-image-with-invalid-data.html
--- a/sdk/tests/conformance/textures/misc/00_test_list.txt
+++ b/sdk/tests/conformance/textures/misc/00_test_list.txt
@@ -20,6 +20,7 @@ origin-clean-conformance.html
 tex-image-and-sub-image-2d-with-array-buffer-view.html
 tex-image-and-uniform-binding-bugs.html
 --min-version 1.0.3 tex-image-canvas-corruption.html
+tex-image-svg-image-no-natural-width-and-height.html
 --min-version 1.0.2 tex-image-webgl.html
 tex-image-with-format-and-type.html
 tex-image-with-invalid-data.html
--- a/sdk/tests/conformance/textures/misc/tex-image-svg-image-no-natural-width-and-height.html
+++ b/sdk/tests/conformance/textures/misc/tex-image-svg-image-no-natural-width-and-height.html
@@ -58,8 +58,10 @@ wtu.loadImageAsync(`data:image/svg+xml,${SVG_IMAGE}`, image => {
 
   test(image, "texImage2D");
   test(image, "texSubImage2D");
+
+  debug('');
+  finishTest();
 });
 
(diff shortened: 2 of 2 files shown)


sdk/tests/conformance/textures/misc/tex-image-svg-image-no-natural-width-and-height.html
```diff
@@ -0,0 +1,65 @@
...
+const SVG_IMAGE = `<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100'>
+                     <rect width='50' height='50' fill='blue'/>
+                     <rect x='50' y='50' width='50' height='50' fill='yellow'/>
+                   </svg>`;
+const gl = wtu.create3DContext();
+const program = wtu.setupTexturedQuad(gl);
+
+wtu.loadImageAsync(`data:image/svg+xml,${SVG_IMAGE}`, image => {
+  image.width = 100;
+  image.height = 100;
```
kainino0x, line 57: Also perhaps due to the ambiguity above we need a test case that doesn't have width/height on the img?

sdk/tests/conformance/textures/misc/tex-image-svg-image-no-natural-width-and-height.html
```diff
@@ -0,0 +1,65 @@
...
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
+  gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
+
+  if (variant === "texSubImage2D") {
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, image.width, image.height, 0, gl.RGBA,
+                  gl.UNSIGNED_BYTE, null);
+    gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGBA, gl.UNSIGNED_BYTE, image);
+  } else {
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, image);
```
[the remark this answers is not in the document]
fsoder: According to spec it should be 100x100 [*]. Adding that check, thanks.

[*] Actually, if reading all the relevant specs to the letter it should be 0x0, but it seems everyone agrees on the 100x100, so we may as well assume that.

sdk/tests/conformance/textures/misc/00_test_list.txt
```diff
@@ -20,6 +20,7 @@ origin-clean-conformance.html
 tex-image-and-sub-image-2d-with-array-buffer-view.html
 tex-image-and-uniform-binding-bugs.html
 --min-version 1.0.3 tex-image-canvas-corruption.html
+tex-image-svg-image-no-natural-width-and-height.html
```
kenrussell: For bookkeeping purposes we'll want to add `--min-version=1.0.4` here. Let me try to add that directly.

sdk/tests/conformance/textures/misc/00_test_list.txt
```diff
@@ -20,6 +20,7 @@ origin-clean-conformance.html
 tex-image-and-sub-image-2d-with-array-buffer-view.html
 tex-image-and-uniform-binding-bugs.html
 --min-version 1.0.3 tex-image-canvas-corruption.html
+--min-version 1.0.4 tex-image-svg-image-no-natural-width-and-height.html
```
kenrussell, line 23: Added `--min-version=1.0.4` here for bookkeeping purposes.

kenrussell (approved):
Thanks for this new test. It passes in Firefox and fails in Chrome without the fix for crbug.com/409203678 . Currently fails in Safari as well; presumably a similar fix is needed there. Merging now and will highlight this to colleagues at Apple.

PR #3730, opened 2025-05-06
Add getAttribLocation and getUniformLocation error tests

### LINKED ISSUE #3727: Clarify error conditions for getAttribLocation and getUniformLocation
The spec has several statements on `getAttribLocation` error conditions with no explicit precedence:
> - If `program` was generated by a different `WebGLRenderingContext` than this one, generates an `INVALID_OPERATION` error and returns -1.
> - If the passed `name` is longer than the restriction defined in Maximum Uniform and Attribute Location Lengths, generates an `INVALID_VALUE` error and returns -1.
> - Returns -1 if `name` starts with one of the reserved WebGL prefixes per GLSL Constructs.
> - Returns -1 if the context's WebGL context lost flag is set.
> - If the invalidated flag of the passed `program` is set, generates an `INVALID_OPERATION` error and returns -1.

These rules are not mutually exclusive so the expected implementation behavior is ambiguous. Specifically: is it an error if the `name` starts with a reserved prefix and some other condition that requires an error is true? For example, the program has its invalidated flag set and the `name` starts with a reserved prefix.

The `getUniformLocation` command has a similar issue.

lexaknyazev:
Fixes #3727.

Implementation-specific issues caught by the new tests:
- Chromium and WebKit do not check if the program is linked when the name starts with a reserved prefix.
- Firefox does not generate an error if the program is not linked.
- Firefox does generate an error if the name starts with a reserved prefix.


kkinnunen-apple (approved):

kenrussell (approved):
Looks good to me.

@kdashg would you still like to review this as well?

kdashg (approved):

PR #3770, opened 2026-04-06
Fix incorrect assumptions in drawingbuffer-storage-test.

kenrussell:
It is not guaranteed that WebGL implementations can allocate a drawing buffer at any specific size. Avoid asserting that allocations up to the maximum renderbuffer size will succeed.

Associated with https://crbug.com/332743717 .

### DIFF
--- a/sdk/tests/js/tests/drawingbuffer-storage-test.js
+++ b/sdk/tests/js/tests/drawingbuffer-storage-test.js
@@ -13,6 +13,8 @@ let height;
 let format;
 let hasDrawingBufferStorage;
 let maxRenderbufferSize;
+let currentWidth;
+let currentHeight;
 
 function runTest(contextVersion) {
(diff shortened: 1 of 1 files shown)


toji (approved):
LGTM!

lexaknyazev (approved):

tidoust:
That's probably triggered by something else, but this seems to have broken the WebGL extensions registry. For example, this used to return an actual spec and currently yields a directory structure:
https://registry.khronos.org/webgl/extensions/EXT_frag_depth/

Last modified time seems to match the time at which this PR was merged.

Same problem for all extensions under https://registry.khronos.org/webgl/extensions/

kenrussell:
@tidoust I couldn't remember where I saw this report and just re-found your comment here :) The WebGL extension registry has been fixed.

PR #3773, opened 2026-04-29
Fix expected error codes for compressed texture updates

lexaknyazev:
Either error code is acceptable because:

> When a command could potentially generate several different errors ..., the GL implementation may choose to generate any of the applicable errors.

Generic rule: 
> If a command that requires an enumerated value is passed a symbolic constant that is not one of those specified as allowable for that command, an `INVALID_ENUM` error is generated.

Command-specific rule:
> An `INVALID_OPERATION` error is generated if format does not match the internal format of the texture image being modified...


kenrussell (approved):
Looks good to me!

PR #3774, opened 2026-05-28
WebGL1: Forbid #extension: disable after non-preprocessor tokens

### LINKED ISSUE #3772: WebGL1: Forbid `#extension: disable` after non-preprocessor tokens
WebGL1 says:

> In the WebGL API, #extension directives may always occur after non-preprocessor tokens in GLSL ES 1.00 shaders. The scope of #extension directives in GLSL ES 1.00 shaders is always the whole shader, and #extension directives that occur later override those seen earlier for the whole shader.

Can we restrict this such that extensions cannot be _disabled_ later in the shader after they were enabled? It's not a useful scenario to support, and it can be complicated to validate, for example if the shader enables an extension, use it (and pass validation) then disable it at the end of the shader.

Optionally (and less ideally), we could disallow extension disables after non-preprocessor tokens only if that extension was previously enabled by the shader, so the shader can't bypass validation like:

```
#extension Foo: enable
    use_tokens_from_foo();
#extension Foo: disable
```

ShabbyX:
Fixes #3772

### DIFF
--- a/specs/latest/1.0/index.html
+++ b/specs/latest/1.0/index.html
@@ -4526,10 +4526,10 @@ extensions.
     The GLSL ES 1.00 <a href="#refsGLES20GLSL">[GLES20GLSL]</a> specification mandates that
     <code>#extension</code> directives must occur before any non-preprocessor tokens unless the
     extension specification says otherwise. In the WebGL API, <code>#extension</code> directives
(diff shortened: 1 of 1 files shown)


ladyhavoc (approved):

kenrussell:
Tests are up for review - merging this spec change now.

PR #3775, opened 2026-06-10
Test PBOs with a couple of invalid texSubImage2D overloads.

kenrussell:
These are expected to fail with INVALID_OPERATION.

Associated with Chromium bug 521757779.


kainino0x (approved):

kenrussell:
Fixing the lint failures.

PR #3776, opened 2026-06-24
Add tests for late extension enabling and disabling.

kenrussell:
These tests apply only to WebGL 1.0, which for historical reasons, allowed extensions to be enabled after non-preprocessor tokens. It was infeasible to allow extensions to be disabled late in shader parsing, so the spec has been revised to forbid this behavior. This does not seem to impact any WebGL applications in production.

Follow-up to #3774.


kainino0x (approved):
LGTM

kenrussell:
Thanks for your review. Merging.

PR #3779, opened 2026-06-30
chore: Update license to correct MIT version

KhronosWebservices:
Update license from MIT-Khronos-old to MIT, and update copyright dates from 2021 to 2021-2026.


kenrussell (approved):
To other WebGL working group members: this upgrades the license from an "old Khronos MIT license" to the current MIT license. It simplifies metadata that describes this repository.

PR #3780, opened 2026-07-06
WEBGL_render_shared_exponent: Update color writemask validation

lexaknyazev:
Original language was based on the Vulkan spec that required only red, green, and blue color writemask values to be the same.

It turned out that some GPUs apply the alpha writemask to the shared exponent component. As a result of that discovery, the Vulkan spec has been updated (1.4.355, dd June 26, 2026) to require all color writemask values to be the same for color attachments with the shared exponent pixel format.

Since the same requirement exists in Metal and Direct3D, the WebGL extension is updated to match native APIs.

New validation is implemented in Chromium since 152.0.7926.0.


kenrussell (approved):
Looks good to me.

kenrussell:
Note: this was associated with Chromium issue 529454067.

PR #3781, opened 2026-07-06
WEBGL_shader_pixel_local_storage: Sync with upstream

lexaknyazev:
Removed implicit enablement of floating-point rendering extensions to match the ANGLE extension language.

Updated and fixed tests.

---

To pass the updated tests, Chromium would have to stop implicitly enabling floating-point rendering extensions when the PLS extension is being enabled.


kenrussell (approved):
Looks good to me.

PR #3782, opened 2026-07-07
Add WebGL CTS test for active transform feedback buffer tracking

zork:
Add WebGL CTS regression test to switching-objects.html to verify active transform feedback buffer tracking when unbound and paused.

Associated with http://crbug.com/523750584 and https://chromium-review.git.corp.google.com/c/chromium/src/+/8015317


kainino0x (approved):

kenrussell (approved):
Thanks @zork for this well-written test case, and linking it to the associated bug and CL!

PR #3784, opened 2026-07-10
Forbid long variable names that start with _

### LINKED ISSUE #3783: Forbid long symbol names staring with `_`
ANGLE's shader translator prefixes symbol names with `_u`, but it can't do that if the symbol name is at least 1022 characters. To avoid name collisions, it [has validation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/angle/src/compiler/translator/ParseContext.cpp;l=1363;drc=04e32f4479e2b4708c4dc6138e6a4e9011063090) that if a name is at least 1022 characters and starts with `_u`, compilation fails.

A while ago, we made that `u` in `_u` [configurable](https://source.chromium.org/chromium/chromium/src/+/main:third_party/angle/include/GLSLANG/ShaderLang.h;l=614;drc=04e32f4479e2b4708c4dc6138e6a4e9011063090), which means the check is not correct if the compiler is invoked with anything other than `u`. I've also recently found a case where it makes sense to use different prefixes for different symbols, so the check would have to verify multiple prefixes, like `_u`, `_b`, etc.

I'd like to change the validation to forbid any too-long name that starts with `_` [like this](https://chromium-review.git.corp.google.com/c/angle/angle/+/8072999/3/src/compiler/translator/ParseContext.cpp). This requires changing the conformance2/glsl3/shader-with-1024-character-identifier.frag.html test because it does use a long name that starts with `_a`.

ShabbyX:
Fixes #3783


kenrussell (changes requested):
Thanks for revising this test. A few pieces of feedback:

* Please make the change against the top-of-tree conformance tests which are under `sdk/tests/` rather than the older conformance suite snapshots under `conformance-suites/`. We can backport the change after it lands, but tend to keep the older snapshots running under both the old and the new behavior.
* To make git logs easier to read and simplify rolls of the WebGL conformance suite into Chromium, please include details from the issue in the pull request's first message, and mention the Chromium issue number, even if it isn't publicly visible yet.
* Please add a new positive test like `shader-with-1022-character-identifier.frag.html‎`. Add it to `00_test_list.txt` with `--min-version=2.0.1`. Recent PRs where you can follow the pattern: https://github.com/KhronosGroup/WebGL/pull/3785 and https://github.com/KhronosGroup/WebGL/pull/3782 .

If any of this is too tedious let me know and I can help prepare the PR.

kenrussell (changes requested):
Could you please also modify `specs/latest/2.0/index.html`, https://registry.khronos.org/webgl/specs/latest/2.0/#5.3.1 , revising it to the new behavior? That way the spec and tests are in sync.

ShabbyX:
Sorry for the delay, all your requests should be done now

PR #3785, opened 2026-07-17
Add WebGL CTS regression test for switching transform feedback programs

kakashidinho:
Add WebGL CTS regression test in switching-programs.html to verify active transform feedback program tracking when switching programs during paused transform feedback.

Associated with http://crbug.com/513760788 and https://chromium-review.git.corp.google.com/c/chromium/src/+/8107215 .


kenrussell (approved):
LGTM - thank you Quyen for upstreaming this test!

PR #3786, opened 2026-07-21
Add test for WebGL background clear after readback crash

lizeb-google:
This is a non-regression test related to Chromium's https://chromium-review.git.corp.google.com/c/chromium/src/+/8112000.


kenrussell:
Thanks Benoit for putting this test here as I asked - I'm sorry, I didn't realize that it requires infrastructure outside of the browser to test the behavior when the tab's backgrounded and foregrounded. We'll need to add this test to Chromium's GPU tests instead, since the WebGL conformance tests don't run in an external harness. I'll contact you on chat to redo the test.

PR #3788, opened 2026-08-04
Wait for first video frame in texture-srgb-upload.html

zijiehe-google-com:
Unlike other WebGL video tests, texture-srgb-upload.html previously called await video.play() without waiting for requestVideoFrameCallback or using wtu.waitVideoUploadable(video). On slow or asynchronous hardware decoders, texImage2D(..., video) could upload before the first frame was ready, reading all zeros ([0, 0, 0, 0]).

Use await wtu.waitVideoUploadable(video) before running texture upload tests to prevent this race condition, matching the standard idiom used in tex-image-and-sub-image-2d-with-video.js and other video tests.


zijiehe-google-com:
cc @kenrussell Could you take a look at this change. I believe this change can fix this test failure on both android and fuchsia.

zijiehe-google-com:
During Ken's OOO, @toji , can you take a look at this change? Thank you.

toji (approved):
LGTM. Thanks!

zijiehe-google-com:
Thank you.

PR #3791, opened 2026-08-22
WEBGL_shader_pixel_local_storage: Fix tests

lexaknyazev:
- Made canvases always visible during tests
- Added resource deallocations


kenrussell (approved):
Looks good, thank you - one small request for a comment.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
```diff
@@ -12,8 +12,9 @@
 </head>
 <body>
 <div id="description"></div>
-<canvas id="canvas" width="128" height="128" style="background-color:#080"> </canvas>
-<canvas id="canvas_no_alpha" width="128" height="128"> </canvas>
+<canvas id="canvas" width="128" height="128" style="background-color:#080;position:fixed;"> </canvas>
```
kenrussell, line 20: Could you add a comment above this saying that the canvases are fixed-position so that they composite reliably, as opposed to them scrolling off the top of the page and being hidden?

kenrussell (approved):

PR #3792, opened 2026-09-04
WEBGL_shader_pixel_local_storage: Add implicit disable tests

lexaknyazev:


sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
```diff
@@ -470,6 +472,393 @@
...
+              gl.INVALID_ENUM);
+
+    // Valid usage
+    checkEnds(() => { gl.invalidateFramebuffer(gl.FRAMEBUFFER, []); },
+              gl.NO_ERROR);
+    checkEnds(() => { gl.invalidateSubFramebuffer(gl.FRAMEBUFFER, [], 0, 0, 1, 1); },
+              gl.NO_ERROR);
+  });
+
+  testCase("bindFramebuffer", gl.DRAW_FRAMEBUFFER, () => {
```
kenrussell: Could you please alphabetize the test cases by the entry point they test?

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
```diff
@@ -470,6 +472,393 @@
...
+    const data = new Uint8Array(4);
+
+    const buf = gl.createBuffer();
+    gl.bindBuffer(gl.PIXEL_PACK_BUFFER, buf);
+    gl.bufferData(gl.PIXEL_PACK_BUFFER, 4, gl.STATIC_DRAW);
+
+    const texAtt = gl.createTexture();
+    gl.bindTexture(gl.TEXTURE_2D, texAtt);
+    gl.texStorage2D(gl.TEXTURE_2D, 1, gl.RGBA8, 1, 1);
+    gl.framebufferTexture2D(gl.DRAW_FRAMEBUFFER, gl.COLOR_ATTACHMENT0, gl.TEXTURE_2D, texAtt, 0);
```
kenrussell: Why does this code block refer to `gl.DRAW_FRAMEBUFFER` rather than `gl.FRAMEBUFFER`? It may have identical behavior because the same framebuffer object is bound to both binding points, but the code would be more readable if it used a consistent enum.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
```diff
@@ -470,6 +472,393 @@
...
+  });
+
+  testCase("readPixels", gl.FRAMEBUFFER, () => {
+    const data = new Uint8Array(4);
+
+    const buf = gl.createBuffer();
+    gl.bindBuffer(gl.PIXEL_PACK_BUFFER, buf);
+    gl.bufferData(gl.PIXEL_PACK_BUFFER, 4, gl.STATIC_DRAW);
+
+    const texAtt = gl.createTexture();
```
kenrussell: Is it possible to reduce code duplication between this and the other test cases with this identical code block?

kenrussell (changes requested):
Thanks for this nice contribution - looks good overall but a few small requests.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
kenrussell: We discussed this offline and decided to leave it as is.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
lexaknyazev: Done.

sdk/tests/conformance2/extensions/webgl-shader-pixel-local-storage.html
lexaknyazev: Done.

kenrussell (approved):
Thank you Alexey for these thorough tests.

